PR #10, opened 2014-02-13
add an RGBD module

vrabaud:
This is previous work from Willow Garage with Maria Dimashova


czgdp1807:
@vrabaud Can you please share the research paper which was used in implementing this module? I have tried to google but unable to find any related reference. Thanks.

PR #132, opened 2014-11-24
Add RGBD registration function that registers depth to an external camera

pokeefe:
Adding a new function to the RGBD module that implements registration (+tests). This is a common task when working with depth cameras and external RGB cameras. A user would have to find the intrinsics and extrinsics of the depth/rgb cameras separately.

Portions of this code were written by @vrabaud. Specifically, this runs 40-50% faster when the external camera doesn't have any distortion!

Here's an image of a raw depth map, and then the same depth map from the perspective of an external camera 40cm to the left of the depth camera:
depthregistration

### DIFF
--- a/modules/rgbd/src/depth_registration.cpp
+++ b/modules/rgbd/src/depth_registration.cpp
@@ -1,0 +1,381 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
+//
+//  IMPORTANT: READ BEFORE DOWNLOADING, COPYING, INSTALLING OR USING.
+//
+//  By downloading, copying, installing or using the software you agree to this license.
+//  If you do not agree to this license, do not download, install,
+//  copy or use the software.
+//
+//
+//                          License Agreement
+//                For Open Source Computer Vision Library
+//
+// Copyright (C) 2000-2008, Intel Corporation, all rights reserved.
+// Copyright (C) 2009, Willow Garage Inc., all rights reserved.
+// Copyright (C) 2014, OpenCV Foundation, all rights reserved.
+// Third party copyrights are property of their respective owners.
+//
+// Redistribution and use in source and binary forms, with or without modification,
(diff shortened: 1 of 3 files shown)


bhack:
@vpisarev What do you think?

vrabaud:
that was goo dlast time I checked. Can you please just change the license to only include the OpenCV foundation ? Thx

Copyright (C) 2014, OpenCV Foundation, all rights reserved.

modules/rgbd/src/depth_registration.cpp
```diff
@@ -0,0 +1,381 @@
...
+                        Mat &registeredDepth)
+    {
+
+        // Create output Mat of the correct type, filled with an initial value indicating no depth
+        registeredDepth = Mat_<DepthDepth>(outputImagePlaneSize, noDepthSentinelValue<DepthDepth>());
+
+        // Figure out whether we'll have to apply a distortion
+        bool hasDistortion = (countNonZero(registeredDistCoeffs) > 0);
+
+        // A point (i,j,1) will have to be converted to 3d first, by multiplying it by K.inv()
```
bhack: Can be re-used  depth utilities already in the [rgbd module](https://github.com/Itseez/opencv_contrib/blob/master/modules/rgbd/src/depth_to_3d.cpp)?

vrabaud: unfortunately no: more computation is done per point here (like multiplying by rbtRgb2Depth) so it actually has to be all done at once for efficieny. But yes, part of it is the same computation.

bhack:
@pokeefe Can be exposed something in openni to accelerate it also on Android? See https://github.com/occipital/OpenNI2/issues/55

vpisarev:
@pokeefe, thank you very much for the contribution! and terribly sorry for a huge delay; it's better late than never, let's finally put it in :+1:

PR #133, opened 2014-11-26
updating github documentation structure

StevenPuttemans:
As the previous PR suggested, it is important to keep a small overview on the github repo of all the contents of the repo. We started with adding all README documents, I will add info into them as soon as possible. But in the meanwhile we should urge every new module PR to include this information themselves!

Spread the word!


vpisarev:
:+1:

PR #136, opened 2014-11-28
line_descriptor module fix vs2013 build

Dikay900:
Could not compile binary_descriptor.cpp due to a variable 'near' which is defined in
"Windows Kits\8.1\Include\shared\minwindef.h:95"

```
#define near
```

Using Visual Studio 2013, found the same file and line in Windows Kit 8.0 for VS 2012 so it would probably fail there too due to this bug

renamed threshold variable since this belongs together

edit: tested with 2012, same error there


vpisarev:
:+1:

PR #137, opened 2014-11-29
[fix Bug #4023] for msm_efl documentation

Dikay900:
fixing bug http://code.opencv.org/issues/4023
change to latest url thinking this is the correct dataset


StevenPuttemans:
@vpisarev why do such little changes do not get merged faster? :D it is quite essential to the OpenCV challenge!

Dikay900:
still not resolved the stated issue 4023 but promoting PR #190 for this seems to be a better link

PR #139, opened 2014-11-29
datasets module: save\load fr_adience dataset to\from hdf5 file

avdmitry:
save\load fr_adience dataset to\from hdf5 file


terfendail:
Transferred to #1424

PR #141, opened 2014-12-01
Implementation of Pei&Lin Image Normalization

majkrzak:
I've implemented transformation described in article [PeiLin] as a part of my master's thesis. I would be grateful if you would consider including it into OpenCV functionalities.

This transformation may be used to get normalized image, which is invariant to translation, rotation, scaling and skew. It may be used in the process of pattern recognition or in other fields of computer vision. Algorithm is quite simple and it don't need much computation.

---

[PeiLin]: Soo-Chang Pei and Chao-Nan Lin. Image normalization for pattern recognition. Image and Vision Computing, Vol. 13, N.10, pp. 711-723, 1995. doi: [0262-8856(95)98753-G](http://dx.doi.org/10.1016/0262-8856%2895%2998753-G)

---

This feature was discused in Itseez/opencv#3465


majkrzak:
@vpisarev, I've moved this isue here and added simple example, would you like me to something else with it?

terfendail:
Transferred to #1379

PR #144, opened 2014-12-06
update mat2xml.m

xfanzone:
Updated mat2xml.m for star-cascade model in voc-dpm (after version 3). Old xml model crash the new LSVMDetector

### DIFF
--- a/modules/latentsvm/testdata/latentsvm/mat2xml.m
+++ b/modules/latentsvm/testdata/latentsvm/mat2xml.m
@@ -2,43 +2,64 @@ function [] = mat2xml(fname_in, fname_out)
 load(fname_in);
 num_feat = 31;
 rootfilters = [];
-for i = 1:length(model.rootfilters)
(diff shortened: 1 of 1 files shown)


GilLevi:
Amazing contribution! In a previous project, me and my colleagues cracked our heads trying to convert from the Matlab models that we trained to the xml's the OpenCV is familiar with. 

Thanks!

congyao:
I found that for cascade models, we should set num_feat = pcax in mat2xml.m, to make the produced .xml file work well. Or some random bounding boxes will be generated in the detection phase.

The code could be like this:
...
pcax = size(model.pca_coeff, 1);
pcay = size(model.pca_coeff, 2);
num_feat = pcax;   % <--- insert this line here
...

mshabunin:
_latentsvm_ module has been replaced with _dpm_ module, the script has been moved to _opencv_extra_: https://github.com/Itseez/opencv_extra/tree/master/testdata/cv/dpm 

It seems that the script has been updated too and should support new models format.

Issue #125, opened 2014-11-07
descriptor.hpp and types.hpp needs inttypes.h


ilkedemir:
Although inttypes.h is a part of C99, VS2010 (and also 2005?) does not come with it. However descriptor.hpp and types.hpp explicitly need inttypes.h (independent of the "HAVE_INTTYPES_H" flag in cvconfig.h).

It would be better to either include inttypes.h as a source file, or make it optional based on HAVE_INTTYPES_H.

I believe the VS impl of inttypes.h can be found here: https://code.google.com/p/msinttypes/downloads/detail?name=msinttypes-r26.zip&can=2&q=

StevenPuttemans:
This has been merged. @mshabunin this can be closed down.

PR #146, opened 2014-12-06
fix issue for vs versions which aren't standarized for the c99 standard

Dikay900:
Reported and investigated in #125 
A developer has to say if this is the correct way to handle this in general.

fixes compilation error for vs2012, 2010, 2008, 2005 and maybe even prior versions

edited: also needed for vs2012


vpisarev:
:+1:

PR #151, opened 2014-12-28
Fixes required to compile MATLAB bindings

rokm:
The following fixes were required to compile MATLAB bindings under linux with MATLAB R2014b.


vpisarev:
:+1:

PR #153, opened 2015-01-03
Rotation invariant brief

GilLevi:
Added rotation invariance to the BRIEF descriptor. The approach is explained and evaluated in the following my blog post:

https://gilscvblog.wordpress.com/2015/01/02/adding-rotation-invariance-to-the-brief-descriptor/

The descriptor itself is implemented in opencv_contrib/xfeatures2d

Gil.

### DIFF
--- a/modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
+++ b/modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
@@ -41,7 +41,7 @@ the use of this software, even if advised of the possibility of such damage.
 
 #include "opencv2/features2d.hpp"
 #include "opencv2/xfeatures2d/nonfree.hpp"
-
+#include <map>
 /** @defgroup xfeatures2d Extra 2D Features Framework
 @{
 @defgroup xfeatures2d_experiment Experimental 2D Features Algorithms
@@ -133,7 +133,12 @@ namespace cv
 		* Rotation Invariant BRIEF Descriptor
 		*/
 
-		/** @ri_brief Class for computing Rotation Invariant BRIEF descriptors described in @TODO: add reference to my blog
+		/** ri_brief Class for computing Rotation Invariant BRIEF descriptors. Rotation Invariance is added by using the Keypoint's
+			detector's estimation of the patch orientation in order to rotate BRIEF's sampling pairs
+
+			@note
+			-   A complete tutorial comparing between original and rotation invariant BRIEF can be found at:
+			opencv_source_code/tutorial/cpp/RI_BRIEF_match.cpp
 
 		*/
 		class CV_EXPORTS RIBriefDescriptorExtractor : public DescriptorExtractor
@@ -143,6 +148,7 @@ namespace cv
 		protected:
 			static int bit_pattern_64_[512 * 4];
 		};
+
 
 		//! @}
 
--- a/modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
+++ b/modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
@@ -3,9 +3,9 @@ By downloading, copying, installing or using the software you agree to this
 license. If you do not agree to this license, do not download, install,
 copy or use the software.
 
-                          License Agreement
-               For Open Source Computer Vision Library
-                       (3-clause BSD License)
+License Agreement
+For Open Source Computer Vision Library
+(3-clause BSD License)
 
 Copyright (C) 2013, OpenCV Foundation, all rights reserved.
 Third party copyrights are property of their respective owners.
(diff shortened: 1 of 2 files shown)


StevenPuttemans:
Crazy stuff but well done! I like your addition alot!

modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
```diff
@@ -58,80 +58,95 @@ known to be patented. Use them at your own risk.
...
+		{
+		public:
+			static Ptr<BriefDescriptorExtractor> create(int bytes = 32);
+		};
+
+		/*
+		* Rotation Invariant BRIEF Descriptor
+		*/
+
+		/** @ri_brief Class for computing Rotation Invariant BRIEF descriptors described in @TODO: add reference to my blog
```
StevenPuttemans: It seems the @ri_brief and @TODO are not recognized. Might remove them?

GilLevi:
Thanks @StevenPuttemans , I actually want to improve the running time by using look up tables, I'll update my code and make another pull request.

StevenPuttemans:
You can just update this branch and push it again. No need for a new PR :)

GilLevi:
OK, thanks!

bmagyar:
I found it a generally good practice to use size_t for the type of running indices of for loops whenever an std vector is involved. This saves the trouble with warnings when compiling for different architectures and also gets rid of casts.
Such a for loop:
`for (int i = 0; i < (int)keypoints.size(); ++i)`
could be written like this:
`for (size_t i = 0; i < keypoints.size(); ++i)`

For your case I see that this is not really an urgent issue but I wanted to make a note :)

Otherwise +1 for the changes, nice work!

GilLevi:
Thanks @bmagyar, I'll change that!

Str3iber:
Excellent work!
As a suggestion for your next iteration, consider avoiding C-style casts if possible. There is more info here: http://en.cppreference.com/w/cpp/language/explicit_cast

GilLevi:
Thanks @Str3iber , I'll keep that in mind.

GilLevi:
Hi,

I just pushed a more recent version, following your comments.

Is there any way to add a link to my blog post on Rotation Invariant BRIEF as a mean of documentation?

Thanks!
Gil.

StevenPuttemans:
Just add it to the documentation itself, but I think you will have to copy it nonetheless.

cbalint13:
@GilLevi,

First, it is a very nice work !

Allow me some comments, and some personal views:

1) Regarding the code:

- Still see spaces/tabs in places where they should't be (e.g. license header). Can squash all patches into one to see better your full  PR ? (https://help.github.com/articles/about-git-rebase/ see "squash") All three commits are quite big against each other its hard to track the full final code. People might see better and help pinpoint potential problems just in very github interface, without download the code. 

- I see that you re-implement some routines already done in original brief.cpp !? I might be wrong in this statement, or did not see well the things, but how about base your rotations routines against brief.cpp, so there would be no duplicate work. I am thinking here to just add the pre-rotation/pre-scale things and let original brief.cpp (or just some routines from original) to work against.

- See some //TODO section, not sure if its allowed, i might be wrong in this statement.

2) Regarding some ways that coud make opencv to benefit better (my opinion, can take it as hint):

- I see whole ri_brief.cpp as much better addition if would be a generic framework, let me reason below.

- [offtopic] Let me sample that this way we could add e.g. D-BRIEF http://cvlab.epfl.ch/research/detect/dbrief and made it rotation invariant given we know the orientation and scale from detector part. Author released the trainer framework too (it's in matlab, but it is easy to implement in opencv) so it could be well done, and would have the first end-to-end trainable very nice descriptor.

- Such additions that are showed in [offtopic] can be extended to e.g. DAISY and any other present/feature descriptors. The whole ri_brief.cpp work could be just a generic rotation/scaler routine, something like helps any descriptor to be invariant given what detector scale/rotation gives (initial image patch normalizer). The contribution could look much better if would be designed as a more generic framework in addition to opencv_contrib, that just extends to any descriptor the final user wants.

GilLevi:
Hi @cbalint13,

Sorry for the late response. I have some questions that I would be very grateful if you could answer, as I really want this code to be integrated:

1. Would it be better if I'll use BRIEF's routines in rotation invariant BRIEF, thus preventing code duplication?


2. I can easily integrate the framework into the original BRIEF descriptor and just add a flag of "IsRotationInvariant" to BRIEF's constructor. Would that be better? 


Regarding the second comment that you made,I don't think integrating this as a general framwork for descriptors would be a good idea. First, all the other descriptors in OpenCV are already rotation invariant (ORB, BRISK, FREAK, AKAZE, KAZE, SURF and SIFT). Second, the method I suggested does not rotate the patch, but instead rotates the sampling pairs. This is much more efficient and also, rotation the patch might produce "black regions" at the borders (which might be problematic for other descriptors). 

Thanks!
Gil.

cbalint13:
Hello @GilLevi ,

I think your mentioned 2nd point would be the best.

  - Since you tweaked the very sampling pattern (sampling pairs as you said) inside brief.cpp it would be better just to issue a patch against original brief.cpp and add some invariance flag option, so the end user can choose. So no duplicate code, users still have 1:1 original BRIEF plus an extra rotation invariance if they want by a simple additional flag.

  - The other way to just publish a generic function that can "sample-out" from whole-image given scale and rotation might not as good for brief scenario. If you sample from the very whole-image given scale+rotation pattern there will be no black edges, but indeed your solution to tweak brief.cpp's own sampling pairs is way better approach, and it's clear that you also gain performance too.

Thanks Gil !

cbalint13:
Hello @GilLevi ,

Based on our discussions I made a proposal as PR #206 . I hope it looks fine for you, the proposed patch is now well readable: https://github.com/cbalint13/opencv_contrib/commit/1fa9fdc126630b1d5bfd8be2765e41bba77859ce

With Respect,
Christian.

mshabunin:
The discussed functionality has been added in #207

PR #155, opened 2015-01-08
Fix for Bug 4074. Both conditions of if are same

nisargthakkar:
The docs specify it to allow 8UC1 and 8UC3 images.


bmagyar:
:+1:

PR #156, opened 2015-01-13
Fix typo in ERFilter

jaybosamiya:
Noticed in http://docs.opencv.org/trunk/modules/text/doc/erfilter.html that "single" has been written as "sinle" instead. Fixed it at both locations found in the repo.


vpisarev:
:+1:

PR #158, opened 2015-01-14
Integrating BGM/BinBoost descriptor into OpenCV 

GilLevi:
Integrating BGM/BinBoost descriptor into OpenCV 

The descriptor is based on:

1. Trzcinski, Tomasz, et al. "Learning image descriptors with the boosting-trick." Advances in neural information processing systems. 2012.‏
2. Trzcinski, Tomasz, et al. "Boosting binary keypoint descriptors." Computer Vision and Pattern Recognition (CVPR), 2013 IEEE Conference on. Ieee, 2013.‏

This is the authors webpage for BinBoost:
http://cvlab.epfl.ch/research/detect/binboost


cbalint13:
@GilLevi,

Allow me few comments on this initial PR:

- Not sure if it is a rule but lower casing in .src/ "BoostDesc.cpp" into "boostdesc.cpp", given fact that all others are lower case would be perhaps more better.
- Are you sure that you want to integrate binary data blobs (e.g. src/bgm.bin) in ./src folder instead of e.g. a data folder ?
- Those data blobs e.g. bgm.bin (i know it contain matrices for pooling) layout should be documented a bit, they could convert to some more opencv's own friendly and readable format e.g. see http://docs.opencv.org/trunk/modules/core/doc/xml_yaml_persistence.html?highlight=filestorage .
- Also, e.g. bgm.bin is derived from a never release framework, its license and origin are questionable, I would kindly ask author if that trainer framework will released or if there are plans for this, or at last instance if there is fine from his side to publish his derived binary data blobs. Also its origin e.g. that it was trained on http://www.cs.ubc.ca/~mbrown/patchdata/patchdata.html dataset should be reflected in doc. Other research groups released their trainer part too e.g. http://www.robots.ox.ac.uk/~vgg/software/learn_desc/
- xfeatures2d.hpp looks very confusing to me at this moment, first i think you use tabs intead of spaces or you just use spaces where there is no need, second you touched other's code (even the very license headers) that is not related to your boostdesc, insted just plain insert your new code in a clearly visible manner. See visualy just right here in github: https://github.com/Itseez/opencv_contrib/pull/158/files#diff-0
- Original code contains more subvariation of boosted descriptor as i found: BGM,BGM-HARD,BGM-BILINEAR,LBGM,BINBOOST,BINBOOST-128,BINBOOST-256. (namings from the paper, 4 euclidian in floats, 3 in hamming space) I would reflect these using some clear types as argument that can be passed, allowing users to construct/call the descriptor with what they want. See e.g. how is fast.cpp (detector for this time) in main repository that is composed with all the predeclared and documented ::TYPE_9_16, ::TYPE_7_12, ::TYPE_5_8 variations that calls different routines using nice c++ templates.
- Also, noticed that you opened https://github.com/Itseez/opencv/pull/3601 with the sample part of BGM against opencv. That one piece from ther with the sample should go here too.


Would be great to see BGM as part of opencv_contrib !

modules/xfeatures2d/src/BoostDesc.cpp
```diff
@@ -0,0 +1,611 @@
...
+	float cosine = (kp.angle >= 0) ? cos(kp.angle*CV_PI / 180) : 1.f;
+	float sine = (kp.angle >= 0) ? sin(kp.angle*CV_PI / 180) : 0.f;
+
+	float M_[] = {
+		s*cosine, -s*sine, (-s*cosine + s*sine) * patchSize / 2.0f + kp.pt.x,
+		s*sine, s*cosine, (-s*sine - s*cosine) * patchSize / 2.0f + kp.pt.y
+	};
+
+	warpAffine(image, patch, Mat(2, 3, CV_32FC1, M_), Size(patchSize, patchSize),
+		CV_WARP_INVERSE_MAP + CV_INTER_CUBIC + CV_WARP_FILL_OUTLIERS);
```
berak: cv::WARP_INVERSE_MAP + cv::INTER_CUBIC + cv::WARP_FILL_OUTLIERS

modules/xfeatures2d/src/BoostDesc.cpp
```diff
@@ -0,0 +1,611 @@
...
+// tomasz.trzcinski@epfl.ch
+
+//M*/
+#include <opencv2/features2d/features2d.hpp>
+#include <bitset>
+#include <stdlib.h>
+#include <cmath>
+
+using namespace std;
+using namespace cv;
```
berak: //using namespace cv;
namespace cv
{
namespace xfeatures2d
{

modules/xfeatures2d/src/BoostDesc.cpp
```diff
@@ -0,0 +1,611 @@
...
+			desc[d / 8] |= (resp >= 0) ? binLookUp[d % 8] : 0;
+		}
+
+		// clean-up
+		patch.release();
+		gradMap.clear();
+		integralMap.clear();
+	}
+}
+
```
berak: } // namespace xfeatures2d
} // namespace cv

modules/xfeatures2d/src/BoostDesc.cpp
```diff
@@ -0,0 +1,611 @@
...
+// and on any theory of liability, whether in contract, strict liability,
+// or tort (including negligence or otherwise) arising in any way out of
+// the use of this software, even if advised of the possibility of such damage.
+//
+//
+// Copyright(C) 2013 by Tomasz Trzcinski
+// tomasz.trzcinski@epfl.ch
+
+//M*/
+#include <opencv2/features2d/features2d.hpp>
```
berak: #include <opencv2/xfeatures2d.hpp>
(opencv2/features2d/features2d.hpp is a compatibility header, don't include that.)

modules/xfeatures2d/src/BoostDesc.cpp
```diff
@@ -0,0 +1,611 @@
...
+				index = int(theta / binSize);
+				index = (index == orientQuant) ? 0 : index;
+
+				switch (gradAssignType)
+				{
+				case ASSIGN_HARD:
+					gradMap[index].at<uchar>(i, j) = 1;
+					break;
+				case ASSIGN_HARD_MAGN:
+					gradMap[index].at<uchar>(i, j) = round(gradMagnitude);
```
berak: ms compilers don't have round.

use cvRound instead

berak:
replace all your std::strings with cv::String, also, add references, where nessecary

GilLevi:
Thanks for your comments @cbalint13 , I'll try to fix it as soon as I can.

Gil.

StevenPuttemans:
@GilLevi still considering in fixing this and getting it integrated?

GilLevi:
Hi @StevenPuttemans , I'm afraid I don't think I could find the time.

StevenPuttemans:
@GilLevi thanks for the feedback! @cbalint13 you might be up to continue fixing this?

cbalint13:
@StevenPuttemans ,

* Yes it would be very easy to add this one, I address to EPFL author(s) if would agree:

1) GPL -> BSD (mandatory) their demonstrator.
2) Publish trainer part too as part of end-to end solution.

- I also ask @GilLevi if willing to publish trainer part too for LATCH (asked, but never got response) ?
- Would reference https://github.com/Itseez/opencv_contrib/pull/486 as work where descriptor has an end-to-end solution (+trainer part).

StevenPuttemans:
@cbalint13 GREAT! Thumbs up for the effort!

cbalint13:
@StevenPuttemans , Cc to very You the proposal for authors. I always did such preliminary where license or legals are questionable/incompatible with OpenCV efforts.

StevenPuttemans:
@cbalint13 seems fine! Lets await the response before continuing the effort :+1:

StevenPuttemans:
@alalek this one can be closed down. @cbalint13 ported this into a new PR, right here: https://github.com/Itseez/opencv_contrib/pull/648!

PR #159, opened 2015-01-16
Locally uniform comparison image descriptor

Str3iber:
Provides a free implementation of LUCID (Ziegler, Andrew, Eric Christiansen, David Kriegman, and Serge J. Belongie. "Locally uniform comparison image descriptor." In Advances in Neural Information Processing Systems, pp. 1-9. 2012.). LUCID should be much faster, while about as robust, or slightly less than, e.g. SURF or BRIEF. This is compatible with BFMatcher and the code was optimized for OpenCV with the help of gliese581h. This commit includes an exposed separable averaging blur function which is required for LUCID.

### DIFF
--- a/modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
+++ b/modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
@@ -130,22 +130,21 @@ public:
 };
 
 
-// Locally Uniform Comparison Image Descriptor
+/** @brief Class implementing the locally uniform comparison image descriptor, described in @cite LUCID
 
-// @brief This class implements the Locally Uniform Comparison Image Descriptor @cite LUCID
+An image descriptor that can be computed very fast, while being
+about as robust as, for example, SURF or BRIEF.
+ */
 class CV_EXPORTS LUCID : public DescriptorExtractor
 {
 public:
+    /**
+     * @param lucid_kernel kernel for descriptor construction, where 1=3x3, 2=5x5, 3=7x7 and so forth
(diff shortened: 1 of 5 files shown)


modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
```diff
@@ -128,7 +128,26 @@ class CV_EXPORTS BriefDescriptorExtractor : public DescriptorExtractor
...
+    static Ptr<LUCID> create(const int lucid_kernel, const int blur_kernel);
+};
+
+/** @brief Separable box filter blur, needed by LUCID, also exposed for the user
+
+@param _src Image on which blur should be applied
+@param _dst Image resulting from _src having blur applied, the output image
+@param kernel Blur kernel size where 1 equates a 3x3 matrix, 2 = 5x5, 3 = 7x7, and so on
+*/
+CV_EXPORTS void separable_blur(const InputArray _src, CV_OUT OutputArray _dst, const int kernel);
```
vpisarev: looks like cv::blur() does the same thing. If yes, could you remove this function and use the standard one?

vpisarev:
thank you for the contribution! Please, check my comment regarding the blur. Also, the RST documentation is deprecated now, would it be possible to put the docs as Doxygen comments?

Str3iber:
You're welcome! Sure, I'll make the changes, thanks.

Str3iber:
@vpisarev docs failed to merge, though I'm not quite sure why. Error looks random. Please let me know if this is what you had in mind regarding the changes.

vpisarev:
:+1:

PR #160, opened 2015-01-16
Bug fix ticket: 4118, 4119 and 4122

Mirceavoievod:
Bug fix 4119 - line end points outside image range
Bug fix 4122 - invalid matches

Signed-off-by: Mircea Popa <mircea.popa@66.com>

### DIFF
--- a/modules/line_descriptor/src/binary_descriptor.cpp
+++ b/modules/line_descriptor/src/binary_descriptor.cpp
@@ -2223,6 +2223,63 @@ int BinaryDescriptor::EDLineDetector::EdgeDrawing( cv::Mat &image, EdgeChains &e
   return 1;
 }
 
+bool BinaryDescriptor::EDLineDetector::clipLine2f(Size img_size, Point2f& pt1, Point2f& pt2)
+{
+	float x1 = pt1.x;
+	float y1 = pt1.y;
+	float x2 = pt2.x;
+	float y2 = pt2.y;
+
+	float right = (float)(img_size.width - 1);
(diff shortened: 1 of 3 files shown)


modules/line_descriptor/src/binary_descriptor.cpp
```diff
@@ -2353,12 +2353,59 @@ int BinaryDescriptor::EDLineDetector::EDline( cv::Mat &image, LineChains &lines
...
-          lineEndP[1] = (float) ( a2 * Py - a3 * Px - a5 );         //y
+		  float x1 = (float) ( a1 * Px - a3 * Py - a4 );         //x
+          float y1 = (float) ( a2 * Py - a3 * Px - a5 );         //y
           Px = pLineXCors[offsetInLineArray - 1];         //last pixel
           Py = pLineYCors[offsetInLineArray - 1];
-          lineEndP[2] = (float) ( a1 * Px - a3 * Py - a4 );         //x
-          lineEndP[3] = (float) ( a2 * Py - a3 * Px - a5 );         //y
+          float x2 = (float) ( a1 * Px - a3 * Py - a4 );         //x
+          float y2 = (float) ( a2 * Py - a3 * Px - a5 );         //y
+		  // clip line to image size.
```
vpisarev: please, make a function and call it, instead of repeating the clipping algorithm twice

vpisarev:
can you resubmit the patch from non-master branch? as you can see, there are too many changes. also, please, preserve indentation of the code that you did not change, otherwise the patch is too difficult to analyze

PR #168, opened 2015-02-05
contrib_world module

mshabunin:
This pull request contains new module (_opencv_contrib_world_) similar to _opencv_world_ but containing only _opencv_contrib_ modules. It uses compound module mechanism implemented in Itseez/opencv#3666.

Some other changes has been made to avoid compilation warnings on iOS platform.


vpisarev:
:+1:

PR #172, opened 2015-02-13
Function cuda::integral() does not require an extra buffer any more.

sh0:
This fix is required to successfully compile against the current OpenCV HEAD.


jet47:
:+1:

PR #176, opened 2015-02-20
add a 'split' train method to datasets/lfw_benchmark

berak:
to get the exact setup for "view2" in http://vis-www.cs.umass.edu/lfw/#views
(10-fold cross validation based on pairs.txt)


vpisarev:
:+1:

PR #178, opened 2015-02-26
Java and python wrappers for contrib modules

mshabunin:
Compatible with Itseez/opencv#3759

### DIFF
--- a/modules/bgsegm/CMakeLists.txt
+++ b/modules/bgsegm/CMakeLists.txt
@@ -1,2 +1,2 @@
 set(the_description "Background Segmentation Algorithms")
-ocv_define_module(bgsegm opencv_core opencv_imgproc opencv_video opencv_highgui)
+ocv_define_module(bgsegm opencv_core opencv_imgproc opencv_video opencv_highgui WRAP python)
--- a/modules/bioinspired/CMakeLists.txt
+++ b/modules/bioinspired/CMakeLists.txt
@@ -1,3 +1,3 @@
 set(the_description "Biologically inspired algorithms")
 ocv_warnings_disable(CMAKE_CXX_FLAGS -Wundef)
-ocv_define_module(bioinspired opencv_core OPTIONAL opencv_highgui opencv_ocl)
+ocv_define_module(bioinspired opencv_core OPTIONAL opencv_highgui opencv_ocl WRAP java)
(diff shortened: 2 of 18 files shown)


StevenPuttemans:
Just wondering, is there a reason why only 1 module gets java wrappers and all the others only python wrappers?

mshabunin:
@StevenPuttemans, I've copied existing scheme:
- java wrappers had __whitelist__ with supported modules
- python wrappers had __blacklist__ with non-supported modules

In reality marked module can be not available in wrappers, it just does not break the build (like _reg_ in python). And non-marked module can be successfully wrapped (like _optflow_ in java). It is matter of further investigations.

StevenPuttemans:
Thank you for the explanation!

albenoit:
Hi,
i just saw that the bioinspired module does not have the Python wrapper.
what should be done to make it available from python ?
If there is a technical reason within this module code, maybe i can do
something.
Thanks.


2015-02-26 14:20 GMT+01:00 Maksim Shabunin <notifications@github.com>:

> @StevenPuttemans <https://github.com/StevenPuttemans>, I've copied
> existing scheme:
>
>    - java wrappers had *whitelist* with supported modules
>    - python wrappers had *blacklist* with non-supported modules
>
> In reality marked module can be not available in wrappers, it just does
> not break the build (like *reg* in python). And non-marked module can be
> successfully wrapped (like *optflow* in java). It is matter of further
> investigations.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Itseez/opencv_contrib/pull/178#issuecomment-76176462>.
>



-- 
Alexandre BENOIT,
Associate Professor / Maître de Conférence
Image processing and visual scene classification,
LISTIC Lab / IUT Annecy
https://sites.google.com/site/benoitalexandrevision/

mshabunin:
@albenoit, looks like python wrapper generator does not support nested classes. I think it is needed  either to implement this support in generator script, either to remove nested classes (move them to the top level or to remove them at all, like I did in _ml_ module last week).

Maybe aliases can help to keep existing C++ code compatible:
```{.cpp}
struct CV_EXPORTS_W  RetinaParam { ... };
struct CV_EXPORTS_W  Retina
{
    typedef RetinaParam RetinaParameters;
};
```

vpisarev:
:+1:

PR #186, opened 2015-03-14
Plot Module

nmoutinho:
Created a Plot2D Module which allows you to easily plot data from a Mat. You can plot 1D or 2D plots, change the window size, the axis limits and the colors of all the elements. 

In C++, you just have to do:
```
Mat data; 
Mat plot_result;
//-------------------------------//
//     fill your data here
//-------------------------------//
Ptr<plot::Plot2d> plot;
plot = plot::createPlot2d(data);
plot->render(plot_result);
//-------------------------------//
//         show the plot
//-------------------------------//
imshow("plot",plot_result );
```
and that's it.

These are two examples of functions being plotted:
sin(x)
sin_x

x^2
x_2


nmoutinho:
Hi @vpisarev, this is my new module now in opencv_contrib. Hope it will be added to the repository.

modules/plot/doc/plot.rst
```diff
@@ -0,0 +1,6 @@
+Plot - Module to Plot Data
```
vpisarev: the docs are now in Doxygen/Markdown format

modules/plot/include/opencv2/plot.hpp
```diff
@@ -0,0 +1,302 @@
...
+#include <iostream>
+#include <stdio.h>
+
+/**
+@defgroup plot Plot function for Mat data
+*/
+
+namespace cv
+{
+    class CV_EXPORTS Plot
```
vpisarev: there should be nested namespace cv::plot2d or something like that, where all the functionality should be placed

modules/plot/include/opencv2/plot.hpp
```diff
@@ -0,0 +1,302 @@
...
+//
+//################################################################################
+
+#ifndef _OPENCV_PLOT_H_
+#define _OPENCV_PLOT_H_
+#ifdef __cplusplus
+
+#include <opencv2/core.hpp>
+#include <opencv2/highgui.hpp>
+#include <opencv2/imgproc.hpp>
```
vpisarev: imgproc and highgui are big headers; if you do not need them to declare things, please, move these includes to precomp.hpp. The same is for stdio.h and maybe iostream too

modules/plot/include/opencv2/plot.hpp
```diff
@@ -0,0 +1,302 @@
...
+/**
+@defgroup plot Plot function for Mat data
+*/
+
+namespace cv
+{
+    class CV_EXPORTS Plot
+{
+    protected:
+        cv::Mat plotDataX;
```
vpisarev: The class does not follow our abstract interface - hidden implementation concept. Please, look at the background subtraction classes in opencv/modules/video or to the stereo matcher classes in opencv/modules/calib3d. Basically, the exposed class should not contain any protected or public data members, no protected methods, only public methods. It should not have constructor. Instead, it should have static create() method that returns smart pointer (Ptr&lt;&gt;) to the class.

modules/plot/include/opencv2/plot.hpp
```diff
@@ -0,0 +1,302 @@
...
+        cv::Mat plotDataY;
```
vpisarev: "cv::" prefix here is redundant and only increases the syntax noise.

modules/plot/include/opencv2/plot.hpp
```diff
@@ -0,0 +1,302 @@
...
+        cv::Scalar plotGridColor;
+        cv::Scalar plotTextColor;
+
+        //the final plot result
+        cv::Mat plotResult;
+
+    public:
+
+        //constructor accepting only Y data to be plotted
+        Plot(cv::Mat _plotData)
```
vpisarev: all of the implementation must be hidden

modules/plot/include/opencv2/plot/plot.hpp
```diff
@@ -0,0 +1,53 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
```
vpisarev: the subdirectory plot/ and nested plot.hpp are only needed for older modules. For new ones they are unnecessary and should be deleted

modules/plot/src/plot.cpp
```diff
@@ -0,0 +1,209 @@
...
+//M*/
+//################################################################################
+//
+//                    Created by Nuno Moutinho
+//
+//################################################################################
+
+#include "precomp.hpp"
+
+using namespace cv;
```
vpisarev: instead of using global "using namespace cv" you should place all the code into namespace cv { namespace plot2d { } }

modules/plot/include/opencv2/plot.hpp
```diff
@@ -0,0 +1,302 @@
...
+                plotSizeHeight = _plotSizeHeight;
+            else
+                plotSizeHeight = 300;
+        }
+
+        //render the plotResult to a Mat
+        void render(cv::Mat &_plotResult);
+
+        //show the plotResult from within the class
+        void show(const char * _plotName);
```
vpisarev: these 2 methods add unnecessary dependency of highgui and are not really needed. Please, remove them and let user explicitly call imread or imwrite.

nmoutinho:
Hi @vpisarev, thanks for your input. I already made some of the changes you suggested. The biggest one is still missing though, to make the class follow the abstract interface - hidden implementation concept.

nmoutinho:
@vpisarev, I am now uploading a new version where the class follows the concept you proposed. Hope it fills the requirements.

nmoutinho:
@vpisarev, the new version of the module is already uploaded. Thanks.

nmoutinho:
Hi @ilya-lavrenov, I am still waiting for a feedback for this new module, to know if there is something I need to change before it can be merged to the opencv_contrib repository. Thanks.

modules/plot/src/plot.cpp
vpisarev: since you already use "namespace cv { namespace plot {", please, remove "using namespace cv; using namespace plot;" put "using namespace std;" inside "namespace cv { namespace plot {"

nmoutinho:
@vpisarev I've already done the changes you suggested. Thanks!

modules/plot/include/opencv2/plot.hpp
```diff
@@ -0,0 +1,91 @@
...
+//                    Created by Nuno Moutinho
+//
+//################################################################################
+
+#ifndef _OPENCV_PLOT_H_
+#define _OPENCV_PLOT_H_
+#ifdef __cplusplus
+
+#include <opencv2/core.hpp>
+#include <opencv2/imgproc.hpp>
```
vpisarev: looks like you may retain ony opencv2/core.hpp include in the header and move other include's to precomp.hpp. It will improve build speed

nmoutinho:
@vpisarev changes committed.

nmoutinho:
@vpisarev is that anything else I have to do?

nmoutinho:
@vpisarev are you considering to add this module to the repository or is there something I have to change? Thank you for your attention.

modules/plot/src/plot.cpp
```diff
@@ -0,0 +1,425 @@
...
+        using namespace std;
+
+        class Plot2dImpl : public Plot2d
+        {
+            public:
+
+            Plot2dImpl(Mat _plotData)
+            {
+                //if the matrix is not Nx1 or 1xN
+                if(_plotData.cols > 1 && _plotData.rows > 1)
```
paroj: this should be a `CV_Assert`

StevenPuttemans:
@nmoutinho might be interesting for you: https://github.com/Itseez/opencv/pull/5386
If we combine both things, the plotting properties can become even better in OpenCV!

StevenPuttemans:
@nmoutinho is it possible to have a look at the failed checks? This would enable us to fix everything and continue adding this. There is a new user base interested in this http://answers.opencv.org/question/73233/how-do-you-plot-graphs-in-opencv-projects/?comment=73401#comment-73401

nmoutinho:
Hi @StevenPuttemans, it is so cool that someone is interested in my Plot! I was waiting for your response since May (long time has passed...). I am a little busy now with my PhD thesis but as soon as I have time I will do all the changes required and hopefully we will get the Plot function in OpenCv. Thanks!

StevenPuttemans:
That is great! Good luck with your PhD thesis, and hoping to hear soon from you!

vpisarev:
@nmoutinho, sorry for a huge delay! let's finally put it in :+1:

StevenPuttemans:
@vpisarev w00t w00t, I have the impressions you guys are doing a huge job at finally merging everything?

nmoutinho:
Thanks a lot @vpisarev ;) I really hope people like it. It still has a lot to improve but it's a good start.

anupamsobti:
Is there a Python API available as well?

StevenPuttemans:
> Is there a Python API available as well?

Since there is not a single CV_WRAP inside the C++ code, I am afraid not. You could have a look at other modules and contribute the wrapping as a pull request?

anupamsobti:
I'm currently a little occupied. I'll try and get back to this sometime next month.

PR #187, opened 2015-03-16
Bug fix #1

Auron-X:
The bug appears on the initialization step, if the initial bounding box width was larger than height. The problem was in function calculating a binary code for ensemble classifiers. In this function “measurement” values were used in wrong order (h1,w1,h2,w2; instead of: w1,w2,h1,h2), generating a access violation error on “patch” data access by the reason that width>height and pointer to data is calculated linearly data.step*height+width


vpisarev:
thanks! :+1:

plutus666:
good!

PR #196, opened 2015-03-25
Bug fixing

Auron-X:
The bug has the same origin as:
https://github.com/Itseez/opencv_contrib/pull/187
The problem was in "fast" version of function generating a binary code for ensemble classifier posterior probability calculation. Bug was eliminated by fixing the order of "values" in offsets calculating function

### DIFF
--- a/modules/tracking/src/precomp.hpp
+++ b/modules/tracking/src/precomp.hpp
@@ -44,6 +44,6 @@
 
 #include "opencv2/tracking.hpp"
 #include "opencv2/core/utility.hpp"
-#include "opencv2/core/private.hpp"
+//#include "opencv2/core/private.hpp"
 
 #endif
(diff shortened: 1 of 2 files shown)


StevenPuttemans:
It is always good to explain why you are changing the values of the tld tracker or link to a issue on the bug tracker!

modules/tracking/src/precomp.hpp
```diff
@@ -44,6 +44,6 @@
 
 #include "opencv2/tracking.hpp"
 #include "opencv2/core/utility.hpp"
-#include "opencv2/core/private.hpp"
+//#include "opencv2/core/private.hpp"
```
iglesias: If it is not needed, then it is not needed. I don't really see the point of having dead code (especially with includes).

StevenPuttemans: Actually this is quite obvious :) The other change isn't :)

iglesias: I agree. Anyway, I meant this line should me removed rather than commented.

Auron-X:
Sorry for late respond. I pulled out the description of bug to comments and deleted unused header

vpisarev:
:+1:

PR #201, opened 2015-04-06
Fix subpixel precision issue and compute line points in original image size

jesusbriales:
Issue 1:
The subpixel precision of LSD was lost in the implicit conversion to Vec4i
Solved:
Line point coordinates are stored in cv::Vec4f
No type conversion needed from extremes to kl float fields

Issue 2:
The keyline point fields for original image were not being correctly filled
Solved:
The points in original image are computed through scaling (using scale factor and octave index)


alalek:
Source branch have been gone, so close PR

PR #202, opened 2015-04-06
Fix (typo) bug in AGAST scoring routines.

cbalint13:
- This small patch fix an obvious bug in AGAST scoring routine. The bug is due to accidental typo during transcript. AGAST is now verified and produce 1:1 same scoring as the original code.


vpisarev:
:+1:

PR #203, opened 2015-04-06
Fix BRISK (grade it over contrib & use AGAST).

cbalint13:
This PR (re)implements BRISK detector/descriptor by regrade it into opencv_contrib.

* Code bases upon opencv mainstream one, see a very short diff [1] for briefing
* Code uses properly (mandatory) AGAST detection and scoring routines see [2] and [3]
* Uses opencv's image resize() and not original one as [4], hence very small difference in results
* Code generates 1:1 same keypoints and descriptors as original, +/- [4] with minuscule diffs
* Needed to rename the class to "Brisk" instead of "BRISK" due to conflicts in python exports
* It was tested in extensive ways with some SfM, no difference in performance against original

Please do not forget apply trivial bugfix PR #202 for AGAST !

[1] http://openrisc.rdsor.ro/brisk.diff
[2] http://code.opencv.org/issues/3976
[3] http://code.opencv.org/issues/2491
[4] https://github.com/Itseez/opencv/blob/master/modules/features2d/src/brisk.cpp#L2290-L2291

Not sure what will happen to main BRISK but that is bad and cannot be fixed without AGAST.


vpisarev:
thanks for the contribution!
If possible, I'd suggest to fix BRISK in the main repository instead of adding parallel implementation.

cbalint13:
@vpisarev ,

- Without AGAST in main repository it cannot be fixed.  Two possible solutions:

a) grade AGAST to main repo (best solution).
b) would you really like to clone some routines (>75% from AGAST) into brisk.cpp in the main repo ?

vpisarev:
well, let's go with a) then, but the API of both AGAST and BRISK will be frozen for at least 1 year

cbalint13:
Then for a):

- I should propose new PR to push AGAST to opencv/main.
- Should propose new PR to fix BRISK given that AGAST will be in opencv/main.
- Booth of them should be a single PR called ("Add AGAST dector and fix BRISK").

@vpisarev , please confirm / decline the above steps and i start to work on it.

OBS:  API will never change for booth for a very long time, i cannot see reasons for this.

cbalint13:
@vpisarev ,

* Can please look at this new PR https://github.com/Itseez/opencv/pull/3919 . Compile is failing because AGAST now needs to be reverted from opencv_contrib at all ! At this moment that PR fails with "ambigous call".

PR #206, opened 2015-04-08
Add rotation invariance option for BRIEF descriptor.

cbalint13:
This is intended to be a clean patch for PR #153 done by @GilLevi . Some changes in contrast with Gil's original proposal:

* It is against existing brief.cpp code, it is clean and well readable, in accord with opencv coding.
* It does not fork into a new ri_brief.cpp file as original proposal, i believe it is better this way.
* It add optional flag "use_orientation" for existing brief.cpp, so users can enable the feature.

Initial tests in real SfM shows significant improvements for use_orientation=true flag enabled.

### DIFF
--- a/modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
+++ b/modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
@@ -120,7 +120,6 @@ public:
 
 @param bytes legth of the descriptor in bytes, valid values are: 16, 32 (default) or 64 .
 @param use_orientation sample patterns using keypoints orientation, disabled by default.
-@param use_scale sample patterns using keypoints size and scale, disabled by default.
 
 @note
    -   A complete BRIEF extractor sample can be found at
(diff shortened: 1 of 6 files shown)


cbalint13:
Please do not commit **YET** .

* Would like to come with sample code and tests too.
* Would debug if patterns sample proper way given orientations.

cbalint13:
Now can be reviewed, if it's fine even merged. For latest d4baba9 + 0cb72d9 we have :

* By default BRIEF::create(32,false,false) behave 1:1 like original one, generates identical results.
* Rewrote the warp engine, now can follow the rotation and also scale from given keypoints.
* Verified the sampling pattern points to properly rotate and scale over image region.
* I am 100% sure sampling is properly done for all modes 16, 32, 64 by getRotationMatrix2D() !
* Did some tests over the "graffiti" image pairs:

Best 500 SURF keypoints (sorted by score) + additional shallow H filtering step:

       1.  found: 38 of 71 good matches BRISK(64) [default mode]
       2.  found: 0 of 53 good matches BRISK(64,true,false)
       3.  found: 38 of 71 good matches BRISK(64,false,true)
       4.  found: 53 of 121 good matches BRISK(64,true,true)
       5.  found: 58 of 100 good matches RIBRISK(64) [Gil's version]

OBS: The failed 0 of 53 is due to fact SURF generates large scale patterns, for BRIEF is not enough to follow only its rotation given by SURF must follow its scale too !

cbalint13:
Still **NOT** pass rotation invariance tests. Will come back bit later with a  **FIX**.

bmagyar:
Hey Balint, 
The easiest way to mark a PR not ready for maintainers is to put WIP at the beginning of the title of the PR. You can edit it at any moment. WIP stands for Work In Progress.

PR #210, opened 2015-04-16
fixed contrib code to match the HAL

vpisarev:
this is complimentary patch to https://github.com/Itseez/opencv/pull/3935


vpisarev:
:+1:

dhood:
I have just had to revert the hal qualifications of this commit to build the master branch of opencv_contrib (at 4d30d09b), FYI

Dikay900:
you will need an update to date opencv source from the master branch. If you used the opencv-beta the current opencv_contrib code will not work.

With up to date opencv and opencv_contrib source it will build without any problems.
Or maybe you have some special options?

dhood:
Yep, sorry for the noise! not sure how I missed the ref.

PR #214, opened 2015-04-23
bugfix4269 included remarks in http://answers.opencv.org/question/59293/...

LaurentBerger:
...problem-with-example-motemplcpp/


Dikay900:
please do not use tabulators for indentations please use 4 spaces instead.
You can simply edit your file and commit to the same branch "bugfix4269".
It will then show up here too.

vpisarev:
:+1:

PR #217, opened 2015-04-28
fixes on compute normals for Linemod and added tests for 16U depth

mtamburrano:
Fixed the computation of normals when the input os a 16U depth map (as in Linemod case) where NaN values could result.
Removed references about the possibilities to pass the missleading input `depth` as CV_16U, because that parameter refers to the Normals depth output (and unsigned short for normals makes no sense) instead to the depth of the input depth map.
Added tests  for the Linemod 16U case.

### DIFF
--- a/modules/rgbd/src/normal.cpp
+++ b/modules/rgbd/src/normal.cpp
@@ -133,6 +133,7 @@ namespace rgbd
       res = -normal_in / norm_vec(normal_in);
     else
       res = normal_in / norm_vec(normal_in);
+      
     normal_out[0] = res[0];
     normal_out[1] = res[1];
     normal_out[2] = res[2];
@@ -432,8 +433,9 @@ multiply_by_K_inv(const Matx<T, 3, 3> & K_inv, U a, U b, U c, Vec<T, 3> &res)
       K_inv(1, 2) = -K(1, 2) / K(1, 1);
 
       Vec3T X1_minus_X, X2_minus_X;
-
+      
       ContainerDepth difference_threshold = 50;
+      normals.setTo(std::numeric_limits<DepthDepth>::quiet_NaN());
       for (int y = r; y < rows_ - r - 1; ++y)
       {
         const DepthDepth * p_line = reinterpret_cast<const DepthDepth*>(depth.ptr(y, r));
@@ -835,7 +837,7 @@ multiply_by_K_inv(const Matx<T, 3, 3> & K_inv, U a, U b, U c, Vec<T, 3> &res)
     }
 
     // Get the normals
-    normals_out.create(points3d_ori.size(), CV_MAKETYPE(depth_, 3));
+    normals_out.create(points3d_ori.size(), CV_MAKETYPE(depth_, 3));      
     if (points3d_in.empty())
       return;
 
--- a/modules/rgbd/test/test_normal.cpp
+++ b/modules/rgbd/test/test_normal.cpp
@@ -35,6 +35,7 @@
 
 #include "test_precomp.hpp"
 #include <opencv2/rgbd.hpp>
+#include <opencv2/calib3d.hpp>
 
 namespace cv
 {
@@ -128,6 +129,35 @@ float cy = H / 2.f + 0.5f;
 
 Mat K = (Mat_<double>(3, 3) << focal_length, 0, cx, 0, focal_length, cy, 0, 0, 1);
 Mat Kinv = K.inv();
+
+void points3dToDepth16U(const Mat_<Vec3f>& points3d, Mat& depthMap);
+
+void points3dToDepth16U(const Mat_<Vec3f>& points3d, Mat& depthMap)
+{
+  std::vector<Point3f> points3dvec;
+  for(int i = 0; i < H; i++)
+    for(int j = 0; j < W; j++)
+      points3dvec.push_back(Point3f(points3d(i,j)[0], points3d(i,j)[1], points3d(i,j)[2]));
+    
+  std::vector<Point2f> img_points;
+  depthMap = Mat::zeros(H, W, CV_32F);
+  Vec3f R(0.0,0.0,0.0);
+  Vec3f T(0.0,0.0,0.0);
+  cv::projectPoints(points3dvec, R, T, K, Mat(), img_points);
+  
+  int index = 0;
+  for(int i = 0; i < H; i++)
+  {
+    
+    for(int j = 0; j < W; j++)
+    {
+      float value = (points3d.at<Vec3f>(i, j))[2]; // value is the z
+      depthMap.at<float>(cvRound(img_points[index].y), cvRound(img_points[index].x)) = value;
+      index++;
+    }
+  }  
+  depthMap.convertTo(depthMap, CV_16U, 1000);
+}
 
 static RNG rng;
 struct Plane
@@ -224,8 +254,8 @@ protected:
         // inner vector: whether it's 1 plane or 3 planes
         // outer vector: float or double
         std::vector<std::vector<float> > errors(2);
-        errors[0].resize(2);
-        errors[1].resize(2);
+        errors[0].resize(4);
+        errors[1].resize(4);
         switch (i)
         {
           case 0:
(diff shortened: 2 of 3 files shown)


modules/rgbd/src/normal.cpp
```diff
@@ -129,13 +129,22 @@ namespace rgbd
   signNormal(const Vec<T, 3> & normal_in, Vec<T, 3> & normal_out)
   {
     Vec<T, 3> res;
-    if (normal_in[2] > 0)
-      res = -normal_in / norm_vec(normal_in);
+    if(norm_vec(normal_in) == 0) { //avoid division by 0
+      normal_out[0] = 0;
```
vrabaud: if normal is undefined: shouldn't it be NaN ?

modules/rgbd/test/test_normal.cpp
```diff
@@ -129,6 +130,40 @@ float cy = H / 2.f + 0.5f;
...
+      float value = (points3d.at<Vec3f>(i, j))[2]; // value is the z
+      depthMap.at<float>(cvRound(img_points[index].y), cvRound(img_points[index].x)) = value;
+      index++;
+    }
+  }
+  depthMap = depthMap*1000; // convert to millimeters
+  
+  Mat depthMap16U;
+  depthMap.convertTo(depthMap16U, CV_16U);
+  depthMap16U.copyTo(depthMap);
```
vrabaud: useless copy no ?

modules/rgbd/test/test_normal.cpp
```diff
@@ -129,6 +130,40 @@ float cy = H / 2.f + 0.5f;
 Mat K = (Mat_<double>(3, 3) << focal_length, 0, cx, 0, focal_length, cy, 0, 0, 1);
 Mat Kinv = K.inv();
 
+void points3dToDepth16U(const Mat_<Vec3f>& points3d, Mat& depthMap);
+
+void points3dToDepth16U(const Mat_<Vec3f>& points3d, Mat& depthMap)
```
vrabaud: why not just take the 3rd channel of points3d in that function ?

modules/rgbd/src/normal.cpp
mtamburrano: not sure about it, it would not be safer to set 0 for all the components?

modules/rgbd/test/test_normal.cpp
mtamburrano: yes is useless, but if I'm not wrong `convertTo` doesn't allow in-place computation

modules/rgbd/test/test_normal.cpp
mtamburrano: could be enough for this test, I thought to generalize it a bit more when R and T are different by 0 for possible future tests ;)

modules/rgbd/src/normal.cpp
vrabaud: not sure it's safer: we just need a convention. And NaN, well it's undefined.
It's like finding a convention for unknown depth: could be 0 in 16_U or NaN with float.

mtamburrano: I don't think I follow what are you saying... When you talk about "unknown depth" do you mean that if the input is a 16U depth than the normal output should have 0 in the undefined locations or NaN when the input is 32F or 64F? I suppose this is the aim of `quiet_NaN`.

Anyway it means that the user needs to handle NaN values in the output Mat, leading to unusual (maybe just for me) checks, for example when quantizing the normals in the linemod case (by the way I intended to send a PR to replace the actual `quantizedNormals` function to use this normal computer)

modules/rgbd/test/test_normal.cpp
bhack: Probably a simply assignment could be enough. Refcounting could maintain "out of the function scope" memory allocated to depthMap16U in covertTo.

modules/rgbd/src/normal.cpp
vrabaud: what I mean is: if ```norm_vec(normal_in)==0```, you are setting the normal to 0, but I think it should be NaN when giving away floats.

modules/rgbd/test/test_normal.cpp
vrabaud: exactly, ```depthMap=depthMap16U``` would be better. But I don't see anything in the doc that says it does not work in place.

bhack: Funny :) Something from the 1st bug tell me that it is allowed: http://code.opencv.org/issues/1

modules/rgbd/src/normal.cpp
mtamburrano: ok restored

modules/rgbd/test/test_normal.cpp
mtamburrano: I encountered this issue some time ago but trying again seems all fine with inplace conversion, so changed to it

modules/rgbd/src/normal.cpp
```diff
@@ -836,6 +837,7 @@ multiply_by_K_inv(const Matx<T, 3, 3> & K_inv, U a, U b, U c, Vec<T, 3> &res)
 
     // Get the normals
     normals_out.create(points3d_ori.size(), CV_MAKETYPE(depth_, 3));
+    normals_out.setTo(0);
```
vrabaud: as every normal is modified, this is useless right ? There fore that file could be left untouched. Thx

vrabaud:
almost looks good. Please fix the last comment and squash your commits, thx (just do a push -f to your branch after squashing, and things will be transparent here).

modules/rgbd/src/normal.cpp
mtamburrano: In the linemod case, don't seems that every normal is modified [in this `for`](https://github.com/Itseez/opencv_contrib/blob/master/modules/rgbd/src/normal.cpp#L437) but there is a border `r` ignored.
On the contrary in the test all the normals are checked [here](https://github.com/Itseez/opencv_contrib/blob/master/modules/rgbd/test/test_normal.cpp#L327) so the `setTo` influences it.
Since normals should have NaN values on undefined values, I think I should change that `setTo` in this way:
```
  if (depth_ == CV_32F)
      normals_out.setTo(std::numeric_limits<float>::quiet_NaN());
    else
      normals_out.setTo(std::numeric_limits<double>::quiet_NaN());
```

do you agree?

vrabaud: ok, it matters but only for linemod so you should actually set that setTo only before https://github.com/Itseez/opencv_contrib/blob/master/modules/rgbd/src/normal.cpp#L437  (it's just a small performance hit to do a getTo when not needed).

mtamburrano:
`setTo` moved and commits squashed.

vrabaud:
+1 on my end, please merge.

vrabaud:
:+1: I meant :)

bhack:
Ping

mshabunin:
@mtamburrano , please fix trailing whitespaces issues:
```
modules/rgbd/src/normal.cpp:136: trailing whitespace.
+      
modules/rgbd/src/normal.cpp:436: trailing whitespace.
+      
modules/rgbd/src/normal.cpp:840: trailing whitespace.
+    normals_out.create(points3d_ori.size(), CV_MAKETYPE(depth_, 3));      
modules/rgbd/test/test_normal.cpp:141: trailing whitespace.
+    
modules/rgbd/test/test_normal.cpp:147: trailing whitespace.
+  
modules/rgbd/test/test_normal.cpp:151: trailing whitespace.
+    
modules/rgbd/test/test_normal.cpp:158: trailing whitespace.
+  }  
modules/rgbd/test/test_normal.cpp:276: trailing whitespace.
+            
modules/rgbd/test/test_normal.cpp:329: trailing whitespace.
+          
modules/rgbd/test/test_normal.cpp:330: trailing whitespace.
+          if(method == RgbdNormals::RGBD_NORMALS_METHOD_LINEMOD) 
modules/rgbd/test/test_normal.cpp:331: trailing whitespace.
+          {            
modules/rgbd/test/test_normal.cpp:344: trailing whitespace.
+            
```

mshabunin:
Moved to #474

PR #219, opened 2015-04-28
Remove AGAST (for push to opencv's main)

cbalint13:
* Remove AGAST in favor of PR: https://github.com/Itseez/opencv/pull/3919


vpisarev:
:+1:

PR #221, opened 2015-04-30
added a calculation of the Fast Hough Transform of an image

xolodilnik:
Moved code from https://github.com/Itseez/opencv/pull/3968 to opencv_contrib

### DIFF
--- a/modules/ximgproc/include/fast_hough_transform.hpp
+++ b/modules/ximgproc/include/fast_hough_transform.hpp
@@ -1,0 +1,165 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
+//
+//  IMPORTANT: READ BEFORE DOWNLOADING, COPYING, INSTALLING OR USING.
+//
+//  By downloading, copying, installing or using the software you agree to this license.
+//  If you do not agree to this license, do not download, install,
+//  copy or use the software.
+//
+//
+//                           License Agreement
+//                For Open Source Computer Vision Library
+//
+// Copyright (C) 2015, Smart Engines Ltd, all rights reserved.
+// Copyright (C) 2015, Institute for Information Transmission Problems of the Russian Academy of Sciences (Kharkevich Institute), all rights reserved.
+// Copyright (C) 2015, Dmitry Nikolaev, Simon Karpenko, Michail Aliev, Elena Kuznetsova, all rights reserved.
+// Third party copyrights are property of their respective owners.
+//
+// Redistribution and use in source and binary forms, with or without modification,
+// are permitted provided that the following conditions are met:
+//
+//   * Redistribution's of source code must retain the above copyright notice,
+//     this list of conditions and the following disclaimer.
+//
+//   * Redistribution's in binary form must reproduce the above copyright notice,
+//     this list of conditions and the following disclaimer in the documentation
+//     and/or other materials provided with the distribution.
+//
+//   * The name of the copyright holders may not be used to endorse or promote products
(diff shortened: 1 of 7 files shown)


cbalint13:
@xolodilnik , very nice contribution ! Congratulations !

* I like your API design, if this PR pass i am thinking extending your work with fast radon transform + inverse radon transform.

modules/ximgproc/include/fast_hough_transform.hpp
```diff
@@ -0,0 +1,165 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
```
vpisarev: each header should be put to opencv2 subdirectory

modules/ximgproc/include/fast_hough_transform.hpp
```diff
@@ -0,0 +1,165 @@
...
+* @param   makeSkew    Specifies to do or not to do image skewing, see cv::HoughDeskewOption
+* @param   rules       Specifies strictness of line segment calculating, see cv::RulesOption
+* @remarks If rules parameter set to RO_STRICT
+           then returned line cut along the border of source image.
+* @remarks If rules parameter set to RO_WEAK then in case of point, which belongs
+           the incorrect part of Hough image, returned line will not intersect source image.
+*
+* The function calculates coordinates of line segment corresponded by point in Hough space.
+*/
+CV_EXPORTS_W void HoughPoint2Line( OutputArray  line,
```
vpisarev: shall we return Vec4i instead of using heavy OutputArray structure for a single line? Or maybe make it a batch function "HoughPoints2Lines"? Also, the output parameters should be put after input ones (and before the algorithm parameters), so if you decide not to return line but use the output parameter, move it after srcImgInfo. And srcImgInfo should preferably be made InputArray

vpisarev:
@xolodilnik, sorry for delay and thank you very much for the contribution! It looks great! I have one request (move header to opencv2 subdirectory) and one suggestion on the API; and then we can merge it in!

xolodilnik:
@vpisarev , i don't like the idea about "HoughPoints2Lines", because "FastHoughTransform" is calculating a dense voting picture along each discrete line of an image, and result can be used not only for line detection on edges images (for example, i used it for tracking oval objects on the structure tensor image in the videostream). I fixed all your notes, and then I have a warning about not defined function 'pyopencv_to<cv::Vec<int, 4> >' and two unsucsessfull builds. Can you help me with it?

vpisarev:
@xolodilnik, I believe, if you change
<pre>
CV_EXPORTS_W void HoughPoint2Line( const Point &houghPoint,
                                   InputArray  srcImgInfo,
                                   Vec4i       &line,
                                   int         angleRange = ARO_315_135,
                                   int         makeSkew = HDO_DESKEW,
                                   int         rules = RO_IGNORE_BORDERS );
</pre>
to
<pre>
CV_EXPORTS_W Vec4i HoughPoint2Line( const Point &houghPoint,
                                   InputArray  srcImgInfo,
                                   int         angleRange = ARO_315_135,
                                   int         makeSkew = HDO_DESKEW,
                                   int         rules = RO_IGNORE_BORDERS );
</pre>
the problem will likely be solved. If not, temporary replace CV_EXPORTS_W with CV_EXPORTS.

xolodilnik:
@vpisarev, hello, i was fixed all your notes more than month ago. What else should i do to merge it in?

vpisarev:
:+1:

PR #229, opened 2015-05-11
Add DAISY descriptor for wide-baseline / keypoints.

cbalint13:
This PR add DAISY [1] descriptor specialized for dense stereo problems. It is very fast for  wide-baseline [2] but can be used just as normal keypoint descriptor generated by some detector, with quite superior performance among other descriptors.

Some more infos:

* Implementation here is properly cited in respect with authors.
* Implementation is derived from latest author's original BSD licensed code available in public.
* Implementation booth follows opencv's coding standard and author original code performance.
* Implementation behave 1:1 numerically and by speed with original.
* Implementation pass well QA with very good rotation invariance properties.
* Implementation pass poorly QA for scale invariance (this is expected !)
* Implementation pass with excellent results real SfM scenarios for dense stereo-matching.
* Implementation is flexible its parameters are well documented and tested.

[1] http://cvlab.epfl.ch/software/daisy/
[2] http://cvlab.epfl.ch/research/surface/emvs/

A cross PR for QA: https://github.com/Itseez/opencv_extra/pull/254

### DIFF
--- a/modules/xfeatures2d/src/daisy.cpp
+++ b/modules/xfeatures2d/src/daisy.cpp
@@ -1,0 +1,1754 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ * Copyright (c) 2009
+ * Engin Tola
+ * web : http://www.engintola.com
+ * email : engin.tola+libdaisy@gmail.com
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of the Willow Garage nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/*
+ "DAISY: An Efficient Dense Descriptor Applied to Wide Baseline Stereo"
+ by Engin Tola, Vincent Lepetit and Pascal Fua. IEEE Transactions on
+ Pattern Analysis and achine Intelligence, 31 Mar. 2009.
+ IEEE computer Society Digital Library. IEEE Computer Society,
+ http:doi.ieeecomputersociety.org/10.1109/TPAMI.2009.77
+
+ "A fast local descriptor for dense matching" by Engin Tola, Vincent
+ Lepetit, and Pascal Fua. Intl. Conf. on Computer Vision and Pattern
+ Recognition, Alaska, USA, June 2008
+
+ OpenCV port by: Cristian Balint <cristian dot balint at gmail dot com>
+ */
+
+#include "precomp.hpp"
+
+
+#include <fstream>
+#include <stdlib.h>
+
+namespace cv
+{
+namespace xfeatures2d
+{
+
+// constants
+const double g_sigma_0 = 1;
+const double g_sigma_1 = sqrt(2.0);
+const double g_sigma_2 = 8;
+const double g_sigma_step = std::pow(2,1.0/2);
+const int g_scale_st = int( (log(g_sigma_1/g_sigma_0)) / log(g_sigma_step) );
+static int g_scale_en = 1;
+
+const double g_sigma_init = 1.6;
+const static int g_grid_orientation_resolution = 360;
+
+static const int MAX_CUBE_NO = 64;
+static const int MAX_NORMALIZATION_ITER = 5;
+
+int g_cube_number;
+int g_selected_cubes[MAX_CUBE_NO]; // m_rad_q_no < MAX_CUBE_NO
+
+/*
+ !DAISY implementation
+ */
+class DAISY_Impl : public DAISY
+{
+
+public:
+    /** Constructor
+     * @param radius radius of the descriptor at the initial scale
+     * @param q_radius amount of radial range divisions
+     * @param q_theta amount of angular range divisions
+     * @param q_hist amount of gradient orientations range divisions
+     * @param norm normalization type
+     * @param H optional 3x3 homography matrix used to warp the grid of daisy but sampling keypoints remains unwarped on image
+     * @param interpolation switch to disable interpolation at minor costs of quality (default is true)
+     * @param use_orientation sample patterns using keypoints orientation, disabled by default.
+     */
+    explicit DAISY_Impl(float radius=15, int q_radius=3, int q_theta=8, int q_hist=8,
+                        int norm = DAISY::NRM_NONE, InputArray H = noArray(),
+                        bool interpolation = true, bool use_orientation = false);
+
+    virtual ~DAISY_Impl();
+
+    /** returns the descriptor length in bytes */
+    virtual int descriptorSize() const {
+        // +1 is for center pixel
+        return ( (m_rad_q_no * m_th_q_no + 1) * m_hist_th_q_no );
+    };
+
+    /** returns the descriptor type */
+    virtual int descriptorType() const { return CV_32F; }
+
+    /** returns the default norm type */
+    virtual int defaultNorm() const { return NORM_L2; }
+
+    /**
+     * @param image image to extract descriptors
+     * @param keypoints of interest within image
+     * @param descriptors resulted descriptors array
+     */
+    virtual void compute( InputArray image, std::vector<KeyPoint>& keypoints, OutputArray descriptors );
+
+    /** @overload
+     * @param image image to extract descriptors
+     * @param roi region of interest within image
+     * @param descriptors resulted descriptors array
+     */
+    virtual void compute( InputArray image, Rect roi, OutputArray descriptors );
+
+    /** @overload
+     * @param image image to extract descriptors
+     * @param descriptors resulted descriptors array
+     */
+    virtual void compute( InputArray image, OutputArray descriptors );
+
+    /**
+     * @param y position y on image
+     * @param x position x on image
+     * @param ori orientation on image (0->360)
+     * @param descriptor supplied array for descriptor storage
+     */
+    virtual void get_descriptor( double y, double x, int orientation, float* descriptor ) const;
+
+    /**
+     * @param y position y on image
+     * @param x position x on image
+     * @param ori orientation on image (0->360)
+     * @param H homography matrix for warped grid
+     * @param descriptor supplied array for descriptor storage
+     * @param get_descriptor true if descriptor was computed
+     */
+    virtual bool get_descriptor( double y, double x, int orientation, double* H, float* descriptor ) const;
+
+    /**
+     * @param y position y on image
+     * @param x position x on image
+     * @param ori orientation on image (0->360)
+     * @param descriptor supplied array for descriptor storage
+     */
+    virtual void get_unnormalized_descriptor( double y, double x, int orientation, float* descriptor ) const;
+
+    /**
+     * @param y position y on image
+     * @param x position x on image
+     * @param ori orientation on image (0->360)
+     * @param H homography matrix for warped grid
+     * @param descriptor supplied array for descriptor storage
+     * @param get_unnormalized_descriptor true if descriptor was computed
+     */
+    virtual bool get_unnormalized_descriptor( double y, double x, int orientation, double* H, float* descriptor ) const;
+
+
+protected:
+
+    /*
+     * DAISY parameters
+     */
+
+    // maximum radius of the descriptor region.
+    float m_rad;
+
+    // the number of quantizations of the radius.
+    int m_rad_q_no;
+
+    // the number of quantizations of the angle.
+    int m_th_q_no;
+
+    // the number of quantizations of the gradient orientations.
+    int m_hist_th_q_no;
+
+    // holds the type of the normalization to apply; equals to NRM_PARTIAL by
+    // default. change the value using set_normalization() function.
+    int m_nrm_type;
+
+    // number of bins in the histograms while computing orientation
+    int m_orientation_resolution;
+
+    // the number of grid locations
+    int m_grid_point_number;
+
+    // the size of the descriptor vector
+    int m_descriptor_size;
+
+    // size of m_hsz layers at a single sigma: m_hsz * m_layer_size
+    int m_cube_size;
+
+    // size of the layer :
+    // m_roi.width*m_roi.height
+    int m_layer_size;
+
+    // the clipping threshold to use in normalization: values above this value
+    // are clipped to this value for normalize_sift_way() function
+    float m_descriptor_normalization_threshold;
+
+    /*
+     * DAISY switches
+     */
+
+    // if set to true, descriptors are scale invariant
+    bool m_scale_invariant;
+
+    // if set to true, descriptors are rotation invariant
+    bool m_rotation_invariant;
+
+    // if enabled, descriptors are computed with casting non-integer locations
+    // to integer positions otherwise we use interpolation.
+    bool m_disable_interpolation;
+
+    // switch to enable sample by keypoints orientation
+    bool m_use_orientation;
+
+    /*
+     * DAISY arrays
+     */
+
+    // holds optional H matrix
+    Mat m_h_matrix;
+
+    // input image.
+    Mat m_image;
+
+    // image roi
+    Rect m_roi;
+
+    // stores the descriptors :
+    // its size is [ m_roi.width*m_roi.height*m_descriptor_size ].
+    Mat m_dense_descriptors;
+
+    // stores the layered gradients in successively smoothed form :
+    // layer[n] = m_gradient_layers * gaussian( sigma_n );
+    // n>= 1; layer[0] is the layered_gradient
+    Mat m_smoothed_gradient_layers;
+
+    // hold the scales of the pixels
+    Mat m_scale_map;
+
+    // holds the orientaitons of the pixels
+    Mat m_orientation_map;
+
+    // Holds the oriented coordinates (y,x) of the grid points of the region.
+    Mat m_oriented_grid_points;
+
+    // holds the gaussian sigmas for radius quantizations for an incremental
+    // application
+    Mat m_cube_sigmas;
+
+    // Holds the coordinates (y,x) of the grid points of the region.
+    Mat m_grid_points;
+
+    // holds the amount of shift that's required for histogram computation
+    double m_orientation_shift_table[360];
+
+
+private:
+
+    // two possible computational mode
+    // ONLY_KEYS -> (mode_1) compute descriptors on demand
+    // COMP_FULL -> (mode_2) compute all descriptors from image
+    enum { ONLY_KEYS = 0, COMP_FULL = 1 };
+
+    /*
+     * DAISY functions
+     */
+
+    // initializes the class: computes gradient and structure-points
+    inline void initialize();
+
+    // initializes for get_descriptor(double, double, int) mode: pre-computes
+    // convolutions of gradient layers in m_smoothed_gradient_layers
+    inline void initialize_single_descriptor_mode();
+
+    // set & precompute parameters
+    inline void set_parameters();
+
+    // image set image as working
+    inline void set_image( InputArray image );
+
+    // releases all the used memory; call this if you want to process
+    // multiple images within a loop.
+    inline void reset();
+
+    // releases unused memory after descriptor computation is completed.
+    inline void release_auxiliary();
+
+    // computes the descriptors for every pixel in the image.
+    inline void compute_descriptors();
+
+    // computes scales for every pixel and scales the structure grid so that the
+    // resulting descriptors are scale invariant.  you must set
+    // m_scale_invariant flag to 1 for the program to call this function
+    inline void compute_scales();
+
+    // compute the smoothed gradient layers.
+    inline void compute_smoothed_gradient_layers();
+
+    // computes pixel orientations and rotates the structure grid so that
+    // resulting descriptors are rotation invariant. If the scales is also
+    // detected, then orientations are computed at the computed scales. you must
+    // set m_rotation_invariant flag to 1 for the program to call this function
+    inline void compute_orientations();
+
+    // computes the histogram at yx; the size of histogram is m_hist_th_q_no
+    inline void compute_histogram( float* hcube, int y, int x, float* histogram );
+
+    // reorganizes the cube data so that histograms are sequential in memory.
+    inline void compute_histograms();
+
+    // computes the sigma's of layers from descriptor parameters if the user did
+    // not sets it. these define the size of the petals of the descriptor.
+    inline void compute_cube_sigmas();
+
+    // Computes the locations of the unscaled unrotated points where the
+    // histograms are going to be computed according to the given parameters.
+    inline void compute_grid_points();
+
+    // Computes the locations of the unscaled rotated points where the
+    // histograms are going to be computed according to the given parameters.
+    inline void compute_oriented_grid_points();
+
+    // normalizes the descriptor
+    inline void normalize_descriptor( float* desc, int nrm_type ) const;
+
+    // applies one of the normalizations (partial,full,sift) to the desciptors.
+    inline void normalize_descriptors( int nrm_type = DAISY::NRM_NONE );
+
+
+    // emulates the way sift is normalized.
+    inline void normalize_sift_way( float* desc ) const;
+
+    // normalizes the descriptor histogram by histogram
+    inline void normalize_partial( float* desc ) const;
+
+    // normalizes the full descriptor.
+    inline void normalize_full( float* desc ) const;
+
+    // normalizes histograms individually
+    inline void normalize_histograms();
+
+    inline void update_selected_cubes();
+
+    // Smooth a histogram by using a [1/3 1/3 1/3] kernel.  Assume the histogram
+    // is connected in a circular buffer.
+    inline void smooth_histogram( Mat hist, int bins );
+
+    // smooths each of the layers by a Gaussian having "sigma" standart
+    // deviation.
+    inline void smooth_layers( Mat layers, int h, int w, int layer_number, float sigma );
+
+    // returns the descriptor vector for the point (y, x) !!! use this for
+    // precomputed operations meaning that you must call compute_descriptors()
+    // before calling this function. if you want normalized descriptors, call
+    // normalize_descriptors() before calling compute_descriptors()
+    inline void get_descriptor( int y, int x, float* &descriptor );
+
+    // does not use interpolation while computing the histogram.
+    inline void ni_get_histogram( float* histogram, int y, int x, int shift, float* hcube ) const;
+
+    // returns the interpolated histogram: picks either bi_get_histogram or
+    // ti_get_histogram depending on 'shift'
+    inline void i_get_histogram( float* histogram, double y, double x, double shift, float* cube ) const;
+
+    // records the histogram that is computed by bilinear interpolation
+    // regarding the shift in the spatial coordinates. hcube is the
+    // histogram cube for a constant smoothness level.
+    inline void bi_get_histogram( float* descriptor, double y, double x, int shift, float* hcube ) const;
+
+    // records the histogram that is computed by trilinear interpolation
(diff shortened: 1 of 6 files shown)


cbalint13:
All passed.

* Current opencv_contrib buildbot have perf/test disabled.
* Squashed all into https://github.com/cbalint13/opencv_contrib/commit/96f1e2566096e3c95098b95273aef8b271a36070

StevenPuttemans:
Nice!

vpisarev:
@cbalint13, thank you a lot for the contribution! We planned to put DAISY into OpenCV for a long time, glad you finally working on it.

I will look into details later, but have a few immediate comments/suggestion regarding the API.
DAISY could be used as sparse descriptor, of course, but its main role is dense descriptor. I think, it would be very, very inefficient to construct a huge array of keypoints and then get descriptors for all of them. More efficient approach would be to add a dedicated method to compute descriptors for each pixel in a ROI of image. Why ROI? It's convenient for parallel processing (by tiles), it's enough information to process to compensate for overhead and it's less space for output descriptors vs the whole image case (but user can always pass ROI covering the whole image if needed).

Of course, before that the image should be passed and preliminary calculations should be made. That is, the following method could be added to DAISY:

class DAISY ... {
...
// stores the collection of filtered images, from which the descriptors can be extracted.
virtual void setImage(InputArray img, <probably some extra params>) = 0;
// stores descriptors for the specified ROI into 2D array (roi.width*roi.height x size_of_DAISY_descriptor). // note the constness, allowing parallel execution.
virtual void compute(Rect roi, OutputArray descriptors, <some more params, perhaps>) const = 0;
};

what do you think?

cbalint13:
@vpisarev ,

Lets clarify DAISY behaviour at this moment of implementation, and its main work-mode flags:

1. DAISY::COMP_FULL when keypoints are ignored and must be _blank_ (there is an assert to check this in this very mode) all pixels are computed, thus in the end the descriptors resulted for Image(x,y) can be accessed by something like this: descriptors.ptr<float>( y*descriptorSize() + x ). I took _special_ care in this dense mode to be as fast as original code and avoid any kind of .clone() or other things, i just reference author's original container into final cv::Mat(), see here: https://github.com/Itseez/opencv_contrib/pull/229/files#diff-453dc15e0b4d66e076bc9e89749e3a13R1970 In this mode DAISY is blazing fast.
  Your statement "very, very inefficient to construct a huge array of keypoints" is not true, i asserted into code,  in this very mode keypoints to be always supplied as _empty_.

2. DAISY::ONLY_KEYS , the classical, when keypoints can be supplied and descriptor is computed _only_ for each keypoint: see  https://github.com/Itseez/opencv_contrib/pull/229/files#diff-453dc15e0b4d66e076bc9e89749e3a13R1980I do not describe anyomore this mode, it is clear its scope. I am impressed about rotation invariance properties in such usage mode, it surpassed my expectations, (never used daisy before in image matching scope).

Lets further improve given your nice suggestions:

1. Agree with "virtual void compute(Rect roi, OutputArray descriptors, ) const = 0;" in DAISY::COMP_FULL mode. Will implement it right now ! Its a brilliant alternative idea for _blank_ keypoints which is ment only for whole image. Obviously  will do some "pragma parallel" for the rectangle collections, only if will not conflict/race with DAISY internals parallel statements (will check the improvements and see if there are races / impacts).
  Technically with ROI we might encounter problems in the code, i am afraid that:
  a) Will have to call daisy for each subimage, then finally merge all descriptor arrays for those regions into single one array. Need to see if array can be reused as buffer at all to avoid final merge and multiple buffers This is most the feasible solution for this. 
  b) Can teach daisy internals to compute only for ROI areas, not so easy really have to *master* original daisy code ! But lets see the code first and then talk, perhaps it is not so hard.
  For booth a. or b. descriptors.ptr<float>( y*descriptorSize() + x ) would return zeros if user is out of ROI. Or, as alternative,  plan to expose public get_descriptor(x,y) that is aware of ROI with hardcoded assert if out of region.

2. One little perhaps uncovered and demanded thing  would be to add a third mode called DAISY::ONLY_QUERY which behave like ONLY_KEYS, but no keys should be supplied at all (will assert this), only some sparse precomputations are done by daisy on whole image and cached in the object (for little speedup) but would allow on-demand computation for any given and queried (x,y) and even using an "r" rotation (x,y,r) factor too. Will expose a public class like get_descriptor(x,y,r). In contrast please notice that COMP_FULL does not allow an extra "r" (due to its speedy mode) , thus why this mode is demanded.

3. A more clear documentation with all above.

4. I plan add a _very_ nice example (promise) of dense-stereo problem (no MRF !) once this PR is merged.

 Will do so, be right back asap. I was using daisy before in dense-basline problems, i really can tell many things about original implementations and authors intentions. In meanwhile i (with polite) suggest any reviewer to read "usage" from daisy original tarball.

vpisarev:
@cbalint13,
thank you for the prompt reply. Unfortunately, github does not allow to answer with citation. So, I just have my answers split into several items that we discuss.

* ah, ok, I now see that there is special mode for this. Yet, I'd prefer to do it in less hacky way, but rather explicitly using the extra method.

* It's good to see that DAISY demonstrates good performance as "sparse" descriptors. BTW, I wondered how they treat the keypoint "scale" and apparently, it's just ignored. I think, this should be mentioned somewhere in the description.

* On parallelizing DAISY. I think, it's not high priority for now, but replacing "#omp pragma parallel for with" with cv::parallel_for_ would be useful at some point, maybe later.

* On the problems with ROI. I shall emphasize that the whole image is first passed (using setImage() method) and preprocessed and then descriptors are extracted only for the ROI - this compute() may be called multiple times after a single setImage(). I think, it should not be difficult to do - modify compute_descriptors:
       1. add "Rect roi" parameter, pass m_descriptors as the output parameter (e.g. "out_descriptors") instead of accessing class member. Call out_descriptors.create(roi.height*roi.width, <DAISY_SIZE>, CV_32F);
       2. instead of the loop over 0..h x 0..w do the loop over roi.y..roi.y+roi.height x roi.x..roi.x+roi.width.
       3. in out_descriptors use the row index "(y - roi.y) * roi.width + (roi. x - roi.x)" to store the computed descriptor. 

   So, on the problem a) - usually, there is no need to keep all the computed descriptors. If you do stereo for a rectified stereo pair, you just need a stripe, even a line at once, which you store to temporary buffer. As long as you've computed the disparity, the descriptors are not needed anymore.

   On the problem b) - the solution is above. 

   The mode 3), as I understood it, is a partial case of both the compute method that I suggested (when (x,y) are integers) - use Rect(x, y, 1, 1)) or the existing compute when there is just one keypoint. I believe, user code can be just adjusted to extract several descriptors at once, this is usually always good idea and helps to decrease the overhead. So, do not worry about ONLY_QUERY.

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,2057 @@
...
+
+ "A fast local descriptor for dense matching" by Engin Tola, Vincent
+ Lepetit, and Pascal Fua. Intl. Conf. on Computer Vision and Pattern
+ Recognition, Alaska, USA, June 2008
+
+ OpenCV port by: Cristian Balint <cristian dot balint at gmail dot com>
+ */
+
+#include "precomp.hpp"
+#include "opencv2/imgproc/imgproc_c.h"
```
vpisarev: could you please modify the code to use the modern C++ API and thus do not use imgproc_c.h?

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,2057 @@
...
+    float* m_image;
+
+    // image height
+    int m_h;
+
+    // image width
+    int m_w;
+
+    // stores the descriptors : its size is [ m_w * m_h * m_descriptor_size ].
+    float* m_dense_descriptors;
```
vpisarev: m_dense_descriptors should not be stored, otherwise the method that extracts descriptors for ROI can not be made "const"

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,2057 @@
...
+
+    // stores the layered gradients in successively smoothed form: layer[n] =
+    // m_gradient_layers * gaussian( sigma_n ); n>= 1; layer[0] is the layered_gradient
+    float* m_smoothed_gradient_layers;
```
vpisarev: use Mat or vector&ltMat&gt; or vector&lt;float&gt; do not use plain arrays and manual memory management

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,2057 @@
...
+    bool m_scale_invariant;
+
+    // if set to true, descriptors are rotation invariant
+    bool m_rotation_invariant;
+
+    // number of bins in the histograms while computing orientation
+    int m_orientation_resolution;
+
+    // hold the scales of the pixels
+    float* m_scale_map;
```
vpisarev: same

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,2057 @@
...
+
+    // holds the orientaitons of the pixels
+    int* m_orientation_map;
```
vpisarev: same

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,2057 @@
...
+
+    // Holds the oriented coordinates (y,x) of the grid points of the region.
+    double** m_oriented_grid_points;
```
vpisarev: same. Also, please change "double" to "float" where possible. I think, the case of DAISY everything could be made "float" without any accuracy problems. float's are several times faster than double's on ARM. On some platforms float's are available in hardware, but double's are not - then the difference becomes catastrophic.

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,2057 @@
...
+
+    // holds the gaussian sigmas for radius quantizations for an incremental
+    // application
+    double* m_cube_sigmas;
```
vpisarev: same

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,2057 @@
...
+    int get_rq() { return m_rad_q_no; }
+    float get_rad() { return m_rad; }
+
+    // sets the type of the normalization to apply out of {NRM_PARTIAL,
+    // NRM_FULL, NRM_SIFT}. Call before using get_descriptor() if you want to
+    // change the default normalization type.
+    void set_normalization( int nrm_type ) { m_nrm_type = nrm_type; }
+
+    // applies one of the normalizations (partial,full,sift) to the desciptors.
+    void normalize_descriptors( int nrm_type = DAISY::NRM_NONE );
```
vpisarev: such methods should include necessary input/output parameters, like "Mat& descriptors"

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,2057 @@
...
+      if( sum != 0 )
+      {
+         for( int x=0; x<fsz; x++ )
+            fltr[x] /= sum;
+      }
+    }
+
+    inline void conv_horizontal( float* image, int h, int w, float* kernel, int ksize )
+    {
+      CvMat cvI; cvInitMatHeader(&cvI, h, w, CV_32FC1, (float*)image);
```
vpisarev: please, use cv::filter2D (or better more efficient cv::sepFilter2D to convolve ((image * kernelX) * kernelY) in one call

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,2057 @@
...
+      CvMat cvI; cvInitMatHeader(&cvI, h, w, CV_64FC1, (double*)image);
+      CvMat cvK; cvInitMatHeader(&cvK, ksize, 1, CV_64FC1, (double*)kernel);
+      cvFilter2D( &cvI, &cvI, &cvK );
+    }
+
+    /*
+     * DAISY utilities
+     */
+
+    template<class T>
```
vpisarev: can it be replaced with cv::Rect?

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,2057 @@
...
+
+      for( int i=0; i<sz; i++ )
+         out[i] = arr[i]*arr[i];
+
+      return out;
+    }
+
+    // computes the l2norm of an array: [ sum_i( [a(i)]^2 ) ]^0.5
+    template<class T> inline
+    float l2norm( T* a, int sz)
```
vpisarev: opencv has such inline functions.

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,2057 @@
...
+    {
+      float d0 = (float) (x0 - x1);
+      float d1 = (float) (y0 - y1);
+
+      return sqrt( d0*d0 + d1*d1 );
+    }
+
+    // allocates a memory of size sz and returns a pointer to the array
+    template<class T> inline
+    T* allocate(const int sz)
```
vpisarev: manual memory allocation should be replaced with std::vector or cv::Mat

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,2057 @@
...
+    template<class T1, class T2> inline
+    void polar2cartesian(T1 r, T1 t, T2 &y, T2 &x)
+    {
+      x = (T2)( r * cos( t ) );
+      y = (T2)( r * sin( t ) );
+    }
+
+
+    template<typename T> inline
+    void convolve_sym_( T* image, int h, int w, T* kernel, int ksize )
```
vpisarev: it can be done with cv::sepFilter2D

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,2057 @@
...
+      float inv_num = (float) (1.0 / num);
+
+      for( int i=0; i<sz; i++ )
+      {
+         arr[i] = (T1)(arr[i]*inv_num);
+      }
+    }
+
+    // returns an array filled with zeroes.
+    template<class T> inline
```
vpisarev: use Mat::zeros() for such things, or std::vector::asign().

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,2057 @@
...
+      return out;
+    }
+
+    // Applies a 2d gaussian blur of sigma std to the input array.  if
+    // kernel_size is not set or it is set to 0, then it is taken as
+    // 3*sigma and if it is set to an even number, it is incremented
+    // to be an odd number.  if in_place=true, then T1 must be equal
+    // to T2 naturally.
+    template<class T1, class T2> inline
+    T1* blur_gaussian_2d( T2* array, int rn, int cn, float sigma, int kernel_size = 0, bool in_place = false )
```
vpisarev: let's replace this with cv::GaussianBlur

cbalint13:
@vpisarev ,

 Agreed with your initial request and dropped the DAISY::COMP_FULL / DAISY::ONLY_KEYS (quite bad to maintain such, and needed a lot of stupid assert to warn users thei are in wrong sittuations) in the favour of elegant overloaded compute()'s. Only one purpose compute() functions which are much better, can be re-entrant and re-useable in place without calling whole constructor/destructor once again (as expected). Exposed a set_image() option for those who want to fast re-iterate and short-circuit compute() call things for a new image and did lot of API scenario tests. Now we have:

1. virtual void compute( InputArray image, std::vector<KeyPoint>& keypoints, OutputArray descriptors ) = 0; // The classic one, no need to comment.

2. virtual void compute( InputArray image, Rect roi, OutputArray descriptors ) = 0; // I agreed with you to have a ROI based one, i think is still a clever idea.

3. virtual void compute( InputArray image, OutputArray descriptors ) = 0; // An this is the full mode one.

*  Also exposed "get_descriptor() const" variants for those who want  further "very fast" query and sample points, and want short-circuit a bit the procedure to fill a keypoints list and call compute() again. These get_descriptor are not for the 1,2,3 mode, they are for the single-key mode (as per original daisy does).

* Regarding const-ness for compute() i think is not feasible , daisy already use internally parallel loops so will not gain any additional speed calling multiple images at once. Plus there is a delicate internal workspace memory management that should be in care. If still insist i can do the const thing, but believe me we will gain nothing.

* Fixed memory handle schema in case user re-entrer into compute() with new image / keypoint, or use set_image() scenarios.

-------------------------
Further more thank you a lot for the deep code review and suggestions ! My intention was in that direction too, but first i had to settle the API down , do more test with 1:1 untouched original code (its done at this moment) and only after then try to optimize and change the code for more opencv style. I will do _ALL_ of your suggested changes here and test&assure that descriptors generated are still the same and behave the same.

modules/xfeatures2d/src/daisy.cpp
cbalint13: DONE.

modules/xfeatures2d/src/daisy.cpp
cbalint13: * m_dense_descriptor is now Mat type. I did extractors as const. [DONE]

modules/xfeatures2d/src/daisy.cpp
cbalint13: No more alloc() / dealloc() or new [] ! Only Mat types, and where was necessary  accessed data via mat.ptr() schema. [DONE]

modules/xfeatures2d/src/daisy.cpp
cbalint13: m_smoothed_gradient_layers is now Mat. It's the most important matrix of floats in daisy. Unfortunate is that is N dimensions (i think N > 5) so in many place i was able to use it as .ptr() reference without undersanding its internal store system. Speed was not affected by conversion at all. [DONE]

cbalint13: m_scale_map too. its now 2d Mat with clear undrstandings whats inside. [DONE]

modules/xfeatures2d/src/daisy.cpp
cbalint13: m_smoothed_gradient_layers, m_scale_map ,m_orientation_map are Mat now (2d, 1d dimensions). I was able to use nice .row() accessors in some cases. In fact we now have only Mat matrices. [DONE]

modules/xfeatures2d/src/daisy.cpp
cbalint13: m_oriented_grid_points is now 2d Mat of CV_64F. I believe it is important to stay CV_64F, however it is small one. This double pointer array was ugly, i agree. [DONE]

modules/xfeatures2d/src/daisy.cpp
cbalint13: m_cube_sigmas too, it is a CV_64F (small one) and is important for precision sampling. [DONE]

modules/xfeatures2d/src/daisy.cpp
cbalint13: These are private internals, will never expose such things for public. Did you really want to pass through them Mat &descriptors ? [PENDING]

modules/xfeatures2d/src/daisy.cpp
cbalint13: [DONE]
* I used now filter2D twice for H,V steps. Was not able to use the nice getGaussianKernel() because it differs in results at 10e-6 and daisy produced different results ! Also was not able to use sepFilter2D(), but filter2D worked perfect using BORDER_REPLIACATE and not default one. 
* Converted to opencv, and we still produce 1:1 at bit level same numerical results.

modules/xfeatures2d/src/daisy.cpp
cbalint13: [DONE], those templates are now 100% gone.

modules/xfeatures2d/src/daisy.cpp
cbalint13: [DONE] no more alloc() / dealloc() or new[]. We have only nice Mat.

modules/xfeatures2d/src/daisy.cpp
cbalint13: [DONE], commented already (upper) on the conv issue, that template too is now gone.

modules/xfeatures2d/src/daisy.cpp
cbalint13: [DONE] we have now Mat(r,c,CV_32F,Scalar(0)) or mat.setTo(Scalar(0)). I did not used std::vector in the code, Mat is much more flexible and can be casted /converted / re-dimed with much more ease.

modules/xfeatures2d/src/daisy.cpp
cbalint13: [DONE] but with 2 x filter2D to be 1:1 bitwise level with same results as original.

cbalint13:
@vpisarev ,

 Work is in progress. I did all your requests upper and commented (for myself to know i did not miss something), plus many more optimizations to the code. It got quite "slim" one after these changes. Later today i push the new code for review, now i am checking through:

1) Should generate 1:1 same code as original [done]
2) Should pass QA as original [done]
3) Should pass real SfM as keypoint descriptor [done]
4) Should pass a real rense-basline densification and same amount of points should be obtained [pending]


Be back,

cbalint13:
@vpisarev ,

All requests done. Will fix win64 warns too later. In the meanwhile you can review once again the code, here is the squashed one: https://github.com/cbalint13/opencv_contrib/commit/fe38c9ef469d079568cd287a1f9c1771fda50bc5

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,1754 @@
...
+   int data_size = data.rows * data.cols;
+   Mat layers( 1, layer_no*data_size, CV_32F, Scalar(0) );
+
+   float kernel[5];
+   gaussian_1d(kernel, 5, 0.5f, 0.0f);
+   Mat Kernel(1, 5, CV_32F, (float*) kernel);
+
+   Mat cvO;
+   // smooth the data matrix
+   filter2D( data, cvO, CV_32F, Kernel, Point( -1, -1 ), 0, BORDER_REPLICATE );
```
vpisarev: could we replace it with GaussianBlur() or sepFilter2D()?

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,1754 @@
...
+   float kernel[5];
+   gaussian_1d(kernel, 5, 0.5f, 0.0f);
+   Mat Kernel(1, 5, CV_32F, (float*) kernel);
+
+   filter2D( cvI, cvI, CV_32F, Kernel, Point( -1, -1 ), 0, BORDER_REPLICATE );
+   filter2D( cvI, cvI, CV_32F, Kernel.t(), Point( -1, -1 ), 0, BORDER_REPLICATE );
+
+   Mat dx(1, data_size, CV_32F);
+   Mat dy(1, data_size, CV_32F);
+   gradient( cvI, data.rows, data.cols, dy, dx );
```
vpisarev: could we replace it with 2 cv::Sobel() calls?

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,1754 @@
...
+
+   Mat cvO;
+   // smooth the data matrix
+   filter2D( data, cvO, CV_32F, Kernel, Point( -1, -1 ), 0, BORDER_REPLICATE );
+   filter2D( cvO, cvO, CV_32F, Kernel.t(), Point( -1, -1 ), 0, BORDER_REPLICATE );
+
+   Mat dx(1, data_size, CV_32F);
+   Mat dy(1, data_size, CV_32F);
+
+   gradient(cvO, data.rows, data.cols, dy, dx);
```
vpisarev: could we replace it with two cv::Sobel() calls?

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,1754 @@
...
+   Mat cvI = data.clone();
+   layers.setTo( Scalar(0) );
+   int data_size = data.rows * data.cols;
+
+   float kernel[5];
+   gaussian_1d(kernel, 5, 0.5f, 0.0f);
+   Mat Kernel(1, 5, CV_32F, (float*) kernel);
+
+   filter2D( cvI, cvI, CV_32F, Kernel, Point( -1, -1 ), 0, BORDER_REPLICATE );
+   filter2D( cvI, cvI, CV_32F, Kernel.t(), Point( -1, -1 ), 0, BORDER_REPLICATE );
```
vpisarev: GaussianBlur? or sepFilter2D?

modules/xfeatures2d/src/daisy.cpp
cbalint13: * As i mention it i was nto able to replace the two filter2D() with single sepFilter2D() without numeric impact in final descriptor. If doing it produced descriptors will not be numericaly 1:1 as original daisy.

 If numerical result slightly differ and this is acceptable here, then  yes i will replace.

modules/xfeatures2d/src/daisy.cpp
cbalint13: I'll try. Only if you confirm that would be acceptable to have slight different numeric results (but obviously same quality) as original daisy.

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,1754 @@
...
+           : m_rad(_radius), m_rad_q_no(_q_radius), m_th_q_no(_q_theta), m_hist_th_q_no(_q_hist),
+             m_nrm_type(_norm), m_disable_interpolation(_interpolation), m_use_orientation(_use_orientation)
+{
+
+    m_image = 0;
+
+    m_descriptor_size = 0;
+    m_grid_point_number = 0;
+
+    m_grid_points.release();
```
vpisarev: the .release() calls are not necessary - all the matrices are empty at the beginning.

modules/xfeatures2d/src/daisy.cpp
```diff
@@ -0,0 +1,1754 @@
...
+    // allocate array
+    _descriptors.create( (int) keypoints.size(), m_descriptor_size, CV_32F );
+
+    // prepare descriptors
+    Mat descriptors = _descriptors.getMat();
+    descriptors.setTo( Scalar(0) );
+
+    // iterate over keypoints
+    // and fill computed descriptors
+    if ( H.empty() )
```
vpisarev: looks like there is almost complete code duplication, whereas the only difference (if I understood correctly) is point_transform_via_homography() call. If you pass NULL pointer as "double* H" and check it before calling point_transform_via_homography() - that will be very efficient, because the condition will be predicted perfectly by CPU and you loose may be just a few CPU clocks per descriptor (and computing a single DAISY descriptor takes 100's of clocks)

vpisarev:
@cbalint13, thank you for working on it! I have a few more comments, regarding elimination of custom gaussian blur & sobel. Switching to standard ones will reduce the further optimization efforts and simplifies long-term support. Also, as I see that the computed descriptors are still kept in the class. It means that you can not extract descriptors from different ROIs in the same image from parallel threads and thus you can not use DAISY in parallel stereo algorithm without obtaining all the descriptors at once (which may take a lot of memory!). It can be addressed later, I just wonder whether you were going to make this change before integration or not.

cbalint13:
@vpisarev 

1. Will try do the proposed replacements. Until now my goal was 1:1 same numerical output as original. If will be acceptable (?!) we will have slightly different numerical output compared to the original DAISY but will try to have same QA , with some chance perhaps slightly better, i wont be surprised :)

2. Will remove m_dense_descriptor out form the class ! I think it is 100% possible to do it.

be back,

vpisarev:
1. DAISY uses floating-point arithmetics. Therefore, you cannot assume to get bit-exact results as in the original code. On different OSes, different compilers, different CPUs with different IEEE754 implementations you can get slightly different descriptors. Besides, as I see, there are many optimization opportunities in this code, so we'll likely diverge from the original code even more. Therefore, I do not see a reason not to do it now :)

2. ok, thank you! I think, this feature (const'ness of the method that computes descriptors for a part of the image) will be really appreciated!

vpisarev:
excellent! :+1:

cbalint13:
@vpisarev ,

Wanted to tell you:

- Converted all arrays into nice Mat arrays and handle them properly to not loose performance.
- Used GaussianBlur, Sobel for more opencv style handling, difference is only at ~10^-7 level.
- We are still in QA for booth opencv's QA and real SfM densification problem, i always checked upon after any small modifications to the code.
- Most important:

 * Could _not_ do "const" for compute(_with_roi), it does *not* make sense at all because DAISY works better in such way:
     1. m_smoothed_gradient_layers, m_image and many more are _pre_computed_ and cached in  the class.
     2. You can ask for full descriptor computation and then destroy your object but most of the time ~95% people will  re-invoke in iterative way exposed  GetDescriptor() const variants with various rotation or various x,y during their agressive but selective baseline scan until 3D model converge, while ptr<DAISY> still caches important pre-computations for each GetDescriptor() calls ! It is hard to tell how many queries will be done, it depends how 3d model converge. Exactly how original daisy was designed and ment for.
     3. If people wants multiple images that construct for themselves multiple ptr<DAISY> classes each with its own caches of point no. 1 and they just will scan using GetDescriptor() const variant within their multiple threads.
    4. Doing compute() const is lousy, will lost all our internal per class caches and all the time daisy will recompute that with precious amount of time, point 2 here is much smarter approach.

- It is trivial to show a sample/tutorial, and i will work one such to show how to use it _efficiently_.
- We are twice slower than original and very little chance to speed up due to the fact raw original OpenMP is more powerful then cv::parallel_for. I tried with untouched original pieces of daisy within cv::parallelel_for and never had the same speed, but the code is very elegant now indeed, and easy to understand the objects there.
- My intuition says that additional speed might be gained when will further use GetDescriptor() const within multiple threads on top of internal cv::xfeat::daisy parallelism but will point it out  when tutorial will be done.

vpisarev:
@cbalint13, thank you!
1) do not worry about speed for now, there are many places in the code where we can compensate for the speed. For example, parallel_for_ has the optional "nstripes" parameter, which you can play with and probably get all the performance back.
2) constness of compute can be provided, but with 2-stage approach that I tried to describe earlier. First, user calls non-constant setImage(image) method, where all the maps are computed for the whole image. And then "compute(Rect roi, OutputArray descritptors) const" method is called without the input image and simply uses the precomputed maps to extract descriptors for the specified ROI.

cbalint13:
@vpisarev,

Vadim,

   I started work on a nice tutorial to add for opencv_contrib, it will be a real post-SfM densification of ~10 images (fountain scene) producing a very dense .PLY model. Will still keep account of what you proposed, will not let those wishes to "just go". I plan to outline the various modes that DAISY can be used. I worked hard for conversion into the nice Mat structures (e.g. used much more optimal ND Mat where was apropiate, and re-casted into sub 2d,3d array views where was needed),  i needed to fully understood booth author papers and the original code and the data structures layout that where quite "entangled". Hope people will like it. I agree with the performance constraints, in fact i like it now much more in this form.

 Thank you **very** much for helping me on it !

vpisarev:
@cbalint13, from my side I would like to thank you for this excellent contribution! In fact, I did not expect that you would make this much refactoring of the code :) and really glad you did it!

edgarriba:
@cbalint13 are you working on this post-Sfm densification with DAISY? I have some lines of code for this following the authors papers.

cbalint13:
@edgarriba,

- Yes. I tried to replicate EMVS (post densification) from author's paper. I also have some lines of code :). Since code is messy would share in private with all the explanations. My goal is to publish it, but not in this state.

Sent you message by e-mail.

PR #232, opened 2015-05-17
Removed HAL namespace

alexanderkoumis:
HAL namespace around fastAtan2 prevents module from building


alexanderkoumis:
Closed fork, saw it wasn't building for build bots, is the hal version of fastAtan2 to be used here? Any idea why it wasn't building w/ the namespace on my end?

PR #234, opened 2015-05-19
wrapping remaining xfeature2d classes to scripting

berak:
this pr adds FREAK, StarDetector, BriefDescriptorExtractor and LUCID to the python wrappers.


vpisarev:
:+1:

PR #237, opened 2015-05-26
always use "opencv2/cvconfig.h" instead of "cvconfig.h"

vpisarev:
this is complementary PR to https://github.com/Itseez/opencv/pull/4059


vpisarev:
ok, Maksim is already working on it

PR #238, opened 2015-05-28
Interface and demo for disparity filtering, basic WLS implementation

sbokov:
Current results on a small dataset (10 stereopairs in total): https://www.dropbox.com/sh/vfwhrcvq37dlqmk/AACo-LFiodnDUtVBqzxzossNa?dl=0


modules/ximgproc/samples/disparity_filtering.cpp
```diff
@@ -0,0 +1,238 @@
+#include "opencv2/calib3d/calib3d.hpp"
```
mshabunin: It should be `opencv2/calib3d.hpp`.

Header files `opencv2/<module>/<module>.hpp` are deprecated.

PR #242, opened 2015-06-02
Bug fix for feature extraction

Aljaksandr:
According to CartToPolar() function documentation, result angles could be in range (0..360). To prevent index overflow this check is important.


bmagyar:
+1

PR #243, opened 2015-06-02
BugFix & Enhance DAISY.

cbalint13:
Went through QA tests and some important fixes are done for DAISY dense mode:

 1. Norm should be "root of sums of squares" not "sums of rooted squares" !
 2. Store and accumulate "sums of squares" into a double not in float, it may overflow !
 3. Get rid of unused spurious function and variable from the code.

* Now opencv DAISY produce better results than original one (due to fix no. 2) !

On castle-p11 image set:

original: 4026934 (3d points)
**opencv**: 4179047 (3d points)


vpisarev:
:+1:

PR #244, opened 2015-06-02
Stereo branch

MMp131316:
@snosov1
Pull request having selected the stereo correspondence files from calib3d module

### DIFF
--- a/modules/README.md
+++ b/modules/README.md
@@ -51,3 +51,5 @@ $ cmake -D OPENCV_EXTRA_MODULES_PATH=<opencv_contrib>/modules -D BUILD_opencv_re
 21. **opencv_xobjdetect**: Integral Channel Features Detector Framework.
 
 22. **opencv_xphoto**: Additional photo processing algorithms: Color balance / Denoising / Inpainting.
+
+23. **opencv_stereo_corresp**: Stereo Correspondence using different descriptors: Block Matching using Census / csCensus / MCT / MV / RT / BRIEF. 
--- a/modules/line_descriptor/src/binary_descriptor.cpp
+++ b/modules/line_descriptor/src/binary_descriptor.cpp
@@ -40,7 +40,6 @@
  //M*/
 
 #include "precomp.hpp"
-
 #define NUM_OF_BANDS 9
 #define Horizontal  255//if |dx|<|dy|;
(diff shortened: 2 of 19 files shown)


modules/README.md
```diff
@@ -1,53 +1,55 @@
-An overview of the contrib modules and a small explanation
```
vpisarev: can you made atomic modification of the file? now it looks like the whole file was modified. I think, this is because of changed line endings. good editors should preserve original line endings

modules/line_descriptor/src/binary_descriptor.cpp
```diff
@@ -1136,9 +1135,9 @@ int BinaryDescriptor::computeLBD( ScaleLines &keyLines, bool useDetectionData )
 
         for ( short wID = 0; wID < lengthOfLSP; wID++ )
         {
-          tempCor = (short) round( sCorX );
```
vpisarev: meaning of the code has changed. use cvRound instead of round.

modules/line_descriptor/src/bitops.hpp
```diff
@@ -96,9 +96,10 @@ inline int match( UINT8*P, UINT8*Q, int codelb )
 }
 
 /* splitting function (b <= 64) */
+
 inline void split( UINT64 *chunks, UINT8 *code, int m, int mplus, int b )
 {
-  UINT64 temp = 0x0;
+ /* UINT64 temp = 0x0;
```
vpisarev: why did you comment off the code?

modules/stereo_corresp/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+set(the_description "Stereo Correspondance Algorithms")
```
vpisarev: since this is the first commit, let's rename the module from stereo_corresp to just "stereo".

modules/stereo_corresp/include/opencv2/calib3d/calib3d.hpp
```diff
@@ -0,0 +1,48 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
```
vpisarev: please, do not copy the headers. create new one "opencv2/stereo.hpp". there is no need in nested "opencv2/stereo/stereo.hpp".

PR #246, opened 2015-06-03
update comments of tracker algorithm

antran89:
* Adding comments of tracker algorithm for the sample program. 
* Beginners don't need to look at contrib_to_opencv/opencv_contrib/modules/tracking/src/tracker.cpp to know which tracker algorithm that opencv offers. 
* The name of tracker algorithm should be uppercase letters.


mshabunin:
:+1:

PR #249, opened 2015-06-05
Bug fix for seeds superpixel

yyisgod:
When useing VS2010 compile seeds.cpp, here will prompt an error because
the overloading problem.Take float to tell the Compiler.


mshabunin:
Fixed in #269

PR #250, opened 2015-06-05
Update README.md to fix type in cmake command

ciscoriordan:


alalek:
We use cmake commands without additional spacing succesfully, so I believe it is not required. What cmake version are you use?

mshabunin:
Actually I had problems when running cmake __with__ space-separated `-D` parameters. As long as there are no evidences of problems in current variant, I suggest to leave it as-is.

parismita:
i did 
cmake \<path\>  #works
cmake \<path\> \<option\>=\<value\> #didnt work
cmake \<option>=\<value\> \<path> #work 
cmake -D \<option\>=\<value\> <path> #work 

i think the problem is not with space and -D, its with providing path befour providing the option , 

-D<option> do not give error as while giving so, the <option> becomes invalid and is not considered as command hencewritting
cmake \<path\> -D\<option\>=\<value\> is same as writting cmake \<path\>

PR #253, opened 2015-06-09
Fix for Bug 4074. This seems to be just a typo-error.

lluisgomez:
Tesseract API can handle correctly with RGB images (double-checked and it works).


vpisarev:
:+1:

PR #262, opened 2015-06-12
Sphereview data generator towards 3D objects

Wangyida:
Sphereview data generator towards 3D objects, the top of the camera is always point towards the z axis.
Such image generator will be used for Triplet training of CNN. the demo use the default sphere iteration depth of '0' and the depth could be changed in the source file.
This is not a function yet, no parameter could be directly type in terminal.


bhack:
As you can see the build failed. Check buildbot log.
Also Please don't create a module for the 3d generator. Use directly a module name that represent your project and put this generator inside it as utility.

Wangyida:
I have updated the model name and rearrange the files suit for cmake

bhack:
Please don't close and open new pull request when you can fix things commitig on the same branch.

Wangyida:
OK, I got it, as you have said that 'Use directly a module name that represent your project and put this generator inside it as utility'.  I have just create a namespace cv::sphereview3d in header files and source files, do you mean that?

PR #263, opened 2015-06-13
CNN for 3D Objects Recognition and Pose Estimation

Wangyida:
Part 1 of CNN for 3D Objects Recognition and Pose Estimation: Sphereview data generator
* Get view towards 3D objects, the top of the camera is always point towards the z axis.
* It will be used for Triplet training of CNN. the demo use the default sphere iteration depth of '0' and the depth could be changed in the source file.
* This is not a function yet, no parameter could be directly type in terminal.


bhack:
Seems that you haven't change the module name. Choiche a module name that could rappresent your gsoc. This need to be an utility inside a module that will contain your gsoc code. Please see again the buildbot log. This PR it is failing.

bhack:
It is a best practice to generate PR from a dedicated branch. See also the other PR in contrib.

Wangyida:
OK, I got it, as you have said that 'Use directly a module name that represent your project and put this generator inside it as utility'. I have just create a namespace cv::sphereview3d in header files and source files, do you mean that? @bhack

bhack:
I don't think that your gsoc module name (and folder) could be shapeview3d. Shapeview3d it is only an utility inside an opencv modules named .....?

Wangyida:
So you mean for the module name should be my GSoC project and the module I just completed could just be represented as a class in my project's namespace, is that right? I used to think that different part in project should be placed in different places. Thanks you.

Wangyida:
I will make a new pull request from my dedicated branch and close this PR @bhack

PR #264, opened 2015-06-14
CNN for 3D Objects Recognition and Pose Estimation

Wangyida:
Part 1 of CNN for 3D Objects Recognition and Pose Estimation: Sphereview data generator
* Get view towards 3D objects, the top of the camera is always point towards the z axis.
* It will be used for Triplet training of CNN. the demo use the default sphere iteration depth of '0' and the depth could be changed in the source file.
* This is not a function yet, no parameter could be directly type in terminal.


Wangyida:
Big changes in minimal dependency building with Caffe. Close for convenient merging.

PR #265, opened 2015-06-16
Convolutional deep neural network module

ludv1x:
libprotobuf required to sucessfuly buid this module


alalek:
Hi!

Generally you shouldn't fail whole OpenCV build if some requirements are missing for module (at first it is needed for other users who uses opencv_contrib). In this case you should simply skip module build with optional message.

Use something like this:
https://github.com/Itseez/opencv/blob/48c9c24da6d8d4c432b674eae8808ee717c422c4/modules/cudev/CMakeLists.txt#L1-3

BTW, This patch is rather big, about 5.5Mb, but I believe it is not very critical. Most part is the `.caffemodel` file.

@vpisarev We can think about installation of the `libprotobuf` on a some builder to test compilation of this patch (I prefer Linux).

bhack:
@vpisarev We will have the same problem of the libprotobuf and probably glog gflag dependency on the other DNN project. Then we will put caffe dependency with minimal cmake flag activated in a submodule if you think that will be buildable on buildbot servers.

modules/matlab/CMakeLists.txt
```diff
@@ -92,6 +92,7 @@ ocv_add_module(matlab   BINDINGS
                                  opencv_calib opencv_calib3d
                                  opencv_stitching opencv_superres
                                  opencv_xfeatures2d
+                                 opencv_ximgproc
```
alalek: Why this change is here?
Similar question is about two ".pyc" files below.

alalek:
Hi,

`libprotobuf` have been installed on Linux slave.

BTW, about big `.caffemodel` file. I believe, that this size is not very critical (I usually saw files about hundreds of megabytes). Also it is very useful to have good completed example and test with data (if there are no licence issues).

modules/matlab/CMakeLists.txt
ludv1x: Sorry. These changes was randomly commited to master branch.

ludv1x:
@alalek I think that 300MB data file of AlexNet network in this repo it's not good idea (opencv_contrib.zip takes 56MB now).

Caffe original repo has special .py script int repo to download trained .caffemodel files.
Maybe add this script and download data using it and CMake add_custom_target()?

So, there is no need to add tests at present time, therefore I delete metioned data file of small network.

alalek:
Right, I mean `gtsrb_iter_36000.caffemodel` file.

Nerei:
Amalgamated protobuf is proposed to include inside the module. Not external protobuf dependency is planned.

ludv1x:
@bhack You can try [this](https://github.com/ludv1x/opencv_contrib/blob/master/modules/dnn/src/caffe/glog_emulator.hpp) simple header to avoid glog depedendecy.

bhack:
@ludv1x We are trying to modularize caffe to let it build with minimal dependencies. See https://github.com/BVLC/caffe/pull/2619

vpisarev:
@Nerei, is it ready for integration?

Nerei:
@vpisarev Any urgency here? The code still has external protobuf dependency. Unsure if cmake is forrect with respect to double pass done by opencv scripts.

modules/dnn/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+else()
+  message(STATUS "PROTOBUF not found. Caffe import function will be disabled.")
+  set(HAVE_PROTOBUF OFF)
+  set(PROTOBUF_LIBRARIES "")
+  add_definitions(-DHAVE_PROTOBUF=0)
+endif()
+
+endmacro(_dnn_find_protobuf)
+
+if(BUILD_opencv_core)#build as OpenCV module
```
alalek: I believe it is better to use here something like this:
```
${CMAKE_PROJECT_NAME} STREQUAL "OpenCV"
```

vpisarev:
@Nerei, ah, it's ok then, no urgency, I've just asked

cbalint13:
@ludv1x , @vpisarev ,

- Cannot see in header the license and authors, from where exactly are these imports ? (I suspect its BVLC's caffe).
- Will be opencv_contrib / itseez able to maintain such a big chunk of complex thing ?
- If its a fork from caffe, is there any ack or discuss with caffe maintainers ? Who's merit is this here ?

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -0,0 +1,101 @@
...
+    public:
+
+        virtual void populateNet(Net net) = 0;
+
+        virtual ~Importer();
+    };
+
+    CV_EXPORTS Ptr<Importer> createCaffeImporter(const String &prototxt, const String &caffeModel = String());
+
+    CV_EXPORTS Ptr<Importer> createTorchImporter(const String &filename, bool isBinary = true);
```
Nerei: No need CV_EXPORTS here

modules/dnn/include/opencv2/dnn/blob.hpp
```diff
@@ -0,0 +1,145 @@
...
+        int num() const;
+        Size size2() const;
+        Vec4i shape4() const;
+
+        //CPU data pointer functions
+        int offset(int n = 0, int cn = 0, int row = 0, int col = 0) const;
+        uchar *ptrRaw(int n = 0, int cn = 0, int row = 0, int col = 0);
+        float *ptrf(int n = 0, int cn = 0, int row = 0, int col = 0);
+        template<typename TFloat>
+        TFloat *ptr(int n = 0, int cn = 0, int row = 0, int col = 0);
```
Nerei: Would you merge ptr() methods similar to Mat::ptr<>

Nerei:
Header for sampe too

vpisarev:
@ludv1x, could you please also fix the error reported by the doc builder?
http://pullrequest.opencv.org/#/summary/contrib

Nerei:
:+1: 

Buildbot failures are not related this this module

alalek:
@ludv1x Please recreate PR with "dnn" branch name (same as the extra PR branch name). Also please squash/replace commits into one (102 commits are too much and there are many useless/nonworking intermediate changes)

alalek:
If you have some problems (there are many "merge" commits from master, squashing will not go well) I prepared a sample squashed commit: https://github.com/Itseez/opencv_contrib/commit/c94fdefcfff9cc5b8300c49052f0c1eee8896f0b

You can get it via:
```
git fetch https://github.com/Itseez/opencv_contrib c94fdefcfff9cc5b8300c49052f0c1eee8896f0b:dnn
git commit --amend # edit commit message
git push -f ...
```

Nerei:
@ludv1x The branch should match with opencv_extra brach so that buildbot  could test them jointly.

ludv1x:
This PR moves to https://github.com/Itseez/opencv_contrib/pull/362

PR #266, opened 2015-06-17
[GSoC 2015] Omnidirectional camera calibration

jiuerbujie:
This is Baisheng Lai's pull request for GSoC 2015. This pull request mainly contains three parts:
- omnidirectional camera calibration toolbox
- random pattern calibration object
- multiple camera calibration toolbox

The omnidirectional camera calibration toolbox uses CMei's model that could model many kinds of omnidirectional cameras, including fisheye and catadioptric ones. Beyond basic calibration toolbox content, several rectification types and stereo reconstruction are supported.

A new calibration object, random pattern is developed. The advantage of this object is that camera can observe only a part of this object.

A multiple camera calibration toolbox based on random pattern calibration object is also developed. It first calibrate each camera individually and use a global optimization to refine extrinsic parameters.  Pinhole and omnidirectional cameras are supported.


vpisarev:
:+1:

PR #269, opened 2015-06-17
remove floor call

berak:
vs2010 does not know, ceil, floor, round and friends.

those are plain integer divisions, that do not need floor at all.

also see https://github.com/Itseez/opencv_contrib/pull/249


vpisarev:
:+1:

PR #271, opened 2015-06-18
This files contains the implemented descriptors

MMp131316:
The #pragma have been commented they will have to be replaced by open cv parallel for
Other files will follow containing the matching sub pixel etc

### DIFF
--- a/modules/stereo/src/descriptor.cpp
+++ b/modules/stereo/src/descriptor.cpp
@@ -110,7 +110,7 @@ namespace cv
             parallel_for_(Range(n2, img1.rows - n2), StarKernelCensus<2>(images, n2,date));
         }
         //single version of star census
-        void stereo::starCensusTransform(const Mat &img1, int kernelSize, Mat &dist)
+        void starCensusTransform(const Mat &img1, int kernelSize, Mat &dist)
         {
             CV_Assert(img1.size() == dist.size());
             CV_Assert(kernelSize % 2 != 0);
@@ -119,7 +119,7 @@ namespace cv
             int n2 = (kernelSize) >> 1;
             Mat images[] = {img1};
             int *date[] = { (int *)dist.data};
-            parallel_for_( Range(n2, img1.rows - n2), StarKernelCensus<1>(images, n2,date));
+            parallel_for_(Range(n2, img1.rows - n2), StarKernelCensus<1>(images, n2,date));
         }
         //Modified census transforms
         //the first one deals with small illumination changes
@@ -167,14 +167,14 @@ namespace cv
             if(type == CV_MODIFIED_CENSUS_TRANSFORM)
             {
                 //MCT
-                parallel_for_( Range(n2, img1.rows - n2),
+                parallel_for_(Range(n2, img1.rows - n2),
                     CombinedDescriptor<2,3,2, 1,MCTKernel<1> >(img1.cols, img1.rows,stride,n2,date,MCTKernel<1>(images,t),n2));
             }
             else if(type == CV_MEAN_VARIATION)
             {
                 //MV
                 int *integral[] = { (int *)IntegralImage.data};
-                parallel_for_(  Range(n2, img1.rows - n2),
+                parallel_for_(Range(n2, img1.rows - n2),
                     CombinedDescriptor<2,3,2,1, MVKernel<1> >(img1.cols, img1.rows,stride,n2,date,MVKernel<1>(images,integral),n2));
             }
         }
@@ -194,11 +194,11 @@ namespace cv
             int stride = (int)img1.step;
             if(type == CV_CS_CENSUS)
             {
-                parallel_for_(  Range(n2, img1.rows - n2), SymetricCensus<2>(imag, n2,date));
+                parallel_for_(Range(n2, img1.rows - n2), SymetricCensus<2>(imag, n2,date));
             }
             else if(type == CV_MODIFIED_CS_CENSUS)
             {
-                parallel_for_(  Range(n2, img1.rows - n2),
+                parallel_for_(Range(n2, img1.rows - n2),
                     CombinedDescriptor<1,1,1,2,ModifiedCsCensus<2> >(img1.cols, img1.rows,stride,n2,date,ModifiedCsCensus<2>(images,n2),1));
             }
         }
@@ -235,7 +235,7 @@ namespace cv
             int height = image.rows;
             cost.setTo(0);
             int *c = (int *)cost.data;
-            parallel_for_( Range(win + 1, height - win - 1),MeanKernelIntegralImage(image,win,scalling,c));
+            parallel_for_(Range(win + 1, height - win - 1),MeanKernelIntegralImage(image,win,scalling,c));
         }
     }
 }
--- a/modules/stereo/src/descriptor.cpp
+++ b/modules/stereo/src/descriptor.cpp
@@ -1,0 +1,418 @@
+//By downloading, copying, installing or using the software you agree to this license.
+//If you do not agree to this license, do not download, install,
+//copy or use the software.
+//
+//
+//                          License Agreement
+//               For Open Source Computer Vision Library
+//                       (3-clause BSD License)
+//
+//Copyright (C) 2000-2015, Intel Corporation, all rights reserved.
+//Copyright (C) 2009-2011, Willow Garage Inc., all rights reserved.
+//Copyright (C) 2009-2015, NVIDIA Corporation, all rights reserved.
+//Copyright (C) 2010-2013, Advanced Micro Devices, Inc., all rights reserved.
+//Copyright (C) 2015, OpenCV Foundation, all rights reserved.
+//Copyright (C) 2015, Itseez Inc., all rights reserved.
+//Third party copyrights are property of their respective owners.
+//
+//Redistribution and use in source and binary forms, with or without modification,
+//are permitted provided that the following conditions are met:
+//
+//  * Redistributions of source code must retain the above copyright notice,
+//    this list of conditions and the following disclaimer.
+//
+//  * Redistributions in binary form must reproduce the above copyright notice,
+//    this list of conditions and the following disclaimer in the documentation
+//    and/or other materials provided with the distribution.
+//
+//  * Neither the names of the copyright holders nor the names of the contributors
+//    may be used to endorse or promote products derived from this software
+//    without specific prior written permission.
+//
+//This software is provided by the copyright holders and contributors "as is" and
+//any express or implied warranties, including, but not limited to, the implied
+//warranties of merchantability and fitness for a particular purpose are disclaimed.
+//In no event shall copyright holders or contributors be liable for any direct,
+//indirect, incidental, special, exemplary, or consequential damages
+//(including, but not limited to, procurement of substitute goods or services;
+//loss of use, data, or profits; or business interruption) however caused
+//and on any theory of liability, whether in contract, strict liability,
+//or tort (including negligence or otherwise) arising in any way out of
+//the use of this software, even if advised of the possibility of such damage.
+
+/*****************************************************************************************************************\
+*   The interface contains the main descriptors that will be implemented in the descriptor class			      *
+*																												  *
+\******************************************************************************************************************/
+
+
+#include "descriptor.hpp"
+
+using namespace cv;
+using namespace stereo;
+
+Descriptor::Descriptor()
+{
+
+}
+//!Implementation for computing the Census transform on the given image
+void Descriptor::applyCensusOnImage(const cv::Mat img, int kernelSize, cv::Mat &dist, const int type)
+{
+	int n2 = (kernelSize - 1) / 2;
+	int height = img.rows;
+	int width = img.cols;
+	uint8_t * image = img.data;
+	int *dst = (int *)dist.data;
+//#pragma omp parallel for
+	for (int i = n2; i <= height - n2; i++)
+	{
+		int rWidth = i * width;
+		for (int j = n2; j <= width - n2; j++)
+		{
+			if(type == CV_SSE_CENSUS)
+			{
+				//to do
+			}
+			else
+			{
+				int c = 0;
+				for (int ii = i - n2; ii <= i + n2; ii++)
+				{
+					int rrWidth = ii * width;
+					for (int jj = j - n2; jj <= j + n2; jj++)
+					{
+						if (ii != i || jj != j)
+						{
+							if (image[(rrWidth + jj)] > image[(rWidth + j)])
+							{
+								c = c + 1;
+							}
+							c = c * 2;
+						}
+					}
+				}
+				dst[(rWidth + j)] = c;
+			}
+		}
+	}
+}
+/**
+Two variations of census applied on input images
+Implementation of a census transform which is taking into account just the some pixels from the census kernel thus allowing for larger block sizes
+**/
+void Descriptor::applyCensusOnImages(const cv::Mat im1, cv::Mat im2, int kernelSize,cv::Mat &dist, cv::Mat &dist2, const int type)
+{
+	int n2 = (kernelSize - 1) / 2;
+	int width = im1.cols;
+	int height = im2.rows;
+	uint8_t * image1 = im1.data;
+	uint8_t * image2 = im2.data;
+	int *dst1 = (int *)dist.data;
+	int *dst2 = (int *)dist2.data;
+//#pragma omp parallel for
+	for (int i =  n2; i <= height - n2; i++)
+	{
+		int rWidth = i * width;
+		int distV = (i) * width;
+		for (int j = n2; j <= width - n2; j++)
+		{
+			int c = 0;
+			int c2 = 0;
+			if(type == CV_DENSE_CENSUS)
+			{
+				for (int ii = i - n2; ii <= i + n2; ii++)
+				{
+					int rrWidth = ii * width;
+					for (int jj = j - n2; jj <= j + n2; jj++)
+					{
+						if (ii != i || jj != j)
+						{
+							if (image1[rrWidth + jj] > image1[rWidth + j])
+							{
+								c = c + 1;
+							}
+							c = c * 2;
+						}
+						if (ii != i || jj != j)
+						{
+							if (image2[rrWidth + jj] > image2[rWidth + j])
+							{
+								c2 = c2 + 1;
+							}
+							c2 = c2 * 2;
+						}
+					}
+				}
+			}
+			else if(type == CV_SPARSE_CENSUS)
+			{
+				for (int ii = i - n2; ii <= i + n2; ii += 2)
+				{
+					int rrWidth = ii * width;
+					for (int jj = j - n2; jj <= j + n2; jj += 2)
+					{
+						if (ii != i || jj != j)
+						{
+							if (image1[(rrWidth + jj)] > image1[(rWidth + j)])
+							{
+								c = c + 1;
+							}
+							c = c * 2;
+						}
+						if (ii != i || jj != j)
+						{
+							if (image2[(rrWidth + jj)] > image2[(rWidth + j)])
+							{
+								c2 = c2 + 1;
+							}
+							c2 = c2 * 2;
+						}
+					}
+				}
+			}
+			dst1[(distV + j)] = c;
+			dst2[(distV + j)] = c2;
+		}
+	}
+}
+/** 
+STANDARD_MCT - Modified census which is memorizing for each pixel 2 bits and includes a tolerance to the pixel comparison
+MCT_MEAN_VARIATION - Implementation of a modified census transform which is also taking into account the variation to the mean of the window not just the center pixel
+**/
+void Descriptor::applyMCTOnImages(const cv::Mat img1, const cv::Mat img2, int kernelSize, int t, cv::Mat &dist, cv::Mat &dist2, const int type)
+{
+	int n2 = (kernelSize - 1) >> 1;
+	int width = img1.cols;
+	int height = img1.rows;
+	uint8_t *image1 = img1.data;
+	uint8_t *image2 = img2.data;
+	int *dst1 = (int *)dist.data;
+	int *dst2 = (int *)dist2.data;
+
+//	#pragma omp parallel for
+	for (int i = n2 + 2; i <= height - n2 - 2; i++)
+	{
+		int rWidth = i * width;
+		int distV = (i) * width;
+		for (int j = n2 + 2; j <= width - n2 - 2; j++)
+		{
+			int c = 0;
+			int c2 = 0;
+			if(type == CV_STANDARD_MCT)
+			{
+				for (int ii = i - n2; ii <= i + n2; ii += 2)
+				{
+					int rrWidth = ii * width;
+					for (int jj = j - n2; jj <= j + n2; jj += 2)
+					{
+						if (ii != i || jj != j)
+						{
+							if (image1[rrWidth + jj] > image1[rWidth + j] - t)
+							{
+								c <<= 2;
+								c |= 0x3;
+							}
+							else if (image1[rWidth + j] - t < image1[rrWidth + jj] && image1[rWidth + j] + t >= image1[rrWidth + jj])
+							{
+								c <<= 2;
+								c = c + 1;
+							}
+							else
+							{
+								c <<= 2;
+							}
+
+						}
+						if (ii != i || jj != j)
+						{
+
+							if (image2[rrWidth + jj]  > image2[rWidth + j] - t)
+							{
+								c2 <<= 2;
+								c2 |= 0x3;
+							}
+							else if (image2[rWidth + j] - t < image2[rrWidth + jj] && image2[rWidth + j] + t >= image2[rrWidth + jj])
+							{
+								c2 <<= 2;
+								c2 = c2 + 1;
+							}
+							else
+							{
+								c2 <<= 2;
+							}
+						}
+					}
+
+				}
+				for (int ii = i - n2; ii <= i + n2; ii += 4)
+				{
+					int rrWidth = ii * width;
+					for (int jj = j - n2; jj <= j + n2; jj += 4)
+					{
+						if (ii != i || jj != j)
+						{
+							if (image1[rrWidth + jj]  > image1[rWidth + j] - t)
+							{
+								c <<= 2;
+								c |= 0x3;
+							}
+							else if (image1[rWidth + j] - t < image1[rrWidth + jj] && image1[rWidth + j] + t >= image1[rrWidth + jj])
+							{
+								c <<= 2;
+								c += 1;
+							}
+							else
+							{
+								c <<= 2;
+							}
+
+						}
+						if (ii != i || jj != j)
+						{
+							if (image2[rrWidth + jj]  > image2[rWidth + j] - t)
+							{
+								c2 <<= 2;
+								c2 |= 0x3;
+							}
+							else if (image2[rWidth + j] - t < image2[rrWidth + jj] && image2[rWidth + j] + t >= image2[rrWidth + jj])
+							{
+								c2 <<= 2;
+								c2 = c2 + 1;
+							}
+							else
+							{
+								c2 <<= 2;
+							}
+						}
+					}
+				}
+
+			}
+			else if(type == CV_MCT_MEAN_VARIATION)
+			{
+				//to do mean variation
+			}
+			dst1[distV + j] = c;
+			dst2[distV + j] = c2;
+		}
+	}
+}
+/**The classical center symetric census
+A modified version of cs census which is comparing a pixel with its correspondent after the center
+**/
+void Descriptor::applySimetricCensus(const cv::Mat img1, const cv::Mat img2,int kernelSize, cv::Mat &dist, cv::Mat &dist2, const int type)
+{
+	int n2 = (kernelSize - 1) / 2;
+	int height = img1.rows;
+	int width = img1.cols;
+	uint8_t *image1 = img1.data;
+	uint8_t *image2 = img2.data;
+	int *dst1 = (int *)dist.data;
+	int *dst2 = (int *)dist2.data;
+	//#pragma omp parallel for
+	for (int i = + n2; i <= height - n2; i++)
+	{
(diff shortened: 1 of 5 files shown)


MMp131316:
@snosov1
A pull request containing the descriptors

modules/stereo/src/Descriptor.cpp
```diff
@@ -0,0 +1,448 @@
...
+//or tort (including negligence or otherwise) arising in any way out of
+//the use of this software, even if advised of the possibility of such damage.
+
+/*****************************************************************************************************************\
+*   The interface contains the main descriptors that will be implemented in the descriptor class			      *
+*																												  *
+\******************************************************************************************************************/
+
+
+#include "Descriptor.hpp"
```
snosov1: Please, use only lower-case letters in file names.

modules/stereo/src/Descriptor.cpp
```diff
@@ -0,0 +1,448 @@
+//By downloading, copying, installing or using the software you agree to this license.
```
snosov1: As a general note, please, use consistent "white-space style" in all of your files. That is,

- Use only spaces for indentation (not tabs). 
- Default indentation width should equal to 4 spaces.
- Don't leave trailing white space.

modules/stereo/src/Descriptor.hpp
```diff
@@ -0,0 +1,92 @@
...
+//the use of this software, even if advised of the possibility of such damage.
+
+/*****************************************************************************************************************\
+*   The interface contains the main descriptors that will be implemented in the descriptor class			      *
+*																												  *
+\******************************************************************************************************************/
+
+#include "precomp.hpp"
+#include <stdint.h>
+#define UNUSED( x ) ( &reinterpret_cast< const char& >( x ) )
```
snosov1: Please, don't declare macros, especially, in headers.  You can use something like `(void) image;` to suppress warnings.

modules/stereo/src/Descriptor.cpp
```diff
@@ -0,0 +1,448 @@
...
+	UNUSED( width );
+	UNUSED( dist );
+	
+}
+//Implementation for computing the Census transform on the given image
+void Descriptor::applyCensusOnImage(uint8_t *image, int height, int width, uint32_t * &dist)
+{
+	int n2 = (this->kernelSize - 1) / 2;
+
+//#pragma omp parallel for
```
snosov1: It's better to use OpenCV parallel for from the very beginning, so we don't forget about it later.

modules/stereo/src/Descriptor.cpp
```diff
@@ -0,0 +1,448 @@
...
+
+using namespace cv;
+using namespace stereo;
+
+Descriptor::Descriptor(int k)
+{
+	this->kernelSize = k;
+}
+//SSE function for computing the census transform on 1 image
+void Descriptor::applyCensusOnImage_SSE(uint8_t * image, int height, int width, uint32_t * &dist)
```
snosov1: Please, use cv::Mat's as parameters. For example, this function should have the following signature: `void Descriptor::applyCensusOnImage_SSE(const cv::Mat &image, cv::Mat &dist)`.

modules/stereo/src/Descriptor.cpp
```diff
@@ -0,0 +1,448 @@
...
+				}
+			}
+			dist[distV + j] = c;
+			dist2[distV + j] = c2;
+		}
+	}
+}
+
+//The classical center symetric census
+void Descriptor::applyCenterSimetricCensus(uint8_t * image1, uint8_t *image2, int height, int width, int padding_top, int padding_bottom, uint32_t * &dist, uint32_t * &dist2)
```
snosov1: Can you, please, consider using more "high-level" functions in implementation? For example, I'm thinking about having a single "census" function for all of the modifications (the needed modification will be specified with input parameter).

modules/stereo/src/Descriptor.cpp
```diff
@@ -0,0 +1,448 @@
...
+*																												  *
+\******************************************************************************************************************/
+
+
+#include "Descriptor.hpp"
+
+using namespace cv;
+using namespace stereo;
+
+Descriptor::Descriptor(int k)
```
snosov1: I believe, we don't need a Descriptor class with a single field. Kernel size could simply be a parameter to the function call.

snosov1:
Thx for the PR, Paul. I believe you did a great job choosing the amount of code to submit. 

I made several comments about the code itself. Most of them are "technicalities", like tabs vs spaces conventions. Let's polish it up, so, you can conform to these conventions from thereon.

modules/stereo/src/Descriptor.cpp
MMp131316: done

modules/stereo/src/Descriptor.cpp
MMp131316: done

modules/stereo/src/Descriptor.cpp
MMp131316: done

modules/stereo/src/Descriptor.cpp
MMp131316: done

modules/stereo/src/Descriptor.hpp
MMp131316: done

modules/stereo/src/Descriptor.cpp
MMp131316: done

modules/stereo/src/Descriptor.cpp
MMp131316: done

modules/stereo/src/descriptor.cpp
```diff
@@ -0,0 +1,112 @@
...
+using namespace cv;
+using namespace stereo;
+
+Descriptor::Descriptor()
+{
+}
+//!Implementation for computing the Census transform on the given image
+void Descriptor::applyCensusOnImage(const cv::Mat img, int kernelSize, cv::Mat &dist, const int type)
+{
+	int n2 = (kernelSize - 1) / 2;
```
snosov1: I'm still seeing that you use tabs for indentation. Please, use spaces instead. 1 tab -> 4 spaces

modules/stereo/src/descriptor.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include "descriptor.hpp"
+
+using namespace cv;
+using namespace stereo;
+
+Descriptor::Descriptor()
+{
+}
+//!Implementation for computing the Census transform on the given image
+void Descriptor::applyCensusOnImage(const cv::Mat img, int kernelSize, cv::Mat &dist, const int type)
```
snosov1: You should pass input cv::Mats as const references (const cv::Mat &), not const values (const cv::Mat).

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,435 @@
...
+											c2 = c2 + 1;
+										}
+										c2 = c2 * 2;
+									}
+								}
+							}
+						}//verifi if user wants a sparser census
+						else if (type == CV_SPARSE_CENSUS)
+						{
+							for (int ii = i - n2; ii <= i + n2; ii += 2)
```
snosov1: As far as I can see, the only difference between 'sparse' and 'dense' versions is that the former performs `ii += 2' while the latter performs `ii++`. If this is so, maybe it will be better to do something like `int step = type == CV_SPARSE_CENSUS ? 2 : 1` ?

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,435 @@
...
+							}
+						}
+						dst1[(rWidth + j)] = c;
+						dst2[(rWidth + j)] = c2;
+					}
+				}
+			}
+		};
+		//!class that implemented the census descriptor on single images
+		class singleImageCensus : public ParallelLoopBody
```
snosov1: I'm a bit confused here. As far as I can see, parallelCensus is the exact same thing as performing singleImageCesus on 2 images consequently. If this is so, why do you need parallelCensus at all?

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,435 @@
...
+*   The interface contains the main descriptors that will be implemented in the descriptor class			      *
+*																												  *
+\******************************************************************************************************************/
+
+#include "precomp.hpp"
+#include <stdint.h>
+#ifndef _OPENCV_DESCRIPTOR_HPP_
+#define _OPENCV_DESCRIPTOR_HPP_
+#ifdef __cplusplus
+#define CV_DENSE_CENSUS 0
```
snosov1: Using macros is a pretty bad practice. Please, use `enum`s instead.

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,435 @@
...
+								}
+							}
+						}
+						dst1[(distV + j)] = c;
+						dst2[(distV + j)] = c2;
+					}
+				}
+			}
+		};
+		class Descriptor
```
snosov1: I don't think we need another nested namespace for those functions. I suppose it's better to simply put them as plain functions in `cv::stereo`

modules/stereo/src/stereo_binary_sgbm.cpp
```diff
@@ -55,904 +55,902 @@ SGBM (Semi-global block matching)
...
-		It contains the minimum current cost, used to find the best disparity, corresponding to the minimal cost.
-		*/
-		static void computeDisparityBinarySGBM( const Mat& img1, const Mat& img2,
-			Mat& disp1, const StereoBinarySGBMParams& params,
-			Mat& buffer )
-		{
+        }
+
+
+        /*
```
snosov1: Something bad is happening with whitespaces here

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,435 @@
...
+						if (type == CV_STANDARD_MCT)
+						{
+							for (int ii = i - n2; ii <= i + n2; ii += 2)
+							{
+								int rrWidth = ii * width;
+								for (int jj = j - n2; jj <= j + n2; jj += 2)
+								{
+									if (ii != i || jj != j)
+									{
+										if (image1[rrWidth + jj] > image1[rWidth + j] - t)
```
snosov1: I wonder if we can reduce the code duplication even further. Like, do you notice a significant speed penality if you place a `type switch` here? I believe that if it's under 10%, we can go for it. At least for now - the code is not optimized yet anyways.

modules/stereo/src/descriptor.cpp
```diff
@@ -0,0 +1,111 @@
...
+using namespace cv;
+using namespace stereo;
+
+Descriptor::Descriptor()
+{
+}
+//!Implementation for computing the Census transform on the given image
+void Descriptor::applyCensusOnImage(const cv::Mat &img, int kernelSize, cv::Mat &dist, const int type)
+{
+    int n2 = (kernelSize - 1) / 2;
```
snosov1: One thing I forgot is that you need to `CV_Assert`s in the beginning of the functions to verify the user input. For example, `CV_Assert(img.size() == dist.size())`, `CV_Assert(img.type() == CV_8UC1`, etc.

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,390 @@
...
+        class parallelCensus :public ParallelLoopBody
+        {
+        private:
+            uint8_t *image1, *image2;
+            int *dst1, *dst2;
+            int n2, width, height, type;
+        public:
+            parallelCensus(uint8_t * img1, uint8_t * img2, int w, int h, int k2, int * distance1, int * distance2,const int t) :
+                image1(img1), image2(img2), dst1(distance1), dst2(distance2), n2(k2), width(w), height(h), type(t){}
+            virtual void operator()(const cv::Range &r) const {
```
snosov1: Why do you need virtual here?

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,390 @@
...
+                        int c2 = 0;
+                        //the classic center symetric census which compares the curent pixel with its symetric not its center.
+                        if (type == SymetricCensus::ClassicCenterSymetricCensus)
+                        {
+                            for (int ii = -n2; ii < 0; ii++)
+                            {
+                                int rrWidth = (ii + i) * width;
+                                for (int jj = -n2; jj <= +n2; jj++)
+                                {
+                                    if (ii != i || jj != j)
```
snosov1: The loop above is `for (int ii = -n2; ii < 0; ii++)`, i.e. `ii` is in the range `[-n2..0)`. Is everything is correct about this condition?

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,390 @@
...
+            int *dst1, *dst2;
+            int n2,t , width, height, type;
+        public:
+            parallelMctDescriptor(uint8_t * img1, uint8_t * img2, int w, int h, int k2,int threshold, int * distance1, int * distance2,const int tip) :
+                image1(img1), image2(img2), dst1(distance1), dst2(distance2), n2(k2), t(threshold), width(w), height(h), type(tip){}
+            virtual void operator()(const cv::Range &r) const {
+                for (int i = r.start; i <= r.end ; i++)
+                {
+                    int rWidth = i * width;
+                    int distV = (i)* width;
```
snosov1: It seems like `rWidth == distV` is always true. Is there an error?

MMp131316: no error, I forgot to erase it

modules/stereo/src/descriptor.hpp
MMp131316: it is correct!

modules/stereo/src/descriptor.hpp
MMp131316: No error, I forgot to delete it after refactoring the padding part
 


     On Tuesday, July 7, 2015 10:32 AM, Sergei Nosov <notifications@github.com> wrote:
   

 In modules/stereo/src/descriptor.hpp:> +        };
> +        //! parallel implementation of MCT type of descriptors
> +        class parallelMctDescriptor:public ParallelLoopBody
> +        {
> +        private:
> +            uint8_t *image1, *image2;
> +            int *dst1, *dst2;
> +            int n2,t , width, height, type;
> +        public:
> +            parallelMctDescriptor(uint8_t * img1, uint8_t * img2, int w, int h, int k2,int threshold, int * distance1, int * distance2,const int tip) :
> +                image1(img1), image2(img2), dst1(distance1), dst2(distance2), n2(k2), t(threshold), width(w), height(h), type(tip){}
> +            virtual void operator()(const cv::Range &r) const {
> +                for (int i = r.start; i <= r.end ; i++)
> +                {
> +                    int rWidth = i * width;
> +                    int distV = (i)* width;
It seems like rWidth == distV is always true. Is there an error?—

modules/stereo/src/descriptor.hpp
snosov1: Well, as far as I can see `ii` is always negative and `i` is always non-negative. If the condition is correct, then why do you need it all?

MMp131316: I need it in order to get the symetric

snosov1: I still don't get it. 

- `ii` is in the range `[-n2, 0)`, i.e. always negative. Agree?
- `i` is in the range `[n2, width - n2)`, i.e. always positive. Agree?
- `ii` is always negative, `i` is always positive, i.e. `ii != i` is always true. Agree?

MMp131316: yes, the condition should not be there

modules/stereo/src/descriptor.cpp
```diff
@@ -0,0 +1,138 @@
...
+        parallel_for_(cv::Range(n2, img.rows - n2),
+            singleDescriptor<1,1,1,CensusKernelSingle>(img.cols, img.rows,n2,(int *)dist.data,CensusKernelSingle(img.data),n2));
+    }
+}
+void cv::stereo::applyCensusOnImages(const cv::Mat &im1,const cv::Mat &im2, int kernelSize, cv::Mat &dist, cv::Mat &dist2, const int type)
+{
+    CV_Assert(im1.size() == im2.size());
+    CV_Assert(im1.type() == CV_8UC1 && im2.type() == CV_8UC1);
+    CV_Assert(type < 2 && type >= 0);
+    CV_Assert(kernelSize <= (type == 0 ? 5 : 10));
```
snosov1: I believe, kernelSize should be also checked for oddity, no?

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,441 @@
...
+#ifdef __cplusplus
+
+namespace cv
+{
+    namespace stereo
+    {
+        enum { Dense_Census, Sparse_Census, StarCensus};
+        enum {ClassicCenterSymetricCensus, ModifiedCenterSymetricCensus};
+        enum {StandardMct,MeanVariation};
+        enum {SSE, NonSSE};
```
snosov1: Conventionally, OpenCV doesn't allow switching between different optimization backends with run-time parameters. Instead, the optimizations are turned on by default if they are available using `#ifdef HAVE_SSE` machinery. I guess, it's safe to completely remove it for now, since we don't have any SSE in this PR anyway. When you decide to add the optimizations, please, see how other functions do this with "HAVE_SSE" macros.
Conventionally, OpenCV doesn't allow switching between different optimization backends with run-time parameters. Instead, the optimizations are turned on by default if they are available using `#ifdef HAVE_SSE` machinery. I guess, it's safe to completely remove it for now, since we don't have any SSE in this PR anyway. When you decide to add the optimizations, please, see how other functions do this with "HAVE_SSE" and the like macros.

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,441 @@
...
+                (void)w2;
+                (void)rWidth;
+                (void)jj;
+                (void)j;
+                (void)c;
+            }
+        };
+        //!kernel that takes the pixels from certain positions from a patch
+        //!offers verry good results
+        struct StarKernel
```
snosov1: Please, remove all the "yet to be implemented" functions. There's no point in adding them without the implementation.

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,441 @@
...
+        struct MCTKernel
+        {
+            uint8_t *image1;
+            uint8_t *image2;
+            int t;
+            MCTKernel(uint8_t * img,uint8_t *img2, int threshold) : image1(img),image2(img2), t(threshold) {}
+            void operator()(int rrWidth,int w2, int rWidth, int jj, int j, int &c, int &c2) const
+            {
+                (void)w2;
+                if (image1[rrWidth + jj] > image1[rWidth + j] - t)
```
snosov1: I think we can make the last step and remove the duplication inside such kernels as well. Something like `int c[2]; uint8_t *images[2];` and a for loop from 0 to 1 will do this.

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,441 @@
...
+            void operator()(const cv::Range &r) const {
+                for (int i = r.start; i <= r.end ; i++)
+                {
+                    int distV = (i)* width;
+                    for (int j = n2; j <= width - n2; j++)
+                    {
+                        int c = 0;
+                        int c2 = 0;
+                        //the classic center symetric census which compares the curent pixel with its symetric not its center.
+                        for (int ii = -n2; ii < 0; ii++)
```
snosov1: I don't really see why symmetricCensus can't be implemented in terms of "kernels". It doesn't seem very different from everything else. Like, we know the coordinates of the center pixel, we know the "current pixel", we can easily evaluate the "symmetric" pixel. Am I missing something?

modules/stereo/src/descriptor.cpp
```diff
@@ -0,0 +1,189 @@
...
+    }
+}
+//computation of the partial sums done on the intensity image
+void cv::stereo::computePartialSumsIntensityImage(const cv::Mat &image, cv::Mat &cost)
+{
+    int width = image.cols;
+    int height = image.rows;
+    uint8_t *img = image.data;
+    int *c = (int *)cost.data;
+    memset(c, 0, sizeof(c[0]) * (width + 1) * (height + 1));
```
snosov1: Consider using cost.setTo(0);

modules/stereo/src/descriptor.cpp
```diff
@@ -0,0 +1,189 @@
...
+{
+    CV_Assert(image.size > 0);
+    CV_Assert(cost.size > 0);
+    CV_Assert(windowSize % 2 != 0);
+    int win = windowSize / 2;
+    float scalling = ((float) 1) / (windowSize * windowSize);
+    int width = image.cols;
+    int height = image.rows;
+    int *c = (int *)cost.data;
+    memset(c, 0, sizeof(c[0]) * (width + 1) * (height + 1));
```
snosov1: memset -> cost.setTo(0);

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,443 @@
...
+                (void)w2;
+                for(int i = 0; i < 2; i++)
+                {
+                    if (image[i][rrWidth + jj] > image[i][rWidth + j])
+                    {
+                        c[i] = c[i] + 1;
+                    }
+                    c[i] = c[i] << 1;
+
+                    if (image[i][rrWidth + jj] > integralImage[i][rWidth + j])
```
snosov1: Shouldn't it be `integralImage[i][rrWidth + jj] > integralImage[i][rWidth + j]`?

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,451 @@
...
+#ifndef _OPENCV_DESCRIPTOR_HPP_
+#define _OPENCV_DESCRIPTOR_HPP_
+#ifdef __cplusplus
+
+namespace cv
+{
+    namespace stereo
+    {
+        //types of supported kernels
+        enum { Dense_Census, Sparse_Census, StarCensus};
```
snosov1: Please, use consistent naming conventions - don't mix CamelCase with underscores. Also, please, name those flags consistently with other OpenCV functions. For example, [calibrateCamera](http://docs.opencv.org/modules/calib3d/doc/camera_calibration_and_3d_reconstruction.html#calibratecamera) uses UPPERCASE identifiers with CV prefix (e.g. CV_CALIB_USE_INTRINSIC_GUESS)

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,451 @@
...
+                            }
+                            dst[d][rWidth + j] = c[d];
+                            dst[d][rWidth + j] = c[d];
+                        }
+                    }
+                }
+            }
+        };
+        //!paralel implementation of the center symetric census
+        class parallelSymetricCensus:public ParallelLoopBody
```
snosov1: `parallel` in the name is misleading. Please, make those descriptors work on 1 or 2 images with compile-time parameter.

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,451 @@
...
+                    }
+                }
+            }
+        };
+        /**
+        Two variations of census applied on input images
+        Implementation of a census transform which is taking into account just the some pixels from the census kernel thus allowing for larger block sizes
+        **/
+        //void applyCensusOnImages(const cv::Mat &im1,const cv::Mat &im2, int kernelSize, cv::Mat &dist, cv::Mat &dist2, const int type);
+        void applyCensusOnImages(const cv::Mat &image1, const cv::Mat &image2, int kernelSize, cv::Mat dist[2], const int type,const int numImages = DoubleImageDescriptor);
```
snosov1: The naming in public API should be terse and consistent. 

- don't use words, like `apply` or `compute` in the beginning
- don't use abbreviations, like MCT
- don't use the `onImages` suffix - the APIs are different, so it will work with overloading

The good names are `censusTransform`, `modifiedCensusTransform`, etc.  

Also, please, use references (&) everywhere you can (e.g. for output Mats). It doesn't make much difference, but is consistent with other functions.

modules/stereo/src/descriptor.cpp
```diff
@@ -0,0 +1,150 @@
...
+*   The file contains the implemented descriptors                                                                  *
+\******************************************************************************************************************/
+#include "descriptor.hpp"
+
+using namespace cv;
+using namespace stereo;
+
+//function that performs the census transform on two images.
+//Two variants of census are offered a sparse version whcih takes every second pixel as well as dense version
+void cv::stereo::censusTransform(const cv::Mat &image1, const cv::Mat &image2, int kernelSize, cv::Mat *&dist, const int type, const int numImages)
```
snosov1: There should be 2 output images, just like there are 2 input images. The array can be created inside the function.

modules/stereo/src/descriptor.cpp
```diff
@@ -0,0 +1,150 @@
...
+            CombinedDescriptor<1,1,1,CV_DUAL_DESCRIPTOR,CensusKernel>(image1.cols, image1.rows,n2,costs,CensusKernel(images, numImages),n2));
+    }
+    else if(type == CV_SPARSE_CENSUS)
+    {
+        parallel_for_(cv::Range(n2, image1.rows - n2),
+            CombinedDescriptor<2,2,1,CV_DUAL_DESCRIPTOR,CensusKernel>(image1.cols, image1.rows,n2,costs,CensusKernel(images, numImages),n2));
+    }
+}
+//in a 9x9 kernel only certain positions are choosen for comparison
+void cv::stereo::starCensusTransform(const cv::Mat &img1, const cv::Mat &img2, int kernelSize, cv::Mat *&dist,const int numImages)
```
snosov1: There seem to be only 2-image functions. Do you plan to add single-image functions later?

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,445 @@
...
+            int **dst;
+            Kernel kernel_;
+            int n2_stop;
+        public:
+            CombinedDescriptor(int w, int h, int k2, int **distance, Kernel kernel,int k2Stop)
+            {
+                width = w;
+                height = h;
+                n2 = k2;
+                dst = new int*[nr_img];
```
snosov1: Let's stick with a static-sized array. As far as I see, there's no need for heap allocations.

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,445 @@
...
+                        }
+                        for(int l = 0; l < nr_img; l++)
+                            dst[l][rWidth + j] = c[l];
+                    }
+                }
+            }
+        };
+        //!calculate the mean of every windowSizexWindwoSize block from the integral Image
+        //!this is a preprocessing for MV kernel
+        class MeanKernelIntegralImage : public ParallelLoopBody
```
snosov1: Is it some kind of a noise-filter, like Gauss or median?

modules/stereo/src/descriptor.hpp
```diff
@@ -0,0 +1,445 @@
...
+        //!not just with the center but also with the mean of the window
+        struct MVKernel
+        {
+            uint8_t **image;
+            int **integralImage;
+            int stop;
+            MVKernel(){}
+            MVKernel(uint8_t **images, int **integral, int numImag = CV_DUAL_DESCRIPTOR)
+            {
+                image = new uint8_t *[numImag];
```
snosov1: Please, remove all the news from the code. Use Mat::create where appropriate.

modules/stereo/src/descriptor.cpp
```diff
@@ -0,0 +1,235 @@
...
+//(including, but not limited to, procurement of substitute goods or services;
+//loss of use, data, or profits; or business interruption) however caused
+//and on any theory of liability, whether in contract, strict liability,
+//or tort (including negligence or otherwise) arising in any way out of
+//the use of this software, even if advised of the possibility of such damage.
+
+/*****************************************************************************************************************\
+*   The file contains the implemented descriptors                                                                  *
+\******************************************************************************************************************/
+#include "descriptor.hpp"
```
vpisarev: each source file should start with
<pre>
#include "precomp.hpp"
</pre>

modules/stereo/src/descriptor.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+using namespace cv;
```
vpisarev: please, do not use "using namespace cv;" etc.
instead, put all the code into
<pre>
namespace cv { namespace stereo {
...
}}
</pre>

modules/stereo/src/descriptor.cpp
```diff
@@ -0,0 +1,235 @@
...
+    CV_Assert(image1.type() == CV_8UC1);
+    CV_Assert(type != CV_DENSE_CENSUS || type != CV_SPARSE_CENSUS);
+    CV_Assert(kernelSize <= ((type == 0) ? 5 : 11));
+    int n2 = (kernelSize) / 2;
+    uint8_t *images[] = {image1.data};
+    int *costs[] = {(int *)dist1.data};
+    int stride = (int)image1.step;
+    if(type == CV_DENSE_CENSUS)
+    {
+        parallel_for_(cv::Range(n2, image1.rows - n2),
```
vpisarev: inside functions there is no need to use "cv::"

vpisarev:
@MMp131316, @snosov1, I left a few technical comments, but the main question is - what's the way to access all this functionality? I only see BinaryStereoSGBM, but I do not see any tracks on how to change the descriptor used, how to set the parameters etc. I do not see any sample code, I do not see any tests. You should add it in any case (or, if you did it already, could you point me to it?)

snosov1:
This PR contains only the descriptors, not the stereo matching algorithms, yet. The algorithms are implemented in the fork repo, but they are not part of this PR. We agreed to merge everything in 4 PRs overall:

1. Descriptors only (this one)
2. BM with binary descriptors 
3. SGBM with binary descriptors
4. Documentation and tests

Personally, I think it eases and improves the quality of the reviews. Are you ok to proceed like this, or would you prefer to have everything in one go?

PR #275, opened 2015-06-20
Point cloud normal computation

FelixMartel:
I had to use the surface_matching module with point clouds without precomputed normals and noticed that the computeNormalsPC3d function wasn't working. These are the changes I had to make so it would work.


bmagyar:
Hello Felix, 

Nice catch, and thanks for the contribution!
Could you please supply files to test this behaviour with? I think it could even be added to the sample folder along with some instructions to raise the attention to the 2 different cases to users.

Which platforms have you tested the code?

If I see well, eigenLowest33() has now lost legitimacy to exist. Feel free to give it the finishing blow and remove it altogether.

FelixMDenis:
Hello!
I have added a function I used to visualise the normals as red points in the cloud and a sample to compute normals for a model.
The computeNormalsPC3d() function can be tested by recomputing the normals of existing models and using these as input to the existing surface matching sample.

I have tested my code on windows only.

bmagyar:
Great! I will compile and test all on Ubuntu & OSX and get back here.

bmagyar:
All good!

:+1:

vpisarev:
@FelixMDenis, thanks for the contribution! Is the root cause in "eigenLowest33" (which was replaced with cv::eigen)? I mean, if eigenLowest33 itself works well, it should be much faster than eigen.

FelixMDenis:
Even though I could find neither Tolga Birdal's original version of this function nor an explanation of how it is supposed to work out the eigenvalues, at first glance you can notice that if sqrt(-1) refers to its standard c++ definition, NaN values would be returned as it was indeed the case in the tests I performed.

modules/surface_matching/src/c_utils.hpp
```diff
@@ -659,100 +659,6 @@ static inline void quatToDCM(double *q, double *R)
...
-  const double t57 = sqrt(3.0);
-  const double t51 = b*c;
-  const double t52 = d*e;
-  const double t53 = e*f;
-  const double t54 = t51+t52+t53;
-  const double t62 = t11+t20+t21+t22-t26-t27-t28;
-  const double t63 = t11+t20+t21+t22-t26-t27-t28;
-  const double t64 = t11+t20+t21+t22-t26-t27-t28;
-  const double t65 = t11+t20+t21+t22-t26-t27-t28;
-  const double t66 = t36+t37+t38-t18*1.0/pow(t20+t21+t22-t26-t27-t28+t39+sqrt(-t41+t62*t62),1.0/3.0)*(1.0/2.0)+t57*(t18*1.0/pow(t20+t21+t22-t26-t27-t28+t39+sqrt(-t41+t64*t64),1.0/3.0)-pow(t20+t21+t22-t26-t27-t28+t39+sqrt(-t41+t65*t65),1.0/3.0))*5.0E-1*sqrt(-1.0)-pow(t20+t21+t22-t26-t27-t28+t39+sqrt(-t41+t63*t63),1.0/3.0)*(1.0/2.0);
```
FelixMDenis: Here.

vpisarev:
@FelixMartel, thank you for the contribution and sorry for delay! @bmagyar, thank you for the review :+1:

PR #276, opened 2015-06-21
GSoC 2015: Caffe integated CNN for 3D Objects Recognition and Pose Estimation

Wangyida:
Work together with Caffe project with triplet training version
* Sphere view data generator for triplet training using CNN.
* Feature extraction pipeline using a constructed model and a net definition file.
* Prediction on Class and Pose of images of 3D objects.


modules/cnn_3dobj/samples/sphereview_3dobj_demo.cpp
```diff
@@ -0,0 +1,124 @@
...
+	float radius = parser.get<float>("radius");
+	//int ite_depth = argv[2][0] - '0';
+	int ite_depth = parser.get<int>("ite_depth");
+	//std::string plymodel = argv[3];
+	string plymodel = parser.get<string>("plymodel");
+	string imagedir = parser.get<string>("imagedir");
+	cv::cnn_3dobj::IcoSphere ViewSphere(10,ite_depth);
+	std::vector<cv::Point3d> campos = ViewSphere.CameraPos;
+	std::fstream imglabel;
+	imglabel.open("../data/label_ape.txt");
```
mtamburrano: maybe the pose.txt can be passed as input

modules/cnn_3dobj/samples/sphereview_3dobj_demo.cpp
```diff
@@ -0,0 +1,124 @@
...
+		return 0;
+	}
+	//float radius = atof(argv[1]);
+	float radius = parser.get<float>("radius");
+	//int ite_depth = argv[2][0] - '0';
+	int ite_depth = parser.get<int>("ite_depth");
+	//std::string plymodel = argv[3];
+	string plymodel = parser.get<string>("plymodel");
+	string imagedir = parser.get<string>("imagedir");
+	cv::cnn_3dobj::IcoSphere ViewSphere(10,ite_depth);
```
mtamburrano: After you create the IcoSphere object, you should check for pose duplicates before iterating over them to render the object. If you see the result images, you'll get a lot of duplicates.
You can easily avoid them using a Map or something like that

mtamburrano:
I see that with the Ape.ply model you obtain rendered images with the ape that is not always centered on the frame, leading to wrong poses.
This is probably due to the pivot point of the object placed on the bottom of the ape instead that at the center of it. Probably the pivot can be moved with meshlab, blender or something similar, so the example will run fine

modules/cnn_3dobj/samples/sphereview_3dobj_demo.cpp
Wangyida: I used to check them in the end of all iteration and some result became very similar though they are the same points because the float coordinate differs more when the iteration process is deeper, I will add a margin towards the coordinate comparison to check the duplicates.

bhack:
@pwohlhart What about background noise? What kind of synthetic background (RBD/DEPTH) do you have experimented in the training phase other that the one in the publication? Do you think that a Perlin Noise could be enough?

pwohlhart:
@bhack I added perlin noise to the background and some small amount of Gaussian noise on the object. 

You can have a look at the section in my code. 
https://cvarlab.icg.tugraz.at/projects/3d_object_detection/ObjRecPoseEst.tar.gz
It's in src/util/misc.py, function addNoiseToTrainSeqs

bhack:
@pwohlhart Yes I know. But in the unmodified dataset you've released I see that objects seems to be placed on a "virtual table" with some kind of "virtual wall" as I can see.

bhack:
@Nerei API it is not finalized buy if you want to make a pass here to let  improve a further switch to the other GSOC infrastructure from caffe it would be very useful.

modules/cnn_3dobj/include/opencv2/cnn_3dobj.hpp
```diff
@@ -0,0 +1,252 @@
...
+About the training and feature extraction process, it is a rough implemetation by using OpenCV
+and Caffe from the idea of Paul Wohlhart. The principal purpose of this API is construting
+a well labeled database from .ply models for CNN training with triplet loss and extracting features
+with the constructed model for prediction or other purpose of pattern recognition, algorithms into two main Class:
+
+**icoSphere**: methods belonging to this class generates a well generated database for CNN
+training together with their class and pose from camera view labels.
+
+**extractor**: methods belonging to this class extract features from 2D images which is
+discriminant on catogery prediction and pose estimation.
```
mtamburrano: replace:
As CNN based learning algorithm shows better performance on the classification issues,
the rich labeled data could be more useful in the training stage. 3D object classification and pose estimation
is a jointed mission aiming at separate different posea apart in the descriptor form.
In the training stage, we prepare 2D training images generated from our module with their
class label and pose label. We fully exploit the information that lies in their labels
by using a triplet and pair-wise jointed loss function in CNN training.
Both class and pose label are in consideration in the triplet loss. The loss score 
will be smaller when features from the same class and same pose is more similar
and features from different classes or different poses will lead to a much larger loss score.
This loss is also jointed with a pair wise component to make sure the loss is never be zero
and have a restriction on the model scale.
About the training and feature extraction process, it is a rough implementation by using OpenCV
and Caffe from the idea of Paul Wohlhart. The principal purpose of this API is constructing
a well labeled database from .ply models for CNN training with triplet loss and extracting features
with the constructed model for prediction or other purpose of pattern recognition, algorithms into two main Class:
**icoSphere**: methods belonging to this class generates 2D images from a 3D model, together with their class and pose from camera view labels.
**extractor**: methods belonging to this class extract descriptors from 2D images which is
discriminant on category prediction and pose estimation.

modules/cnn_3dobj/include/opencv2/cnn_3dobj.hpp
```diff
@@ -0,0 +1,252 @@
...
+    {
+        private:
+        /** @brief X position of one base point on the initial Icosohedron sphere,
+          Y is set to be 0 as default.
+         */
+        float X;
+        /** @brief Z position of one base point on the initial Icosohedron sphere.
+         */
+        float Z;
+        /** @brief A threshold for the duplex point elimination.
```
mtamburrano: replace:
A threshold for the dupicated points elimination.

modules/cnn_3dobj/include/opencv2/cnn_3dobj.hpp
```diff
@@ -0,0 +1,252 @@
...
+         */
+        float diff;
+        /** @brief Temp camera position for duplex position elimination.
+         */
+        std::vector<cv::Point3d> CameraPos_temp;
+        /** @brief Make all view points having the some distance from the focal point used by the camera view.
```
mtamburrano: replace:
Make all view points having the same distance from the focal point used by the camera view.

modules/cnn_3dobj/include/opencv2/cnn_3dobj.hpp
```diff
@@ -0,0 +1,252 @@
...
+         */
+        CV_WRAP void norm(float v[]);
+        /** @brief Add new view point between 2 point of the previous view point.
```
mtamburrano: replace:
Add a new view point

modules/cnn_3dobj/include/opencv2/cnn_3dobj.hpp
```diff
@@ -0,0 +1,252 @@
...
+         */
+        CV_WRAP void add(float v[]);
+        /** @brief Generating new view points from all triangles.
```
mtamburrano: replace:
Generate new view points from all triangles.

modules/cnn_3dobj/include/opencv2/cnn_3dobj.hpp
```diff
@@ -0,0 +1,252 @@
...
+         */
+        CV_WRAP void subdivide(float v1[], float v2[], float v3[], int depth);
+
+        public:
+        /** @brief Camera position on the sphere after duplex position elimination.
```
mtamburrano: replace:
Camera position on the sphere after duplicated points elimination.

modules/cnn_3dobj/include/opencv2/cnn_3dobj.hpp
```diff
@@ -0,0 +1,252 @@
...
+         */
+        std::vector<cv::Point3d> CameraPos;
+        /** @brief Generating a sphere by mean of a iteration based points selection process.
+        @param radius_in Another radius used for adjusting the view distance.
+        @param depth_in Number of interation depth for increasing the points on sphere.
```
mtamburrano: replace:
Number of interations for increasing the points on sphere.

modules/cnn_3dobj/include/opencv2/cnn_3dobj.hpp
```diff
@@ -0,0 +1,252 @@
...
+        @param x Pose label of X.
+        @param y Pose label of Y.
+        @param z Pose label of Z.
+        @param isrgb Option for choice of using RGB images or not.
+         */
+        CV_WRAP static void writeBinaryfile(string filenameImg, const char* binaryPath, const char* headerPath, int num_item, int label_class, int x, int y, int z, int isrgb);
+    };
+
+/** @brief Caffe based 3D images descriptor.
+ The class extract feature which is discriminant for class and pose from given samples @cite wohlhart15.
```
mtamburrano: Replace:
A class to extract features from an image. The so obtained descriptors can be used for classification and pose estimation goals @cite wohlhart15.

modules/cnn_3dobj/include/opencv2/cnn_3dobj.hpp
```diff
@@ -0,0 +1,252 @@
...
+        /** @brief Set the device for feature extraction, if the GPU is used, there should be a device_id.
+        @param device_type CPU or GPU.
+        @param device_id ID of GPU.
+         */
+        descriptorExtractor(const string& device_type, int device_id);
+        /** @brief Get device information for feature extraction.
+         */
+        std::vector<string> getDevice();
+        /** @brief Set device information for feature extraction, if the descriptor is already loaded and
+         you want to change the device for featrue extraction, this function is convenient.
```
mtamburrano: replace:
Set device information for feature extraction. Useful to change device without the need to reload the net

modules/cnn_3dobj/include/opencv2/cnn_3dobj.hpp
```diff
@@ -0,0 +1,252 @@
...
+         */
+        void setDevice(const string& device_type, const string& device_id = "");
+        /** @brief Initiate a classification structure, the net work parameter is stored in model_file,
+         the network structure is stored in trained_file, you can decide whether to use mean images or not.
+        @param model_file Path of caffemodel which including all parameters in CNN.
+        @param trained_file Path of prototxt which defining the structure of CNN.
+        @param mean_file Path of mean file(option).
+         */
+        void loadNet(const string& model_file, const string& trained_file, string mean_file = "");
+        /** @brief Extract features from a single image or a set of images which is collected in a vector.
```
mtamburrano: replace:
Extract features from a single image or from a vector of images. If loadNet was not called before, this method invocation will fail

modules/cnn_3dobj/samples/classifyIMG_demo.cpp
```diff
@@ -0,0 +1,144 @@
...
+    {
+        img_gallery.push_back(cv::imread(name_gallery[i], -1));
+    }
+    descriptor.extract(img_gallery, feature_reference, feature_blob);
+
+    std::cout << std::endl << "---------- Prediction for " << target_img << " ----------" << std::endl;
+
+    cv::Mat img = cv::imread(target_img, -1);
+    // CHECK(!img.empty()) << "Unable to decode image " << target_img;
+    std::cout << std::endl << "---------- Featrue of gallery images ----------" << std::endl;
```
mtamburrano: replace "Featrue" with "Features"

modules/cnn_3dobj/samples/classifyIMG_demo.cpp
```diff
@@ -0,0 +1,144 @@
...
+    std::vector<std::pair<string, float> > prediction;
+    for (unsigned int i = 0; i < feature_reference.rows; i++)
+        std::cout << feature_reference.row(i) << endl;
+    cv::Mat feature_test;
+    descriptor.extract(img, feature_test, feature_blob);
+    cv::BFMatcher matcher(NORM_L2);
+    std::vector<std::vector<cv::DMatch> > matches;
+    matcher.knnMatch(feature_test, feature_reference, matches, num_candidate);
+    std::cout << std::endl << "---------- Featrue of target image: " << target_img << "----------" << endl << feature_test << std::endl;
```
mtamburrano: replace "Featrue" with "Features"

modules/cnn_3dobj/src/cnn_feature.cpp
```diff
@@ -0,0 +1,244 @@
+#include "precomp.hpp"
+using namespace caffe;
+using std::string;
+
+namespace cv
+{
+namespace cnn_3dobj
+{
+    descriptorExtractor::descriptorExtractor(const string& device_type, int device_id)
```
mtamburrano: the device_id should be optional (bacause in the CPU case, we don't need it), with a default of 0.

modules/cnn_3dobj/src/cnn_feature.cpp
```diff
@@ -0,0 +1,244 @@
...
+            net_set = true;
+        }
+        else
+        {
+            std::cout << "Error: Device name must be 'GPU' together with an device number or 'CPU'." << std::endl;
+            net_set = false;
+        }
+    };
+
+    std::vector<string> descriptorExtractor::getDevice()
```
mtamburrano: split this method in 2: getDeviceType() and getDeviceId().
So you can return a single value instead than a vector of strings. It makes more sense due to the changes to setDevice (see below)

modules/cnn_3dobj/src/cnn_feature.cpp
```diff
@@ -0,0 +1,244 @@
...
+    {
+        std::vector<string> device_info_out;
+        device_info_out = device_info;
+        return device_info_out;
+    };
+
+    void descriptorExtractor::setDevice(const string& device_type, const string& device_id)
```
mtamburrano: split this method in 2: setDeviceType(const string& type) and getDeviceId(const int& id).
Doesn't make much sense that deviceId is a string here and an int in the constructor.

modules/cnn_3dobj/src/cnn_feature.cpp
```diff
@@ -0,0 +1,244 @@
...
+    {
+        if (strcmp(device_type.c_str(), "CPU") == 0 || strcmp(device_type.c_str(), "GPU") == 0)
+        {
+            if (strcmp(device_type.c_str(), "CPU") == 0)
+            {
+                caffe::Caffe::set_mode(caffe::Caffe::CPU);
+                device_info.push_back("CPU");
```
mtamburrano: Multiple invocations to setDevice will increase the size of device_info? And also dev_id should not be saved in private variables?
Anyway because you are going to split the set and get methods, you can remove this vector and simply have a private variable for device type and another one for device_id

modules/cnn_3dobj/src/cnn_feature.cpp
```diff
@@ -0,0 +1,244 @@
...
+        else
+        {
+            std::cout << "Error: Device name must be 'GPU' together with an device number or 'CPU'." << std::endl;
+            net_set = false;
+        }
+    };
+
+    void descriptorExtractor::loadNet(const string& model_file, const string& trained_file, string mean_file)
+    {
+        net_ready = 0;
```
mtamburrano: this variable should be initialized to zero into the constructor, not here

modules/cnn_3dobj/src/cnn_feature.cpp
```diff
@@ -0,0 +1,244 @@
...
+            net_set = true;
+        }
+        else
+        {
+            std::cout << "Error: Device name must be 'GPU' together with an device number or 'CPU'." << std::endl;
+            net_set = false;
+        }
+    };
+
+    void descriptorExtractor::loadNet(const string& model_file, const string& trained_file, string mean_file)
```
mtamburrano: why the parameter mean_file is not "const string&" like others?

modules/cnn_3dobj/src/cnn_feature.cpp
```diff
@@ -0,0 +1,244 @@
...
+                net_ready = 2;
+            }
+            else
+            {
+                net_ready = 1;
+            }
+        }
+        else
+        {
+            std::cout << "Error: Device must be set in advance using SetNet function" << std::endl;
```
mtamburrano: SetNet function doesn't exist anymore

modules/cnn_3dobj/src/cnn_feature.cpp
```diff
@@ -0,0 +1,244 @@
...
+                        int dim_feature = feature_vector.cols;
+                        feature_vector.resize(img.size(), dim_feature);
+                    }
+                    feature_vector.row(i) = cv::Mat(featureVec, true).t();
+                }
+                feature_vector.copyTo(feature);
+            }
+        }
+        else
+          std::cout << "Network must be set properly using SetNet and LoadNet in advance.";
```
mtamburrano: SetNet function doesn't exist anymore

modules/cnn_3dobj/test/test_cnn_3dobj_feature_extract.cpp
```diff
@@ -0,0 +1,63 @@
...
+ * This test checks the following:
+ * Feature extraction by the triplet trained CNN model
+ */
+void CV_CNN_Feature_Test::run(int)
+{
+    string caffemodel = "/home/wangyida/Desktop/opencv_contrib/modules/cnn_3dobj/samples/data/3d_triplet_iter_5000.caffemodel";
+    string network_forIMG = "/home/wangyida/Desktop/opencv_contrib/modules/cnn_3dobj/samples/data/3d_triplet_testIMG.prototxt";
+    string mean_file = "no";
+    std::vector<string> ref_img;
+    string target_img = "/home/wangyida/Desktop/opencv_contrib/modules/cnn_3dobj/samples/data/images_all/1_8.png";
```
mtamburrano: absolute paths are not very good

Wangyida:
Thanks, All fixed with your commit @mtamburrano

vpisarev:
:+1:

bhack:
@Wangyida @vpisarev is Trying to merge here. Can you try to rebase?

Wangyida:
@bhack @vpisarev It is just been rebased and some shell scripts have been modified these days, they are not necessary if the module need to be tiny enough.

bhack:
I see you have introduced some code for Pascal 3D+ plus dataset. @vpisarev Cause this is "a real dataset" do you think that he need to contribute in [datasets module](https://github.com/Itseez/opencv_contrib/tree/master/modules/datasets)?

bhack:
@Wangyida do you have got some interesting results on Pascal3d+?

Wangyida:
@bhack Yes, I have done some modification in the code samples both in Caffe PR and this OpenCV PR, these code samples could simply generating some proper training samples for PASCAL3D+ models.

bhack:
Do you have already some interesting results training this model on PASCAL3D+?

Wangyida:
@bhack The result could be shown as a different distribution against some softmax like distribution, The feature of generated images from PASCAL3D+ could be shown using PCA to a 3 dimensional space and the separation is more clear than softmax like distribution. Do you think this is an useful result? I am doing experiment on this dataset now.

bhack:
@Wangyida  I mean if you have already some performance result on PASCAL3D+ evaluation metrics

Wangyida:
@bhack I trained on all generated images from CAD models and get an accuracy of 30% around using norm2 distance for nearest neighbour classification on selected images from Imagenet samples from PASCAL3D+ dataset.

bhack:
See [recent state of the art](https://shapenet.cs.stanford.edu/projects/RenderForCNN/).

pwohlhart:
@bhack @Wangyida concerning the ICCV paper: from what I understand by quickly browsing over the paper, they dont do recognition, only pose estimation. The class class has to be known and effectively there is even a separate branch for each class in the final network (Sec 4.2). That's a bit lame :)

bhack:
@pwohlhart It is a bit lame for a ICCV selected "oral presentation" :).
But this kind of architectures could be assembled in an unified network in Caffe with the filter layer if you can find a common shared architecture that can solve classification.

Wangyida:
@bhack Thanks, This paper shows something meaningful and clear indeed, it is a good analysis on model based training.
@pwohlhart Though my experiment on the PASCAL3D+ dataset, I can find that even the triplet loss function using the class label and pose label at the same time, the training information is till a weaker supervised label using triplet(show some comparatively information in the triplet set). Though the trained model could do both pose estimation and classification, the classification result itself is usually worse than a softmax loss function using the strongly surprised class label, is any work have a balance on these two direction?

bhack:
Caffe + generation code is at https://github.com/ShapeNet/RenderForCNN

pwohlhart:
@Wangyida I llike the "strongly surprised" :-)   What you are saying is that the triplets are a weaker information than the class info? That's true in a sense, but also the task for the CNN is quite different ... computing descriptors vs. directly delivering the final output. There are pros and cons for both approaches.
Anyway, I think you should be able to get better results if you do a clever sampling of the triplets. I'm currently implementing another round of hard negative mining, where I search for samples that are close to "wrong" templates and add them together with their "target template" in new minibatches.

Wangyida:
@pwohlhart  Through my experiment I find that this 'multi-label' task which both including class and pose could be optimised by fine-tuning after your origin arrangement of triplet which have 3 negative samples(2 negative class and 1 worse pose with the same class), if I will do a simple classification issue, these could be all set to negative class if the pose differs little with each other.

edgarriba:
hi guys! I've been checking this PR and if I'm not wrong the class descriptorExtractor is concerned to return 2D features which makes me wonder why did you decide to include it in this module instead of including it into xfeatures2d as an inherited class of Features2D?

PR #279, opened 2015-06-22
Fix for Itseez/opencv_contrib/issues/278 - core dump in the case of no match(...) results.

buzziam:
Modified reported poses by constraining to the number of poses found - thus avoiding core dump.


bmagyar:
Hi, thanks for the fix!

Just being pedantic:
* You should exit with EXIT_FAILURE, 0 is for successful execution. 
* I have no problems with using exit, @vpisarev may have a different opinion on that.

vpisarev:
for a sample it's fine to use exit()

buzziam:
Your call folks - my opinion when looking at it was that exit was already being used in a different case and that its okay not too have found a match - in fact it may even be considered successful to have not found one. But if the underlying assumption is that a match will/must/should be found - then clearly no matches would be a failure. Let me know - I'll be happy either way.

bmagyar:
You are right, I was being too pedantic.

+1

PR #282, opened 2015-06-26
GSoC 2015: Projected Augmented Reality

micuat:
Hi,

This is a pull request for a GSoC project, Projected Augmented Reality.

Todo:
- ✖ more efficient segmentation/clustering algorithm  
  ➔ **haven't been done yet since this turned out to be complicated; currently inefficient but at least works**
- ✖ make test (stubbed)  
  ➔ **no tests since contributions are mostly sample codes**
- ✖ projector-camera (Kinect) calibration
  - <del>what is the status of the structured light project?</del>  
    ➔ **merged from Roberta's branch, but the stub version**
- ✓ a depth-map capture program
- ✓ uv texture coordinate mapping
- ✓ upload the Unity renderer app
- ✖ Processing/openFrameworks renderer for projection mapping w/o Unity
  ➔ **haven't been done since now it's hugely relying on Unity features**
- ✖ integrate with `surface_matching`  
  ➔ **will be a post-GSoC project**
- update tutorials


micuat:
[PR 275](https://github.com/FelixMDenis/opencv_contrib/blob/master/modules/surface_matching/src/ppf_helpers.cpp) has a .ply exporter. Implementing .ply and .obj exporter in separate modules is not good, I think

bmagyar:
@micuat Indeed. The student preferred to have it this way last year and there wasn't more time to move it outside. Those should be a library call, or we should make it a new module "meshtools" ?

micuat:
Thanks @bmagyar, I'm even thinking of merging `surface_matching` to `rgbd` since they have similar features (`rgbd` has ICP for odometry).

Currently I'm trying to refactor `rgbd` so that `depthTo3d` and other functions can be applied to `RgbdFrame` directly, instead of passing several matrices as function arguments. I'm trying to keep backward compatibility but probably need other peoples' ideas as well (in `rgbd_projection_dev` branch, as I don't want to mess up this PR before midterm evaluation).
https://github.com/micuat/opencv_contrib/blob/rgbd_projection_dev/modules/rgbd/include/opencv2/rgbd.hpp

bmagyar:
Yes! I did some exploratory work before thinking about how to make the API of surface matching more convenient and I found the need to use depthTo3d to move from depth images to a set of 3d points. If you have time for it, feel fee to merge it. I may do it myself by the end of summer if all goes well.

PR #285, opened 2015-06-28
[GSoC 2015] ArUco Module 

sergarrido:
[GSoC '15] ArUco / ChArUco Project - Midterm PR

Current main features of the aruco module:
- Detection of ArUco square fiducial markers.
- Pose estimation from one marker or board/set of markers.
- Camera calibration using markers
- Detection of high accurate Charuco corners
- Pose estimation from Charuco corners
- Camera calibration using Charuco corners.
- Samples and tests included.


modules/aruco/samples/board_detector.cpp
```diff
@@ -0,0 +1,227 @@
...
+any direct, indirect, incidental, special, exemplary, or consequential damages
+(including, but not limited to, procurement of substitute goods or services;
+loss of use, data, or profits; or business interruption) however caused
+and on any theory of liability, whether in contract, strict liability,
+or tort (including negligence or otherwise) arising in any way out of
+the use of this software, even if advised of the possibility of such damage.
+*/
+
+
+#include <opencv2/highgui.hpp>
```
vpisarev: each .cpp should include "precomp.hpp" as the first header, in order to make the precompiled header mechanism work properly on Windows

modules/aruco/include/opencv2/aruco/charuco.hpp
```diff
@@ -0,0 +1,269 @@
...
+or tort (including negligence or otherwise) arising in any way out of
+the use of this software, even if advised of the possibility of such damage.
+*/
+
+#ifndef __OPENCV_CHARUCO_HPP__
+#define __OPENCV_CHARUCO_HPP__
+
+#include <opencv2/core.hpp>
+#include <vector>
+#include "../aruco.hpp"
```
vpisarev: it's not good practice: 1) to use relative path in #include 2) to include the main module header. Instead, aruco.hpp should include charuco.hpp (as "opencv2/aruco/charuco.hpp")

modules/aruco/src/aruco.cpp
```diff
@@ -0,0 +1,1038 @@
...
+(including, but not limited to, procurement of substitute goods or services;
+loss of use, data, or profits; or business interruption) however caused
+and on any theory of liability, whether in contract, strict liability,
+or tort (including negligence or otherwise) arising in any way out of
+the use of this software, even if advised of the possibility of such damage.
+*/
+
+#ifndef __OPENCV_ARUCO_CPP__
+#define __OPENCV_ARUCO_CPP__
+#ifdef __cplusplus
```
vpisarev: this file will always be built using C++ compiler, so this #ifdef __cplusplus is unnecessary

modules/aruco/src/aruco.cpp
```diff
@@ -0,0 +1,1038 @@
...
+disclaimed. In no event shall copyright holders or contributors be liable for
+any direct, indirect, incidental, special, exemplary, or consequential damages
+(including, but not limited to, procurement of substitute goods or services;
+loss of use, data, or profits; or business interruption) however caused
+and on any theory of liability, whether in contract, strict liability,
+or tort (including negligence or otherwise) arising in any way out of
+the use of this software, even if advised of the possibility of such damage.
+*/
+
+#ifndef __OPENCV_ARUCO_CPP__
```
vpisarev: why do you use these guards in .cpp files?

sergarrido: Thanks for the comments.
It has been corrected.

PR #259, opened 2015-06-12
added INRIA pedestrian dataset

Jiaolong:
@bmagyar Could you please help me to review this patch?

PR #240, opened 2015-06-01
added PASCAL VOC dataset

Jiaolong:
Tested on VOC2012 only.

PR #286, opened 2015-06-29
GSoC 2015: improving part-based detector

Jiaolong:
OpenCV GSoC 2015 project: improving deformable part-based detector

- Youtube video:
https://youtu.be/Q6sf9zp5ltY

- Octave training code:
https://github.com/Jiaolong/dpm-octave

- Corresponding models and data:
https://github.com/Itseez/opencv_extra/pull/264


modules/dpm/src/dpm_cascade.hpp
```diff
@@ -0,0 +1,135 @@
...
+//M*/
+
+#ifndef __DPM_CASCADE__
+#define __DPM_CASCADE__
+
+#include "dpm_model.hpp"
+#include "dpm_feature.hpp"
+#include "dpm_convolution.hpp"
+
+#include "opencv2/opencv.hpp"
```
bmagyar: Including this here should be avoided. Apart from not compiling now on the server it would mean including the entire OpenCV universe for this project.

In general: try to minimize includes. Only include headers that have types you use _in this_ file.

modules/dpm/src/dpm_cascade_detector.cpp
```diff
@@ -0,0 +1,159 @@
...
+//
+//  By downloading, copying, installing or using the software you agree to this license.
+//  If you do not agree to this license, do not download, install,
+//  copy or use the software.
+//
+//
+//                           License Agreement
+//                For Open Source Computer Vision Library
+//
+// Copyright (C) 2010-2013, University of Nizhny Novgorod, all rights reserved.
```
bmagyar: This license could be updated to reflect the significant amount of work done on it by you. You can add the standard 2015 OpenCV note on top of this statement

modules/dpm/src/dpm_cascade_detector.cpp
```diff
@@ -0,0 +1,159 @@
...
+    const std::vector<std::string>& getClassNames() const;
+    size_t getClassCount() const;
+    std::string extractModelName( const std::string& filename );
+
+private:
+    std::vector<DPMCascade*> detectors;
+    std::vector<std::string> classNames;
+};
+
+cv::Ptr<DPMDetector> DPMDetector::create(std::vector<std::string> const &filenames,
```
bmagyar: You are in namespace cv, no need to write cv:: while inside.

modules/dpm/src/dpm_cascade_detector.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+DPMDetectorImpl::DPMDetectorImpl( const std::vector<std::string>& filenames, const std::vector<std::string>& _classNames )
+{
+    for( size_t i = 0; i < filenames.size(); i++ )
+    {
+        const std::string filename = filenames[i];
+        if( filename.length() < 5 || filename.substr(filename.length()-4, 4) != ".xml" )
+            continue;
+
+        DPMCascade* detector = new DPMCascade();
```
bmagyar: In general whereever you say "new", there should be a corresponding delete. Now you have a tiny memory leak. 

Using here Ptr<DPMCascade> would solve it simply. Whenever you have raw pointers, make an effort to replace them with Ptr<...>

Jiaolong:
Tested on INRIA Pedestrian dataset:

Current version: dpm-cascade				
Interval	TBB	FPS	AP
10	        yes	1.5	0.86
10	        no	1.0	0.86			
5	       yes	2.7	0.80
5	       no	1.9	0.80
				
Previous version: latentsvm (cascade)				
Interval	TBB	FPS	AP
10	        yes	1.4	0.87
10	        no	0.9	0.87
5	        yes	2.4	0.81
5	        no	1.7	0.81

Interval: number of levels per octave in the feature pyramid
TBB: using Intel Threading Building Block (parallel)
FPS: frame per second
AP: average precision

Jiaolong:
Added samples.

Jiaolong:
- Added new example: detection from web camera:
`./example_dpm_detect_camera <model_path>`

- Added OpenMP support. If TBB is not available, OpenMP can also be used.

vpisarev:
@Jiaolong, @bmagyar, is the PR ready for integration? From my side I do not have any objections. The patch grew quite big, so probably it makes sense to merge it in and then refine with subsequent PR's. What do you think?

bmagyar:
I would like to run the new samples today, but in general I think it is ready. 

According to the test results above, the new module has reached and exceeded the capabilities of the old one already, and it has a much nicer API + support for paralellization.

bmagyar:
Done, all went fine and fast.

:+1:

Jiaolong:
@vpisarev please check and merge this patch, thanks!

Jiaolong:
@vpisarev, @bmagyar, Once this patch is merged, could you please also check the DPM Octave training code? It is currently in [my repository](https://github.com/Jiaolong/dpm-octave). I just wonder where to put it in opencv_contrib? Thanks!

bmagyar:
This is my final :+1:

Jiaolong:
Updated README

vpisarev:
:+1:

Jiaolong:
Thanks! :+1:

bmagyar:
Adding references to by-products of the GSoC project for future reference #240 #259

adekurniawan44:
hi...recently i'm doing research about dpm algorithm, i should train a new model for dpm, so how to do that? i tried to make it in Pascaldevkit 2007 and voc-release 3.1 for windows version, it didn' success, any suggestion or some tutorial how to train a new model for windows version, for your kindness i say thankyou so much...

PR #291, opened 2015-07-01
Waldboost with LBP

shahurik:
Waldboost improvement project.
Feature quantization was added to training procedure, so now training is much faster.
Waldboost was debugged for face detection with LBP features from OpenCV cascade. Currently it shows similar to rigid cascade quality, but is 1.5-2 times faster.


vpisarev:
@shahurik, thanks for the PR!
Here are some requests:
1) Please, add the OpenCV license (in addition to OpenCV Foundation, there can also be your copyright) in the beginning of each file.
2) start each source file (.cpp) with #include "precomp.hpp" to enable the precompiled header mechanism on Windows
3) instead of putting "using namespace cv" it's preferable to enclose all the functionality in cv::xobjdetect namespace.
4) I believe, xobjdetect is normal OpenCV module, but one of the files contains main() function. It needs to be moved into a sample.

teng88:
@shahurik, in file "modules/xobjdetect/src/main.cpp", from line  238 - 266: 
In some cases, max_col is finally equal to neg_trace.cols. Could you confirm this problem?

teng88:
@shahurik, in file "modules/xobjdetect/src/main.cpp", from line 498 to 512:
bboxes.size() may not equal to confidences.rows, because bboxes are grouped in function boost.detect, this leads to an access exception in line 509

modules/xobjdetect/src/main.cpp
```diff
@@ -0,0 +1,556 @@
...
+            if (confidences.rows > 0) {
+                Mat1i indices;
+                sortIdx(confidences, indices,
+                        CV_SORT_EVERY_COLUMN + CV_SORT_DESCENDING);
+
+                int win_count = min(max_per_image, confidences.rows);
+                win_count = min(win_count, stage_neg - bootstrap_count);
+                Mat window;
+                for (int k = 0; k < win_count; ++k) {
+                    resize(img(bboxes[indices(k, 0)]), window, Size(24, 24));
```
teng88: bboxes.size() may not equal to confidences.rows, because bboxes are grouped in function boost.detect, this leads to an access exception in line 509, e.g. bboxes.size() is 0 and confidences.rows is 1

modules/xobjdetect/src/main.cpp
```diff
@@ -0,0 +1,556 @@
...
+                        max_col = col_neg;
+                    max_pos = pos;
+                }
+            }
+
+            if (max_pos) {
+                cascade_thresholds_.push_back(pos_trace(0, max_col));
+            } else {
+                cascade_thresholds_.push_back(neg_trace(0, max_col));
+            }
```
teng88: In some cases, max_col is finally equal to neg_trace.cols, leading to an access exception in line 265

modules/xobjdetect/src/main.cpp
```diff
@@ -0,0 +1,556 @@
...
+            int col_pos = 0, col_neg = 0;
+            int cur_pos = 0, cur_neg = 0;
+            int max_col = -1;
+            bool max_pos = false;
+            for (;
+                 col_pos < data_pos.cols &&
+                 col_neg < data_neg.cols; ) {
+                bool pos = false;
+                if (data_pos.at<uchar>(min_feature_ind, col_pos) <
+                    data_pos.at<uchar>(min_feature_ind, col_neg)) {
```
teng88: should this line be data_neg.at<uchar>(min_feature_ind, col_neg)?

modules/xobjdetect/include/opencv2/xobjdetect.hpp
```diff
@@ -49,203 +49,29 @@ the use of this software, even if advised of the possibility of such damage.
...
+    virtual ~WBDetector(){}
 };
 
-CV_EXPORTS void write(FileStorage& fs, String&, const ICFDetector& detector);
-
-CV_EXPORTS void read(const FileNode& node, ICFDetector& d,
-    const ICFDetector& default_value = ICFDetector());
-
-//! @}
+CV_EXPORTS Ptr<WBDetector> create_wbdetector();
```
vpisarev: please, make static WBDetector::create() method instead

modules/xobjdetect/src/waldboost.cpp
```diff
@@ -28,360 +32,416 @@ are permitted provided that the following conditions are met:
...
+        }
     }
-    return trace;
+    *h = res;
+    return res > cascade_thresholds_[count - 1] ? +1 : -1;
+}
+
+void WaldBoost::write(FileStorage &fs) const
+{
+    fs << "waldboost" << "{";
```
vpisarev: the method should not write any name. Delete `<< "waldboost"`

modules/xobjdetect/tools/waldboost_detector/waldboost_detector.cpp
```diff
@@ -0,0 +1,45 @@
...
+        return 0;
+    }
+
+    string mode = argv[1];
+    Ptr<WBDetector> detector = create_wbdetector();
+    if (mode == "train") {
+        assert(argc == 5);
+        detector->train(argv[3], argv[4]);
+        FileStorage fs(argv[2], FileStorage::WRITE);
+        detector->write(fs);
```
vpisarev: use `fs << "waldboost" << detector`

modules/xobjdetect/tools/waldboost_detector/waldboost_detector.cpp
```diff
@@ -0,0 +1,45 @@
...
+    } else if (mode == "detect") {
+        assert(argc == 6);
+        vector<Rect> bboxes;
+        vector<double> confidences;
+        Mat img = imread(argv[3], CV_LOAD_IMAGE_GRAYSCALE);
+        FileStorage fs(argv[2], FileStorage::READ);
+        detector->read(fs["waldboost"]);
```
vpisarev: use `fs.getFirstTopLevelNode()` instead of `fs["waldboost"]`

vpisarev:
@shahurik, the code looks pretty good now and the API is much closer to our style than before. May I ask you to add some simple regression test using our googletest-based engine - put some pre-trained cascade for faces and within the test run it on a few pictures and check that it found certain number of faces around certain locations?

vpisarev:
@shahurik, thanks! can we expect a regression test as well?

vpisarev:
:+1:

gabrielkdc:
In the CMakeLists.txt I got in the module root folder, there is next single line:
add_subdirectory(waldboost_detector)

but this directory do not exists and CMake complains about it, I just changed it to

add_subdirectory(tools)

And OpenCV generated correctly, maybe you want to check it.

LorenaGdL:
Are the ICF/ACF detectors completely gone now?

StevenPuttemans:
@vpisarev @shahurik could any one of you tell us why ICF/ACF has been deleted by this PR? This seems nonsense? Even if it was slow and underperforming...

PR #298, opened 2015-07-09
Adds createOCRHMMTransitionsTable() utility function

lluisgomez:
allows to create a tailored language model transitions table from a given list of words (lexicon).


vpisarev:
:+1:

PR #299, opened 2015-07-11
Fixed bug in LSDDetector where mask doesn't remove all undesired lines

atomala:
I noticed the mask feature in LSDDetector::detect wasn't working so I looked at the code and realized that the keyCounter isn't decremented when an line is erased. This causes the loop to skip over a line which ends up screwing the mask feature. Adding the decrement fixed the problem.


vpisarev:
:+1:

PR #304, opened 2015-07-22
make sources compile again on MSVC 2012 (VC 11) by adding round()

AlexanderStohr:
MSVC had no implemenation for round() in earlier versions.
support for round() was probably first added with MSVC 2013.


alalek:
:+1:

alalek:
@AlexanderStohr Some note: please don't use `master/2.4` as a branch name. It may cause issues with our build infrastructure.

PR #308, opened 2015-07-24
Fix bug #4373: 

lluisgomez:
Error (Assertion failed in resize) when passing very elongated contours to the recognition module.
See http://code.opencv.org/issues/4373


vpisarev:
:+1:

PR #320, opened 2015-07-31
White Space missing in cmake command

der-Daniel:
The command is not executed correctly if copy & paste.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Here is the CMake command for you:
 
 ```
 $ cd <opencv_build_directory>
-$ cmake -DOPENCV_EXTRA_MODULES_PATH=<opencv_contrib>/modules <opencv_source_directory>
+$ cmake -D OPENCV_EXTRA_MODULES_PATH=<opencv_contrib>/modules <opencv_source_directory>
 $ make -j5
 ```
 
@@ -27,7 +27,7 @@ modules from `opencv_contrib` repository. If you don't want all of the modules,
 use CMake's `BUILD_opencv_*` options. Like in this example:
 
 ```
-$ cmake -DOPENCV_EXTRA_MODULES_PATH=<opencv_contrib>/modules -DBUILD_opencv_legacy=OFF <opencv_source_directory>
+$ cmake -D OPENCV_EXTRA_MODULES_PATH=<opencv_contrib>/modules -D BUILD_opencv_legacy=OFF <opencv_source_directory>
 ```
 
 If you prefer using the gui version of cmake (cmake-gui), then, you can add `opencv_contrib` modules within `opencv` core by doing the following:


alalek:
We use CMake command without spaces for years.
Could you provide more details about problem cases (CMake version, OS, specifies paths with spaces or not)?

BTW, CMake commands example: http://pullrequest.opencv.org/buildbot/builders/master-win64-vc12/builds/413/steps/cmake/logs/stdio

der-Daniel:
Wow, if you are using it for years it may have been caused by a different source...
I have never used white spaces ever :p 

I will try to reproduce the error, I got yesterday. I was kinda sure my failed build was caused by the missing white spaces / bad cmake command. Maybe the error was caused by another source.

Anyway isn't it official to have white space between the D flag and its value:
http://www.cmake.org/cmake/help/cmake2.4docs.html

StevenPuttemans:
@alalek and @der-Daniel actually this pull request is following the correct CMAKE convention. I have been doing it all my programming life with the space. Some Linux based systems, like Ubuntu14.04 do not agree if you manually pass  `-DOPENCV_EXTRA_MODULES_PATH` but they request you to pass `-D OPENCV_EXTRA_MODULES_PATH`. Once it is not manually passed but through the cmake-gui then it doesn't mind.

So :+1: for this PR!

alalek:
Actually it works on Ubuntu 14.04 also: http://master.ocv/buildbot/builders/master_noOCL_noICV-lin64/builds/424/steps/cmake/logs/stdio

I don't see any problems to accept this patch, but It will be good to know more details about real reason of this problem in case of usage without space (spaces in path / unicode chars / some symbols / slash direction / something else).

BTW, True official command line parameter is (with `:PATH` type):
```
-D OPENCV_EXTRA_MODULES_PATH:PATH=<opencv_contrib>/modules
```

StevenPuttemans:
For some reason I cannot access that link? I guess it is a local server?
I can confirm that my Ubuntu 14.04 does indeed accept both now :) It didn't when it was just released, nor did my 12.04 implementation.

alalek:
@StevenPuttemans Link is fixed

der-Daniel:
I tried to reproduce the error. I was not able to create the complete workflow which led me to this pull req. The whole installation was a bit tricky on my machine..


This may be the wrong place to start a fundamental discussion about CMake syntax, but why is a missing space ok in the first place.
The general Terminal/Shell syntax always uses spaces the separate options identifiers from their values.

This could be a critical scenario for not having white spaces:

If you are using ```-DXYZ``` you cannot predict what is the option identifier and what is the utilized value:
There are 3 possibilities:
```-D XYZ```
```-DX YZ```
```-DXY Z```

So if the are writing a script to build a software and you want to parameterise its options into a .ini file, you would definitely need to able to make this distingution. 

It may be obvious to a person to correctly guess that -DPYTH.. is a -D PYTH... option, but it would be hard to guess for a parser given that input.

PS: Does the missing white space also work on Windows machines?

alalek:
All builders from here (http://pullrequest.opencv.org/#/summary/contrib) uses `-D` parameter without space.

der-Daniel:
If this works, it's fine
This is more a fundamental discussion, why CMake is allowing this.

I am arguing from a 'never-used-NOT-white-spaces-between-options-and-value' guy. 

You neither do the following:

rm -rmydir (is it -r mydir or -rm ydir
cp -rfdir1 dir2 (is it -rf dir1 or -r fdir1?)


But like I said, that may not be the right place.
I was thought this was a Bug, because of my 'never-used-NOT-white-spaces-between-options-and-value' guy point o view. Turns out it was not, so this Pull Req. is kinda weird :D

Mathematics always like clean gramma.

Cheers

StevenPuttemans:
I agree with @der-Daniel that it is not a question of functionality but a question of following some specific universal guidelines. And I do think that his examples are the exact reason why we should stick to those official guidelines/rules.

mshabunin:
@der-Daniel , please see my comment in #250.

StevenPuttemans:
@mshabunin I understand your approach, but I rather investigate why a universal guideline is not working on your system :D that might uncover the real problem.

PR #321, opened 2015-08-03
BaseOCR / OCRHMMDecoder refactor

lluisgomez:
This is an important change in classes BaseOCR / OCRHMMDecoder / OCRTesseract / OCRBeamSearchDecoder.
But it does NOT break compatibility with previous version.

Overload the run() method in BaseOCR class in order to adapt to different classifier callbacks. The original run() method accepts only one Mat input image, this is expected to be a binarzed image with black and white text and works both with the OCRTesseract class and the OCRHMMDecoder class when the character classifier callback works with binary images (e.g. NM). The new run() method accepts two Mat input parameters. One for the gray scale (or color) source image and the other for a binary mask where each connected component corresponds to a pre-segmented character in the input image. This way the OCRHMMDecoder is able to work with character classifiers that operate in grey scale (or color) images (e.g. a CNN).

Also, updates the CNN model for character recognition. Trained with an augmented dataset.

Also, Adds example on segmented word recognition.


vpisarev:
:+1:

Issue #167, opened 2015-02-03
rst_parser module not found when compiling matlab binding


anuppari:
I'm trying to follow the instructions here http://seevisionc.blogspot.com/2014/03/compiling-opencv-30-with-matlab-support.html for compiling with matlab bindings. In cmake, I get errors that the rst_parser is not found. I see the hdr_parser module is in the modules/python/src2 folder but I don't see any rst_parser. Is this a separate python module? If so, where is it available ( a quick google search didn't provide any results)?

mshabunin:
@anuppari, rst documentation has been removed from OpenCV along with _rst_parser_ script. 
You can try one of the following solutions:
- remove _rst_parser_ dependency from Matlab binding
- take older version of OpenCV sources (before [this commit](https://github.com/Itseez/opencv/commit/d01bedbc61b7febbb0d35a5978efb4c30709af5d))

anuppari:
It looks like gen_matlab.py in the matlab module is still heavily dependent on rst_parser. What can be used to replace it?

talbym:
Could someone drop a comment on this issue, when it is solved, over in the openCV user forum in the comment area of this [thread](http://answers.opencv.org/question/54990/bindings-to-access-opencv-from-matlab-do-not-generate/), please? I feel to still be too much beginner for bothering you here at github, because I could only mention problems I found, but not constructively add to the discussion are problem solving. Thanks to all of you who work on openCV, you are just great! Talby

yifita:
if someone downloaded the earlier version of rst_parser, could you share it somewhere? i can no longer retrieve that file from old commits

howtobeahacker:
@yifita @talbym  You can download rst_parser here https://github.com/Itseez/opencv/blob/d01bedbc61b7febbb0d35a5978efb4c30709af5d/modules/java/generator/rst_parser.py.
I downloaded and put it in my local corresponding folder, according to this solution [http://code.opencv.org/issues/4175]. And it works

StevenPuttemans:
temporary solution can be found above and PR for adding it back is suggested for the moment :)

StevenPuttemans:
PR merged and build just fine now! You can close this issue down!

PR #324, opened 2015-08-05
Remove dependency on RST parser for the matlab module

StevenPuttemans:
Removing the dependencies with help of @berak 
should solve https://github.com/Itseez/opencv_contrib/issues/167

### DIFF
--- a/modules/matlab/CMakeLists.txt
+++ b/modules/matlab/CMakeLists.txt
@@ -101,7 +101,6 @@ string(REGEX REPLACE "(\r?\n)+$" "" GIT_COMMIT "${GIT_COMMIT}")
 # set the path to the C++ header and doc parser, and template engine
 set(JINJA2_PATH ${CMAKE_SOURCE_DIR}/3rdparty)
 set(HDR_PARSER_PATH ${CMAKE_SOURCE_DIR}/modules/python/src2)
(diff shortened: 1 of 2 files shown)


berak:
imho, the dependancy should be removed instead, no ?

if you actually got matlab (no such thing here), do me a favour , and try [an attempt at fixing it](https://github.com/berak/opencv_contrib/tree/fix_matlab_rst)

StevenPuttemans:
@berak yep thats true, but since I am not realy into that parsing stuff, I didn't dare to try it myself. This fix would temporarily fix it.

I will see, never built matlab bindings before, but will see where it gets me! :) Thanks!

StevenPuttemans:
It works! I will add the changes :)

StevenPuttemans:
@berak: I changed the python generator files, but the pyc files, how do I generate those? Because you cannot open them in an editor

berak:
oh, nice to hear that. i originally planned to push the button myself, but you beat me to that ;)

the pyc files are locally generated, when running the py script, they should not go into the codebase at all.

StevenPuttemans:
O great!

StevenPuttemans:
Btw thdrksdfthmn might also be a good choice. He is quite active also!

alalek:
:+1:

StevenPuttemans:
thanks!

PR #326, opened 2015-08-06
Matching operations

MMp131316:
This class contains the most important functions used in stereo correspondence
(StereoBM and StereoSGBM) call methods from these classes


MMp131316:
@snosov1
A pull request containing the matching class

modules/stereo/src/matching.cpp
```diff
@@ -0,0 +1,338 @@
...
+        }
+        Matching::~Matching(void)
+        {
+        }
+        //constructor for the matching class
+        //maxDisp - represents the maximum disparity
+        Matching::Matching(int maxDisp, int scalling, int confidence)
+        {
+            CV_Assert(maxDisp > 10);
+            CV_Assert(scalling != 0);
```
snosov1: You have one set of checks here, and another one in functions, like setScalingFactor. As a matter of fact, the checks are slightly different, which, I assume, shouldn't be like that. Why duplicate the checks? Let's just use set* functions here.

modules/stereo/src/matching.cpp
```diff
@@ -0,0 +1,338 @@
...
+        void Matching::hammingDistanceBlockMatching(const Mat &leftImage, const Mat &rightImage, Mat &cost, const int kernelSize)
+        {
+            CV_Assert(leftImage.cols == rightImage.cols);
+            CV_Assert(leftImage.rows == rightImage.rows);
+            CV_Assert(kernelSize % 2 != 0);
+            CV_Assert(cost.rows == leftImage.rows);
+            CV_Assert(cost.cols / (maxDisparity + 1) == leftImage.cols);
+            // cost.setTo(0);
+            int *c = (int *)cost.data;
+            memset(c, 0, sizeof(c[0]) * leftImage.cols * leftImage.rows * (maxDisparity + 1));
```
snosov1: Sorry, Paul, I forgot, what's wrong with setTo?

Also, you're using .cols instead of stride here as well. It's probably, not exactly an error right here, but anyway, later on it can shoot you in the foot. That's why I'm saying that messing directly with pointers is error-prone, you have to be very careful. Once again, I would suggest to use "politically correct" OpenCV ways of doing things to avoid such errors (like, setTo in this case). 99% of the time there are no good reasons to not do this.

modules/stereo/src/matching.hpp
```diff
@@ -0,0 +1,364 @@
...
+            //!method for setting the scalling factor
+            void setScallingFactor(int val);
+            //!method for getting the scalling factor
+            int getScallingFactor();
+            //!setter for the confidence check
+            void setConfidence(double val);
+            //!getter for confidence check
+            double getConfidence();
+            //!method for computing the hamming difference
+            void hammingDistanceBlockMatching(const Mat &left, const Mat &right, Mat &c, const int kernelSize = 9);
```
snosov1: Are all of these methods supposed to be used by the client code? I don't think it's a good idea. 

Also, it seems like a simple "here's 2 images, give me the disparity" is missing, isn't it?

modules/stereo/src/matching.hpp
```diff
@@ -0,0 +1,364 @@
...
+                uint8_t *harta;
+                uint8_t *mapModified;
+                int height, width;
+            public:
+                Median9x1(const Mat &hartaOriginala, Mat &map)
+                {
+                    harta = hartaOriginala.data;
+                    mapModified = map.data;
+                    height = hartaOriginala.rows;
+                    width = hartaOriginala.cols;
```
snosov1: See, width instead of stride again. 

Please, double-check that everything is correct with all the other code in that regard.

modules/stereo/src/matching.hpp
```diff
@@ -0,0 +1,364 @@
...
+            };
+            //!median 9x1 paralelized filter
+            class Median9x1:public ParallelLoopBody
+            {
+            private:
+                uint8_t *harta;
+                uint8_t *mapModified;
+                int height, width;
+            public:
+                Median9x1(const Mat &hartaOriginala, Mat &map)
```
snosov1: `hartaOriginala` is this English? =)

Anyways, median filter is a very generic function - it takes a Mat and returns it slightly modified. The physical meaning of the values in Mat can be arbitrary. So, there's no reason to use specific words, like "harta". It can as well be an image, not a "disparitate harta" =)

And of course, using both "harta" and "map" (in the same function signature, by the way!) doesn't seem like a buna idee =)

modules/stereo/src/matching.hpp
```diff
@@ -0,0 +1,364 @@
...
+
+namespace cv
+{
+    namespace stereo
+    {
+        class Matching
+        {
+        private:
+            //arrays used in the region removal
+            int specklePointX[1000001];
```
snosov1: Why statically sized arrays?

modules/stereo/src/matching.cpp
```diff
@@ -0,0 +1,338 @@
...
+                hamLut[i] = dist;
+            }
+        }
+        //!remove small regions that have an area smaller than t, we fill the region with the average of the good pixels around it
+        void Matching::smallRegionRemoval(const Mat &currentMap, int t, Mat &out)
+        {
+            CV_Assert(currentMap.cols == out.cols);
+            CV_Assert(currentMap.rows == out.rows);
+            CV_Assert(t > 0);
+            memset(pus, 0, 2000000 * sizeof(pus[0]));
```
snosov1: Just a imagine how disastrous will it be if you'll decide to shrink the array a bit and forget to fix this memset.

BTW, for a static array `sizeof` returns the length of the whole array in bytes.

modules/stereo/samples/Sample1.cpp
```diff
@@ -0,0 +1,78 @@
...
+    sbm->setDisp12MaxDiff(0);
+    sbm->setScalleFactor(4);//the scalling factor
+    sbm->setBinaryKernelType(CV_MODIFIED_CENSUS_TRANSFORM);//binary descriptor kernel
+    sbm->setAgregationWindowSize(9);
+    sbm->setSpekleRemovalTechnique(CV_SPECKLE_REMOVAL_AVG_ALGORITHM);//speckle removal algorithm
+    sbm->setUsePrefilter(false);//prefilter or not the images prior to making the transformations
+
+    for(int i = 0 ; i < 200; i++)
+    {
+        string path = "D:\\WorkingSec";
```
snosov1: The paths shouldn't be hard-coded. The application should be configurable via command line. Also, cv::FrameSource can read "image directories" if you provide the filename as a "printf pattern". Please, see the documentation (http://docs.opencv.org/modules/highgui/doc/reading_and_writing_images_and_video.html#videocapture)

modules/stereo/include/opencv2/stereo.hpp
```diff
@@ -47,211 +47,229 @@
...
+            @param mode Set it to StereoSGBM::MODE_HH to run the full-scale two-pass dynamic programming
+            algorithm. It will consume O(W\*H\*numDisparities) bytes, which is large for 640x480 stereo and
+            huge for HD-size pictures. By default, it is set to false .
+
+            The first constructor initializes StereoSGBM with all the default parameters. So, you only have to
+            set StereoSGBM::numDisparities at minimum. The second constructor enables you to set each parameter
+            to a custom value.
+            */
+            CV_EXPORTS static Ptr<cv::stereo::StereoBinarySGBM> create(int minDisparity, int numDisparities, int blockSize,
+                int P1 = 0, int P2 = 0, int disp12MaxDiff = 0,
```
snosov1: The default values should be sensible, i.e. they should give more or less decent results on any more or less good stereo images.

modules/stereo/samples/Sample2.cpp
```diff
@@ -0,0 +1,61 @@
+#include <iostream>
```
snosov1: Why 2 almost identical samples?

modules/stereo/src/matching.hpp
```diff
@@ -0,0 +1,624 @@
...
+
+#ifndef _OPENCV_MATCHING_HPP_
+#define _OPENCV_MATCHING_HPP_
+#ifdef __cplusplus
+
+namespace cv
+{
+    namespace stereo
+    {
+        template <int imageDimension>
```
snosov1: Is this value always available at compile-time?

modules/stereo/src/stereo_binary_bm.cpp
```diff
@@ -46,693 +46,481 @@
...
+                preFilterType = StereoBinaryBM::PREFILTER_XSOBEL;
+                preFilterSize = 9;
+                preFilterCap = 31;
+                kernelSize = _kernelSize;
+                minDisparity = 0;
+                numDisparities = _numDisparities > 0 ? _numDisparities : 64;
+                textureThreshold = 10;
+                uniquenessRatio = 15;
+                speckleRange = speckleWindowSize = 0;
+                roi1 = roi2 = Rect(0, 0, 0, 0);
```
snosov1: What are those used for?

modules/stereo/src/stereo_binary_bm.cpp
```diff
@@ -46,693 +46,481 @@
...
+            void setSpekleRemovalTechnique(int factor = CV_SPECKLE_REMOVAL_AVG_ALGORITHM) { params.regionRemoval = factor; }
+
+            bool getUsePrefilter() const { return params.usePrefilter;}
+            void setUsePrefilter(bool value = false) { params.usePrefilter = value;}
+
+            int getScalleFactor() const { return params.scalling;}
+            void setScalleFactor(int factor) {params.scalling = factor; setScallingFactor(factor);}
+
+            int getMinDisparity() const { return params.minDisparity; }
+            void setMinDisparity(int minDisparity) { params.minDisparity = minDisparity; }
```
snosov1: Have you removed all the "contracts" checks (like, `CV_Assert(minDisparity > 0);`?

vpisarev:
@snosov1, please, let me know when/if you are fine with the content

vpisarev:
:+1:

PR #327, opened 2015-08-08
Fix SVT dataset standard evaluation protocol

lluisgomez:
Changes SVT recognition evaluation to be Case Insensitive.
Report the total F1 score (0.37) and not the mean-f1.


vpisarev:
:+1:

PR #328, opened 2015-08-08
ICDAR2015 dataset

lluisgomez:
Adds interface for the ICDAR2015 dataset and benchmark code using the text module.


vpisarev:
:+1:

PR #329, opened 2015-08-12
Optimization

kurnianggoro:
Included in this pull request:
- optimize the Mat variable declaration, use global variable instead of re-allocating in each functions
- optimization on the per-element division of complex number
- enable features merging


iglesias:
Hi Laksono. It may be a good idea to squash some of the commits in this pull request.

modules/tracking/src/trackerKCF.cpp
```diff
@@ -87,18 +87,20 @@ namespace cv{
     * KCF functions and vars
     */
     void createHanningWindow(OutputArray _dst, const cv::Size winSize, const int type) const;
-    void inline fft2(const Mat src, std::vector<Mat> & dest) const;
-    void inline fft2(const Mat src, Mat & dest) const;
+    void inline fft2(const Mat src, std::vector<Mat> & _dest, std::vector<Mat> & _layers) const;
```
iglesias: Why some of the arguments starting with _ and others no? Are you using the leading _ to denote the arguments that are passed by reference?

iglesias: I believe inline does not make much of a difference these days in C++. Anyway, I think it does not hurt having it either. I had never used it in methods/functions like these ones that are not precisely very short (few lines without control flow and loops).

modules/tracking/src/trackerKCF.cpp
```diff
@@ -626,36 +709,32 @@ namespace cv{
...
-      for(int j=0;j<_k.cols;j++){
-        temp=std::complex<double>(spec.at<Vec2d>(i,j)[0],spec.at<Vec2d>(i,j)[1])/(std::complex<double>(_alphaf_den.at<Vec2d>(i,j)[0],_alphaf_den.at<Vec2d>(i,j)[1])/*+std::complex<double>(0.0000000001,0.0000000001)*/);
-        spec2.at<Vec2d >(i,j)[0]=temp.real();
-        spec2.at<Vec2d >(i,j)[1]=temp.imag();
+  void TrackerKCFImpl::calcResponse(const Mat _alphaf, const Mat _alphaf_den, const Mat _kf, Mat & _response, Mat & _spec, Mat & _spec2) const {
+
+    mulSpectrums(_alphaf,_kf,_spec,0,false);
+
+    //z=(a+bi)/(c+di)=[(ac+bd)+i(bc-ad)]/(c^2+d^2)
+    double den;
```
iglesias: Just for the sake of good practice, initialise the variable upon declaration (e.g. to zero).

modules/tracking/src/trackerKCF.cpp
```diff
@@ -486,23 +570,23 @@ namespace cv{
     if(patch.rows==0 || patch.cols==0)return false;
 
     // extract the desired descriptors
-    switch(params.descriptor){
-      case GRAY:
-        if(img.channels()>1)cvtColor(patch,patch, CV_BGR2GRAY);
-        patch.convertTo(patch,CV_64F);
-        patch=patch/255.0-0.5; // normalize to range -0.5 .. 0.5
+    switch(desc){
+      default:
```
iglesias: I find this a bit funny, a switch that starts with a default ending with a break. Can you please elaborate?

iglesias: So, to me it seems that you wanted to write GRAY here.

iglesias:
Good job, @kurnianggoro. Have a look at the comments above, most of them are minor details.

modules/tracking/src/trackerKCF.cpp
kurnianggoro: i forget to clean up this mess, because i redefined several variables with duplicated name (from inside function become public variable), i will clean this mess soon

modules/tracking/src/trackerKCF.cpp
kurnianggoro: yes the default descriptor will be GRAY, i also surprised that this code working, ok i will rewrite in standard form with default in the last to avoid any unwanted results.

iglesias: If `desc` can only be either `GRAY` or `CN`, then there is no real need for `default` except for capturing errors.

vpisarev:
thanks! :+1:

kurnianggoro:
@vpisarev I think because this branch contains a commit from multitracker branch, and then I deleted those modifications, see /commit b245d3142fdb57510efc98fb5eaea85dbf7929a0 and 68f8387b6d153c434d8a773fcf56402f6bb0e6f0
shoud I make a new branch that not conflicted? our maybe you have a simpler solution?

vpisarev:
I would make a fresh PR, especially given that the patch is not that big

iglesias:
@kurnianggoro, why do you think that should be problem if they are changes you undid?

Anyhow, just sync your branch with Itseez:master and push again. Something in the following lines should work:

```
$ git checkout optimize # make sure you are in this branch
$ git fetch Itseez
$ git rebase Itseez/master
$ git push origin optimize
```

Note that you might need to force (`--force` or `-f`) the last push since you will be rewriting the history of your branch. This is in general not a good practice. However, if you are **sure** that none else forked your branch and plan to merge it in the feature, it is ok.

iglesias:
You could also just go with a more traditional:

```
$ git checkout optimize
$ git pull Itseez/master # this will probably lead you to conflicts, as the rebase above (forgot to mention)
# solve the conflicts locally
$ git push origin optimize # no need to force in this case, you will be adding an extra commit due to the merge
```

kurnianggoro:
@iglesias i think because i deleted a file that needed by the multitracker class, 
the error in building bot says: CONFLICT (modify/delete): modules/tracking/src/multiTracker.cpp deleted in 89897c0d5d5783db4a350da4edf5d607ccde87a4 and modified in HEAD. Version HEAD of modules/tracking/src/multiTracker.cpp left in tree.

@vpisarev moved to #352

vpisarev:
superseded by #352

PR #331, opened 2015-08-14
TLD Multi Object tracking support

Auron-X:
1. Extended Long term API with Multi-tracker classes
2. Added an optimization in detector part of MO TLD (~20% speedup)
3. Added a functionality to access VOT 2015 dataset
3. Fixed number of bugs


modules/tracking/include/opencv2/tracking/tracker.hpp
```diff
@@ -564,7 +565,7 @@ class CV_EXPORTS_W Tracker : public virtual Algorithm
   virtual void read( const FileNode& fn )=0;
   virtual void write( FileStorage& fs ) const=0;
 
- protected:
+ public:
```
lenlen: It should be protected... you can create a method
```

vpisarev:
ok, since #341 includes this, let's close it

PR #334, opened 2015-08-17
TrackerTLD export parameters & fix OCL detection

paroj:


lenlen:
/cc @Auron-X 
There may be conflicts with another pull request (https://github.com/Itseez/opencv_contrib/pull/331). There, the kernel arguments were fixed.

paroj:
the kernel arguments fix is a separate commit. can back it out if needed. I think it depends on which PR should get merged first.

PR #341, opened 2015-08-21
TLD/VOT2015 Datasets Support

Auron-X:
This PR includes previous multi tracker PR: https://github.com/Itseez/opencv_contrib/pull/331
-Added support of VOT2015 Dataset
-Added VOT2015 dataset usage example
-Tracking module now has dependency on Datasets module
-Fixed tracking and multitracking examples to read from datasets


modules/datasets/src/track_vot.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+inline bool fileExists(const std::string& name)
+{
+	struct stat buffer;
+	return (stat(name.c_str(), &buffer) == 0);
+}
+
+void TRACK_votImpl::loadDataset(const string &rootPath)
+{
+	ifstream namesList(rootPath + "/list.txt");
```
lenlen:
```
ifstream namesList((rootPath + "/list.txt").c_str());
```

modules/datasets/src/track_vot.cpp
```diff
@@ -0,0 +1,231 @@
...
+		while (getline(namesList, datasetName))
+		{
+			currDatasetID++;
+			vector <Ptr<TRACK_votObj> > objects;
+
+			//All frames/images loop
+			Ptr<TRACK_votObj> currDataset(new TRACK_votObj);
+
+			//Open dataset's ground truth file
+			ifstream gtList(rootPath + "/" + datasetName + "/groundtruth.txt");
```
lenlen:
```
ifstream gtList((rootPath + "/" + datasetName + "/groundtruth.txt").c_str());
```

lenlen:
@Auron-X can you check the indentation in these files? (4 spaces)

modules/datasets/samples/track_vot.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+#include <string>
+#include <vector>
+#include <set>
+
+using namespace std;
+using namespace cv;
+using namespace cv::datasets;
+
+#define DATASET_ID 1
```
lenlen: Maybe it could be useful to pass the dataset index as an optional argument with default to 1, what do you think?

lenlen:
@Auron-X can you add the usage documentation in dataset.hpp as for the others datasets?

Auron-X:
Yes, sure!

modules/tracking/samples/tracker_dataset.cpp
```diff
@@ -0,0 +1,224 @@
...
+			}
+			break;
+		}
+	}
+}
+
+static void help()
+{
+	cout << "\nThis example shows the functionality of \"Long-term optical tracking API\""
+		"TLD dataset ID: 1~10, VOT2015 dataset ID: 1~60\n"
```
lenlen: Add a row of the usage for example in the help string
````
"./example_tracking_tracker_dataset <tracker_algorithm> <dataset_path> <dataset_id>\n"
```

modules/datasets/samples/track_vot.cpp
Auron-X: Hmm.. I think it is better to leave it as it is - IMO user should know exact dataset he wants to use)

lenlen: But for each test must be recompiled all code. So we can leave it as optional, not mandatory, with default to 1.

modules/datasets/include/opencv2/datasets/track_vot.hpp
```diff
@@ -0,0 +1,102 @@
...
+// and on any theory of liability, whether in contract, strict liability,
+// or tort (including negligence or otherwise) arising in any way out of
+// the use of this software, even if advised of the possibility of such damage.
+//
+//M*/
+
+#ifndef OPENCV_DATASETS_TRACK_VOT_HPP
+#define OPENCV_DATASETS_TRACK_VOT_HPP
+
+#include <string>
```
vpisarev: please, move ununsed headers, such as iostream, map etc. to precomp.hpp or to the .cpp files

modules/datasets/include/opencv2/datasets/track_vot.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+    virtual int getDatasetLength(int id) = 0;
+
+    virtual bool initDataset(int id) = 0;
+
+    virtual bool getNextFrame(Mat &frame) = 0;
+
+    virtual vector <Point2d> getGT() = 0;
+
+protected:
```
vpisarev: since you will have derived class that will actually implement those methods, please, move the protected section to the derived class

modules/datasets/src/track_vot.cpp
```diff
@@ -0,0 +1,233 @@
...
+// indirect, incidental, special, exemplary, or consequential damages
+// (including, but not limited to, procurement of substitute goods or services;
+// loss of use, data, or profits; or business interruption) however caused
+// and on any theory of liability, whether in contract, strict liability,
+// or tort (including negligence or otherwise) arising in any way out of
+// the use of this software, even if advised of the possibility of such damage.
+//
+//M*/
+
+#include "opencv2/datasets/track_vot.hpp"
```
vpisarev: put `#include "precomp.hpp"` as the first directive

vpisarev:
@Auron-X, thanks! I noticed that the code produces some warnings in iOS build:
http://pullrequest.opencv.org/#/summary/contrib
could you please look at it?

lukacu:
Hello, so since I have been included in this discussion I will give some comments on behalf of VOT committee regarding the OpenCV/VOT integration.

We have been discussing the integration since VOT2014 workshop and we are very grateful for the support and interest, but I am worried about the current state of integration. VOT challenge uses a dedicated toolkit that handles experiments since their execution is complex and includes potential reinitializations and many other possible scenarios and manipulations of data. Because of this we offer a pretty clean way of integrating trackers into the toolkit using a dedicated protocol. We thought that OpenCV/VOT integration will look like an executable that will support the protocol and will enable using any OpenCV single-target tracker to be used in the evaluation. Instead, based on what I can see from the code, the current support somehow deals with loading the entire VOT dataset into OpenCV. I am not sure how this will contribute to the integration, since the experiments cannot be performed according to VOT methodology using only the sequences. I am also afraid that it could spread misinformation in the research community therefore I would like to know if the current status of OpenCV/VOT integration is considered final or if I am worrying too quickly?

lenlen:
@lukacu the work of this GSoC is focused on the cpu/gpu optimization of the TLD algorithm and the design and the development of multitarget tracker. This pull request is only a way to use the VOT dataset in OpenCV, is not the integration that we had discussed together, on which we will return to work soon.

lukacu:
OK, great. Thanks for clarification. I am aware of the primary task of the GSoC project, I was just unsure where does the VOT support fit in.

vpisarev:
@Auron-X, sorry once again; there are 2 GSoC projects related to the tracking module, so the previous PR that I merged now yields merge conflicts with your PR. Could you please rebase it and I will merge it in immediately.

Auron-X:
No prob

On Wed, Aug 26, 2015 at 11:40 PM, Vadim Pisarevsky <notifications@github.com

lenlen:
@vpisarev is it possible merge this pull request? Thanks.

vpisarev:
terribly sorry for delay! yes, definitely :+1:

PR #344, opened 2015-08-21
Ocr character recognizer

lluisgomez:
Adds demo program and sample images for scene text Character Recognition using the CNN classifier.

Add benchmark code for the Chars74k dataset. Using the CNN character classifier reaches 75% and 84% accuracy for case-sensitive and case-insensitive recognition respectively.


vpisarev:
thank you! :+1:

PR #346, opened 2015-08-22
[GSoC 2015] ArUco Module

sergarrido:
[GSoC '15] ArUco / ChArUco Project - Final PR

Main features:
- Detection of ArUco square fiducial markers.
- Pose estimation from one marker or board/set of markers.
- Detection of highly accurate Charuco corners
- Pose estimation from Charuco corners
- Camera calibration using Aruco markers and Charuco corners.
- Detection of Charuco diamond markers
- Samples, tutorials, doxygen documentation and tests included.

Sample video:
https://www.youtube.com/watch?v=f1LCDsPjobo


modules/aruco/test/test_charucodetection.cpp
```diff
@@ -0,0 +1,567 @@
...
+
+#include "test_precomp.hpp"
+#include <opencv2/aruco/charuco.hpp>
+#include <string>
+
+using namespace std;
+using namespace cv;
+
+
+const double PI = 3.141592653589793238463;
```
prasannavk: there is a CV_PI that can be used if I right.

modules/aruco/include/opencv2/aruco.hpp
```diff
@@ -0,0 +1,502 @@
...
+ * - maxErroneousBitsInBorderRate: maximum number of accepted erroneous bits in the border (i.e.
+ *   number of allowed white bits in the border). Represented as a rate respect to the total
+ *   number of bits per marker (default 0.5).
+ * - minOtsuStdDev: minimun standard deviation in pixels values during the decodification step to
+ *   apply Otsu thresholding (otherwise, all the bits are set to 0 or 1 depending on mean higher
+ *   than 128 or not) (default 5.0)
+ * - errorCorrectionRate error correction rate respect to the maximun error correction capability
+ *   for each dictionary. (default 0.6).
+ */
+struct DetectorParameters {
```
vpisarev: CV_EXPORTS needs to be put here: struct CV_EXPORTS DetectorParameters { ...

modules/aruco/include/opencv2/aruco.hpp
```diff
@@ -0,0 +1,502 @@
...
+
+    CV_EXPORTS DetectorParameters();
```
vpisarev: here you should put CV_WRAP instead of CV_EXPORTS

vpisarev:
@sergarrido, the code looks great! @prasannavk, if/when you are fine with the content, I can merge it :+1:

modules/aruco/test/test_charucodetection.cpp
sergarrido: Fixed

modules/aruco/include/opencv2/aruco.hpp
sergarrido: Fixed

vpisarev:
ok, let's put it in and then make subsequent PRs if necessary

prasannavk:
Will do more detailed check over weekend. Sorry couldnt check immediately.

modules/aruco/samples/detect_markers.cpp
```diff
@@ -0,0 +1,234 @@
...
+/**
+ */
+static void help() {
+    cout << "Basic marker detection" << endl;
+    cout << "Parameters: " << endl;
+    cout << "-d <dictionary> # DICT_4X4_50=0, DICT_4X4_100=1, DICT_4X4_250=2, "
+         << "DICT_4X4_1000=3, DICT_5X5_50=4, DICT_5X5_100=5, DICT_5X5_250=6, DICT_5X5_1000=7, "
+         << "DICT_6X6_50=8, DICT_6X6_100=9, DICT_6X6_250=10, DICT_6X6_1000=11, DICT_7X7_50=12,"
+         << "DICT_7X7_100=13, DICT_7X7_250=14, DICT_7X7_1000=15, DICT_ARUCO_ORIGINAL = 16" << endl;
+    cout << "[-v <videoFile>] # Input from video file, if ommited, input comes from camera" << endl;
```
prasannavk: *omitted

modules/aruco/samples/detect_markers.cpp
```diff
@@ -0,0 +1,234 @@
...
+    cout << "[-ci <int>] # Camera id if input doesnt come from video (-v). Default is 0" << endl;
+    cout << "[-c <cameraParams>] # Camera intrinsic parameters. Needed for camera pose" << endl;
```
prasannavk: should the camera parameters be separated by space?

modules/aruco/samples/detect_markers.cpp
```diff
@@ -0,0 +1,234 @@
...
+    cout << "[-l <markerLength>] # Marker side lenght (in meters). Needed for correct"
```
prasannavk: *length

modules/aruco/samples/detect_markers.cpp
```diff
@@ -0,0 +1,234 @@
...
+    aruco::Dictionary dictionary =
+        aruco::getPredefinedDictionary(aruco::PREDEFINED_DICTIONARY_NAME(dictionaryId));
+
+    bool showRejected = false;
+    if(isParam("-r", argc, argv)) showRejected = true;
+
+    bool estimatePose = false;
+    Mat camMatrix, distCoeffs;
+    if(isParam("-c", argc, argv)) {
+        bool readOk = readCameraParameters(getParam("-c", argc, argv), camMatrix, distCoeffs);
```
prasannavk: oh I see. So it is a camera parameter file. Need to mention that in the help.

modules/aruco/samples/detect_markers.cpp
```diff
@@ -0,0 +1,234 @@
...
+        if(!readOk) {
+            cerr << "Invalid camera file" << endl;
+            return 0;
+        }
+        estimatePose = true;
+    }
+    float markerLength = (float)atof(getParam("-l", argc, argv, "0.1").c_str());
```
prasannavk: why do you cast the output of atof?

prasannavk: specify units of ```markerLength``` in variable name as ```markerLength_m``` and specify it in comments too.

modules/aruco/src/aruco.cpp
```diff
@@ -0,0 +1,1573 @@
...
+                              OutputArrayOfArrays _contours, DetectorParameters params) {
+
+    Mat image = _image.getMat();
+    CV_Assert(image.total() != 0);
+
+    /// 1. CONVERT TO GRAY
+    Mat grey;
+    _convertToGrey(image, grey);
+
+    vector< vector< Point2f > > candidates;
```
prasannavk: bad naming: ```candidates``` and ```_candidates``` are so close. I cant tell the difference. Similarly, ```contours``` and ```_contours```. Also, why are candidates floating point while the contours are integers? Aren't they both pixel coordinates?

prasannavk: ok I guess you are trying to relate them somehow. make ```candidates``` descriptive, say, ```candidates_initial```

modules/aruco/src/aruco.cpp
```diff
@@ -0,0 +1,1573 @@
...
+        // all black or all white, depending on mean value
+        if(mean.ptr< double >(0)[0] > 127)
+            bits.setTo(1);
+        else
+            bits.setTo(0);
+        return bits;
+    }
+
+    // now extract code, first threshold using Otsu
+    threshold(resultImg, resultImg, 125, 255, THRESH_BINARY | THRESH_OTSU);
```
prasannavk: 125 and 255 are magic numbers. Can they be parameters or atleast named constants?

modules/aruco/src/aruco.cpp
```diff
@@ -0,0 +1,1573 @@
...
+    // check if standard deviation is enough to apply Otsu
+    // if not enough, it probably means all bits are the same color (black or white)
+    Mat mean, stddev;
+    // Remove some border just to avoid border noise from perspective transformation
+    Mat innerRegion = resultImg.colRange(cellSize / 2, resultImg.cols - cellSize / 2)
+                          .rowRange(cellSize / 2, resultImg.rows - cellSize / 2);
+    meanStdDev(innerRegion, mean, stddev);
+    if(stddev.ptr< double >(0)[0] < minStdDevOtsu) {
+        // all black or all white, depending on mean value
+        if(mean.ptr< double >(0)[0] > 127)
```
prasannavk: magic number.

modules/aruco/src/aruco.cpp
```diff
@@ -0,0 +1,1573 @@
...
+        vector< vector< Point2f > > currentCandidates;
+        vector< vector< Point > > currentContours;
+        _findMarkerContours(thresh, currentCandidates, currentContours,
+                            params.minMarkerPerimeterRate, params.maxMarkerPerimeterRate,
+                            params.polygonalApproxAccuracyRate, params.minCornerDistanceRate,
+                            params.minDistanceToBorder);
+
+        // join candidates
+        for(unsigned int i = 0; i < currentCandidates.size(); i++) {
+            candidates.push_back(currentCandidates[i]);
```
prasannavk: Why do you have a join candidates in the single scale case? You could just pass a reference.

modules/aruco/src/aruco.cpp
```diff
@@ -0,0 +1,1573 @@
...
+    _accepted.create((int)accepted.size(), 1, CV_32FC2);
+    for(unsigned int i = 0; i < accepted.size(); i++) {
+        _accepted.create(4, 1, CV_32FC2, i, true);
+        Mat m = _accepted.getMat(i);
+        accepted[i].copyTo(m);
+    }
+
+    _ids.create((int)ids.size(), 1, CV_32SC1);
+    for(unsigned int i = 0; i < ids.size(); i++)
+        _ids.getMat().ptr< int >(0)[i] = ids[i];
```
prasannavk: Why don't you build ```_ids``` matrix when you check ```validCandidates``` avoid the temporary ```ids``` vector?

modules/aruco/src/aruco.cpp
```diff
@@ -0,0 +1,1573 @@
...
+                    break;
+                }
+            }
+            if(inside) {
+                toRemove[j] = true;
+                atLeastOneRemove = true;
+                continue;
+            }
+
+            // check the second marker
```
prasannavk: code repetition. make a function.

prasannavk: you need not check both actually I think. Depends on the speed of pointPolygonTest. I would assume it is more expensive than calculating pixel distances. You could check the diagonal distances perhaps to see which is higher and just check once. Just a thought.

modules/aruco/src/aruco.cpp
```diff
@@ -0,0 +1,1573 @@
...
+        for(unsigned int i = 0; i < toRemove.size(); i++) {
+            if(!toRemove[i]) {
+                filteredCorners.push_back(_inCorners.getMat(i).clone());
+                filteredIds.push_back(_inIds.getMat().ptr< int >(0)[i]);
+            }
+        }
+
+        _outIds.create((int)filteredIds.size(), 1, CV_32SC1);
+        for(unsigned int i = 0; i < filteredIds.size(); i++)
+            _outIds.getMat().ptr< int >(0)[i] = filteredIds[i];
```
prasannavk: There is a way to set a Mat to a vectors memory. It s part of the Mat constructor. You wouldn't have to iterate.

modules/aruco/samples/detect_markers.cpp
sergarrido: The cast of atof is to avoid the compilation warning (conversion to 'float' from 'double').

The units of markerLength do not have to be meters (although people usually use them because they are international units). 
For instance you can perfectly provide centimeters or whatever unit you want. The main difference would be that the estimated pose of the marker/board will be in the same unit (more concretely the Tvec array will be in the same unit).

modules/aruco/src/aruco.cpp
sergarrido: I have use the _variable nomenclature since I have seen it in other OpenCV function. For instance solvePnP function:
https://github.com/Itseez/opencv/blob/master/modules/calib3d/src/solvepnp.cpp#L55
The purpose is to relate the proxy variables (InputArray, OutputArray, ...) to their respective variables with the contained data (Mat, vector<Mat>, ...)

The candidates are floating points because it is necessary for subpixel refinement. It is true that in this stage it is maybe too early and integer values could be used until corner refinement is performed.

modules/aruco/samples/detect_markers.cpp
prasannavk: I don't think it is a good idea to leave it open to any units because the consequence of the Tvec array being in the same units is hidden. This might create confusion. I would prefer choosing meters and mentioning that in help and in documentation. That way there is clarity.

modules/aruco/src/aruco.cpp
prasannavk: I see. Because the candidates are outputs from your function I would say enforce that they are integers post the corner subpixel. Also, you may want to assert types of the matrix values you get if float or integer type matrices are assumptions. This allows for an early crash if someone is trying to use this function.

modules/aruco/src/aruco.cpp
```diff
@@ -0,0 +1,1573 @@
...
+    DetectorParameters *params;
+};
+
+
+/**
+ * @brief Initial steps on finding square candidates
+ */
+static void _detectInitialCandidates(const Mat &grey, vector< vector< Point2f > > &candidates,
+                                     vector< vector< Point > > &contours,
+                                     DetectorParameters params) {
```
prasannavk: ```params``` should be const reference.

modules/aruco/src/aruco.cpp
```diff
@@ -0,0 +1,1573 @@
...
+            _rejected.create(4, 1, CV_32FC2, i, true);
+            Mat m = _rejected.getMat(i);
+            rejected[i].copyTo(m);
+        }
+    }
+}
+
+
+/**
+  * @brief Final filter of markers after its identification
```
prasannavk: The comment should have why the filtering is done too. This is an obvious comment. Does not add new information.

modules/aruco/include/opencv2/aruco.hpp
```diff
@@ -0,0 +1,502 @@
...
+    // array of object points of all the marker corners in the board
+    // each marker include its 4 corners, i.e. for M markers, the size is Mx4
+    std::vector< std::vector< Point3f > > objPoints;
+
+    // the dictionary of markers employed for this board
+    Dictionary dictionary;
+
+    // vector of the identifiers of the markers in the board (same size than objPoints)
+    // The identifiers refers to the board dictionary
+    std::vector< int > ids;
```
prasannavk: I really like the simple design of Board and related data. But, I think Board itself could be an ```std::map``` or ```std::unordered_map``` with key being id and value being ```objPoints[index_corresponding_to_key]``` to allow for fast query of 3D objectPoints. Did you have a particular reason in terms of memory efficiency to choose vectors? We could measure how memory performance is when Charuco runs.

modules/aruco/src/dictionary.cpp
```diff
@@ -0,0 +1,473 @@
...
+
+
+/**
+ */
+bool Dictionary::identify(const Mat &onlyBits, int &idx, int &rotation,
+                          double maxCorrectionRate) const {
+
+    CV_Assert(onlyBits.rows == markerSize && onlyBits.cols == markerSize);
+
+    int maxCorrectionRecalculed = int(double(maxCorrectionBits) * maxCorrectionRate);
```
prasannavk: *```maxCorrectionRecalculated```

modules/aruco/include/opencv2/aruco/dictionary.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+
+/**
+ * @brief Dictionary/Set of markers. It contains the inner codification
+ *
+ */
+class CV_EXPORTS Dictionary {
+
+    public:
+    Mat bytesList;         // marker code information
```
prasannavk: Isnt there a convention to indicate what members are in a class? It is probably easier to read if you have something like ```bytesList_``` for members, where there is an underscore for members or ```m_bytesList```. Not sure what opencv recommends as the http://code.opencv.org/projects/opencv/wiki/CodingStyleGuide does not specifically talk about it. Probably worth checking one of the style guide they reference there.

modules/aruco/src/dictionary.cpp
```diff
@@ -0,0 +1,473 @@
...
+            for(int k = 0; k < 4; k++)
+                bytesList.at< Vec4b >(i, j)[k] = bytes[i * (4 * nbytes) + k * nbytes + j];
+        }
+    }
+}
+
+
+
+/**
+ */
```
prasannavk: Need comments here. Please explain why do it in this particular way.

modules/aruco/src/dictionary.cpp
```diff
@@ -0,0 +1,473 @@
...
+    int nbytes = (bits.cols * bits.rows) / 8;
+    if((bits.cols * bits.rows) % 8 != 0) nbytes++;
+
+    Mat candidateByteList(1, nbytes, CV_8UC4, Scalar::all(0));
+    unsigned char currentBit = 0;
+    int currentByte = 0;
+    for(int row = 0; row < bits.rows; row++) {
+        for(int col = 0; col < bits.cols; col++) {
+            // circular shift
+            candidateByteList.ptr< Vec4b >(0)[currentByte][0] =
```
prasannavk: Minor optimization: You could create a temp variable/reference to ```candidateByteList.ptr< Vec4b >(0)[currentByte]``` and manipulate it inside the loop. Here you are accessing the same thing again. The processor caching would definitely help you in the current state. But this way it s easy to read as well.

modules/aruco/src/dictionary.cpp
```diff
@@ -0,0 +1,473 @@
...
+
+    // resize tiny marker to output size
+    cv::resize(tinyMarker, _img.getMat(), _img.getMat().size(), 0, 0, INTER_NEAREST);
+}
+
+
+
+
+/**
+  * @brief Transform matrix of bits to list of bytes in the 4 rotations
```
prasannavk: What is happening here? Please give a simple example in the comments.

modules/aruco/include/opencv2/aruco/dictionary.hpp
sergarrido: As you have said, this is not mentioned in the OpenCV Coding Style. Thus I have followed the same convention I have seen in other parts of the library.
The convention is not full clear, but in general it depends on whether the member variable is ```public``` or ```private```.
Private member variables sometimes include an underscore at the beginning ( ```_variable``` ), while public member variables do not have any mark.

There are several samples here ( classes ```Mat```, ```UMat```, ```SparseMat```, ```RNG```, ```SVD```, etc ):
https://github.com/Itseez/opencv/blob/master/modules/core/include/opencv2/core/mat.hpp
https://github.com/Itseez/opencv/blob/master/modules/core/include/opencv2/core.hpp

In this case, since variables are public I have not put any mark. For instance, in the ```CharucoBoard``` class, private variables have an underscore at the beginning.

Many style guides encourage not to use public variables inside classes, so maybe a better approach would be setting these variables as ```private``` (with their corresponding underscores) and creating the corresponding getters and setters.

PR #348, opened 2015-08-25
Contrib tests

mshabunin:
fixing tests for contrib modules
- test data moved to opencv_extra
- some paths fixed
- some tests disabled (DAISY)


modules/xfeatures2d/src/daisy.cpp
```diff
@@ -1156,7 +1156,7 @@ struct ComputeHistogramsInvoker : ParallelLoopBody
                Rect( 0, 0, layers->at(r).size[2]-1, layers->at(r).size[1]-1 ) )
              ) continue;
 
-          float* hist = layers->at(r).ptr<float>(0,y*_hist_th_q_no,x*_hist_th_q_no);
+          float* hist = layers->at(r).ptr<float>(0,y,x);
```
mshabunin: @cbalint13, can you please take a look at these changes (in daisy.cpp). I had to fix the coordinates to avoid assertions in debug mode. After fix some tests had failed (RotationInvariance_Descriptor, ScaleInvariance_Descriptor).

vpisarev:
thank you! :+1:

PR #354, opened 2015-08-28
remove viz/vtk dependancy

berak:
currently the structured_light module does not build without viz/vtk, 
but actually only the pointcloud sample depends on viz

### DIFF
--- a/modules/structured_light/samples/pointcloud.cpp
+++ b/modules/structured_light/samples/pointcloud.cpp
@@ -45,7 +45,12 @@
 #include <opencv2/calib3d.hpp>
 #include <opencv2/imgproc.hpp>
 #include <opencv2/structured_light.hpp>
+#include <opencv2/opencv_modules.hpp>
+
+// (if you did not build the opencv_viz module, you will only see the disparity images)
+#ifdef HAVE_OPENCV_VIZ
 #include <opencv2/viz.hpp>
+#endif
 
 using namespace std;
 using namespace cv;
@@ -73,13 +78,13 @@ static bool readStringList( const string& filename, vector<string>& l )
   if( !fs.isOpened() )
   {
     cerr << "failed to open " << filename << endl;
-    return -1;
+    return false;
   }
   FileNode n = fs.getFirstTopLevelNode();
   if( n.type() != FileNode::SEQ )
   {
     cerr << "cam 1 images are not a sequence! FAIL" << endl;
-    return -1;
+    return false;
   }
 
   FileNodeIterator it = n.begin(), it_end = n.end();
@@ -92,7 +97,7 @@ static bool readStringList( const string& filename, vector<string>& l )
   if( n.type() != FileNode::SEQ )
   {
     cerr << "cam 2 images are not a sequence! FAIL" << endl;
-    return -1;
+    return false;
   }
 
   it = n.begin(), it_end = n.end();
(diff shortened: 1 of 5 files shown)


berak:
i'll have to find out, how the pointcloud example is actually built (there's no CmakeLists.txt in the samples folder)

mshabunin:
@berak, all files from `<module>/samples` folder are added by [this](https://github.com/Itseez/opencv/blob/784f77a4a5669101c8c773611080d8837ae068a4/cmake/OpenCVModule.cmake#L1073) cmake function. Maybe it could be called with additional dependencies for all samples: `ocv_add_samples(<extra dependencies>)`

berak:
ah, thanks, i'll try that.

modules/structured_light/samples/pointcloud.cpp
```diff
@@ -45,7 +45,14 @@
 #include <opencv2/calib3d.hpp>
 #include <opencv2/imgproc.hpp>
 #include <opencv2/structured_light.hpp>
+
+// if you have build the opencv_viz module,
+// uncomment the next line (else you will only see the disparity images)
+#define HAVE_OPENCV_VIZ
```
alalek: Use `#include "opencv-modules.hpp"`

berak:
ahh, sure, thanks !

berak:
looks like "opencv_modules.hpp" is not available for contrib modules ?

`/build-2/precommit-contrib_macosx/opencv_contrib/modules/structured_light/samples/pointcloud.cpp:48:10: fatal error: 'opencv_modules.hpp' file not found
#include "opencv_modules.hpp"`

berak:
In file included from /build-3/precommit-contrib_macosx/opencv_contrib/modules/structured_light/samples/pointcloud.cpp:43:
In file included from /build-3/precommit-contrib_macosx/opencv/modules/core/include/opencv2/core.hpp:3108:
/build-3/precommit-contrib_macosx/opencv/modules/core/include/opencv2/core/utility.hpp:726:55: error: no member named 'type' in 'cv::ParamType<unsigned long>'
        getByIndex(index, space_delete, ParamType<T>::type, (void*)&val);
                                                      ^
/build-3/precommit-contrib_macosx/opencv_contrib/modules/structured_light/samples/pointcloud.cpp:141:27: note: in instantiation of function template specialization 'cv::CommandLineParser::get<unsigned long>' requested here
    white_thresh = parser.get<size_t>( 4 );
                          ^

this probably means it needs another specialization for 'unsigned long [here](https://github.com/Itseez/opencv/blob/master/modules/core/include/opencv2/core.hpp#L3078) to make clang happy ?

berak:
should i fix the winx64 warnings as well (while i'm at it..) ? 
(easy, just not sure if it should go into this commit)

mshabunin:
@berak, thank you for contibution! Yes, can you please fix them? We can not merge PR which introduces (or reveals) build warnings.

berak:
@mshabunin , just to report, i tried adding `SAMPLE_DEPS opencv_viz` to ocv_define_module(), then splitting it off the $ARGN inside, and pass it to ocv_add_samples(), 

but then we end up [here](https://github.com/Itseez/opencv/blob/784f77a4a5669101c8c773611080d8837ae068a4/cmake/OpenCVModule.cmake#L1081) and still *all* samples are dependant in the same way, so *none* will get built without that dependancy.

the conditional is clumsy, but at least module and all samples will build (the pointcloud one just won't show the viz without vtk)

alalek:
:+1:

PR #356, opened 2015-08-28
aruco: the nScales == 1 is already handled inside parallel_for_

paroj:


sergarrido:
This is right, however it is done this way in purpose.
The use of parallel_for for a single scale reduces the performance, so I decided to handle this case separately without using paralel_for

paroj:
this is strange, as nothing in the code would suggest a slowdown. Which compiler do you use? And do you have a performance test for this?

paroj:
I performed a performance evaluation on my machine:
* i7-4770K
* 64bit Ubuntu 14.04
* Release Build

there was no measurable performance difference to one another. As `parallel_for` is the natural place for this kind of optimizations I would suggest merging my changes.

sergarrido:
Finally I have found some time to get a look at this.
I have done a thorough test and you are right. I am having an average difference of only 0.2 ms on my machine.
So yes, I also suggest merging this PR.

alalek:
:+1:

PR #358, opened 2015-08-31
Parallelize edges detection

DirtyMaster:
On large images, edge detection can be rather long. I propose to parallelize the two for loops, which can drastically speed up this routine.


vpisarev:
:+1:

PR #359, opened 2015-09-02
face::FaceRecognizer now returns a map associating labels with their confidence

Cylix:
I'm currently working on a face recognition software based on OpenCV face recognition engine.

However, openCV face recognition library `predict` functions return only the label with the minimum distance to the unknown face. There is no way to get the list of all the labels with the result of the prediction (distance) for each of these labels.

I've added this feature that can be quite useful (it is in my case) by changing the prototype of `void predict(InputArray src, CV_OUT int &label, CV_OUT double &confidence) const` to `std::map<int, double> predict(InputArray src, CV_OUT int &label, CV_OUT double &confidence) const`.

Predict function now return an std::map associating a label to its distance to the face on which we are making prediction.


berak:
unfiortunately now, this breaks the python bindings, because there is no pyopencv_from/pyopencv_to for map<int,double> in cv2.cpp

Cylix:
Ok, so what is the best option? Adding the binding or changing the container used to return the results?

In the second case, what are the containers with a binding (for example, can we replace `std::map` by `std::list<std::pair<int, double>>` or `std::vector<std::pair<int, double>>`);

berak:
hmm, using something, which is already supported from cv2.cpp ofc. is the most easy option, like returning a `vector<Vec2f>` , but it will make the internal handling very clumsy (since you have to access/search for the label quite often there). 

no real good idea atm. here.

berak:
btw, you could overload it instead of changing the existing signature, and remove the CV_WRAP .

it yould not be available in python/java/matlab, but at least the c++ functionality would compile.

comdiv:
Have alternation look on this problem with same tasks: https://github.com/Itseez/opencv_contrib/issues/401 
with implementation in #402 

The only difference that i sugguset to use "collector delegate" in form "void(int,double)" that caller side can supply to "predict" - this notation cover all scenarios (current "best" and your "all") but allow use statistical( min-max, avg) and relevance (top 10 and so on) without need to create full list of results.

Good luck.

mshabunin:
Another implementation has been accepted: #402

PR #360, opened 2015-09-03
Fix contrib tests

mshabunin:
- removed perf sanity check for some tests
- reduced test sets


mshabunin:
Some test data fixed in Itseez/opencv_extra#273

alalek:
:+1:

PR #361, opened 2015-09-03
Graphsegmentation: First implementation

the-glu:
This is a proposition of an implementation of a very fast segmentation algorithm, described in [Efficient graph-based image segmentation](http://cs.brown.edu/~pff/papers/seg-ijcv.pdf).

This is my first opencv module - I hope everything is correct :)

### DIFF
--- a/modules/ximgproc/README.md
+++ b/modules/ximgproc/README.md
@@ -7,3 +7,4 @@ Extended Image Processing
 4. Adaptive Manifold Filter
 5. Joint Bilateral Filter
 6. Superpixels
+7. Graph segmentation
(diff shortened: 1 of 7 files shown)


vpisarev:
thank you for the contribution! could you please put it to ximgproc instead of creating a separate module?

cbalint13:
@the-glu , very nice work ! I am also interested getting your PR passed.

* Also could merge into single .cpp file inside /ximgproc/src ? I mean merge your actual graphsegmentation.cpp & utils.cpp . A good example how to expose things would be existing SEEDS's api (see seeds.cpp & seeds.hpp). I also follow SEEDS example/standard in my open PRs for SLIC and  LSC superpixels, so would be great if this one follows too. I would be glad to help out, feel free ping me if need help.

the-glu:
Okay, I updated the PR with code inside ximgproc and merged everything in a single file!

Others things to fix ? :)

vpisarev:
thank you! :+1:

PR #364, opened 2015-09-06
Fixing uninitialized memory access problem in DeepFlow

sbokov:
On my system DeepFlow sometimes produced invalid results. Proposed change seems to fix the problem.


vpisarev:
thanks! :+1:

PR #369, opened 2015-09-09
FHOG class added

joaofaro:


StevenPuttemans:
Could you maybe point out the difference between the FHOG and the HOG implementation?

joaofaro:
Yes, I can update the docs to explain the difference. The idea was to have a simple and easy-to-use FHOG extractor but apparently the this can be done with dpm module (dpm_feature). Thanks for pointing that, I have never used this module.

StevenPuttemans:
Then I am afraid that your PR will be closed down, since the functionality is already provided.
@mshabunin what do you think?

mshabunin:
@joaofaro, can you please check if same functionality could be exposed from _dpm_ module (https://github.com/Itseez/opencv_contrib/blob/master/modules/dpm/src/dpm_feature.hpp)?

vpisarev:
it looks like the author confirmed that this is already implemented

PR #372, opened 2015-09-13
OCL: fix "off" mode

alalek:
Main repo: https://github.com/Itseez/opencv/pull/5340


vpisarev:
:+1:

lenlen:
@alalek this macro if/else is missing in this file starting from: https://github.com/Itseez/opencv_contrib/blob/master/modules/tracking/src/multiTracker.cpp#L218

PR #377, opened 2015-09-15
Edge boxes

ultrafro:
This is an implementation of the EdgeBoxes Saliency algorithm. It is not fully working yet, however I would appreciate commentary. Please see: modules/saliency/README.md

The purpose of this module is to create, group and make available to the users, different saliency algorithms, belonging to different categories.

The EdgeBoxes algorithm (Dollar et al, 2014) utilizes edge information in an image to propose a set of bounding boxes which enclose objects. The algorithm searches for windows with fully-enclosed edge-groups, and penalizes windows with edge-groups which cross the boundary of the box.
This is an initial implementation in OpenCV


modules/saliency/include/opencv2/saliency/saliencySpecializedClasses.hpp
```diff
@@ -405,6 +405,173 @@ class CV_EXPORTS ObjectnessBING : public Objectness
...
+ * [4] Zitnick, C. Lawrence, and Piotr Dollár. "Edge boxes: Locating object proposals from edges." Computer Vision–ECCV 2014
+ */
+
+/** @brief the EdgeBoxes algorithm from @cite EdgeBoxes
+ */
+class CV_EXPORTS ObjectnessEdgeBoxes : public Objectness
+{
+public:
+
+	const int EDGE_IMAGE = 0;
```
mshabunin: @ultrafro, we are testing OpenCV without C++11 support and these constructions does not work in older standard. It seems that these constants are not used anywhere, so the can be deleted.

modules/saliency/include/opencv2/saliency/saliencySpecializedClasses.hpp
```diff
@@ -405,6 +405,173 @@ class CV_EXPORTS ObjectnessBING : public Objectness
...
+    @param objectnessBoundingBox objectness Bounding Box vector. According to the result given by this
+    specialized algorithm, the objectnessBoundingBox is a *vector\<Vec4i\>*. Each bounding box is
+    represented by a *Vec4i* for (minX, minY, maxX, maxY).
+     */
+  bool computeSaliencyImpl( InputArray image, OutputArray objectnessBoundingBox );
+
+private:
+
+
+	bool _bScoresReady = false;
```
mshabunin: This variable should be initialized in class constructor to conform to C++03 standard.

modules/saliency/include/opencv2/saliency/saliencySpecializedClasses.hpp
mshabunin: If you plan using them in future, you can add these constants via enum like this:
```.cpp
class ObjectnessEdgeBoxes
{
public:
    enum { EDGE_IMAGE = 0, ORIGINAL_IMAGE = 1 }
...
```

futurely:
https://github.com/samarth-robo/edges

cjnolet:
Any chance I could get an update on where this implementation is? I don't mind helping to finish this up because I separately need it.

sovrasov:
Already implemented in #1215

PR #384, opened 2015-09-18
aruco: simplify Dictionary by not interleaving the marker rotations

paroj:
store marker internally as 
- `r0_b0, r0_b1, .. , r0_bn, r1_b0, ..` instead of 
- `r0_b0, r1_b0, .. , rn_b0, r1_b0, ..`

i.e. the same as in predefined_dictionaries.cpp.
This makes loading a dictionary a simple memcpy and also allows to get
rid of the custom hammingDistance implementation in favor of
`hal::normHamming`.

### DIFF
--- a/modules/aruco/src/dictionary.cpp
+++ b/modules/aruco/src/dictionary.cpp
@@ -48,24 +48,6 @@ namespace aruco {
 
 using namespace std;
 
-
-
-/**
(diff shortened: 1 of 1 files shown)


sergarrido:
Nice :+1:

alalek:
:+1:

PR #385, opened 2015-09-19
Pull Request 4219 from main opencv repo

Dikay900:
see https://github.com/Itseez/opencv/pull/4219


alalek:
:+1:

PhilLab:
Just out of pure curiosity: why do we have two FREAK implementations?

PR #386, opened 2015-09-21
aruco: create dictionary from Mat instead of raw pointer

paroj:
also avoids manual memcpy: the predefined dictionary data is not
duplicated any more.


alalek:
@sergarrido Could you please take a look on these changes?

sergarrido:
great job :+1:

alalek:
:+1:

PR #388, opened 2015-09-26
Draw keylines in random color by default as documentation depicts

Milania1:
Fix bug that color is not chosen randomly when no color argument is supplied (default) as the documentation depicts and also is the case in drawLineMatches.


alalek:
:+1:

Issue #383, opened 2015-09-18
Segfault in dnn module


inferrna:
Just import cv2 from python console and crash happened.

```
~$ gdb python3
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from python3...Reading symbols from /usr/lib/debug//usr/bin/python3.4m...done.
done.
(gdb) run
Starting program: /usr/bin/python3 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Python 3.4.3 (default, Jul 28 2015, 18:20:59) 
[GCC 4.8.4] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import cv2
[New Thread 0x7fffd0991700 (LWP 4811)]

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff21b1ddb in cv::dnn::LayerFactory::registerLayer(cv::String const&, cv::Ptr<cv::dnn::Layer> (*)(cv::dnn::LayerParams&)) ()
   from /usr/local/lib/libopencv_dnn.so.3.0
(gdb) bt
#0  0x00007ffff21b1ddb in cv::dnn::LayerFactory::registerLayer(cv::String const&, cv::Ptr<cv::dnn::Layer> (*)(cv::dnn::LayerParams&)) ()
   from /usr/local/lib/libopencv_dnn.so.3.0
#1  0x00007ffff21aa350 in cv::dnn::initModule() () from /usr/local/lib/libopencv_dnn.so.3.0
#2  0x00007ffff7dea13a in call_init (l=<optimized out>, argc=argc@entry=1, argv=argv@entry=0x7fffffffd928, env=env@entry=0x7fffffffd938) at dl-init.c:78
#3  0x00007ffff7dea223 in call_init (env=<optimized out>, argv=<optimized out>, argc=<optimized out>, l=<optimized out>) at dl-init.c:36
#4  _dl_init (main_map=main_map@entry=0xaa48c0, argc=1, argv=0x7fffffffd928, env=0x7fffffffd938) at dl-init.c:126
#5  0x00007ffff7deec70 in dl_open_worker (a=a@entry=0x7fffffffbf18) at dl-open.c:577
```

inferrna:
Additional info.
OS: Ubuntu 14.04

**gcc -v**
```
Using built-in specs.
COLLECT_GCC=/usr/bin/gcc
[...]

PR #391, opened 2015-10-04
opencv_dnn: replaced static LayerFactory::impl member with a static function that constructs the object on the first use (fix for bug #383)

rokm:
This way, we are certain not to depend on the order of static
variable initialization (LayerFactory::impl in dnn.cpp, vs. init
in init.cpp), which is causing bug #383.


alalek:
:+1:

PR #397, opened 2015-10-08
aruco: use densely packed array for pose computation

paroj:
allows using Mat and vector<Vec3d> as argument for rvecs, tvecs that are
allocated at once.


sergarrido:
The sample code inside the tutorials 'aruco_detection' and 'charuco_diamond_detection' should also be updated.
And I would also change the board samples (and tutorials) to use ```Vec3d``` for ```rvec``` and ```tvec```, just for consistency.

paroj:
done

sergarrido:
:+1:

mshabunin:
:+1:

Issue #401, opened 2015-10-14
Custom handling of FaceRecognizer results


comdiv:
For now FaceRecognizer allows only one type of prediction result handling - "max confedence". This type of result handling is not all that needed from recognition.

We have many cases where we must to check:
1) Average confidence over trained set (for example we try to decide women/men, child/adult) so we haven't exactly same faces, but average confidence over groups can give us answer
2) Top N "most similar" - in scenario where we have photos with bad quality (from secure cams and so on) and trained base with 1-2 image per person - so it's frequently that "lightly similar" faces are faces of same persons
3) Sharding of training set with following aggregation over shard results - while shard can be not well balanced their confidences can be not directly comparable - so we require use some logic over raw per-shard result

Suggestion:
1. Allow caller side to supply own result handling delegate with addition method of FaceRecognizer
2. It must work same way as usual "predict" but without checking of best result

Implementation:
I have implement it myself in 
https://github.com/comdiv/opencv_contrib/commit/a3ca6d7bddc96530370e882cc842490c4c7aab9d
and it has correspond pull request #402 

It's very simple implementation:
1) std::function used as delegate base class (so it allows to use lambdas and method references clear way)
2) predict_collect method added for FaceRecognizer and all it's descendants - I thought to use existed "predict" name and overload it, but while it's forked code - decide that better use special name
3) all implementations of FaceRecognizer implements this method - it's just copy of existed "predict" - but it send all result to delegate - alternately, while collect sygnature is more generic than exited predict it was able to make predict_collect method main and predict as overload to it, but as in (2) it could be not much friendly in contributed code

Conclusion:
1) Think that predict->(src, collector) notation is very usefull in many scenarios
2) It's simply to implement and use and doesn't require breaking changes to API
3) If there are some suggestions I can fix my code and remake pull request

berak:
there is a related pr [here](https://github.com/Itseez/opencv_contrib/pull/359) , (also related problems)

comdiv:
Related problem can lead to large memory consumption. Think that collector pattern is safer and ease to use in aggregation and map-reduce scenarios.

berak:
in general , do we **want** c++11 dependancy here ? older compilers won't be able to build this module any more.

comdiv:
Heh, haven't check it out. 
Think best solution is:
1) Add overloaded sygnature with (void*)(int,double) instead of std::funciton - template may be?
[...]

PR #402, opened 2015-10-14
Custom handling of FaceRecognizer results

comdiv:
It's code that can close https://github.com/Itseez/opencv_contrib/issues/401 or can be used as suggestion to close this ticket

### DIFF
--- a/modules/face/include/opencv2/face.hpp
+++ b/modules/face/include/opencv2/face.hpp
@@ -48,6 +48,7 @@ the use of this software, even if advised of the possibility of such damage.
 */
 
 #include "opencv2/core.hpp"
+#include "face/predict_collector.hpp"
 #include <map>
 
 namespace cv { namespace face {
@@ -255,7 +256,8 @@ public:
     CV_WRAP virtual void update(InputArrayOfArrays src, InputArray labels);
 
     /** @overload */
-    virtual int predict(InputArray src) const = 0;
+    CV_WRAP int predict(InputArray src) const;
+
 
     /** @brief Predicts a label and associated confidence (e.g. distance) for a given input image.
 
@@ -292,7 +294,18 @@ public:
     model->predict(img, predicted_label, predicted_confidence);
     @endcode
      */
-    CV_WRAP virtual void predict(InputArray src, CV_OUT int &label, CV_OUT double &confidence) const = 0;
+    CV_WRAP void predict(InputArray src, CV_OUT int &label, CV_OUT double &confidence) const;
+
+
+    /** @brief - if implemented - send all result of prediction to collector that can be used for somehow custom result handling
+    @param src Sample image to get a prediction from.
+    @param collector User-defined collector object that accepts all results
+    @param state - optional user-defined state token that should be passed back from FaceRecognizer implementation
+
+    To implement this method u just have to do same internal cycle as in predict(InputArray src, CV_OUT int &label, CV_OUT double &confidence) but
+    not try to get "best@ result, just resend it to caller side with given collector
+    */
+    CV_WRAP virtual void predict(InputArray src, Ptr<PredictCollector> collector, const int state = 0) const = 0;
 
     /** @brief Saves a FaceRecognizer and its model state.
 
@@ -345,7 +358,8 @@ public:
     info.
      */
     CV_WRAP virtual std::vector<int> getLabelsByString(const String& str) const;
-
+    /** @brief threshhold parameter accessor - required for default BestMinDist collector */
+    virtual double getThreshold() const = 0;
 protected:
     // Stored pairs "label id - string info"
     std::map<int, String> _labelsInfo;
--- a/modules/face/src/predict_collector.cpp
+++ b/modules/face/src/predict_collector.cpp
@@ -1,0 +1,86 @@
+/*
+By downloading, copying, installing or using the software you agree to this license.
+If you do not agree to this license, do not download, install,
+copy or use the software.
+
+
+                          License Agreement
+               For Open Source Computer Vision Library
+                       (3-clause BSD License)
+
+Copyright (C) 2000-2015, Intel Corporation, all rights reserved.
+Copyright (C) 2009-2011, Willow Garage Inc., all rights reserved.
+Copyright (C) 2009-2015, NVIDIA Corporation, all rights reserved.
+Copyright (C) 2010-2013, Advanced Micro Devices, Inc., all rights reserved.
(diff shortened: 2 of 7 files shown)


modules/face/include/opencv2/face.hpp
```diff
@@ -257,6 +266,15 @@ class CV_EXPORTS_W FaceRecognizer : public Algorithm
...
     virtual int predict(InputArray src) const = 0;
 
+	/** @brief - if implemented - send all result of prediction to collector that can be used for somehow custom result handling
+	@param src Sample image to get a prediction from.
+	@param collector User-defined collector function that is acceptor for all prediction-checking results
+
+	To implement this method u just have to do same internal cycle as in predict(InputArray src, CV_OUT int &label, CV_OUT double &confidence) but
+	not try to get "best@ result, just resend it to caller side with given collector
+	*/
+	CV_WRAP virtual void predict_collect(InputArray src, PredictCollector collector) const = 0;
```
berak: since the automated scripting generator (for python java,etc)  can't handle callbacks (save c++11 things), this might have to go without the CV_WRAP tag

modules/face/include/opencv2/face/facerec.hpp
```diff
@@ -123,6 +123,7 @@ class CV_EXPORTS_W LBPHFaceRecognizer : public FaceRecognizer
     CV_WRAP virtual void setThreshold(double val) = 0;
     CV_WRAP virtual std::vector<cv::Mat> getHistograms() const = 0;
     CV_WRAP virtual cv::Mat getLabels() const = 0;
+
```
berak: please remove trailing whitespace (docbot complains)

modules/face/include/opencv2/face.hpp
comdiv: OK will fix it

comdiv:
suggestions fixed

berak:
the linux/android builds will need enabling c++11, similar to [this](https://github.com/Itseez/opencv_contrib/blob/master/modules/cvv/CMakeLists.txt#L7)

mshabunin:
@comdiv, why didn't you use object as a handler? Something similar to:
```.cpp
struct Collector {
    virtual bool collectOne(int, double, int) = 0;
}
...
virtual void predict_collect(InputArray src, Collector & collector, const int state = 0) const = 0;
```

I think it is not a good idea to have different interfaces for different C++ standards. If one have library compiled with C++03 compiler than all applications using this library will be restricted to use same standard, otherwise linking errors will occur.

comdiv:
I think that really better is to use Iterator patter - 
```
PredictItertator FaceRecognizer.open(...) ;
PredictRecord PredictIterator::next();
void PredictIterator::close();
struct PredictRecord  { int label, double dist } ; // or {int* label, double* dist}
```
As it used to be in .NET world - so client side decide itself how to deal with that.

While may be Collector class should be more usefull while we can predefine MaxCollector, AvgCollector, FirstByTreshholdCollector, Top10Collector and so on with ready to use aggregations.

Such set of classes could be Templated and be capable work with different collect_one notations.

The only thing is not good - while i'm using (bool*) and std::funciton it's just overload of predict, expressed in other C++ standard things. When we add Collector class family we grow entropy of opencv + opencv_contrib design. Where it could be defined? In face? In shared code for contrib? In opencv? Somwhere else?

comdiv:
Think I'll rewrite on collector as class design and replace existed master overload of predict (while it's more generic).
The only implemented in core type of Collector will be MostCloseCollector that will be used internally by FaceRecognizer.
On public API level only abstract Collector will be provided.
So any brainf**ks about C++11 will be fixed.

comdiv:
Have totally rewrite facerecognizer internals to use with PredicCollector class,
not breaking change - older API still in game
no c++ 11 dependency or other conditional compilation

comdiv:
require suggestion - how to make it Pythonable?
When I make CV_WRAP onto method - it error that PredicCollector is unknown.
How to make it well?

mshabunin:
Try to change to `Ptr<PredictCollector>`:
```.cpp
CV_WRAP virtual void predict(InputArray src, Ptr<PredictCollector> collector, const int state = 0) const = 0;
 ```
Also, users will need to create instances of PredictCollector in python, for this factory method should be implemented (static class method or separate function):
```.cpp
Ptr<MinDistancePredictCollector> MinDistancePredictCollector::create() {}
```

comdiv:
All green but builds was renamed so warning in github still occures

mshabunin:
@comdiv, please replace tabs with 4 spaces.

comdiv:
done

comdiv:
Is it usefull to add some custom PredictCollector implementations?
For example i have TopNPredictCollector.

mshabunin:
@comdiv , please revert all unrelated changes e.g. code formatting.

comdiv:
Have fixed

modules/face/src/predict_collector.cpp
```diff
@@ -0,0 +1,60 @@
...
+*     notice, this list of conditions and the following disclaimer.
+*   * Redistributions in binary form must reproduce the above copyright
+*     notice, this list of conditions and the following disclaimer in the
+*     documentation and/or other materials provided with the distribution.
+*   * Neither the name of the organization nor the names of its contributors
+*     may be used to endorse or promote products derived from this software
+*     without specific prior written permission.
+*
+*   See <http://www.opensource.org/licenses/bsd-license>
+*/
```
mshabunin: I think this license is wrong. Please replace it with text from [the main license](https://github.com/Itseez/opencv/blob/master/LICENSE).

mshabunin:
I have no other comments.

comdiv:
yes, copyed license from somewhere in opencv or opencv-contrib, will place your one

comdiv:
Have applyed valid license.
Another question - i can add some PredictCollector subclasses that can be usefull - is it pleased?

mshabunin:
@comdiv , yes it could be useful. I think it is better to create new PR for this. Maybe we could also implement a simple example of using new interface.

:+1:

comdiv:
What is better?
1) New ISSUE + new PR
2) Existed ISSUE + new PR
3) Continue this pair of ISSUE+PR

mshabunin:
Just create a new PR with the proposed functionality, new issue is not necessary.

comdiv:
Have question - can std::list be used with CV_WRAP , if not what is common type and conversion for such return values.

PR #405, opened 2015-10-19
Added getters to the points of the KeyLine class

Milania1:
Useful since it is often needed to have the points from the keyline directly in the cv::Point2f format.


mshabunin:
:+1:

PR #412, opened 2015-10-25
Fix comple error with opencv 3.0 on Nvidia Jetson

cpsandbox:
This re-creates a fix for a bug reported here: https://github.com/CVVisualPSETeam/CVVisual/issues/46. 

The original fix is show here: https://github.com/CVVisualPSETeam/CVVisual/commit/f7b1c68cb693167c6fd8656e91fab3d1da07d1bb

This fix applies the replacement of 10,50lu with std::size_t{10},  std::size_t{50} in the three files.


alalek:
Please fix commits hell: 1 commit is enough for this patch.
Please use more traditional format for type casts without curly brackets: `(std::size_t)50`
Remove unnecessary `+`.

PR #417, opened 2015-10-31
Adding a fast and parallel function for computing the estimated covar…

ogreen:
…iance matrix

### DIFF
--- a/modules/ximgproc/src/estimated_covariance.cpp
+++ b/modules/ximgproc/src/estimated_covariance.cpp
@@ -103,8 +103,7 @@ private:
     int pc;
     int threads;
 
-//    Combination* combinationsTable;
-    std::vector<Combination> combinationsTable;    
+    std::vector<Combination> combinationsTable;
 };
 
 
--- a/modules/ximgproc/src/estimated_covariance.cpp
+++ b/modules/ximgproc/src/estimated_covariance.cpp
@@ -185,7 +185,7 @@ void EstimateCovariance::iterateCombinations(Mat inputData,Mat outputData)
 
 #ifdef HAVE_OPENMP
     #pragma omp parallel
-#endif  
+#endif
     {
         int idx;
         int combs;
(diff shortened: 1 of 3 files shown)


modules/ximgproc/src/estimated_covariance.cpp
```diff
@@ -0,0 +1,415 @@
...
+            std::complex<float>& dst){dst = a*b;}
+
+private:
+    int nr;
+    int nc;
+    int pr;
+    int pc;
+    int threads;
+
+    Combination* combinationsTable;
```
vpisarev: please, use std::vector<Combination> instead of manual memory allocation

modules/ximgproc/src/estimated_covariance.cpp
```diff
@@ -0,0 +1,415 @@
...
+A previous and less efficient version of the algorithm can be found:
+ * O. Green, L. David, A. Galperin, Y. Birk, "Efficient parallel computation of the estimated covariance matrix", arXiv, 2013
+
+
+*/
+
+#include "precomp.hpp"
+//#include <opencv2/opencv.hpp>
+
+#ifdef HAVE_OPENMP
```
vpisarev: it's recommended to use cv::parallel_for_ instead of OpenMP etc. The parallel_for_ function has numerous backends, so that the parallelization is valid on practically any platform and any compiler, including iOS, Android etc.

modules/ximgproc/src/estimated_covariance.cpp
ogreen: That has been completed.

modules/ximgproc/src/estimated_covariance.cpp
ogreen: I will look into this, not entirely sure that the cv:parallel_for_ is ideal for this case for a large number of reason. Probably the most important of them is that in the current implementation the vectors finalMatPosR and finalMatPosC are allocated exactly once per thread. Whereas using the cv::parallel_for would require allocating this memory once per iteration. If that is the case, then I might as well remove the parallelization because the overhead memory allocation will out weigh the benefit of the optimization for many cases.

ogreen:
@vpisarev , it seems that my latest push is not building yet the error is breaking the build is unrelated to my commit.
Any advice on how to resolve that?

vpisarev:
ok, thanks! let's first merge this code and then think of further optimizations

vpisarev:
:+1:

PR #420, opened 2015-11-01
Code and documentation for the CVPR '15 paper "DNNs are easily fooled"

anguyen8:
Code and documentation for the CVPR '15 paper "DNNs are Easily Fooled" are placed in the modules/dnns_easily_fooled directory. 

This PR is for the Vision Challenge. http://code.opencv.org/projects/opencv/wiki/VisionChallenge


vpisarev:
sorry, having caffe to be put inside OpenCV is not an option, it's way too big dependency. Could you please consider using our dnn module? (if you do not need training). If you need training, it's better to detect and link Caffe as external library, without putting it inside OpenCV

anguyen8:
@vpisarev : Hi Vadim, thanks for your review. I see your point, but for this paper, we actually extended Caffe to make the experiments work.

I will try to transfer the extended features to the Caffe in the DNN module then.

vpisarev:
@anguyen8, thank you too for understanding, I was going to write a separate e-mail to you, explaining our policies, but since you are reacting fast and wish to cooperate, I can try to do it here. If you need patched Caffe, it may be a too big effort to convert all our code to the new API. Instead, may I suggest you to put a description (or may be a script that does that?) on how to download certain snapshot of Caffe, apply your patch to it (the patch itself can be made a part of your contribution) and then run with proper flags?

anguyen8:
@vpisarev : Thanks. That sounds great. I will write the script, and submit a different PR for your review.

PR #422, opened 2015-11-03
a python script to build iOS framework with extra modules added

virt87:
When updating the iOS project from opencv2.9 to 3.0 was stumped  with the `non free` folder containing `SURF`. The default recommendation described here didn't work for me and the build script from the OpenCV folder with modification to use extra modules crashed when launched. Have slightly modified the default build script in order to build an iOS OpenCV framework with OpenCV3.0 and some extra modules, in particular with xfeatures2d module that has replaced the old `non free`  folder.

### DIFF
--- a/samples/ios/README.md
+++ b/samples/ios/README.md
@@ -1,0 +1,29 @@
+## A sample build script for iOS
+
+This folder contains some useful scripts and samples using "extra" modules for the iOS platform.
+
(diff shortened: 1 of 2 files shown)


mshabunin:
@virt87 , please check the updated variant of `build_framework.py` script: Itseez/opencv#5654
It should gather all modules into one framework.

Issue #415, opened 2015-10-29
Trackers are not exposed to python


WestFlame:
The Trackers API is not accesible in generated bindings for python (I'm using python3).

berak:
yes, that's true, before we can do that, some issues have to get solved:

* `cv::Rect2d` has to get wrapped into python (cv2.cpp in main trunk)
* there is no `cv::tracking` namespace in this module, so e.g. `Tracker_create()` will go into the 'global' cv2 namespace. not a bug, but definitely ugly.
* at the bottom of tracker.hpp, there are some `selectROI` exposed to python, even if `Rect2d` works, the `std::string` is problematic (maybe should get changed to cv::String)

berak:
you might want to help, and try the Rect2d branch in main, and an early , [experimental branch](https://github.com/berak/opencv_contrib/tree/python_tracking) for the tracking module.

WestFlame:
Thank you for such fast response :). I will definitely try it.

berak:
just don't forget to setup proper branches for both main & contrib before you merge anything, to keep master clean !

berak:
sorry, first attempt was terribly sloppy. can you try again ?

WestFlame:
I was finally able to compile and try it. It seems that it works fine. But I've currently tryied only the basic functionality (the `Tracker_create()`, `tracker.init(...)` and `tracker.update(...)` loop).

bhack:
Can we close this?

PR #425, opened 2015-11-03
add python bindings for Tracker and MultiTracker

berak:
see [#415](https://github.com/Itseez/opencv_contrib/issues/415)

also added 2 samples, tracker.py and multitracker.py


berak:
@kurnianggoro , what do you think about wrapping all of it into a namespace `cv::tracking` ?

lenlen:
@berak thank for the pull request. I'm agree with the namespace cv::tracking. I think that maybe the first step to do is merge this pending pull request (https://github.com/Itseez/opencv_contrib/pull/341/files) in which there are some important stuff for this module and then complete the python binding. What do you think?

berak:
@lenlen , yes, so the namespace will have to wait for #341 to get merged.

should i wait with this pr, too ?

vpisarev:
:+1:

PR #426, opened 2015-11-04
Add macro definition to avoid building error.

wangyan42164:
Otherwise, it will block building of opencv_contrib in my platform. Please help check it. Thanks


wangyan42164:
Seem not this mortification to make windows building failed?

vpisarev:
it looks like the patch is correct, but the build failed because of insufficient disk space. @mshabunin, @alalek, could you please cleanup the windows buildbot slave?

alalek:
Thanks! :+1:

Issue #428, opened 2015-11-04
Dataset module dependency create too many low correlated dependencies


bhack:
When a module depend on dataset it require too many other modules to be compiled.

PR #431, opened 2015-11-08
datasets: make most dependancies optional

berak:
the datasets module itself only depends on core, all other dependancies are for the samples only , and should be optional

also, since nothing is exposed to script wrappers, remove python bindings

see [428](https://github.com/Itseez/opencv_contrib/issues/428)


berak:
hmm, this does not really solve the issue

PR #433, opened 2015-11-10
SURF OCL: now enabled

PolarNick239:
Changes:
- Uncommented status=1 - the only way to set "SURF can use OpenCL" to
  True"
- To match with Transparent-API: OpenCL version will run only if UMat
  was passed

It seems, that OpenCL version of SURF was unavailable at all.

Can somebody explain, why that line with status=1 was commented?

### DIFF
--- a/modules/xfeatures2d/src/surf.cpp
+++ b/modules/xfeatures2d/src/surf.cpp
@@ -893,7 +893,7 @@ void SURF_Impl::detectAndCompute(InputArray _img, InputArray _mask,
     CV_Assert(_descriptors.needed() || !useProvidedKeypoints);
 
 #ifdef HAVE_OPENCL
-    if( ocl::useOpenCL() )
+    if( ocl::useOpenCL() && _img.isUMat())
     {
         SURF_OCL ocl_surf;
         UMat gpu_kpt;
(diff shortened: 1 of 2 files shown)


yvo2m:
Probably because SURF OCL is buggy :
https://github.com/Itseez/opencv/issues/4685
> Unfortunately OpenCL implementation of SURF not available in master branch now, we found out that current implementation have errors.

PolarNick239:
Hm, I missed that. I see worse results in comparison with CPU too, also OCL implementatino does not deterministic.

But this way (suggested in T-API, and I followed it) this change looks safe - because if user would not explicitly use UMat - he will run CPU version.

Also, is there a way to document this problem? Some warnings or at least correct place to document inconsistency with CPU version?

yvo2m:
Well, the CUDA version is not so bad so if you specify UMat, it will run CPU or CUDA version and not OpenCL buggy version.
I think it should be fixed first, before enabling it again.

PolarNick239:
Hm, I agree if T-Api was transparent in that way, but unfortunately CUDA can be used only explicitly (with GpuMat and SURF_CUDA).

alalek:
Unfortunately, there is a low chance to enable this without correct T-API tests.

PolarNick239:
Some day I want to try to revert [old commit](https://github.com/Itseez/opencv/pull/2014), that seems to be the reason of problems, update implementation to support T-API (I think it should be quite straightforward?), and fix tests. Would not you disagree with revert of that pull request?

alalek:
PR #2014 is for 2.4 branch. You trying to fix opencv_contrib, which is for OpenCV 3.x/master branch only.
OpenCL-related code is different in 2.4 and 3.x branches (explicit vs T-API calls).

PolarNick239:
Yes, I know that, and mention it in words "update implementation (from old 2.4) to support T-API"

sovrasov:
#931 is more relevant than this one.

PR #434, opened 2015-11-11
Fix iOS framework issues

mshabunin:
!!! Merge with Itseez/opencv#5654
- disabled dnn module (can not build protobuf)
- disabled tools build from xobjdetect module


alalek:
:+1:

mtngld:
@mshabunin, Is it really not possible to build protobuf on iOS? No work-around for this limitation?

PR #437, opened 2015-11-12
Fixes cvv compilation for Apple's clang

fran6co:
It doesn't support thread_local.

### DIFF
--- a/modules/cvv/src/impl/call.cpp
+++ b/modules/cvv/src/impl/call.cpp
@@ -1,4 +1,5 @@
 #include "call.hpp"
+#include <atomic>
 
 namespace cvv
 {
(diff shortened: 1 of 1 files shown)


modules/cvv/src/impl/call.cpp
```diff
@@ -7,7 +7,7 @@ namespace impl
 
 size_t newCallId()
 {
-	thread_local size_t nextId = 1;
+	static std::atomic_size_t nextId (1);
```
mshabunin: You have to `#include <atomic>` to use it.

fran6co: Done.

mshabunin:
:+1:

PR #438, opened 2015-11-14
Add support for selecting ROI while keeping the video running in the roiSelector.cpp.

superxingzheng:
Add support for selecting ROI while keeping the video running in the roiSelector.cpp.

-- Fix misleading message and keyboard input printed by roiSelector.cpp.
-- Fix wrong comments.
-- Fix python complain about passing a reference to VideoCapture.

### DIFF
--- a/modules/tracking/include/opencv2/tracking/tracker.hpp
+++ b/modules/tracking/include/opencv2/tracking/tracker.hpp
@@ -50,7 +50,6 @@
 #include "onlineBoosting.hpp"
 #include <iostream>
 
-
 #define BOILERPLATE_CODE(name,classname) \
     static Ptr<classname> createTracker(const classname::Params &parameters=classname::Params());\
     virtual ~classname(){};
@@ -65,8 +64,7 @@
  *
  */
 
-namespace cv
-{
+namespace cv {
 
 //! @addtogroup tracking
 //! @{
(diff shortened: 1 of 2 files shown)


modules/tracking/include/opencv2/tracking/tracker.hpp
```diff
@@ -1366,9 +1370,9 @@ class ROISelector {
 };
 
 Rect2d CV_EXPORTS_W selectROI(Mat img, bool fromCenter = true);
-Rect2d CV_EXPORTS_W selectROI(const cv::String& windowName, Mat img, bool showCrossair = true, bool fromCenter = true);
-void CV_EXPORTS_W selectROI(const cv::String& windowName, Mat img, std::vector<Rect2d> & boundingBox, bool fromCenter = true);
-
+Rect2d CV_EXPORTS_W selectROI(const cv::String &windowName, Mat img, bool showCrossair = true, bool fromCenter = true);
+void CV_EXPORTS_W selectROI(const cv::String &windowName, Mat img, std::vector<Rect2d> &boundingBox, bool fromCenter = true);
+void CV_EXPORTS_W selectROI(const cv::String &windowName, cv::VideoCapture &cap, std::vector<Rect2d> &boundingBox, bool fromCenter = true);
```
berak: does it really need this function ? 

* imho, it's not practical to draw a boundingrect while the video moves, there's good reason to do this on a still image instead.

* the python wrappers complain about passing a reference to VideoCapture, so it probably should be CV_EXPORTS (w/o _W) , and so not wrapped to python

superxingzheng: @berak 
Thanks for your time.
I added the selecting with video running function so that the tracker doesn't have to search far away if the target is moving while user is selecting. If we select target on the static frame, the target may have already moved away when the tracking gets started.

I dont use python often so I didn't know the complain about passing a reference to VideoCapture.

berak:
almost done ! you've seen the [doc warnings](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_docs/builds/1947/steps/whitespace%20contrib/logs/stdio) ?

modules/tracking/include/opencv2/tracking/tracker.hpp
berak: helo Zhen Ying, 
>>I added the selecting with video running function so that the tracker doesn't have to search far away if the target is moving while user is selecting. If we select target on the static frame, the target may have already moved away when the tracking gets started.
ah, i see what you mean now. sorry for not getting it instantly.
just fix the doc errors (trailing spaces and such), and squash your commits to a single one, then - best luck (and patience, things on contrib take their time...)



    On Monday, November 23, 2015 5:38 AM, Zheng Xing <notifications@github.com> wrote:
 

 In modules/tracking/include/opencv2/tracking/tracker.hpp:> @@ -1366,9 +1370,9 @@ class ROISelector {
>  };
>  
>  Rect2d CV_EXPORTS_W selectROI(Mat img, bool fromCenter = true);
> -Rect2d CV_EXPORTS_W selectROI(const cv::String& windowName, Mat img, bool showCrossair = true, bool fromCenter = true);
> -void CV_EXPORTS_W selectROI(const cv::String& windowName, Mat img, std::vector<Rect2d> & boundingBox, bool fromCenter = true);
> -
> +Rect2d CV_EXPORTS_W selectROI(const cv::String &windowName, Mat img, bool showCrossair = true, bool fromCenter = true);
> +void CV_EXPORTS_W selectROI(const cv::String &windowName, Mat img, std::vector<Rect2d> &boundingBox, bool fromCenter = true);
> +void CV_EXPORTS_W selectROI(const cv::String &windowName, cv::VideoCapture &cap, std::vector<Rect2d> &boundingBox, bool fromCenter = true);
@berak 
Thanks for your time.
I added the selecting with video running function so that the tracker doesn't have to search far away if the target is moving while user is selecting. If we select target on the static frame, the target may have already moved away when the tracking gets started.I dont use python often so I didn't know the complain about passing a reference to VideoCapture.—

PR #441, opened 2015-11-15
Add HDF module & implement HDF version 5.

cbalint13:
This PR adds **HDF** (Hierarchical Data Format) module and for now implements I/O for HDF5 (version 5). 

* Full support for ```cv::Mat``` data model of any type and any channel amount.
* Support optimizations like internal hdf5 data **chunking** and **compression**.
* Support **sub-windowed** reads, writes, inserts or appends even in mutithreaded scenario.
* Support **unlimited** mode/size **property** on any of desired dataset dimension.
* Support for **insert**, **append** over and **by expanding** dataset using e.g. dsinsert().
* Extended support for ```std::vector<cv:KeyPoint>``` storage (as HDF compounds).
* Support for creation of groups, and create dataset into some group/subgroup.
* API expose well documented simple routines with examples that cover any demand or scenario.

For reviewers, API docs are here: 
* http://openrisc.rdsor.ro/hdf5docs/d6/dee/group__hdf5.html
* http://openrisc.rdsor.ro/hdf5docs/d3/d23/classcv_1_1hdf_1_1HDF5.html

HDFView
--------------
Once reviewed with full cv::Mat support, implementation can be extended:

* Same API for the other HDF4 (the older format)
* Full support for other types: Vec, Matx, Point
* Support HDF attributes over dataset/group in form of Matx, Vec, Point.


cbalint13:
@mshabunin ,
- Can help me with these warns:
http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_android/builds/1461/steps/compile%20release/logs/warnings%20%2833%29

I don't believe is coused by this PR here.

mshabunin:
@cbalint13, these warning will be fixed soon. After that I will restart all affected builds.

vpisarev:
@cbalint13, thank you for this very nice contribution!

Overall, it looks terrific and is pretty much ready for integration!

May I suggest you to replace "const char*" with String to make it more C++ and wrapper friendly?
Possibly, const int* offset/counts could be replaced with vector<int>& too, but it's not critical.

cbalint13:
@vpisarev ,

 First, thank you for the review ! I am glad that you like this contrib, hope community will appreciate too.

* (1) I wanted ```std::string``` right from begin but had problem with CV_WARP. Now I realize OpenCV has its own ```String``` !. Will update right now !!

* (2) Yes, thinked to use std::vector instead ```const int *```, But just got inspired by how ```cv::Mat``` itself uses same kind of ```int *``` array for n-dim initialization. I believe i could do an overloaded version with ```std::vector``` too.

Will do the requested steps, in shortest possible time.

cbalint13:
@vpisarev , I am pretty done with requests, squashed all into one (force push).

(1) Replaced all ```const char *``` with ```cv::String``` .
(2) Added overloaded variants where ```const int*``` is now ```std::vector<int>&``` .

Docs updated accordingly. Let me know if have more ideas of improvement. Regardless it is merged or not, will work to add those proposals too (e.g. ```std::vector<cv::Keypoints>``` and so on) I believe room is now open for further improvements.

-----<<<-----

I've tested this implementation to some edge limit. Was able to append using dsinsert() until obtained ~10Gb of dataset with a very large matrix. Read is also very fast, if all done in subwindow mode one can follow process using nice progress bar (cases of predictible ```for``` cycles).

cbalint13:
* On top https://github.com/Itseez/opencv_contrib/pull/441#issuecomment-159236121 now ```vector<KeyPoint>``` handlers are implemented and documented too, just like ones for ```cv::Mat```.

vpisarev:
looks great, thanks a lot! :+1:

PR #445, opened 2015-11-17
aruco: port samples to cv::CommandLineParser

paroj:
results in less code and more consistent parameter handling.


vpisarev:
thanks! :+1:

PR #447, opened 2015-11-18
Fixed the issue #5032 posted in opencv

stoneyang:
This commit aims to fix issue [#5032](https://github.com/Itseez/opencv/issues/5032) posted in opencv.

The missing line causes the following error message when compiling using VS2012:

    fatal error C1083: Cannot open include file: 'opencv2/hal/defs.h': No such file or directory
which fails the compilation of matlab binding.


vpisarev:
:+1:

PR #448, opened 2015-11-19
Fixed issue caused by missing round() in VS before 2013

stoneyang:
This commit fixed issue caused by missing round() in VS before 2013.
> According to [discussion on MSDN](https://social.msdn.microsoft.com/Forums/vstudio/en-US/260e04fc-dd05-4a96-8953-9c6ea1ad62fb/cant-find-stdround-in-cmath), ``round()`` function is not implemented in Microsoft Visual Studio before 2013.

The added lines decides whether applying the implementation according to the version of Visual Studio on one's machine. Thus, it suppress the error caused by the missing identifier.

### DIFF
--- a/modules/dpm/src/dpm_feature.cpp
+++ b/modules/dpm/src/dpm_feature.cpp
@@ -47,6 +47,7 @@ namespace cv
 {
 namespace dpm
 {
+
 Feature::Feature()
 {
 }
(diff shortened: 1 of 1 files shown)


stoneyang:
Could anyone tell me is DOC in build farm means Documentation? ....

alalek:
Yes, this builders checks documentation build and some coding-style rules.
There is problem with "trailing whitespace":
http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_docs/builds/1929/steps/whitespace%20contrib/logs/stdio
Please remove these spaces, run `git commit --amend` and `git push -f origin stone_port`. After that this PR will be automatically rechecked.

 P.S. More info about OpenCV contribution process and some relates notes is here: https://github.com/Itseez/opencv/wiki/How_to_contribute (whitespace problem helper is placed under "Q3" section)

stoneyang:
@alalek , thanks for your kind reminder!  Problem solved. :)

vpisarev:
@stoneyang, @alalek, this is okay solution for the problem, but I actually do not like the fact that we try to replicate semi-standard C function. As I see, this round() is used just twice in the code and each time the result is immediately converted to integer. So, may I suggest to replace (int)round(...) with cvRound() in those 2 places, which would be much better solution, IMO?

stoneyang:
@vpisarev , suggestion noted. ``cvRound()`` 's better than this handly made ``round()``. Thanks! :)

vpisarev:
@stoneyang, great! thanks! :+1:

PR #449, opened 2015-11-19
Expose Error Cross-Covariance in Uncented Kalman Filters

Audenrith:
For some applications it is useful to have an estimate of how uncertain
the specific variable is estimated. This could help to act accordingly
e.g. increase the measurement zone if the current estimate is very
uncertain.


paroj:
:+1:

vpisarev:
:+1:

PR #450, opened 2015-11-19
Enabled FAST feature detection on specific image locations

Audenrith:
Some applications only want to know if there are feature at specific locations. To fit these needs the FAST approach is extended in FASTForPointSet to recieve a vector of locations and calculates the FAST response on these positions. 
If it is below the threshold, it will be removed from the list.

### DIFF
--- a/modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
+++ b/modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
@@ -259,6 +259,32 @@ public:
 };
 
 
+
+/** @brief Estimates cornerness for prespecified KeyPoints using the FAST algorithm
+
+@param image grayscale image where keypoints (corners) are detected.
+@param keypoints keypoints which should be tested to fit the FAST criteria. Keypoints not beeing
+detected as corners are removed.
+@param threshold threshold on difference between intensity of the central pixel and pixels of a
+circle around this pixel.
+@param nonmaxSuppression if true, non-maximum suppression is applied to detected corners
+(keypoints).
+@param type one of the three neighborhoods as defined in the paper:
+FastFeatureDetector::TYPE_9_16, FastFeatureDetector::TYPE_7_12,
+FastFeatureDetector::TYPE_5_8
+
+Detects corners using the FAST algorithm by @cite Rosten06 .
+ */
+CV_EXPORTS void FASTForPointSet( InputArray image, CV_IN_OUT std::vector<KeyPoint>& keypoints,
+                      int threshold, bool nonmaxSuppression=true, int type=FastFeatureDetector::TYPE_9_16);
+
+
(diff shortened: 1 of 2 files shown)


StevenPuttemans:
Shouldn't it be better to implement it more as a universal strategy that allows you to call any feature descriptor on a given point set? This seems to specific for only a single descriptor.

cbalint13:
@StevenPuttemans , i share the same viewpoint. E.g. something like teach the very FAST/ORB/AGAST compute() to reconsider if keypoints are already filled.

Audenrith:
I proposed this in opencv but the problem is, that some code expects the FAST function to clear the keypoint vector. See [https://github.com/Itseez/opencv/pull/5585].

StevenPuttemans:
@vpisarev as suggested by you in the previous PR and as not agreed on by me and @cbalint13, could you shed some light why a more universal approach would not work?

vpisarev:
@StevenPuttemans, trying to implement some universal feature in such a big library as opencv often means quite big efforts on implementing, debugging, testing, documenting and supporting that feature, and so we should always weight such proposals and, because of all the above, we actually try to reject most of such proposals or redirect them to opencv_contrib, where they can prove their usefulness with time.

In this particular case, in opencv we already have 3 mechanisms to specify an image region when dealing with keypoints:
1) just like in almost any other OpenCV function or method, we can pass ROI to the keypoint detector instead of a whole image
2) and/or we can also specify mask where to find the keypoints.
3) if we know exactly what are our keypoints, we can pass the flag "useProvidedKeypoints" and just compute descriptors for them.

in my opinion, that covers most of users needs, but apparently there are always users who need special features. I would not call them "common" or "universal" though.

@Audenrith, thanks for the PR. please, see my comments below :)

modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
```diff
@@ -259,6 +259,32 @@ class CV_EXPORTS_W DAISY : public Feature2D
...
+FastFeatureDetector::TYPE_9_16, FastFeatureDetector::TYPE_7_12,
+FastFeatureDetector::TYPE_5_8
+
+Detects corners using the FAST algorithm by @cite Rosten06 .
+ */
+CV_EXPORTS void FASTForPointSet( InputArray image, CV_IN_OUT std::vector<KeyPoint>& keypoints,
+                      int threshold, bool nonmaxSuppression=true, int type=FastFeatureDetector::TYPE_9_16);
+
+
+void makeOffsets(int pixel[25], int row_stride, int patternSize);
```
vpisarev: these are helper functions and they should not be put into the header

StevenPuttemans:
@vpisarev thanks for the update!

terfendail:
Transferred to #1387

PR #453, opened 2015-11-24
add HDF5_INCLUDE_DIR for compiled with Caffe

mylxiaoyi:
add `HDF5_INCLUDE_DIR` and `HDF5_INCLUDE_LIBS` to `modules/cnn_3dobj/CMakeLists.txt` for compile OpenCV with Caffe.

For  Debian-based Linux Systems, `HDF5_INCLUDE_DIR` may be `/usr/include/hdf5/serial` or `/usr/include/hdf5/openmpi`. If Caffe is compiled with HD5, this may be a problem.

### DIFF
--- a/modules/cnn_3dobj/CMakeLists.txt
+++ b/modules/cnn_3dobj/CMakeLists.txt
@@ -24,6 +24,13 @@ else()
   message(STATUS "Glog:   NO")
 endif()
 
+find_path(HDF5_INCLUDE_DIR NAMES hdf5.h
+    HINTS
+    /usr/include/hdf5)
+find_library(HDF5_LIBS NAMES hdf5
+    HINTS
+    /usr/lib)
+
 if(HAVE_CAFFE)
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/cnn_3dobj_config.hpp.in
                ${CMAKE_CURRENT_SOURCE_DIR}/include/opencv2/cnn_3dobj_config.hpp @ONLY)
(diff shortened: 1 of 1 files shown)


vpisarev:
@mylxiaoyi, thanks! may I ask you (and @mshabunin) - what happends if Caffe is compiled without HDF5, i.e. Caffe is there, but HDF5 is not? will cnn_3dobj still compile?

mshabunin:
@vpisarev , this module is broken and should be fixed first. We does not build it on buildbot because it depends on caffe which is not installed on the buildslave.
Some of the problems:
- some samples include wrong headers ([example](https://github.com/Itseez/opencv_contrib/blob/24d9e616f0d3ed81d451316da68409ddc24f7b48/modules/cnn_3dobj/samples/demo_classify.cpp#L41))
- some samples probably depend on opencv_features2d and opencv_calib3d which are not added to the module dependencies, thus are not added to the samples dependencies (same example)
- one of the source files is being rewritten during build process ([cmake command](https://github.com/Itseez/opencv_contrib/blob/24d9e616f0d3ed81d451316da68409ddc24f7b48/modules/cnn_3dobj/CMakeLists.txt#L29))
- one of the included caffe headers is obsolete ([here](https://github.com/Itseez/opencv_contrib/blob/24d9e616f0d3ed81d451316da68409ddc24f7b48/modules/cnn_3dobj/include/opencv2/cnn_3dobj.hpp#L68))

AFAIK, caffe has switched to HDF5 format by default and probably can not be built without HDF5 library (https://github.com/BVLC/caffe/pull/2836).

There are two build systems available for caffe: cmake and Makefile. There is a chance that cmake will be chosen as the default (and the only?) BS at some time (https://github.com/BVLC/caffe/issues/3351). Maybe using standard cmake _find_package_ for linking with caffe could be better choice.

mylxiaoyi:
@vpisarev , I test again, and find that `caffe` header file `include/caffe/data_layers.hpp` includes `hdf5.h`, so can not find `hdf5.h` when compiling and can not find `libhdf5.so` when linking.

I also noticed that the header files in `caffe` have been changed, so the error message now is `opencv_contrib/modules/cnn_3dobj/include/opencv2/cnn_3dobj.hpp:68:35: fatal error: caffe/vision_layers.hpp: no such file or directory`

paroj:
using this version of caffe: https://github.com/BVLC/caffe/pull/5019 with HDF5 enabled and compiled using the CMake build system, I did not have to add the above includes to opencv.

AFAIK @sovrasov compiled the module as well.

mshabunin:
@paroj , thank you!

PR #459, opened 2015-11-30
Add support for selecting ROI while keeping the video running in the roiSelector.cpp. 

superxingzheng:
-- Fix misleading message and keyboard input printed by roiSelector.cpp. 
-- Fix wrong comments.
-- Fix python complain about passing a reference to VideoCapture.

I added the support for selecting ROI with video running so that the tracker doesn't have to search far away if the target is moving while the user is selecting. If we select a target on a static frame, the target may have already moved away when the tracking gets started.

### DIFF
--- a/modules/tracking/include/opencv2/tracking/tracker.hpp
+++ b/modules/tracking/include/opencv2/tracking/tracker.hpp
@@ -50,7 +50,6 @@
 #include "onlineBoosting.hpp"
 #include <iostream>
 
-
 #define BOILERPLATE_CODE(name,classname) \
     static Ptr<classname> createTracker(const classname::Params &parameters=classname::Params());\
     virtual ~classname(){};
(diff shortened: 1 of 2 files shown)


vpisarev:
@superxingzheng, thank you for the contribution! As you may see by looking at diff, because of the changed indentation, it's very difficult to figure out what you actually have changed. Could you please restore the formatting so that the difference is obvious?

superxingzheng:
@vpisarev Dose OpenCV have a preferred "Indent Style"? Some of the ones listed in wiki are built in Eclipse. (https://en.wikipedia.org/wiki/Indent_style#Allman_style)
Probably the original codes this pull request is working on was edited by some editor other than Eclipse.
I think formatting should be unified before codes are merged into official branches.

sovrasov:
ROI selector has been moved to highgui module, so this PR is outdated.

PR #460, opened 2015-12-01
Core implementation of the fine-grained search (FGS) method

YutingZhang:
Core implementation of the fine-grained search (FGS) method based on the Gaussian process regression with latent variables.

Implementation of the Gaussian process regression with latent variables.

Local FGS is to be added.

### DIFF
--- a/modules/fgs/README.md
+++ b/modules/fgs/README.md
@@ -1,0 +1,6 @@
+Fine-grained search based on Gaussian process regression with latent variables for object detection
+=======================================================
+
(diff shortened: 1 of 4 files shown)


vpisarev:
@YutingZhang, thank you for the contribution! I would probably put this functionality into some larger module, not sure which one though. What do you use this algorithm for?
also, may I ask you to fix inconsistent whitespaces, found by our buildbot (pullrequest.opencv.org)?

terfendail:
Transferred to #1386

PR #463, opened 2015-12-06
surface_matching loadPLYSimple() now reports that only ascii format is supported

ahundt:
in addition to ascii there is a binary .ply format, which this function does not support. I added code to report that unsupported data is being read.

### DIFF
--- a/modules/surface_matching/src/ppf_helpers.cpp
+++ b/modules/surface_matching/src/ppf_helpers.cpp
@@ -69,10 +69,13 @@ std::vector<std::string> split(const std::string &text, char sep) {
 
 
 
-Mat loadPLYSimple(const char* fileName, int withNormals)
+Mat loadPLYSimple(const char* fileName, int withNormals /* = 0 */)
 {
   Mat cloud;
   int numVertices=0;
+  int numCols=3;
+  bool with_color = false;
+  bool with_alpha = false; // alpha transparency
 
   std::ifstream ifs(fileName);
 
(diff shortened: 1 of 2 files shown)


ahundt:
@bmagyar this is probably relevant to you

modules/surface_matching/src/ppf_helpers.cpp
```diff
@@ -54,6 +54,20 @@ void getRandQuat(double q[4]);
 void getRandomRotation(double R[9]);
 void meanCovLocalPC(const float* pc, const int ws, const int point_count, double CovMat[3][3], double Mean[4]);
 void meanCovLocalPCInd(const float* pc, const int* Indices, const int ws, const int point_count, double CovMat[3][3], double Mean[4]);
+std::vector<std::string> split(const std::string &text, char sep);
```
bmagyar: This forward declaration is not really doing anything here for us, since the definition comes right after it, you may remove it.

bmagyar:
I was wondering if there is a better way to report this error than a printf. Frankly, I don't like the printf since it may get lost if you are using this code in a bigger application. 

Maybe an error like this?
https://github.com/Itseez/opencv/blob/master/modules/videoio/src/cap_qt.cpp#L314

Or the CV_ERROR macro?

Apart from this the patch looks good :+1: 
Thanks for contributing!

ahundt:
I used a printf because the other error output in the same file is a printf. Actually if the licenses are compatible perhaps it would be best to simply incorporate the PCL ply loader.

ahundt:
cv_error does seem appropriate in the case a full ply loader isn't being used

bmagyar:
I'm actually not sure if a standard ply loader would take less lines of code than this. There we would need to integrate the reader, then convert from it's format to the one surface_matching is using (probably they are not the same). Are you familiar with the PCL ply loader?

ahundt:
@bmagyar I tried switching to CV_ERROR but it seems CV_StsBadArg is not declared, and I can't find it.

modules/surface_matching/src/ppf_helpers.cpp
ahundt: It didn't compile for me without the forward declaration, strangely enough.

mshabunin:
There are two compiler warnings:
```
.../surface_matching/src/ppf_helpers.cpp:77:8: warning: variable 'with_color' set but not used [-Wunused-but-set-variable]
.../surface_matching/src/ppf_helpers.cpp:78:8: warning: variable 'with_alpha' set but not used [-Wunused-but-set-variable]
```
And a trailing whitespace:  `modules/surface_matching/src/ppf_helpers.cpp:135: trailing whitespace.`

mshabunin:
Try to use `cv::Error::StsBadArg` instead ([doc](http://docs.opencv.org/master/db/de0/group__core__utils.html#gga759fa1af92f7aa7377c76ffb142abccaaf587497af64537041ee62c04a92b755d)).

terfendail:
Transferred to #1393

PR #465, opened 2015-12-08
revised PredictCollector with set of implementations

comdiv:
In prolongation for issue #401 and PR #402
Ported from real project.
Improves previously provided #402 on following manner: filtering, unification, better class model, collector set for buffering set of result and for statistical threshold metric count. 

* PredictCollector base class remaked:
	* templated `collect` method 
	* `distanceKoef` to allow compare results of distinct algorithms
	* `minThreshold` to allow filter out duplicates and keep only similar responses
	* `excludeLabel` to allow cross check images in train set without checking with self
	* `defaultFilter` that checks not just minthresh but maxthresh, guard label and state, applys `distanceKoef` to dist
	* `filter` method for overload to provide cusom filtering logic (non-filtered predicts will not send to emit)
	* filters can overwrite label/distance pair
	* `emit` method now not require to perform any checkings and it's implementation should just somehow store prediction in internal result (mapper)
* `PredictCollector` implementations added:
	* `MinDistancePredictCollector` - default for existed API - search for single search result
	* `TopNPredictCollector` - collects top N (5 by default) most closes result, `std::list` returned, sorted
	* `VectorPredictCollector` - stores all predict results in `std::vector` (for #359 usage), no double checkings - raw results
	* `MapPredictCollector` - stores all predict results in `std::map` where label is key - if doublers - most closed distance will be value
	* `StatPredictCollector` - allows collect base statistics about image prediction over train set (min, max, sum, count, avg) - usefull to determine best thresholds for certain task/train set and account koefs to unify different algorithm response
	* `StdPredictCollector` - collects standard diviation with given AVG (can be got from StatPredictCollector ) - usefull to determine best thresholds for certain task/train set

### DIFF
--- a/modules/face/include/opencv2/face/predict_collector.hpp
+++ b/modules/face/include/opencv2/face/predict_collector.hpp
@@ -45,9 +45,12 @@ the use of this software, even if advised of the possibility of such damage.
 #ifndef __OPENCV_PREDICT_COLLECTOR_HPP__
 #define __OPENCV_PREDICT_COLLECTOR_HPP__
 #include <cfloat>
+#include <list>
+#include <vector>
+#include <map>
 #include "opencv2/core/cvdef.h"
 #include "opencv2/core/cvstd.hpp"
-#undef emit  //fix for qt
+#undef emit //fix for qt
 namespace cv {
 namespace face {
 //! @addtogroup face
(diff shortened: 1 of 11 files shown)


comdiv:
question, how to deal with:
`/usr/bin/ld: CMakeFiles/opencv_python3.dir/__/src2/cv2.cpp.o: relocation R_X86_64_PC32 against undefined symbol `_Z13pyopencv_fromIN2cv3PtrISt4listISt4pairIidESaIS4_EEEEEP7_objectRKT_' can not be used when making a shared object; recompile with -fPIC`

it was result of
`CV_WRAP Ptr<std::list<std::pair<int, double> > > getResult();`

What to do in this case?

1) Do not try to pytonish and remove CV_WRAP 
2) Try to provide special pythonable  overload - what is best variant? what is compatible with Python?

comdiv:
Have stabilized it. Ready for revision.

comdiv:
have accpeted https://github.com/Itseez/opencv_contrib/pull/501

alalek:
@comdiv Could you please rebase changes.

comdiv:
Have rebased

modules/face/include/opencv2/face/predict_collector.hpp
```diff
@@ -45,9 +45,12 @@ the use of this software, even if advised of the possibility of such damage.
 #ifndef __OPENCV_PREDICT_COLLECTOR_HPP__
 #define __OPENCV_PREDICT_COLLECTOR_HPP__
 #include <cfloat>
+#include <list>
+#include <vector>
+#include <map>
 #include "opencv2/core/cvdef.h"
 #include "opencv2/core/cvstd.hpp"
-#undef emit  //fix for qt
```
alalek: Could you leave this line as is. It is added here: #515

comdiv:
have fix invalid rebase

alalek:
:+1:

soft4good:
hey guys, I want to use this but cant see the new code on master? where is it? thanks!

PR #466, opened 2015-12-09
Fixed include error

maccesch:
Fixed an include error in the stereo module

### DIFF
--- a/modules/stereo/include/opencv2/stereo.hpp
+++ b/modules/stereo/include/opencv2/stereo.hpp
@@ -47,8 +47,8 @@
 #include "opencv2/core.hpp"
 #include "opencv2/features2d.hpp"
 #include "opencv2/core/affine.hpp"
-#include "stereo/descriptor.hpp"
-#include "stereo/matching.hpp"
+#include "opencv2/stereo/descriptor.hpp"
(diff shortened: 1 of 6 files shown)


vpisarev:
it's not quite clear why some implementation-related headers have been moved from src to include subdirectory? If API is not exposed, the headers should not be put into include

maccesch:
That was actually the problem. These aren't implementation specific but are included in stereo.hpp.

vpisarev:
oops, sorry, I missed that

modules/stereo/include/opencv2/stereo.hpp
```diff
@@ -47,8 +47,8 @@
 #include "opencv2/core.hpp"
 #include "opencv2/features2d.hpp"
 #include "opencv2/core/affine.hpp"
-#include "../../stereo/src/descriptor.hpp"
-#include "../../stereo/src/matching.hpp"
+#include "stereo/descriptor.hpp"
```
vpisarev: could you then replace "stereo/descriptor.hpp" with "opencv2/stereo/descriptor.hpp" and do the same with subsequent include too for better clarity?

maccesch:
updated :)

vpisarev:
thanks! :+1:

PR #339, opened 2015-08-20
Fuzzy module added

tucna:
opencv_extra: https://github.com/Itseez/opencv_extra/pull/272

New module for image processing using fuzzy mathematics. Current version includes basic F0 operations, inpainting and filtering. Usage is tested in ***test_image*** and demonstrated in samples ***fuzzy_filtering.cpp*** and ***fuzzy_inpainting.cpp*** .

**Related papers:**

Perfilieva, Irina, and Pavel Vlašánek. "Image Reconstruction by means of F-transform." Knowledge-Based Systems 70 (2014): 55-63.

Perfilieva, Irina. "Fuzzy transforms: Theory and applications." Fuzzy sets and systems 157.8 (2006): 993-1023.

( new version and branch for the PR #224 )

PR #467, opened 2015-12-11
Fuzzy module

vpisarev:
this is the same as https://github.com/Itseez/opencv_contrib/pull/339, but with the merge problem fixed


vpisarev:
:+1:

PR #471, opened 2015-12-14
Fixed linker error in OSX for module sfm

maccesch:


alalek:
Please remove unnecessary merge commits.

maccesch:
Oh, sorry. How do I do that?

alalek:
1) `git rebase -i upstream/master` (or in your case this may work: `git rebase -i Itseez/master`).
2) Leave only 1 line with required commit (possible there is already 1 line, so nothing to edit here)
3) Save changes.
4) `git push -f origin master` to update source branch

maccesch:
I did as you said, and it sold me that the rebase was successful. But even with `push -f` git tells me that everything is up to date.

alalek:
It is strange, try `git rebase -i HEAD~3`

alalek:
BTW, there is also similar thing with `GLOG_LIBRARIES`. I believe it also should be moved out of "if" and placed near `GFLAGS_LIBRARIES` fallback.

maccesch:
You're probably right

maccesch:
Hehe after changing that running the rebase routine again, it worked :)
Thanks

alalek:
:+1:

Issue #481, opened 2015-12-21
[ximgproc] SLIC superpixels: contour mask contains {-1,1} instead of {-1,0}


ehku:
The mask returned by getLabelContourMask() contains -1 at the segmentation contours and 1 elsewhere, which is not a valid mask. A valid mask should contains 'a' (where a != 0) at the contours and 0 elsewhere (e.g. in the case of SEEDS: a = 255).

cbalint13:
Thanks a lot @ehku !

PR https://github.com/Itseez/opencv_contrib/pull/483 is onlgoing to address the issue. I made SLIC & LSC masks just like SEEDS one. Also matrix should be CV_8UC1 type.

ehku:
Thanks, @cbalint13 !!

cbalint13:
@ehku ,

Revisitied the patch. Matrix should be 8UC1, so i corrected ```seeds.cpp``` too. See here the reason:
* https://github.com/Itseez/opencv/blob/master/modules/imgproc/src/contours.cpp#L197

PR #483, opened 2015-12-21
BugFix: Issue #481 related to SLIC & LSC contour mask values.

cbalint13:
* This fix Issue https://github.com/Itseez/opencv_contrib/issues/481 . Thanks a lot @ehku !


cbalint13:
Also, matrix should be 8UC1 for ```seeds.cpp``` too:
https://github.com/Itseez/opencv/blob/master/modules/imgproc/src/contours.cpp#L197

alalek:
:+1:

AndrewChenUoA:
(y) x100. Just spent three hours trying to make the OpenCV3.1.0 code work and noticed that the outputs were not in line with the documentation...

PR #484, opened 2015-12-22
CompileFix: SetCommandLineOption may be in gflags:: not google::

cbalint13:
* ```SetCommandLineOption()``` may be in ```gflags::``` not ```google::``` namespace on newer version of libs. This simple patch allows the compiler to choose the appropriate namespace.


alalek:
Try to add a dummy namespace before added `using namespace` statements:
```
namespace gflags { }
```

alalek:
:+1:

PR #495, opened 2015-12-28
Java wrapping added for text module

NickyWeber:
Commit 67a2066c4b411897f7a4f12c9ae921cfec45e660 states "Java and python wrappers for contrib modules" for the CMakeLists.txt file. This commit adds java actually. Just tested, jar includes all text module wrappers.

### DIFF
--- a/modules/text/CMakeLists.txt
+++ b/modules/text/CMakeLists.txt
@@ -18,7 +18,7 @@ include_directories(${Tesseract_INCLUDE_DIR})
 endif()
 
 set(the_description "Text Detection and Recognition")
-ocv_define_module(text opencv_ml opencv_highgui opencv_imgproc opencv_core opencv_features2d WRAP python)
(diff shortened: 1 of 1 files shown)


sturkmen72:
i tried this patch. but could not generate `opencv-310.jar` maybe it needs Tesseract to be installed.

marcogovo:
I too tried to recompile opencv with this change but it gets me the same @sturkmen72 's error..
Do you have some news to resolve this? I'm thinking to modify nativeObj attribute from 'protected' to 'public' but i don't know if it would be a safe move.
Thank you all!!

puneet336:
Could you please share the resolution.
I am facing same issue with opencv-3.2.0!

NickyWeber:
Sorry, it's been too long.

abratchik:
well, I dont have any issues with this module - did you update from the latest master?

PR #498, opened 2016-01-01
opencv_text: src/erfilter.cpp: replaced bit and/or with their logical counterparts

rokm:
Variables p2-p9 are booleans, so combination of logical negation
and use of bit and/or operations in lines 2823 and 2824 results
in the following warning from gcc:
warning: suggest parentheses around operand of ‘!’ or change ‘&’
to ‘&&’ or ‘!’ to ‘~’ [-Wparentheses]


arrybn:
:+1:

PR #501, opened 2016-01-05
face: change the 'not found' label back to -1

berak:
let's keep consistancy with the docs and recent behaviour

cc @comdiv


alalek:
:+1:

PR #502, opened 2016-01-06
Bugfix & Enhance: Fix close() issue and add H5_GETCHUNKDIMS option.

cbalint13:
This PR address:

(1) Fix annoying ```close()``` issue when multiple datasets are opened at same time. Invoking ```close()``` **should not** alter other datasets status !

* It fix this scenario when two sets are opened at same time like below:
```
  cv::Ptr<cv::hdf::HDF5> h5io = cv::hdf::open( "test1.h5" );
  cv::Ptr<cv::hdf::HDF5> h5iw = cv::hdf::open( "test2.h5" );
  h5io->close();
  h5iw->close();
HDF5-DIAG: Error detected in HDF5 (1.8.15-patch1) thread 0:
  #000: ../../src/H5F.c line 781 in H5Fclose(): invalid file identifier
    major: Invalid arguments to routine
    minor: Inappropriate type
```
*  First ```h5io->dsclose()``` instance force booth datasets to close due to code bug caused by superfluous call to HDF's internal *H5close()* that shutdown all open sessions: https://github.com/cbalint13/opencv_contrib/commit/f468c1a634e4756d811a515ab96832b63d3af24f#diff-af5f4404415cf174c84ecdcc91e6cea1L286 .

* File status in ```open()``` is updated to reflect all possible states: https://github.com/cbalint13/opencv_contrib/commit/a01002f62b32293c44f20effbba6ff3f55d97901#diff-af5f4404415cf174c84ecdcc91e6cea1R273

(2) Add possibility to get properties *about chunk dimensions* not only about size or maxsize. Introduced the new **H5_GETCHUNKDIMS** flag among H5_GETDIMS, H5_GETMAXDIMS. This is an enhanchement and is documented too.

### DIFF
--- a/modules/hdf/include/opencv2/hdf/hdf5.hpp
+++ b/modules/hdf/include/opencv2/hdf/hdf5.hpp
@@ -63,7 +63,7 @@ public:
 
     CV_WRAP enum
     {
-      H5_UNLIMITED = -1, H5_NONE = -1, H5_GETDIMS = 100, H5_GETMAXDIMS = 101,
+      H5_UNLIMITED = -1, H5_NONE = -1, H5_GETDIMS = 100, H5_GETMAXDIMS = 101, H5_GETCHUNKDIMS = 102,
     };
 
     virtual ~HDF5() {}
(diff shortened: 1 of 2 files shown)


alalek:
:+1:

PR #505, opened 2016-01-06
Make aruco module exportable

jfolz:
References and raw pointers are replaced by shared Ptrs to facilitate the generation of bindings and respective macros are added to enable the export.
Also fixes a copy problem observed while testing in Python where returned corners pointed to uninitialized memory.
Remaining problem: drawCharucoDiamond cannot be exported, because core is missing conversion functions for Vec4i.

### DIFF
--- a/modules/aruco/include/opencv2/aruco.hpp
+++ b/modules/aruco/include/opencv2/aruco.hpp
@@ -121,30 +121,32 @@ namespace aruco {
  * - errorCorrectionRate error correction rate respect to the maximun error correction capability
  *   for each dictionary. (default 0.6).
  */
-struct CV_EXPORTS DetectorParameters {
+struct CV_EXPORTS_W DetectorParameters {
 
     DetectorParameters();
 
-    int adaptiveThreshWinSizeMin;
-    int adaptiveThreshWinSizeMax;
-    int adaptiveThreshWinSizeStep;
-    double adaptiveThreshConstant;
-    double minMarkerPerimeterRate;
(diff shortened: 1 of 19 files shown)


jfolz:
Forgot about samples, working on it.

joshdoe:
I take it this will lead to the aruco module being available in Python, something I'm very interested in. Need any help? Could your branch be used as-is to build Python bindings, except for drawCharucoDiamond? Thanks for your work.

jfolz:
@joshdoe This branch builds against opencv Itseez/opencv@345678770b3e4cdb0054c818e41469b1c0993c96 (haven't tested newer) and aruco can be used from Python with the mentioned exception.
I haven't been able to test all of the functionality, so maybe you can provide some feedback based on your use case.

joshdoe:
@jfolz Yesterday I built this branch against the latest opencv, and did some simple tests generating and detecting markers, and it seemed to work great. I won't be doing anything much more complicated, but will report back if I have any issues. Thanks!

alalek:
Please squash commits and rebase patch to resolve merge conflicts

joshdoe:
@jfolz I went to try out `CharucoBoard_create` and the `draw()` method was missing, so I added `CV_WRAP` and recompiled to get it to work. A quick check shows all methods of `CharucoBoard` and `GridBoard` are missing. I didn't think `CV_WRAP` had to be added to every member function...

jfolz:
@alalek done :)

alalek:
:+1:

PR #506, opened 2016-01-06
Fix type/signedness mismatches in aruco

jfolz:
Fix some signed/unsigned comparisons and unsafe casts I stumbled upon while working on the code.


alalek:
:+1:

Issue #513, opened 2016-01-12
Build fails in stereo/include/opencv2/stereo/matching.hpp - Incorrect test for existence of POPCNT


SSteve:
matching.hpp incorrectly tests for the existence of POPCNT with `#if CV_SSE4_1`. This causes a build failure on a Core2Duo MacBook Pro. According to [Wikipedia (with a link to the Intel reference)](https://en.wikipedia.org/wiki/SSE4#POPCNT_and_LZCNT) "*although introduced by AMD with the SSE4a instruction set, they are counted as separate extensions with their own dedicated CPUID bits to indicate support*". Support for POPCNT is indicated via the CPUID.01H:ECX.POPCNT[Bit 23] flag, not `CV_SSE4_1`.

SSteve:
I think the fix for this is to replace `#if CV_SSE4_1` with `#if CV_POPCNT` in matching.hpp. I'm using Homebrew to install opencv3. When I forked opencv and opencv_contrib and did the build manually I didn't get the error that happens when Homebrew does the build so I'm unable to test this change. But the fix seems obvious.

PR #514, opened 2016-01-13
Use correct test for existence of CPU's POPCNT feature

SSteve:
Existence of POPCNT is indicated by `CV_POPCNT`, not `CV_SSE4_1`. Fix for [https://github.com/Itseez/opencv_contrib/issues/513].


alalek:
In general, runtime check should be added also via this pattern: 
```
#if CV_POPCNT
    if (checkHardwareSupport(CV_CPU_POPCNT)
    {
        ... popcnt call ...
    }
    else
#endif
    {
        ... fallback ...
    }
````

alalek:
BTW, This runtime check is called on every iteration of 3 nested loops. It will be great to cache value of `checkHardwareSupport()` call.

SSteve:
I think I will leave that to the original author since I'm unable to run and test the code.

alalek:
OK, :+1:

PR #518, opened 2016-01-18
MSVC 2008 Python fixes

patricksnape:
Just a couple of commits adding simple casts that enable building of OpenCV 3.1 for Python 2.7 (which implicitly requires MSVC 2008).


alalek:
:+1:

PR #520, opened 2016-01-19
Implementation of SelectiveSearchSegmentation

the-glu:
This is a proposition of an implementation of a segmentation algorithm, described in [Selective search for object recognition](http://koen.me/research/pub/uijlings-ijcv2013-draft.pdf).

It use #361 as a base algorithm, I put it in the same module, I don't know if it's ok but I can move it :)

### DIFF
--- a/modules/ximgproc/include/opencv2/ximgproc/segmentation.hpp
+++ b/modules/ximgproc/include/opencv2/ximgproc/segmentation.hpp
@@ -67,55 +67,183 @@ namespace cv {
                         @param min_size The minimum size of segments
                      */
                     CV_EXPORTS_W Ptr<GraphSegmentation> createGraphSegmentation(double sigma=0.5, float k=300, int min_size=100);
+
+                    /** @brief Strategie for the selective search segmentation algorithm
+                        The class implements a generic stragery for the algorithm described in @cite uijlings2013selective.
+                     */
+                    class CV_EXPORTS_W SelectiveSearchSegmentationStrategy : public Algorithm {
+                        public:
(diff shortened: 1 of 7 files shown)


modules/ximgproc/include/opencv2/ximgproc/segmentation.hpp
```diff
@@ -117,6 +296,67 @@ namespace cv {
 
             };
 
+            // SelectiveSearch helpers
```
alalek: It is better to move implementation-specific declarations from public header file. You can place them in .cpp file.

the-glu:
I moved helpers from the headers to the cpp, updated the readme and (should have) fixed the warnings in non-linux builds :)

alalek:
Thanks!

You need to change location of the sample (compilation of this sample is not tested now):
```
git mv modules/ximgproc{/src,}/selectivesearchsegmentation.cpp
```

Could you please squash commits into one after all fixes (to remove unnecessary intermediate changes from the git history).

the-glu:
The command doesn't seems to move the location of the sample. Do you want me to move them from `samples/cpp/` to `samples` ?

alalek:
@the-glu Right

alalek:
add "static" to `hsv_to_rgb()` and `color_mapping()`

the-glu:
Yep I'm fixings warnings and I will squash everything when everything is green :P

the-glu:
Okay, everything should be ok now :)

alalek:
:+1:

PR #521, opened 2016-01-20
Update shape_transformation.cpp

sturkmen72:


alalek:
There is a conflict with branch names (patch-2 exists in the main OpenCV repo fork too)

PR #522, opened 2016-01-20
Update shape_transformation.cpp

sturkmen72:


alalek:
`patch-3` is not an unique name also. May be use `update_shape_trasformation` name for this branch?

PR #524, opened 2016-01-21
Fix memory leaks in GraphSegmentation

the-glu:
This fix obvious and sads memory leaks in `GraphSegmentation`.


alalek:
:+1:

BTW, `std::vector<>` may be used instead of raw pointers.

PR #527, opened 2016-01-22
Update plot.cpp

sturkmen72:
patch for making **setPlotBackgroundColor** functional

### DIFF
--- a/modules/plot/src/plot.cpp
+++ b/modules/plot/src/plot.cpp
@@ -182,7 +182,7 @@ namespace cv
             void render(Mat &_plotResult)
             {
                 //create the plot result
-                plotResult = Mat::zeros(plotSizeHeight, plotSizeWidth, CV_8UC3);
+                plotResult = Mat(plotSizeHeight, plotSizeWidth, CV_8UC3, plotBackgroundColor);
 
                 int NumVecElements = plotDataX.rows;
 


StevenPuttemans:
Euhm it seems reasonable that you would want to choose the background color yourself, but for now, you only change the function calling itself, not the actual implementation. There must be a part missing right?

sturkmen72:
**setPlotBackgroundColor** defined before but not functional
https://github.com/sturkmen72/opencv_contrib/blob/update_plot_cpp/modules/plot/src/plot.cpp#L152-L155

StevenPuttemans:
BUT what you are doing is incorrect, you are creating some sort of own matrix constructor, which is not what you want right?

modules/plot/src/plot.cpp
```diff
@@ -182,7 +182,7 @@ namespace cv
             void render(Mat &_plotResult)
             {
                 //create the plot result
-                plotResult = Mat::zeros(plotSizeHeight, plotSizeWidth, CV_8UC3);
+                plotResult = Mat(plotSizeHeight, plotSizeWidth, CV_8UC3, plotBackgroundColor);
```
StevenPuttemans: There is no MAT constructor that takes a Scalar as an extra argument inside OpenCV...

StevenPuttemans: And the Mat::zeros is there to be sure that there is no random initialized crap inside the matrix...

sturkmen72: http://docs.opencv.org/master/d3/d63/classcv_1_1Mat.html#a3620c370690b5ca4d40c767be6fb4ceb

StevenPuttemans: Aha did not know about that :D

alalek:
:+1:

PR #528, opened 2016-01-22
ERFilter python bindings

lluisgomez:
first time I try to expose this code to Python. I did it by looking how other modules do it. Since ERStat is a kind of "low-level" struct with no much interest for final users I have wrapped all the process in a single function cv2.text.detectRegions() which returns a vector< vector<Point> > with all regions. (Similarly as MSER function does).

Provided sample script detect_er_chars.py works ok for me in Linux (Ubuntu 15). 

@vpisarev can you please take a look at it? thnkx!

### DIFF
--- a/modules/text/include/opencv2/text/erfilter.hpp
+++ b/modules/text/include/opencv2/text/erfilter.hpp
@@ -225,7 +225,7 @@ classifier uses all the features calculated in the first stage and the following
 features: hole area ratio, convex hull ratio, and number of outer inflexion points.
  */
 CV_EXPORTS_W Ptr<ERFilter> createERFilterNM2(const Ptr<ERFilter::Callback>& cb,
-                                                  float minProbability = 0.3);
+                                                  float minProbability = (float)0.3);
 
 
 /** @brief Allow to implicitly load the default classifier when creating an ERFilter object.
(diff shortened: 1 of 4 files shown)


bharris47:
Would be nice to have `erGrouping` support.

lluisgomez:
I've another PR prepared for the erGrouping python binding, but waiting for this one to be more mature (i.e. merged :)...

Omnipresent:
@lluisgomez I am trying this PR on my local. I've copied 3 files in your commit and replaced them with the files that were there:

- (replaced) `erfilter.hpp`
- (replaced) `erfilter.cpp`
- (created new) `detect_er_chars.py`

But when I try to run `detect_er_chars.py` I get an error:

```
$ python detect_er_chars.py ~/my_pic.jpg
detect_er_chars.py
       A simple demo script using the Extremal Region Filter algorithm described in:
       Neumann L., Matas J.: Real-Time Scene Text Localization and Recognition, CVPR 2012
  File "detect_er_chars.py", line 25, in <module>
    erc1 = cv2.text.loadClassifierNM1(pathname+'/trained_classifierNM1.xml')
AttributeError: module 'cv2.text' has no attribute 'loadClassifierNM1'
```

Am I missing something? Do I need to build opencv_contrib again making these changes?

Omnipresent:
Had to rebuild opencv (w/ opencv_contrib) and then I was able to execute `detect_er_chars`

@lluisgomez +1 to having support for `erGrouping` as @bharris47 mentioned. `erGrouping` is being used in `textdetection.cpp` sample https://github.com/Itseez/opencv_contrib/blob/master/modules/text/samples/textdetection.cpp#L65 and would be really useful to use from Python. 

Can I help in any way to get python bindings for `erGrouping` so that a sample similar to `textdetection.cpp` can be run from Python?

vpisarev:
:+1:

PR #533, opened 2016-01-27
Updated bioinspired OpenCL implementation to OpenCV v3's ocl API

IrmatDen:
Hi,

I went ahead and updated bioinspired's ocl implementation to use the new opencv v3's API. That's my first time using opencv's v3 opencl api. So, thorough review is required!
Regarding results' differences, I get at most 1/255th of difference on parvo output (sometimes on 2 channels at once), and I did a video blend/difference for the magno outputs, and those seems to be the very same.

I tested mainly with defaults parameters, and for the parvo used the cat sample found under "Preliminary illustration" in [the beta v3 docs](http://docs.opencv.org/3.0-beta/modules/bioinspired/doc/retina/index.html#demos-and-experiments). Although the parvo outputs using the recommended settings differs from what is captured under this page, the CPU vs OpenCL implementation outputs are almost identical (within the 1/255th error mentionned earlier).

Edit: this has been only tested on VS2013 though


IrmatDen:
It seems auto-generated python and java bindings keep failing the build. Could you please give me a hint as to how this could be solved please?

modules/bioinspired/include/opencv2/bioinspired/retina.hpp
```diff
@@ -446,9 +447,9 @@ the log scale that is applied
  */
 CV_EXPORTS_W Ptr<Retina> createRetina(Size inputSize, const bool colorMode, int colorSamplingMethod=RETINA_COLOR_BAYER, const bool useRetinaLogSampling=false, const float reductionFactor=1.0f, const float samplingStrenght=10.0f);
 
-#ifdef HAVE_OPENCV_OCL
-Ptr<Retina> createRetina_OCL(Size inputSize);
-Ptr<Retina> createRetina_OCL(Size inputSize, const bool colorMode, int colorSamplingMethod=RETINA_COLOR_BAYER, const bool useRetinaLogSampling=false, const float reductionFactor=1.0f, const float samplingStrenght=10.0f);
+#ifdef HAVE_OPENCL
+CV_EXPORTS_W Ptr<Retina> createRetina_OCL(Size inputSize);
+CV_EXPORTS_W Ptr<Retina> createRetina_OCL(Size inputSize, const bool colorMode, int colorSamplingMethod = RETINA_COLOR_BAYER, const bool useRetinaLogSampling = false, const double reductionFactor = 1.0f, const double samplingStrenght = 10.0f);
 #endif
```
alalek: Problem with binding is here. `#ifdef` should be removed from public headers (with `#include "cvconfig.h"`)

BTW, 3.x OpenCL API is "transparent" (T-API). It means that the same user code should work on platforms with and without OpenCL support. So public API should not contain separate "_OCL" functions.

IrmatDen:
Thanks for the explanation of the failures. I was a little bit afraid of really "merging" the CPU and OpenCL code behind the two basics createRetina(...) methods, in case anything was wrong with the OpenCL implementation. I'll take a bit of time to do that soon.

sovrasov:
These changes are seems to be already introduced by Li Peng, so let's close this PR.

PR #536, opened 2016-01-31
datasets: make or_pascal fail gracefully on absent tags

berak:
* check for null pointer dereference in parseNodeText()
* pass string references, not copies, to avoid 'dangling pointer' problem


alalek:
:+1:

PR #540, opened 2016-02-04
fix plot2d orientation problem

sturkmen72:


sturkmen72:
@nmoutinho could you review this fix

modules/plot/src/plot.cpp
```diff
@@ -291,12 +291,12 @@ namespace cv
                 //setting the min and max values for each axis
                 plotMinX = MinX;
                 plotMaxX = MaxX;
-                plotMinY = MinY;
-                plotMaxY = MaxY;
+                plotMinY = MaxY;
+                plotMaxY = MinY;
```
StevenPuttemans: We should pay attention to constructs like this. We are assigning min and maximum values to plot max and min values ... why is this needed? Could you explain in more detail. If this is the case, then the functionality should be rewritten in order to get better readable code.

modules/plot/src/plot.cpp
```diff
@@ -291,12 +291,12 @@ namespace cv
...
                 plotMinX_plusZero = MinX_plusZero;
                 plotMaxX_plusZero = MaxX_plusZero;
-                plotMinY_plusZero = MinY_plusZero;
-                plotMaxY_plusZero = MaxY_plusZero;
+                plotMinY_plusZero = MaxY_plusZero;
+                plotMaxY_plusZero = MinY_plusZero;
```
StevenPuttemans: Same comments here

sturkmen72:
apologize for delay to adding more info
test code:
```
#include <opencv2/highgui.hpp>
#include <opencv2/plot.hpp>
#include <iostream>

using namespace cv;

{
    Mat data_x( 1, 11, CV_64F );
    Mat data_y( 1, 11, CV_64F );

    for( int i = 0;  i < 11; i++)
    {
        data_x.at<double>(0,i) = i-5;
        data_y.at<double>(0,i) = i-5;
    }

    std::cout <<  "data_x : " << data_x << std::endl;
    std::cout <<  "data_y : " << data_y << std::endl;

    Mat plot_result;

    Ptr<plot::Plot2d> plot = plot::createPlot2d( data_x, data_y );
    plot->render( plot_result );

    imshow( "plot", plot_result );
    waitKey();

    return 0;
}

```
current result:
plot

result after proposed fix:
aplot

StevenPuttemans:
Actually, I am of the opinion that the original one is the only correct one. OpenCV uses the 0,0 point at the top left corner and thus the initial image is actually just the same. In the second case you simple repositioned the origin point to the bottom left corner. Since non of the OpenCV functions expect this, I am afraid that changing this will add more confusion then necessary. A solution could be to put arrows on the axis end points to declare the direction of the axis!

sturkmen72:
it seems fix is crude
but i will not close the PR immediately

alalek:
What is about to add an option (or separate configuration calls) which setups coordinate system?

StevenPuttemans:
That would indeed be a good option. A function `setOrigin(parameter)` and then the options `ORIGIN_TOP_LEFT` or `ORIGIN_BOTTOM_LEFT` could easily do the trick!

nmoutinho:
@sturkmen72 and @StevenPuttemans sorry for the late reply guys. I was checking my original code and I confirm there is a mistake that leads to this sign switch. Where we have `Mat InterpYdata = linearInterpolation(plotMinY, plotMaxY, 0, plotSizeHeight, plotDataY);` we should have `Mat InterpYdata = linearInterpolation(plotMaxY, plotMinY, 0, plotSizeHeight, plotDataY);` and where we have `Mat InterpYdataFindZero = linearInterpolation(plotMinY_plusZero, plotMaxY_plusZero, 0, plotSizeHeight, plotDataY_plusZero);` we should have `Mat InterpYdataFindZero = linearInterpolation(plotMinY_plusZero, plotMaxY_plusZero, 0, plotSizeHeight, plotDataY_plusZero);`. @sturkmen72 can you switch to the original code, apply this changes and see if the plot becomes correct?

StevenPuttemans:
:+1:

StevenPuttemans:
@sturkmen72 could you confirm that this results in the desired figure?

sturkmen72:
i think Orientation issue is solved by this change.
what is your opinions about removing zeros ( "0"  there are four of them ). i think no need to display

StevenPuttemans:
I agree that the zeros might be clugging up the image, so removing them seems fine for me. However if possible it would be nice to have arrows at the end of the axis so that people immediatly know where the origin is and how the range is defined. Could you add this?

nmoutinho:
@StevenPuttemans as a first and fast solution we can keep only one zero for now (lower left corner) and delete all the other 3. Then I will think of how to implement the arrows and possible, put interval values on the axis which may help on the plot reading.

nmoutinho:
I think we can close this error fix for now. The arrows and add-ons must be included in another pull request.

StevenPuttemans:
:+1: @alalek could you close this issue down?

PR #545, opened 2016-02-09
LATCH: remove unnecessary gaussian blur.

ofirbb:
LATCH descriptor does not require a gaussian blur and the operation is being overwritten anyways in case of a color image.
The removed line causes excess runtime for no reason.

### DIFF
--- a/modules/xfeatures2d/src/latch.cpp
+++ b/modules/xfeatures2d/src/latch.cpp
@@ -500,20 +500,21 @@ namespace cv
             if ( keypoints.empty() )
                 return;
 
-
             Mat grayImage;
-            GaussianBlur(image, grayImage, cv::Size(3, 3), 2, 2);
-
-            if (image.type() != CV_8U) cvtColor(image, grayImage, COLOR_BGR2GRAY);
-
-
+            if (image.type() != CV_8U)
+            {
+                cvtColor(image, grayImage, COLOR_BGR2GRAY);
+            }
+            else
+            {
+                grayImage = image;
+            }
 
             //Remove keypoints very close to the border
             KeyPointsFilter::runByImageBorder(keypoints, image.size(), PATCH_SIZE / 2 + half_ssd_size_);
             
             bool _1d = false;
             Mat descriptors;
-
             _1d = _descriptors.kind() == _InputArray::STD_VECTOR && _descriptors.type() == CV_8U;
             if( _1d )
             {


StevenPuttemans:
@GilLevi could you check this?

It seems to me that the code on line 502-503 should simply be placed behind the check of the color image. Because now it indeed adds nothing when a color image is added. 

As to the use of gaussian blur, a small filter can remove small noise artefacts, so I do not think that this is completely random and useless.

modules/xfeatures2d/src/latch.cpp
```diff
@@ -500,20 +500,21 @@ namespace cv
             if ( keypoints.empty() )
                 return;
 
-
             Mat grayImage;
-            GaussianBlur(image, grayImage, cv::Size(3, 3), 2, 2);
-
-            if (image.type() != CV_8U) cvtColor(image, grayImage, COLOR_BGR2GRAY);
```
StevenPuttemans: I would simply reposition line 505 to line 512 and further! In that case the blur is added in both color and grayscale image case.

ofirbb:
In my experience there is no sense in treating color and grayscale images differently, one just needs to be converted.
As for the blurring reducing noise, that's true but irrelevant, the LATCH descriptor works in a three way comparison, so the noise level is always compared to the same reference and is reduced anyways.
Also, in some tests I ran, the descriptor accuracy is lower when blurring the image, which makes sense given the way it's built.
@GilLevi I would appreciate your opinion.

StevenPuttemans:
> In my experience there is no sense in treating color and grayscale images differently, one just needs to be converted.

Agreed completely, that is exactly why I would suggest to move it further down, to treat both cases equally.

ofirbb:
But you suggest leaving the gaussian blur, I maintain that it ruins the descriptor and the only reason it wasn't noticed may be that usually people work with color images as input.

StevenPuttemans:
> But you suggest leaving the gaussian blur, I maintain that it ruins the descriptor and the only reason it wasn't noticed may be that usually people work with color images as input.

Ow I just wanted to point out that I agree on the handling of color and BW images. As too the blurring or not, I am still eager to know why Gil added it in the first place. But I do understand your reasoning behind not doing it at all.

Furthermore, could you maybe add some tests proving that with and without blur it actually makes an accuracy difference? This could solidify your case :)

cbalint13:
@StevenPuttemans , @ofirbb ,

 - Its a well known fact GaussianBlur helps descriptors by smoothing out variance. In my opinion it should be there. But it should act on the gray image (if rgb then after the very grey conversion and not before). The speed impact should be really little.

ofirbb:
It is a well known fact and it is essential when comparing pixel to pixel, but that's part of the point in LATCH, you compare patches and not a single point so this actually is not necessary.
In processing terms, the gaussian blur step accounted for over 80% of the runtime in my experiments.

GilLevi:
Hi everyone:)

Thank you all for your interest in LATCH, I never thought it would get such attention that people will actually go over those subtleties. Specially, thank you @ofirbb for point out those problems.

First, if gaussian blur is needed then it should be done for both color and grayscale images.

Now, in my experiments I found that a small gaussian blur (3x3) actually improves performance, but I will glad to look into it again. Regarding running times, I don't remember that it accounted for so much of the running times (given all the comparisons LATCH makes), but I be glad to look into that as well.

If the blurring is indeed so costly, than it should be removed. Saving so much running time can allow one to use larger mini-patches that improves performance far more than the blurring.

Gil

cbalint13:
@ofirbb , @GilLevi ,

How about adding a switch like 'sigma' where sigma of kernel can be specified (now is 2.0f). Sigma 0.0f means no blur at all, higher sigma means more blur. If it would be also documented then all people should be happy.

StevenPuttemans:
@cbalint13 I kind of like your suggestion. But I am still convinced that in order to get this change added we should first ask @ofirbb to supply
1) timing results on a test code and test data (please share so we can reproduce)
2) accuracy result on test code and test data with and without blur

Certainly because most of you are pointing out that blurring should NOT be the bottleneck, I am curious if the bottleneck is just not due to the programming style used.

StevenPuttemans:
To all of you, no interest anymore in getting this fixed correctly?

ofirbb:
I will move it forward sometime next week.
Open source curse so I have a major deadline this week which kept me from this task.

StevenPuttemans:
:+1:

terfendail:
Transferred to #1419

Issue #546, opened 2016-02-11
Cannot build hdf modules on windows


stereomatchingkiss:
OS : Win8 64bits
Compiler : vc2013 64bits

Install hdf5 1.8.16 from [(Windows (64-bit)   [1]   Compilers:   CMake VS 2013 C, C++, IVF 15 )](https://www.hdfgroup.org/HDF5/release/obtain5.html)

The first issues is the cmake will complain it can not find the HDF5_LIBRARIES

Second issues is visual c++ expected constant expression

   ```
 // fetch rank
    int n_dims = H5Sget_simple_extent_ndims( fspace );

    // dims storage
    hsize_t dims[n_dims];
```

I guess the problem is n_dims cannot determine at compile time, maybe cv::AutoBuffer could solve this problem?
`
cv::AutoBuffer<hsize_t, 64>`

cbalint13:
@stereomatchingkiss ,

* I'll address today a patch for this issue, i am aware of this. It's already done just need to PR.
* At the moment of intial PR the hdf library was missing on msvc buildbot so there where no feedback.


I'll notice you here when done.

cbalint13:
@stereomatchingkiss ,

- An alpha patch: http://openrisc.rdsor.ro/hdf5-msvc.cpp.diff
- Need to check it in extensive ways and then pass with PR (today).

stereomatchingkiss:
So fast, thanks for the patch. I upload the fix too, I use cv::AutoBuffer to fix the issues rather than raw array and new the memory on heap. It works. but I do not know how to deal with cmake issues yet

[fix](https://github.com/stereomatchingkiss/opencv_contrib/commit/afc41621ef6b14b214a9b96c513e51263987cc23)

cbalint13:
@stereomatchingkiss , as original author:

* With respect, for HDF _low_ level back-end I believe _no_ C+++ (triple plus) style is opportune !
* Allow me a proper msvc fix (just for MSVC), keeping the HDF philosophy, of c-ish style as backend.
* As remainder you can create cv::Mat N-d matrix just passing same simple array int[] to ::Mat.

Lets keep it simple, RTM about cv::Mat.

stereomatchingkiss:
Since you are the author and prefer c-ish style, I respect your decisions :). About the patch, I cannot find delete being called, is it a mistake or some sort of trick?

[...]

Issue #497, opened 2015-12-30
opencv_waldboost_detector build failure against opencv_world


lych77:
Visual Studio 2015
opencv_contrib v3.1.0
BUILD_opencv_world enabled (This issue won't occur if this is disabled)

Getting the following error:

    Severity	Code	Description	Project	File	Line	Suppression State
    Error	LNK1104	cannot open file 'opencv_core.lib'	opencv_waldboost_detector	C:\work\opencv_3.1\build_static\modules\xobjdetect\tools\waldboost_detector\LINK	1	

Clue:

In the Property Pages of project opencv_waldboost_detector, Linker -> Input -> Additional Dependencies, the content is:

    kernel32.lib
    user32.lib
    gdi32.lib
    winspool.lib
    shell32.lib
    ole32.lib
    oleaut32.lib
    uuid.lib
    comdlg32.lib
    advapi32.lib
    opencv_core.lib
    opencv_imgproc.lib
    opencv_imgcodecs.lib
    opencv_videoio.lib
    opencv_highgui.lib
    ..\..\..\..\lib\Release\opencv_xobjdetect310.lib
    ..\..\..\..\lib\Release\opencv_xfeatures2d310.lib
    ..\..\..\..\lib\Release\opencv_world310.lib
    C:\work\opencv_3.1\opencv\sources\3rdparty\ippicv\unpack\ippicv_win\lib\intel64\ippicvmt.lib
    ..\..\..\..\3rdparty\lib\Release\libjpeg.lib
    ..\..\..\..\3rdparty\lib\Release\libwebp.lib
    ..\..\..\..\3rdparty\lib\Release\libpng.lib
    ..\..\..\..\3rdparty\lib\Release\libtiff.lib
    ..\..\..\..\3rdparty\lib\Release\libjasper.lib
    ..\..\..\..\3rdparty\lib\Release\IlmImf.lib
    ..\..\..\..\3rdparty\lib\Release\zlib.lib
    vfw32.lib
    comctl32.lib
    setupapi.lib
    ws2_32.lib

The 5 lines starting with 'opencv_' looks strange. they lack a path prefix and the version suffix '310', or they should not even exist as there is already '..\\..\\..\\..\\lib\\Release\\opencv_world310.lib'.

PR #547, opened 2016-02-12
HDF: Fix MSVC build. (Issues #546, #497, #552).

cbalint13:
* This patch fixes hdf5 build for MSVC: https://github.com/Itseez/opencv_contrib/issues/546
* At the PR time the msvc buildbot lacked hdf5 lib thus build verification was not possible.

### DIFF
--- a/modules/hdf/CMakeLists.txt
+++ b/modules/hdf/CMakeLists.txt
@@ -1,6 +1,24 @@
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR})
 
-find_package(HDF5)
+if(WIN32)
+  # windows cmake internal lookups are broken for now
+  # will lookup for headers and shared libs given HDF_DIR env
+  find_path(HDF5_INCLUDE_DIRS hdf5.h HINTS "$ENV{HDF5_DIR}\\..\\include")
+  find_library(HDF5_C_LIBRARY NAMES hdf5 HINTS "$ENV{HDF5_DIR}\\..\\lib")
+  if(HDF5_INCLUDE_DIRS AND HDF5_C_LIBRARY)
+    set(HDF5_FOUND "YES")
+    set(HDF5_LIBRARIES ${HDF5_C_LIBRARY})
+    mark_as_advanced(HDF5_LIBRARIES)
+    mark_as_advanced(HDF5_C_LIBRARY)
+    mark_as_advanced(HDF5_INCLUDE_DIRS)
+    add_definitions(-DH5_BUILT_AS_DYNAMIC_LIB -D_HDF5USEDLL_)
+  else()
+     set(HDF5_FOUND "NO")
+  endif()
+else()
+  find_package(HDF5)
+endif()
+
 if(HDF5_FOUND)
     set(HAVE_HDF5 1)
     message(STATUS "HDF5:   YES")
@@ -9,13 +27,13 @@ else()
     message(STATUS "HDF5:   NO")
 endif()
 
-if(${HDF5_FOUND})
+if(HDF5_FOUND)
   include_directories(${HDF5_INCLUDE_DIRS})
 endif()
 
 set(the_description "Hierarchical Data Format I/O")
 ocv_define_module(hdf opencv_core WRAP python)
 
-if(${HDF5_FOUND})
+if(HDF5_FOUND)
   target_link_libraries(opencv_hdf ${HDF5_LIBRARIES})
 endif()
--- a/modules/hdf/src/hdf5.cpp
+++ b/modules/hdf/src/hdf5.cpp
@@ -329,7 +329,7 @@ vector<int> HDF5Impl::dsgetsize( String dslabel, int dims_flag ) const
     int n_dims = H5Sget_simple_extent_ndims( fspace );
 
     // dims storage
-    hsize_t dims[n_dims];
+    hsize_t *dims = new hsize_t[n_dims];
 
     // output storage
     vector<int> SizeVect(0);
@@ -367,6 +367,8 @@ vector<int> HDF5Impl::dsgetsize( String dslabel, int dims_flag ) const
     H5Dclose( dsdata );
     H5Sclose( fspace );
 
+    delete dims;
+
     return SizeVect;
 }
 
@@ -386,7 +388,7 @@ int HDF5Impl::dsgettype( String dslabel ) const
       // fetch channs
       hsize_t ardims[1];
       H5Tget_array_dims( dstype, ardims );
-      channs = ardims[0];
+      channs = (int)ardims[0];
       // fetch depth
       hid_t tsuper = H5Tget_super( dstype );
       h5type = H5Tget_native_type( tsuper, H5T_DIR_ASCEND );
@@ -481,9 +483,9 @@ void HDF5Impl::dscreate( const int n_dims, const int* sizes, const int type,
 
     int channs = CV_MAT_CN( type );
 
-    hsize_t chunks[n_dims];
-    hsize_t dsdims[n_dims];
-    hsize_t maxdim[n_dims];
+    hsize_t *chunks = new hsize_t[n_dims];
+    hsize_t *dsdims = new hsize_t[n_dims];
+    hsize_t *maxdim = new hsize_t[n_dims];
 
     // dimension space
     for ( int d = 0; d < n_dims; d++ )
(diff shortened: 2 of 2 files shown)


cbalint13:
@mshabunin , @vpisarev , @alalek 

 Could please help me with the buildbots, seems that on several hdf5-lib is missing:
```-- Could NOT find HDF5 (missing:  HDF5_LIBRARIES HDF5_INCLUDE_DIRS)```

* Only the **Linux** buildbot have it, so e.g. I cannot proof this PR is valid.
* Would be nice to be available for VS2013, VS2014, Mac (perhaps android too ?!) 

Thanks !

**Comment by alalek** [comment]
Almost all 3rdparty dependencies are optional:
1) There is not easy to maintain them all, especially patched versions.
2) We can't check all possible configurations (in your case, with and without hdf5 libraries), so we check "default" configuration at least.
3) Windows environment is not user friendly to manage 3rdparty dependencies. But I will try to find some simple guide how to install HDF5 on Windows.
4) Missed dependencies should not break existed builds (all is fine here). Users should be able to build OpenCV without installing "unused" 3rdparty software.

**Comment by alalek** [comment]
1) I have installed HDF5 on Win10 slave (MSVS 2015 / Win64 only)
2) Added [valid](https://www.hdfgroup.org/ftp/HDF5/current/src/unpacked/release_docs/USING_HDF5_CMake.txt) `HDF5_DIR` environment variable.

But this doesn't work due some [unexpected failures](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_windows_ten/builds/313/steps/cmake/logs/stdio).

**Comment by cbalint13** [comment]
@alalek ,

For 1,2,3 understood the points, there is no problem. However this PR still can be merged for MSVC users.

 New issue, would require your help:

* Can help me with this issue that appeared today on VS2015 buildbot ?
* Logs are here: http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_windows_ten/builds/313/steps/cmake/logs/stdio

It appears that on VS2015 CMake's internal FindHDF.cmake module see the header files (H5version.h) since its able to detect the version but not the library itself (probably cannot find the path).

```
-- Could NOT find HDF5 (missing:  HDF5_LIBRARIES) (found version "1.8.16")
CMake Warning (dev) at C:/utils/cmake-3.4.1/share/cmake-3.4/Modules/FindHDF5.cmake:234 (get_target_property):
  Policy CMP0045 is not set: Error on non-existent target in
  get_target_property.  Run "cmake --help-policy CMP0045" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.

  get_target_property() called with non-existent target "hdf5".
Call Stack (most recent call first):
  C:/builds/precommit-contrib_windows_ten/opencv_contrib/modules/hdf/CMakeLists.txt:3 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
-- Checking SFM deps... FALSE
```

**Comment by alalek** [comment]
@cbalint13 I installed HDF5 on Windows slave as requested, errors(warnings) are coming from this change.

**Comment by alalek** [comment]
@cbalint13 Could you replace `find_package(HDF5)` statement to this?
```
find_package (HDF5 NAMES hdf5 COMPONENTS C static)
```

**Comment by cbalint13** [comment]
@alalek ,


* I did not saw that you added HDF5 to VS2015 build. As described by me (upper) I think cmake cannot see the very library at all, but sees the header files. In my experience HDF5_LIBRARIES is needed to be declared too manually, not only the generic HDF_DIR.

**Comment by cbalint13** [comment]
@aalek,
* I replace this now: ```find_package (HDF5 NAMES hdf5 COMPONENTS C static)```

Lets see.

**Comment by cbalint13** [comment]
@alalek 

* find_package (HDF5 NAMES hdf5 COMPONENTS C static) fails on local:
```
CMake Warning at opencv_contrib/modules/hdf/CMakeLists.txt:3 (find_package):
  Could not find a package configuration file provided by "HDF5" with any of
  the following names:

    hdf5Config.cmake
    hdf5-config.cmake

  Add the installation prefix of "HDF5" to CMAKE_PREFIX_PATH or set
  "HDF5_DIR" to a directory containing one of the above files.  If "HDF5"
  provides a separate development package or SDK, be sure it has been
  installed.
```
There are problems with such statement:
* 1 CMake internal is called `/usr/share/cmake/Modules/FindHDF5.cmake` not hdf5Config.cmake.
* 2 Forcing ```static``` would break linux distros where hdf comes in dynamic form.

This looks fine: ```find_package(HDF5 COMPONENTS C)```
I believe buildbot can be curred only if HDF5_LIBRARIES is declared manually in case of MSVC build.

**Comment by alalek** [comment]
@cbalint13 Please note: I follow an official guide of HDF5 library and use `HDF5_DIR`:
https://www.hdfgroup.org/ftp/HDF5/current/src/unpacked/release_docs/USING_HDF5_CMake.txt

It is not easy to pass `HDF5_LIBRARIES` on slave via setting environment variable: http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_windows_ten/builds/319/steps/cmake/logs/stdio
```
HDF5_LIBRARIES=C:/utils/HDF5/1.8.16/lib
...
Could NOT find HDF5 (missing:  HDF5_LIBRARIES HDF5_INCLUDE_DIRS)
```
Also it can't be passed via `cmake` command line due some problems (build commands are shared between slaves (we have several Windows slaves), so absolute paths are smell there). Also there is our policy of testing "default" setup, so commands should be short as possible.

You can try to play with it (insert before find package):
```
ocv_check_environment_variables(HDF5_LIBRARIES HDF5_INCLUDE_DIRS)
```

Also this  approach works on Linux:
```
find_package (HDF5 QUIET NAMES hdf5 COMPONENTS C static)
if(NOT HDF5_FOUND)
  find_package(HDF5 QUIET)
endif()
```

Anyway I believe this switch helps in all other cases:
```
if (WIN32)
  ...
else()
  ...
endif()
```

**Comment by cbalint13** [comment]
@alalek , 

* First, many thanks for the time spent helping me !
* I'll try tackle more as suggested, plus few my ideas, but will end up with IF WIN32 things if I fail.
* Will look into a real MSVC  machine too to better understand the failure.

Be back,
~cristian.

**Comment by cbalint13** [comment]
@alalek ,

* Can put back the package to the path C:/utils/HDF5/1.8.16 ?
* Can send me in private the full dir layout of ```ls -lR C:/utils/HDF5/1.8.16/*``` ?

**Comment by alalek** [comment]
Done: [hdf5_setup.txt](https://github.com/Itseez/opencv_contrib/files/131006/hdf5_setup.txt)

**Comment by cbalint13** [comment]
@alalek ,

* Can remove enviroment "HDF5_DIR=C:/utils/HDF5/1.8.16/cmake" ?

**Comment by cbalint13** [comment]
@alalek ,

Lets have HDF5_DIR in final, don't remove:

* Here we go: https://github.com/cbalint13/opencv_contrib/commit/b9e6e0ebe671445f4d2bbfc997a0a19cde2b24bf , i just mimic official HDF5_DIR way on MSVC.
* I hope will be acceptable to declare cmake's internal HDF5 routine broken on MSVC.
* Seems that now have in buildbot 'vars' all the proper paths:
```
HDF5_INCLUDE_DIRS=C:/utils/HDF5/1.8.16/include
HDF5_LIBRARY=C:/utils/HDF5/1.8.16/lib/hdf5.lib
```
I'll wait until it builds fine.
On any other issues will fix (i expect none but it may require szlib.lib dependency).

cbalint13:
@alalek ,

By hereby patch I propose a simple fix for MSVC land. Few thoughts over HDF5 & CMake especially on MSVC:

* CMake internal FindHDF5.cmake is clearly broken, it even don't handle HDF5_DIR at all, i checked it source.

* In contrast CMake internal FindHDF5.cmake may work better, but booth in fact are looking over env${PATH} as HINT to find folders, but not fully working if one tries it out as documentation states. The idea is to install their .msi installer and just to work out of the box via installed system PATH. Doesn't work either (only partially), at least for several latest hdf5 binary releases.

* Original FindHDF5.cmake from hdf5 tarball (might be better then CMake's) package seems ever changing thus would be hard to maintain, not counting that I believe is over-sophisticated.
 
All three statements are valid even for older hdf5 binary release as all my tests concluded. My final conclusion is to handle with separate simple & maintainable routines via env{HDF5_DIRS} just like actual patch does. Or perhaps even accounting the PATH environment that is always settled by .msi installers. Actual patch can be extended with ease to handle more HINTS if will be necessary at any time.

Let me know if would like to add to HINT the +env{PATH} too (a bit regexped and filtered). Perhaps it is a good idea to have booth HDF5_DIR and PATH lookup.

alalek:
:+1:

PR #550, opened 2016-02-14
Add bilateral texture filter code

jhlee525:
Bilateral texture filter [1], along with its test code, is added.

Bilateral texture filter is one of its kind for texture smoothing and edge preserving. This method blurs high-frequency texture area while preserving structural edges including object borders. It is done with a simple modification of the original bilateral filter. This method can be applicable to many other researches including feature extraction, depth image noise reduction and detail manipulation. This method was accepted in Siggraph 2014 as an oral presentation.

[1] Cho et al., Bilateral Texture Filtering, in proceedings of SIGGRAPH 2014


modules/ximgproc/perf/perf_bilateral_texture_filter.cpp
```diff
@@ -0,0 +1,83 @@
...
+typedef TestBaseWithParam<BTFTestParam> BilateralTextureFilterTest;
+
+PERF_TEST_P(BilateralTextureFilterTest, perf,
+    Combine(
+    Values(2, 3),
+    Values(0.5, 1.0),
+    Values(0.5, 1.0),
+    SZ_TYPICAL,
+    Values(CV_8U, CV_32F),
+    Values(1, 3))
```
alalek: Could you please reduce number of performance tests? 128 tests are too much for single function.

jhlee525: I reduced number of test to 4. Thank you for comment.

modules/ximgproc/test/test_bilateral_texture_filter.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+INSTANTIATE_TEST_CASE_P(
+  TypicalSet1,
+  BilateralTextureFilterTest,
+  Combine(
+    Values(2, 3),
+    Values(0.5, 1.0),
+    Values(0.5, 1.0),
+    Values(CV_8U, CV_32F),
+    Values(1, 3)
```
alalek: Could you please reduce this test set also (it runs for 2 min - it is about 50% of whole [test_ximgproc](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_linux64/builds/2293/steps/test_ximgproc))

alalek:
Looks good to me! Thanks!
Could you please squash all commits into one (intermediate history is not necessary)?

PR #554, opened 2016-02-18
Aruco: completed wrapping for Python bindings

cr333:


paroj:
:+1:

adrianheron:
I was using the aruco bindings in your branch (thank you!) and encountered an issue with the estimatePoseSingleMarkers function.  I've tried to describe it here https://github.com/Itseez/opencv_contrib/issues/574

cr333:
The bindings for estimatePoseSingleMarkers already existed before my changes, and I didn't try them myself.

jfolz:
Good job on finding all the stuff that I missed!
One question: Do the Dictionary properties still work with ``CV_PROP`` instead of ``CV_PROP_RW``? I encountered some problems with accessibility from Python using the former.
I have also yet to find a good explanation for how they differ. Maybe you know more?

cr333:
@jfolz Sorry, I'm not really familiar with the internals of the wrapping process either. My reference was http://docs.opencv.org/3.1.0/da/d49/tutorial_py_bindings_basics.html

It suggests that `CV_PROP` is used for fields of classes wrapped with `CV_EXPORTS_W`, and `CV_PROP_RW` for `CV_EXPORTS_W_SIMPLE` (small classes passed by value) or `CV_EXPORTS_W_MAP` (wrapped as dictionary), which is the case you're interested in.

jfolz:
@cr333 Yes, that's my source as well and I find the explanation a bit lacking. Looking at the parser and generator code suggests that there is exactly one difference: ``CP_PROP_RW`` creates a writeable property and ``CV_PROP`` does not. You most likely don't need writeable properties for the Dictionary in Python, Java etc., but its properties are public in C++, so why not?

paroj:
@alalek

mshabunin:
:+1:

PR #555, opened 2016-02-18
Add option to create GridBoard starting at user-specified marker id

SSteve:
For our application, we want to be able to make a number of uniquely-identifiable boards. I added an optional parameter to GridBoard::create to make this possible. Omitting the optional parameter reproduces existing behavior.


alalek:
Please squash all commits into one (to eliminate unnecessary intermediate changes from git history).

SSteve:
Done.

alalek:
:+1:

PR #556, opened 2016-02-20
Compiler error for VC10 removed

polarfuchs99:
modified:   modules/line_descriptor/perf/perf_matching.cpp
	modified:   modules/line_descriptor/test/test_matcher_regression.cpp

If have observed two compiler errors when compiling with MS Visual Studio 2010. The 
reason is the function  pow( 2, i ) . The first argument should not be an integer but rather 
a float or double number. For this reason, I have replaced this function by  pow( 2.0, i ) ,
I have done this in two cases. The changes should not have any effect on the calculation.

### DIFF
--- a/modules/line_descriptor/perf/perf_matching.cpp
+++ b/modules/line_descriptor/perf/perf_matching.cpp
@@ -101,7 +101,7 @@ uchar invertSingleBits( uchar dividend_char, int numBits )
   /* reconvert to decimal */
   uchar result = 0;
   for ( int i = (int) bin_vector.size() - 1; i >= 0; i-- )
-    result += (uchar) ( bin_vector[i] * pow( 2, i ) );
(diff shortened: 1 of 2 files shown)


alalek:
Any ideas why "(2 << i)" can't be used here?

polarfuchs99:
@alalek 
This is a good idea. However, it must be (1<<i). I have checked that pow(2,i) is 
equivalent to (1<<i). Thus I have made one more change to replace  pow( 2.0, i )
by ( 1 << i ). This has been done in the two cases as before.

polarfuchs99:
@alalek 
This is a good idea. However, it must be (1<<i). I have checked that pow(2,i) is 
equivalent to (1<<i). Thus I have made one more change to replace  pow( 2.0, i )
by ( 1 << i ). This has been done in the two cases as before.

alalek:
:+1:

Issue #166, opened 2015-01-31
CMakeLists.txt in cvv module


TNick:
```
-- General configuration for OpenCV 3.0.0-dev 
--   Version control:               3.0.0-beta-466-g97aa8d3
-- 
--   Platform:
--     Host:                        Windows 6.2 AMD64
--     CMake:                       3.1.0-rc1
--     CMake generator:             NMake Makefiles
--     CMake build tool:            nmake
--     MSVC:                        1800
--     Configuration:               Debug
```
Because MSVC compiler has only partial support for c++11 the compilation of *cvv module* fails on Windows with this compiler.

As a side note the flags
```cmake
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra -pedantic")
```
should be changed to something like:
```cmake
if(MSVC)
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W3")
endif()
if(CMAKE_COMPILER_IS_GNUCC)
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra -pedantic")
endif()
```
in [CMakeLists.txt](https://github.com/Itseez/opencv_contrib/blob/master/modules/cvv/CMakeLists.txt).

PS Some keywords to help Google locate this bug report:
`... is not a special member function which can be defaulted`

nstiurca:
I patched things up to compile with Visual Studio 2013 and Intel 16.0 compiler. I will try and send a pull request this weekend.

nstiurca:
This should be closed now that #557 was merged.

PR #557, opened 2016-02-23
some fixes to get CVV module compiling with Visual Studio 2013

nstiurca:
This should address #166 . I tested it on Windows 10 with Visual Studio 2013 with both Microsoft and Intel compilers. I have not tested the patch on other OS's, but hopefully the build bot will succeed.


alalek:
Actually buildbot doesn't test this module, because there is no QT5 on Windows slaves.

alalek:
:+1:

PR #559, opened 2016-02-27
OpenCV dnn

miraj7789:
Opencv dnn


alalek:
You need to remove unrelated commits.

PR #566, opened 2016-03-05
Fixed segmentation fault problem of sift.cpp

kmyid:
Boundary check for orientation histogram calculation was wrong.


kmyid:
Sorry, did not check thoroughly. The original code had no problems.

PR #575, opened 2016-03-12
Pyramid

LaurentBerger:
Hi,

This module is used to play with pyramid structure. I have never created a module and this one is not finished. But I need a reviewer to check first if it is useful and based class is compliant with opencv.


modules/pyramid/src/pyramid.cpp
```diff
@@ -0,0 +1,321 @@
+#include "precomp.hpp"
+
+using namespace std;
+using namespace cv;
```
berak: the code should not be `using name space cv; using namespace pyramid;` but should be inside cv::pyramid

like:

    namespace cv {
    namespace pyramid {
    ... // code
    }
    }

this is also, why the linker does not find `cv::pyramid::LaplacianPyramid::LaplacianPyramid(Mat &,int)` and others

LaurentBerger: Thanks for your comment.
I think that I didn't understand your comment. I have tried  
`namespace cv
{
namespace pyramid
{

/* Reference
This code is patened http://www.google.com/patents/US20140072229
1 ICCP 2014 Riesz Pyramids for Fast Phase-Based Video Magnification  http://people.csail.mit.edu/nwadhwa/riesz-pyramid/
2 In "Supplemental for Riesz Pyramids for Fast Phase-Based Video Magnification" supplemental material : http://people.csail.mit.edu/nwadhwa/riesz-pyramid/RieszPyrSupplemental.zip


*/

Pyramid::Pyramid(Pyramid const &p)
{
`
but there is still a link error. I have got VS2013 and 2015 and there is no error. I don't know how to reproduce this error. Don't you know an option in VS to have same error?

LaurentBerger:
@alalek Is it possible to use this code http://www.iowahills.com/ inside opencv and this one https://www-users.cs.york.ac.uk/~fisher/software/mkfilter/Vsn1/mkfilter.c
thanks for your answer

alalek:
Not sure, I can't find a normal license description on these links.

LaurentBerger:
@alalek 
about iowallis "Filter design was one of my specialties as an RF engineer. Now that I am retired, I work on these filter design programs when it is too cold in Iowa to ride bicycle. The programs are free." on first page and I'm waiting an answer from this site.

About mkfilter author is dead (https://www-users.cs.york.ac.uk/~fisher/)  and it's free for for non-profit making purposes only.

LaurentBerger:
@alalek There is no problem with iowallis.com I have got an answer from DanielK@IowaHills.com. If you want I can ask him to insert a license in his code.

PR #579, opened 2016-03-15
Kcf

gongzg:
This PR is to optimize the kcf algorithm and mainly consist of two optimization:
1. Use float data type rather than double type which get about 20% performance gain.
2. Use OCL to optimize a transpose matrix mutiplication in updateProjectMatrix which get about 10% - 20% performance gain on Intel iGPU platform.


arrybn:
Hi, @gongzg! Could you resolve conflicts between your branch and current master?

Issue #581, opened 2016-03-15
Add an option to show points only for plot module


jasjuang:
It would be great to have an option to not connect all the data points to form a line but simply show as a point at their coordinates (perhaps as a circle or cross) in the plot.

This is super helpful if the data is not a continuous function and/or randomly distributed.

PR #582, opened 2016-03-15
add in option to show points only for plot module

jasjuang:
In order to solve https://github.com/Itseez/opencv_contrib/issues/581

If the input of setPlotLineWidth is not larger than zero, show data as points only instead of connecting them to form a line

### DIFF
--- a/modules/plot/src/plot.cpp
+++ b/modules/plot/src/plot.cpp
@@ -199,23 +199,35 @@ namespace cv
                 double CurrentX = plotDataX.at<double>(NumVecElements-1,0);
                 double CurrentY = plotDataY.at<double>(NumVecElements-1,0);
 
-                //Draw the plot by connecting lines between the points
-                Point p1;
-                p1.x = (int)InterpXdata.at<double>(0,0);
-                p1.y = (int)InterpYdata.at<double>(0,0);
-
                 drawAxis(ImageXzero,ImageYzero, CurrentX, CurrentY, plotAxisColor, plotGridColor);
 
-                for (int r=1; r<InterpXdata.rows; r++){
+                if(plotLineWidth > 0)
+                {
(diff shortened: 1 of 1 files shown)


StevenPuttemans:
Could you also document this option in the function description?

jasjuang:
could you let me know where exactly should I add the documentation? Is it plot.hpp?

StevenPuttemans:
Sorry for the delay, kind of forgot about this. Yes documentation goes above functions inside the header file. So indeed, you need to update plot.hpp to incorporate information about this.

jasjuang:
Just to make sure I am looking at the right file, is it opencv_contrib/modules/plot/include/opencv2/plot.hpp that I should update? The weird part is there is no documentation for the original functions

StevenPuttemans:
Ow that is actually correct ... http://docs.opencv.org/master/d0/d1e/classcv_1_1plot_1_1Plot2d.html#gsc.tab=0

Would you mind adding documentation for the complete plot module if you are up to it?  And yes that is the file. It should start to look like other classes, e.g. https://github.com/Itseez/opencv_contrib/blob/master/modules/aruco/include/opencv2/aruco.hpp

If not give me a shout and I will add docs for those classes!

jasjuang:
I prefer not to add the entire documentation because I am not familiar with the entire code. Please go ahead and do it!

StevenPuttemans:
I will add it to my todo list!
@nmoutinho are you maybe up to helping us out with some documentation on the main plot module?

nmoutinho:
Hi @StevenPuttemans. I'll do that. No problem ;)

StevenPuttemans:
@nmoutinho thank you for doing that!

sovrasov:
Replaced by #787

Issue #548, opened 2016-02-12
roiSelector function seems to be platform dependent (Tracking module)


benv2k6:
The function roiSelector used in the tracking module seems to be platform dependent since the following code while(!(key==32 || key==27 || key==13)) is considering these integers to represent the keys Enter, Esc and Space. For example, on my Ubutnu 14.04 64bit system, Esc= 1048603, Space = 1048608

Aravind-Suresh:
The lower 8 bits of those numbers match ( 27 and 1048603, 32 and 1048608 ). I think comparing `lower 8/16 bits of key` will fix the problem. Please tell me whether this is correct. I will go ahead and fix this.

kurnianggoro:
@Aravind-Suresh I guess that also the correct solution, you can try modify all occurrence of key==27 to key== 27&0xFF or (uint8_t)key== 27, but make sure you test it on Ubuntu 64bit system, or maybe @benv2k6 you can try to change it on your system and then make pull request for the bug fix, thank you for mentioning the bug :)

PR #593, opened 2016-03-24
roiSelector platform dependency bug

Aravind-Suresh:
Added lower 8-bit comparison for key presses.

Fixes https://github.com/Itseez/opencv_contrib/issues/548


mshabunin:
:+1:

PR #607, opened 2016-03-27
aruco::estimatePoseSingleMarkers rvec/ tvec memory layout fixes

### LINKED ISSUE #574: python binding for aruco estimatePoseSingleMarkers broken
Configuration:
opencv 3.1.0
opencv_contrib #554
python 2.7.8 (tried 32 and 64 bit)
tried Windows and Ubuntu

I've been trying to use the python bindings for aruco with some success but when calling the estimatePoseSingleMarkers function it crashes on https://github.com/Itseez/opencv_contrib/blob/3.1.0/modules/aruco/src/aruco.cpp#L866 with:
```cv2.error: ...\opencv\modules\core\src\matrix.cpp:1201: error: (-215) 0 <= i && i < (int)v.size() in function cv::_InputArray::getMat_```

There appears to be some problem with the way the OutputArrayofArrays _rvecs and _tvecs are handled.  After some trial and error I was able to make it work with the following code, but I'm pretty sure this is not the right way to do it.  It would be great if someone who knew what they were doing had a look!
```c++
void estimatePoseSingleMarkers(InputArrayOfArrays _corners, float markerLength,
                               InputArray _cameraMatrix, InputArray _distCoeffs,
                               OutputArrayOfArrays _rvecs, OutputArrayOfArrays _tvecs) {

    CV_Assert(markerLength > 0);

    Mat markerObjPoints;
    _getSingleMarkerObjectPoints(markerLength, markerObjPoints);
    int nMarkers = (int)_corners.total();
    
	//_rvecs.create(nMarkers, 1, CV_64FC3);
    //_tvecs.create(nMarkers, 1, CV_64FC3);

	// This is the old line that crashed
    //Mat rvecs = _rvecs.getMat(), tvecs = _tvecs.getMat();

	// Instead, create regular Mats to pass into the next call
	Mat rvecs(nMarkers, 1, CV_64FC3);
	Mat tvecs(nMarkers, 1, CV_64FC3);
	
    //// for each marker, calculate its pose
    // for (int i = 0; i < nMarkers; i++) {
    //    solvePnP(markerObjPoints, _corners.getMat(i), _cameraMatrix, _distCoeffs,
    //             _rvecs.getMat(i), _tvecs.getMat(i));
    //}

    // this is the parallel call for the previous commented loop (result is equivalent)
    parallel_for_(Range(0, nMarkers),
                  SinglePoseEstimationParallel(markerObjPoints, _corners, _cameraMatrix,
                                               _distCoeffs, rvecs, tvecs));


	// Then copy the new Mats to the Output arguments
	vector< Mat > vrvecs;
	vrvecs.push_back(rvecs);
	_copyVector2Output(vrvecs, _rvecs);
	vector< Mat > vtvecs;
	vtvecs.push_back(tvecs);
	_copyVector2Output(vtvecs, _tvecs);
}
```

paroj:
type should be OutputArray instead of OutputArrayOfArrays.
at<Vec3d> is one dimensional. (fixes debug assert)
fixes #574

### DIFF
--- a/modules/aruco/include/opencv2/aruco.hpp
+++ b/modules/aruco/include/opencv2/aruco.hpp
@@ -192,9 +192,9 @@ CV_EXPORTS_W void detectMarkers(InputArray image, Ptr<Dictionary> &dictionary, O
  * \f$A = \vecthreethree{f_x}{0}{c_x}{0}{f_y}{c_y}{0}{0}{1}\f$
  * @param distCoeffs vector of distortion coefficients
  * \f$(k_1, k_2, p_1, p_2[, k_3[, k_4, k_5, k_6],[s_1, s_2, s_3, s_4]])\f$ of 4, 5, 8 or 12 elements
- * @param rvecs array of output rotation vectors (@sa Rodrigues) (e.g. std::vector<cv::Vec3d>>).
+ * @param rvecs array of output rotation vectors (@sa Rodrigues) (e.g. std::vector<cv::Vec3d>).
  * Each element in rvecs corresponds to the specific marker in imgPoints.
- * @param tvecs array of output translation vectors (e.g. std::vector<cv::Vec3d>>).
+ * @param tvecs array of output translation vectors (e.g. std::vector<cv::Vec3d>).
  * Each element in tvecs corresponds to the specific marker in imgPoints.
  *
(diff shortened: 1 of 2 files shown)


alalek:
:+1:

PR #616, opened 2016-03-28
Extensive changes to master README.md

garybradski:
Extensive changes to README.md opencv_contrib file to bring it up to date.

### DIFF
--- a/modules/README.md
+++ b/modules/README.md
@@ -1,59 +1,71 @@
-An overview of the contrib modules and a small explanation
-----------------------------------------------------------
+An overview of the opencv_contrib modules
+-----------------------------------------
 
 This list gives an overview of all modules available inside the contrib repository.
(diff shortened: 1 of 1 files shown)


alalek:
Could you please fix [trailing whitespace issues](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_docs/builds/2316/steps/whitespace%20opencv_contrib/logs/stdio)

StevenPuttemans:
:+1: thank you :D since the addition of this files, it was indeed hard to force people to update it or make it match once they add modules. I still think this should be enforced by assignees every time some commits extra functionality OR a new module, because for now, this is still not the case.

vpisarev:
@garybradski,
it looks like there are some extra whitespaces. Click on the details and then on the failed doc builder to see where they are

PR #620, opened 2016-04-01
no trailing rect capture on escape

ezavesky:
- no ambiguity about last rect (complement #459 and bd20972847615184811553768e2d35b3c4c78451)
also [answer self request](https://github.com/Itseez/opencv_contrib/pull/539#commitcomment-16864422)

### DIFF
--- a/modules/tracking/src/roiSelector.cpp
+++ b/modules/tracking/src/roiSelector.cpp
@@ -161,8 +161,9 @@ namespace cv {
     printf("Finish the selection process by pressing ESC button!\n" );
 
     // while key is not ESC (27)
-    while(key!=27){
+    for(;;) {
       temp=select(windowName, img, true, fromCenter);
+      if(key==27) break;
       if(temp.width>0 && temp.height>0)
         box.push_back(temp);
     }


modules/tracking/src/roiSelector.cpp
```diff
@@ -163,7 +163,7 @@ namespace cv {
     // while key is not ESC (27)
     while(key!=27){
       temp=select(windowName, img, true, fromCenter);
-      if(temp.width>0 && temp.height>0)
+      if(key!=27 && temp.width>0 && temp.height>0)
```
StevenPuttemans: Isn't this enforced by the while loop? It is a matter of micro seconds of difference at most ...

alalek: I was catched on this issue also, but all is fine here: `key` is field of `ROISelector` class and it is changed during `select` call (another `select` method).

BTW, There is "nanoseconds" optimization (removed while condition):
```
// remove key = 0 on line 157
...
while(true)
{
    temp=select(windowName, img, true, fromCenter);
    if (key == 27)
        break;
    if(temp.width>0 && temp.height>0)
        box.push_back(temp);
}
```

StevenPuttemans: Ow well I didnt look close enough to the fact that there was indeed a function call that could change the key value. In that case it would indeed be better to switch the checks to after the function call like you suggested. That solves the problem at all.

ezavesky: Thanks!  I just pushed a cleaner version with your suggestion.  You're right that some people would never be caught by this, but when I handed the controls over to a novice, they were consistently confused about getting an extra rect if they had selected something *and then* hit escape.

modules/tracking/src/roiSelector.cpp
```diff
@@ -161,8 +161,9 @@ namespace cv {
     printf("Finish the selection process by pressing ESC button!\n" );
 
     // while key is not ESC (27)
-    while(key!=27){
+    while(true) {
```
alalek: MSVC compiler dislikes this, please use `for (;;)` here. Thanks!

ezavesky: What a challenging place ;)  Done.

alalek:
Thanks! Could you please remove unrelated file from the last commit.

ezavesky:
Indeed.  Last change wasn't properly stashed before commit.

PR #627, opened 2016-04-06
quick fix to avoid CMake 3.5 crashing in cl2cpp.cmake

amroamroamro:
see https://github.com/Itseez/opencv/issues/6370


alalek:
Thanks for investigation and workaround!
:+1:

Issue #630, opened 2016-04-14
Contrib modules getting built perfectly fine, even when CMAKE says their dependencies are not available


StevenPuttemans:
Just grabbed latest master from both main and contrib repository and tried to build all the contrib modules also. What baffled me is the following.

The Cmake procedure clearly looks for dependencies and states that for example the following dependencies are not found

```
Caffe:   NO
Tesseract:   NO
```

And there comes the catch, it then states, ow lets build the modules depending on it

```
To be built:                 cudev core cudaarithm flann imgproc ml reg surface_matching video cudabgsegm cudafilters cudaimgproc cudawarping dnn fuzzy imgcodecs photo shape videoio cudacodec highgui objdetect plot ts xobjdetect xphoto bgsegm bioinspired dpm face features2d line_descriptor saliency text calib3d ccalib cudafeatures2d cudalegacy cudaobjdetect cudaoptflow cudastereo datasets rgbd stereo structured_light superres tracking videostab xfeatures2d ximgproc aruco optflow stitching matlab python2
    Disabled:                    world contrib_world
    Disabled by dependency:      -
    Unavailable:                 java python3 viz cvv hdf sfm
```

Which means in my opinion that the CMake is wrongly configured. 

So my suggestions

- If dependencies are not found on the system, do not build the modules itself. That will force people to install the dependencies first and ensure that they know what they are doing.

- A small notice is that the `Tesseract:   NO` is being projected twice in the output, so I am guessing it is checked twice also, which is redundant.

alalek:
It is just a report of checks. It is not error report.
Most part of OpenCV dependencies is optional.

> If dependencies are not found on the system, do not build the modules itself

It works if dependency is required for module, for example, see "hdf" or "viz" module (or set of cuda modules).

> Tesseract:   NO

Agree, double check is redundant.
"text" module is available in this case, but it is not fully functional (it is not disabled totally because there was an assumption that there are many detection backends and "tesseract" is only one of them)

StevenPuttemans:
I know they are checks. So if I understand it correctly, even though torch and caffe are for example not on my system, i am still able to completely use the dnn module?

alalek:
In general, it depends on module and its design.
For example, dnn module can read caffe models and execute forward pass processing without using of anything from caffe (dnn module doesn't require caffe at all).

StevenPuttemans:
:+1: okay then I guess simply providing a small fix for the redundant check would be enough for closing this down!

[...]

PR #632, opened 2016-04-18
fix the CMAKE order for issue #630

StevenPuttemans:
Ensure that the double pass through CMakeLists.txt files do not generate double output
More details discussed in issue https://github.com/Itseez/opencv_contrib/issues/630


StevenPuttemans:
@alalek these are the 4 modules in opencv_contrib suffering from this issue

StevenPuttemans:
@alalek it seems that the proposed fixed where not as straightforward as considered :P it leads to a shitload of errors. Any idea why?

modules/cvv/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
+set(the_description "Debug visualization framework")
+ocv_define_module(cvv opencv_core opencv_imgproc opencv_features2d ${CVV_LIBRARIES} WRAP python)
```
alalek: This module requires QT5, so there 2 lines should be placed after next `if` block

modules/cnn_3dobj/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
+set(the_description "CNN for 3D object recognition and pose estimation including a completed Sphere View on 3D objects")
+ocv_define_module(cnn_3dobj opencv_core opencv_imgproc opencv_viz opencv_highgui OPTIONAL WRAP python)
```
alalek: This module requires caffe. See the line 27: `if(HAVE_CAFFE)`

modules/hdf/CMakeLists.txt
```diff
@@ -19,6 +19,9 @@ else()
   find_package(HDF5)
 endif()
 
+set(the_description "Hierarchical Data Format I/O")
+ocv_define_module(hdf opencv_core WRAP python)
```
alalek: Should be after block with `ocv_module_disable(hdf)` statement (line 26/29)

alalek:
:+1:

modules/cvv/CMakeLists.txt
mshabunin: I think this file should be reverted:

- variable `CVV_LIBRARIES` is used before being set
- other preparation steps (`add_definitions`, `include_directories`, set compiler flags)  do not affect some subtargets, probably leading to #659

Issue #640, opened 2016-04-21
KCF tracker returns wrong bounding box in Python


alessiodore:
When using the KCF tracker from Python the returned bounding box always has size=(0,0) and the position is wrong. For example:

    import cv2
    firstFrame = True
    pathVid =  '<path to vid or im seq>' # e.g. vidName.mp4 or images/%04.jpg (if image name is 0001.jpg...)
    vidReader = cv2.VideoCapture(pathVid)
    initBbox = (30, 60, 60, 70 )  # (x_tl, y_tl, w, h)
    tracker = cv2.Tracker_create('KCF')
    while vidReader.isOpened():
        ok, image=vidReader.read()
        if firstFrame:
            ok = tracker.init(image, initBbox)
            firstFrame = False
        ok, bbox = tracker.update(image)
        print ok, bbox

prints wrong bbox coordinates and width and height are always 0. The same code with:

    tracker = cv2.Tracker_create('MIL')

works.

alessiodore:
The problem is in the updateImpl function. When the function is called from Python from:

    tracker.update(image)

the variable boundingBox which is the output of the tracker is initialize with 0s whereas in the C++ implementation the variable is passed by reference. Therefore the bounding box update implemented as:

       // update the bounding box
      boundingBox.x=(resizeImage?roi.x*2:roi.x)+boundingBox.width/2;
      boundingBox.y=(resizeImage?roi.y*2:roi.y)+boundingBox.height/2;  

will give wrong values in Python as width and height are 0s.

HaiyangLiu711:
DetectMultiScale (InputArray image, vector <rect> & objects, double scaleFactor = 1.1, int minNeighbors = 3, int flags = 0, Size minSize = Size (), Size maxSize = Size ()Where the target is vector <rect> type, however，the tracking function MultiTracker :: add (const String & trackerType, const Mat & image, const Rect2d & boundingBox) is the target of the Rect2d type. Do you know how to deal with it?
thank you very much

PR #641, opened 2016-04-21
Fix KCF tracker when used in Python

alessiodore:
Description of the issue in https://github.com/Itseez/opencv_contrib/issues/640

To solve the problem I used the member variable roi to set the boundingBox variable  as follows:

    boundingBox.x=(resizeImage?roi.x*2:roi.x)+(resizeImage?roi.width*2:roi.width)/4;
    boundingBox.y=(resizeImage?roi.y*2:roi.y)+(resizeImage?roi.height*2:roi.height)/4;
    boundingBox.width = (resizeImage?roi.width*2:roi.width)/2;
    boundingBox.height = (resizeImage?roi.height*2:roi.height)/2;

It is worth noticing that roi width and hight are double the boundingBox size as roi is used as searching area for the next frame. 
I have moved the update bounding box code ouside the block if( frame > 0) because in the examples:

modules/tracking/samples/tracker.py
modules/tracking/samples/multitracker.py

the update function is called after init even in the first frame.

### DIFF
--- a/modules/tracking/src/trackerKCF.cpp
+++ b/modules/tracking/src/trackerKCF.cpp
@@ -336,13 +336,13 @@ namespace cv{
       minMaxLoc( response, &minVal, &maxVal, &minLoc, &maxLoc );
       roi.x+=(maxLoc.x-roi.width/2+1);
       roi.y+=(maxLoc.y-roi.height/2+1);
-
-      // update the bounding box
-      boundingBox.x=(resizeImage?roi.x*2:roi.x)+(resizeImage?roi.width*2:roi.width)/4;
-      boundingBox.y=(resizeImage?roi.y*2:roi.y)+(resizeImage?roi.height*2:roi.height)/4;
-      boundingBox.width = (resizeImage?roi.width*2:roi.width)/2;
-      boundingBox.height = (resizeImage?roi.height*2:roi.height)/2;
     }
+
+    // update the bounding box
+    boundingBox.x=(resizeImage?roi.x*2:roi.x)+(resizeImage?roi.width*2:roi.width)/4;
+    boundingBox.y=(resizeImage?roi.y*2:roi.y)+(resizeImage?roi.height*2:roi.height)/4;
+    boundingBox.width = (resizeImage?roi.width*2:roi.width)/2;
+    boundingBox.height = (resizeImage?roi.height*2:roi.height)/2;
 
     // extract the patch for learning purpose
     // get non compressed descriptors
--- a/modules/tracking/src/trackerKCF.cpp
+++ b/modules/tracking/src/trackerKCF.cpp
@@ -338,8 +338,10 @@ namespace cv{
       roi.y+=(maxLoc.y-roi.height/2+1);
 
       // update the bounding box
-      boundingBox.x=(resizeImage?roi.x*2:roi.x)+boundingBox.width/2;
-      boundingBox.y=(resizeImage?roi.y*2:roi.y)+boundingBox.height/2;
+      boundingBox.x=(resizeImage?roi.x*2:roi.x)+(resizeImage?roi.width*2:roi.width)/4;
+      boundingBox.y=(resizeImage?roi.y*2:roi.y)+(resizeImage?roi.height*2:roi.height)/4;
+      boundingBox.width = (resizeImage?roi.width*2:roi.width)/2;
+      boundingBox.height = (resizeImage?roi.height*2:roi.height)/2;
     }
 
     // extract the patch for learning purpose


roganov:
Don't know why, but your patch didn't fix the bug for me.

alessiodore:
@roganov Can you make sure the version with the fix is correctly compiled and linked?

roganov:
@alessiodore Sorry, I indeed failed to recompile properly. Now working as expected, thanks for the patch.

AlexTS1980:
I followed all your suggestions, ran cmake and make on opencv 3.1.0 and keep getting the exact same error, including initialization at (0,0). Any suggestions?

alessiodore:
It is a bit difficult to understand what is wrong. Possibly the patch version is not compiled or linked correctly. Does you cmake configure the project correctly to build opencv_contrib too where the tracker is? Are you compiling my fork of opencv_contrib or did you edit opencv_contrib yourself?

AlexTS1980:
If you mean this fork: 
https://github.com/alessiodore/opencv_contrib/blob/master/modules/tracking/src/trackerKCF.cpp

then I don't see anything different here from the ones in the original repository. Where did you add the changes

alessiodore:
You can see the changes in the ["File changed" tab](https://github.com/opencv/opencv_contrib/pull/641/files)

AlexTS1980:
Ok, downloaded your fork, chaned the code in trackerKCF.cpp o te one you wrote and fully recompiled opencv 3.1.0 with our contrib module, but nothing changes. Here's the code snippet

import numpy as np
import cv2
cap = cv2.VideoCapture('snippet1.mp4')
track_window = (c,r,w,h)
tracker = cv2.Tracker_create('KCF')
init_once = False
while(True):
   ret, frame = cap.read()
   if not init_once:
        ok = tracker.init(frame, track_window)
        init_once = True

    if ok:
        p1 = (int(newbox[0]), int(newbox[1]))
        p2 = (int(newbox[0] + newbox[2]), int(newbox[1] + newbox[3]))       
        cv2.putText(frame,"KCF", (p1[0],p1[1]+25), cv2.FONT_HERSHEY_SIMPLEX, 0.75, (200,0, 200), 2)
        img2 = cv2.rectangle(frame, p1, p2, (200,0,200), thickness=2, lineType=2)

alessiodore:
What is "newbox"? Something like this should work:

    ...
    if init_once:
         ok, track_window = tracker.update(frame)
         p1 = (int(track_window[0]), int(track_window[1])) 
    ....

AlexTS1980:
sorry,  I skipped a line: 

`ok, newbox = tracker.update(frame)`

and it still doesn't work. all other trackers work fine. The `newbox` initializes to  ((0.0, 0.0, 0.0, 0.0), which is strange because I fixed the code exactly the way you  did, outside of if (frame>0) loop

AlexTS1980:
...and then it umps to (975.0, 175.0, 0.0, 0.0), i.e bounding box size 0. Perhaps more of the code needs to be changed in trackerKCF.cpp? Unfortunately I don't know C++ too well to understand it.

alessiodore:
This the wrong behaviour described [here](https://github.com/opencv/opencv_contrib/issues/640). I still think there is something wrong on how you re building/linking the patched tracking library from opecv_contrib. Just as a sanity check add a print statement in the patched code (and build again) to make sure that the one you are running from Python is the correct version. I hope it helps.

AlexTS1980:
OK I'll try this. Perhaps my compilation is incorrect? What I do is 
sudo cmake -DOPENCV_EXTRA_MODULES_PATH=path_to_opencv_contrib_git_clone/modules path_to_opencv-3.1.0/

then I cd to /build directory and 

sudo make all -j5

This should be correct though...

alessiodore:
Shouldn't you do:

    sudo make install

too?

AlexTS1980:
Thanks! That's silly, but it still doesn't work and it doesn't print out the text, so something's somewhere is wrong. But what? the first command adds `/module` directory to configuration file, the second builds the new files and the third installs them. 

If the text isn't printed it means the source file isn't compiled, but why? I added the dir with modules the way I described earlier. Anthing else I should do mb?

AlexTS1980:
OK, now on top of this nothing works and I get a crazy error message 

ImportError: /usr/local/lib/python2.7/dist-packages/cv2.so: undefined symbol: _ZNK2cv4face17StandardCollector10getMinDistEv

AlexTS1980:
OK, so I got it to work, but it was no easy thing. Essentially I deleted and fully reinstalled opencv 3.10 as per tutorial (https://github.com/milq/scripts-ubuntu-debian/blob/master/install-opencv.sh) from `build` dir: 

```
cmake -DWITH_QT=ON -DWITH_OPENGL=ON -DFORCE_VTK=ON -DWITH_TBB=ON -DWITH_GDAL=ON -DWITH_XINE=ON -DBUILD_EXAMPLES=ON ..
sudo make -j4
sudo make install
sudo ldconfig
```

then added contrib package:

```
sudo cmake -DOPENCV_EXTRA_MODULES_PATH=/home/alex/Downloads/opencv_git_clone/opencv_contrib/modules ../
sudo make -j8
sudo make install
sudo ldconfig
```
Just in case, I added a path to export: 

`NEWPWD="/home/alex/Downloads/opencv-3.1.0"
`
Not sure if it is necessary though. And it works now. Phew!

terfendail:
:+1:

PR #651, opened 2016-05-01
Allow vector<Mat> raster bands for SEEDS.

cbalint13:
This tiny patch allows to pass image in ```std::vector<cv::Mat>``` not only in ```cv::Mat``` thus enabling additional raster image layout for SEEDS superpixles algorithm. 

It is enhancement not bugfix, some reasons:

* It is useful e.g. for very large satellite imagery where stack of  2d ```cv::Mat``` rasterbands layout is much more memory friendly since read of large bands happen in sequential way and bands are just concatenated at the end into ```std::vector``` array.

* LSC and SLIC already handle booth ```std::vector<cv::Mat>``` not only in ```cv::Mat``` layouts, so expecting SEEDS too. In fact ```std::vector<cv::Mat>``` is the internal default schema for these two since each bands (it can be multispectral) represents a separate band matrix.

### DIFF
--- a/modules/ximgproc/src/seeds.cpp
+++ b/modules/ximgproc/src/seeds.cpp
@@ -375,7 +375,31 @@ void SuperpixelSEEDSImpl::initImageBins<float>(const Mat& img, int)
 
 void SuperpixelSEEDSImpl::initImage(InputArray img)
 {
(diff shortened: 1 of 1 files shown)


alalek:
:+1:

PR #656, opened 2016-05-03
structured_light:  python / java bindings

### LINKED ISSUE #652: Java wrappers for structured_light module cannot be generated
An attempt to generate Java wrappers for the `structured_light` module by appending `WRAP java` to the `structured_light/CMakeLists.txt` file fails with the following error:
```
opencv-3.1.0/build/src/org/opencv/structured_light/GrayCodePattern.java:34: error: method create() is already defined in class GrayCodePattern
    [javac]     public static GrayCodePattern create()
    [javac]                                   ^
Linking CXX executable ../../bin/opencv_test_structured_light
    [javac] 1 error
Target 'jar' failed with message 'Compile failed; see the compiler error output for details.'.
```
Are there any guidelines on how to write OpenCV module API so that Java wrapping works well? I would like to test the module from Java.
Also, the generated Java `create` method does not have any parameters while in the C++ code the method takes one parameter: `GrayCodePattern::Params &parameters = GrayCodePattern::Params()`.

berak:
should fix #652

### DIFF
--- a/modules/structured_light/CMakeLists.txt
+++ b/modules/structured_light/CMakeLists.txt
@@ -1,2 +1,2 @@
 set(the_description "Structured Light API")
-ocv_define_module(structured_light opencv_core opencv_calib3d opencv_imgproc opencv_highgui opencv_features2d opencv_rgbd OPTIONAL opencv_viz)
+ocv_define_module(structured_light opencv_core opencv_calib3d opencv_imgproc opencv_highgui opencv_features2d opencv_rgbd OPTIONAL opencv_viz WRAP python java)
--- a/modules/structured_light/include/opencv2/structured_light/graycodepattern.hpp
+++ b/modules/structured_light/include/opencv2/structured_light/graycodepattern.hpp
@@ -43,6 +43,7 @@
 #define __OPENCV_GRAY_CODE_PATTERN_HPP__
 
 #include "opencv2/core.hpp"
+#include "opencv2/structured_light/structured_light.hpp"
 
 namespace cv {
 namespace structured_light {
@@ -72,21 +73,21 @@ class CV_EXPORTS_W GrayCodePattern : public StructuredLightPattern
    *  @param width Projector's width. Default value is 1024.
    *  @param height Projector's height. Default value is 768.
    */
-  struct CV_EXPORTS_W_SIMPLE Params
+  struct CV_EXPORTS Params
   {
-    CV_WRAP
     Params();
-    CV_PROP_RW
     int width;
-    CV_PROP_RW
     int height;
   };
 
   /** @brief Constructor
    @param parameters GrayCodePattern parameters GrayCodePattern::Params: the width and the height of the projector.
    */
-  CV_WRAP
   static Ptr<GrayCodePattern> create( const GrayCodePattern::Params &parameters = GrayCodePattern::Params() );
+
+  // alias for scripting
+  CV_WRAP
+  static Ptr<GrayCodePattern> create( int width, int height );
 
   /** @brief Get the number of pattern images needed for the graycode pattern.
    *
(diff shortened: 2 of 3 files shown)


berak:
warnings due to uninitialized `size_t value` will go away, once [this patch](https://github.com/Itseez/opencv/pull/6498) gets accepted.

dharezlak:
The generated  Java API now compiles, however, when invoking `GrayCodePattern.create(1024, 768)` the following link error shows up:

```
java.lang.UnsatisfiedLinkError: org.opencv.structured_light.GrayCodePattern.create_0(II)J
	at org.opencv.structured_light.GrayCodePattern.create_0(Native Method)
	at org.opencv.structured_light.GrayCodePattern.create(GrayCodePattern.java:28)
	at pl.danhar.scan.StructuredLightTest.testStructruredLight(StructuredLightTest.java:18)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
```
When looking at the native code the create method looks like this (modules/java/structured_light.cpp):
```
JNIEXPORT jlong JNICALL Java_org_opencv_structured_light_GrayCodePattern_create_10 (JNIEnv*, jclass, jint, jint);

JNIEXPORT jlong JNICALL Java_org_opencv_structured_light_GrayCodePattern_create_10
  (JNIEnv* env, jclass , jint width, jint height)
{
    static const char method_name[] = "structured_light::create_10()";
//...
```

berak:
@dharezlak , thanks for trying, nonetheless. 

i'll reopen this, when i got a solution (and the python warnings are fixed)

dharezlak:
Hello,

The test code is very simple for now: https://gist.github.com/dharezlak/3bed0c603d1066c59ca0be970315c171.

The build was done by using the following commands:

```
export JAVA_HOME=/usr/lib/jvm/java-8-oracle
cmake -D CMAKE_BUILD_TYPE=Release -D CMAKE_INSTALL_PREFIX=`pwd` -D BUILD_SHARED_LIBS=OFF -D OPENCV_EXTRA_MODULES_PATH=../opencv_contrib/modules ..
make -j8
```

dharezlak:
@berak: Could you provide any pointers on what could be the problem?

berak:
@dharezlak , unfortunately, my former dev box went up in flames (literally), and the current one does not have any java dev env, atm.

dharezlak:
Can the problem be connected with the JNI method name resolving: https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/design.html#resolving_native_method_names? If there is an underscore in the package name (here: `structured_light`) the native method name should contain `_1` where the underscore is present.

dharezlak:
Yes, that seems to resolve the linking problem. After manually changing the method signatures to `Java_org_opencv_structured_1light...` in the structured_light.cpp file Java linking problem does not show again. Will report this in the original issue.

mshabunin:
I think it is better to refactor class code a little: get rid of `Params` nested class and add several get/set methods for each property. We had similar problem with _ml_ wrappers because C++ parser and bindings generators need to be updated to understand differences between namespaces and nested classes. It seemed to be quite hard problem, so we had chosen different way - simplifying class interface (https://github.com/Itseez/opencv/pull/3679/files#diff-587e0476e652d7fcd6b954e175884346).

dharezlak:
After @berak's modifications the `Params` class is not exposed to the Java layer at all. The only problem that was left was a proper building of the name of the package in the JNI layer which can be easily fixed by providing the fix described in https://github.com/Itseez/opencv_contrib/issues/652#issuecomment-220623626. Should we proceed with the fix?

mshabunin:
@dharezlak , yes, lets try this fix.

Issue #655, opened 2016-05-02
selectROI() does not return after hitting space/esc/enter


hccheng:
The problem being that the condition of while loop is never checked.

In roiSelector.cpp, line 113
```
while(!(key_==32 || key_==27 || key_==13)){
...
}
```
The condition checks the value of key_, which is never set in the while loop.

A simple fix is adding the following line by the end of loop.
`key_ = key & 0xFF;`

Might be a duplicate of #635

PR #657, opened 2016-05-03
tracking: fix selectRoi()

### LINKED ISSUE #635: selectROI("tracker", frame, objects)  cannot work
I used the latest code,and when I press the ESC key,there is no response？

...
	// get bounding box
	cap >> frame;
	selectROI("tracker", frame, objects);
	//quit when the tracked object(s) is not provided
	if (objects.size()<1)
		return 0;
	// initialize the tracker
	trackers.add(frame, objects);
...

berak:
`key_` variable was not updated in loop, this should fix #635 and #655


alalek:
:+1:

Issue #654, opened 2016-05-02
model filename for structured_edge_detection


sushruta:
[structured_edge_detection program](https://github.com/Itseez/opencv_contrib/blob/master/modules/ximgproc/samples/structured_edge_detection.cpp#L52) seems to take a model filename as input. Where can I find this file? The original implementation of the paper - https://github.com/pdollar/edges has a model but it's a matlab `.mat` file and I'm not sure what to do of it.

Any help or pointers would be greatly appreciated. thanks!

StevenPuttemans:
A small google search led me to [this SO topic](http://stackoverflow.com/questions/33317152/model-file-for-opencvs-structured-edge-detector) telling you that the opencv extra repo has the [necessary data](https://github.com/Itseez/opencv_extra/blob/master/testdata/cv/ximgproc/model.yml.gz) for testing the sample.

I guess it could be added more clearly to the sample! I will add an extra description through a PR.

StevenPuttemans:
There you go!

sushruta:
thanks a lot. This helps a lot

PR #660, opened 2016-05-04
add model location info

StevenPuttemans:
The exact model location was missing and thus people do not know how to run the sample.
See https://github.com/Itseez/opencv_contrib/issues/654


sushruta:
looks good to me! 💯

alalek:
:+1:

Issue #644, opened 2016-04-24
haarcascade_mcs_eyepair_small.xml


sturkmen72:
Test code

```

#include <iostream>

#include "opencv2/highgui.hpp"
#include "opencv2/objdetect.hpp"
#include "opencv2/imgproc.hpp"

using namespace std;
using namespace cv;

int main( int argc, const char** argv )
{
   	if (argc < 3)
    {
    cerr << "usage:\n" << argv[0] << " <image_file_name> <model_file_name>" << endl;
    return 0;
    }

    // Read in the input arguments
    string model = argv[2];

    CascadeClassifier detector(model);
    if(detector.empty())
    {
        cerr << "The model could not be loaded." << endl;
    }

    Mat current_image, grayscale, equalize;

    // Read in image and perform preprocessing
    current_image = imread(argv[1]);
    cvtColor(current_image, grayscale, CV_BGR2GRAY);
    equalizeHist(grayscale,equalize);

    vector<Rect> objects;
    detector.detectMultiScale(equalize, objects, 1.05, 1);

    for(int i = 0; i < objects.size(); i++)
    {
        rectangle(current_image, objects[i], Scalar(0, 255, 0),2);
    }

    imshow("result",current_image);
    waitKey();
    return 0;
}

```
calling this code like `testcode oscarSelfie.jpg haarcascade_mcs_eyepair_small.xml` gives bad output like

![result_current](https://cloud.githubusercontent.com/assets/12151541/14764650/a085cf64-09c8-11e6-964b-7109178ce510.png)

result of another `haarcascade_mcs_eyepair_small.xml` ( i found it on my computer, don't remember from where i get it )

![result_new](https://cloud.githubusercontent.com/assets/12151541/14764666/75798846-09c9-11e6-872b-87a4c9baa1ed.png)

StevenPuttemans:
Ow can you see how your metadata says that the image is oriented? OpenCV reads that data now since a month or 5 and thus interprets the landscape/portrait mode

sturkmen72:
i think the problem is here.

```
<cascade type_id="opencv-cascade-classifier"><stageType>BOOST</stageType>
  <featureType>HAAR</featureType>
  <height>22</height>
  <width>5</width>
```

https://github.com/Itseez/opencv_contrib/blob/master/modules/face/data/cascades/haarcascade_mcs_eyepair_small.xml#L124-L125

 maybe there is a bug on the code for creating or converting cascade-classifier

[...]

PR #678, opened 2016-05-22
fix for width&height problem in haarcascade_mcs_xxx.xml files

sturkmen72:
https://github.com/Itseez/opencv_contrib/issues/644


StevenPuttemans:
This seems fine by me! @alalek can you also confirm this? This bug was already solved in the normal modules containing these models, but seem to be forgetten in the face module. This also leads me to the question, why are we keeping doubles of the same model in the repo? This seems a lot of overburden to maintain!

alalek:
:+1:

PR #682, opened 2016-05-24
Charuco doc patch

jokereactive:
I was working with contrib modules, the AruCo module mostly, for one of my projects. While I was using them, I found some discrepancies between the documentation/tutorials and the code besides some general suggestions. This patch covers some of them - 

- **[Main] Repository README.md**:
   - _all contrib modules do not compile with the 2.4 branch, it is advised to use latest releases or the master branch of OpenCV when working with contrib modules._
   - _additional linker flags will be added to your IDE/compile statement for the module one chooses to use._ Although obvious, it is one specific change one has to make in their compile process after adding an extra module, which could be highlighted.
- **[AruCo] Command line Arguments**:
   - While the tutorial code still cites direct usage examples, the code is parameter based. This makes it difficult to run a sample, just by following the tutorial. Example: To run [detect_diamonds](https://github.com/Itseez/opencv_contrib/blob/master/modules/aruco/samples/detect_diamonds.cpp) following the tutorial given [here](https://github.com/Itseez/opencv_contrib/blob/master/modules/aruco/tutorials/charuco_diamond_detection/charuco_diamond_detection.markdown) requires one to make out what the default parameters must be from different parts of the tutorial and then create the params statement. 
   - Sample command line arguments in the tutorial could make this a lot easier. Ideally the code examples in the tutorial should be snippets from the sample but since in this case the sample files are command line based, additional information could be provided in the format mentioned in this patch.  
- **[AruCo] YML File**:  
   - The [YML file](https://github.com/Itseez/opencv_contrib/blob/master/modules/aruco/samples/detector_params.yml) under modules/aruco/samples is outdated and does not allow a run of sample files, such as detect_diamonds.cpp, which use it, due to missing parameters. 
   - This patch adds those missing parameters in the YML file which allows all the samples to run, thereafter.
- **[AruCo] Example Boards**:
   - Throughout the tutorials the dictionary DICT_6x6_250 is used while the boards in the tutorial use some other dictionary. 
   - This adds a dependency of creating a board first before testing functionality such as calibration and pose estimation or getting the impression of something is wrong, if one prints the ones present on the repository and tests the code.
   - Presence of boards consistent with the tutorial in the resources of the tutorials will be very helpful.
   - Only the ChAruCo diamond board seems to be consistent with the tutorial, the other 3 - AruCo Marker, AruCo board and ChAruCo board are not consistent with the dictionary used in the tutorial. 
- **[AruCo] Parameter Description**
   - Description of parameters accepted in 4 of the files is wrong.
   - _detect_board_charuco.cpp, caliberate_camera_charuco.cpp & detect_diamonds.cpp_, the parameters _sl, ml_ are not in pixels, but accept float and are in meters. 

@sergarrido  @prasannavk


modules/aruco/samples/calibrate_camera_charuco.cpp
```diff
@@ -57,8 +57,8 @@ const char* about =
 const char* keys  =
         "{w        |       | Number of squares in X direction }"
         "{h        |       | Number of squares in Y direction }"
-        "{sl       |       | Square side lenght (in pixels) }"
-        "{ml       |       | Marker side lenght (in pixels) }"
+        "{sl       |       | Square side length (in meters) }"
+        "{ml       |       | Marker side length (in meters) }"
```
StevenPuttemans: Actually this is incorrect as far as I understand it. A calibrated setup always works on sizes in pixels of the camera coordinate system, not in real measurements. This is the same for a stereo setup calibration...

jokereactive: But this is the code that generates the calibration parameters. To generate these for a camera system you need the units in the real world context only then will we get a relation between pixels and real world units like millimetres. 

The input parameters too down in the parser code, accept **float** instead of **int**.

StevenPuttemans: Ow I actually mixed things up, you are correct, this should indeed be units of length. But is meters the right parameter then? Shouldnt we use cm or mm?

sergarrido: Indeed, to calibrate intrinsic parameters, the unit you use is completely irrelevant (you can use mm, m, inches or anything). This is simply because, in the pinhole model, what you use is the ratio X/Z and Y/Z respectively for x and y. So as far as all the points are in the same unit, it is not relevant. In this concrete sample, as far as -sl and -ml are in the same unit, it should work ok.
On the contrary, when you are estimating the pose (the extrinsic parameters), the unit you use is relevant since the translation vector will be estimated in the same unit (for instance in the detect_markers or detect_board samples).

However, to simplify this for the users and avoid confusions/errors, we just decided to "force" them to use always the same unit by specifying "(in meters)" in all the samples. It seems that, in a later PR, someone changed it to "(in pixels)" by mistake and I missed that in the revision.

StevenPuttemans: :+1:

prasannavk:
Any idea how to fix the failed check? http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_docs/builds/2408/steps/whitespace%20opencv_contrib

jokereactive:
fixed.

prasannavk:
looks good to me. I think if you merge in the latest master, you won't have any conflicts.

alalek:
@jokereactive Please squash commits into one and rebase to latest master branch (there is merge conflict)

jokereactive:
@alalek I have squashed the changes into one commit, merged the branch with master, and resolved the conflicts.

alalek:
@jokereactive Something is wrong here with commits squashing, because GitHub shows 4 commits for this PR.

jokereactive:
Hey @alalek! I hope it looks fine now. @prasannavk helped me sort it out.

alalek:
Looks good to me!
:+1:

PR #688, opened 2016-05-29
Implement image hash modules

stereomatchingkiss:
I intend to implement different image hashing algorithms and port the image hashing algorithms of [PHash library](http://www.phash.org/docs/) to opencv.  there are three algorithms already implemented by this pull request, they are "average hash", "phash" and "Marr-Hildreth Operator Based Hash".

Test case of average mean and phash are based on golden models(calculated by hand), test case of Marr-Hildreth Operator is based on the golden model compare with origin implementation(I make sure the steps of fillBlocks and createHash generate same value as PHash did).

In the future, I will port another algorithms from PHash library, like Radial Variance based Hash and block Mean Value Based Hash.


stereomatchingkiss:
I am confuse with the [warning message](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_docs/builds/2422/steps/make%20doxygen/logs/stdio),  how could I solve this problem?Any pointer?Thanks

Dikay900:
interesting module indeed, the warning seems to be a problem with linking your module doc to the main list of modules. Looking at the current doc e.g. the fuzzy module it seems that there needs to be a img_hash group added in doxygen.
`http://docs.opencv.org/master/df/d5b/group__fuzzy.html#gsc.tab=0`
Since your namespace is named ihash you only have this group added.
Maybe you can simply change the ihash namespace to the img_hash namespace.

stereomatchingkiss:
@Dikay900 Thanks for your helps, the warning has gone.

I will try to implement another image hash algorithm on this weekend

stereomatchingkiss:
@Dikay900 I change the namespace from ihash to img_hash, but it still show me the same [warning message](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_docs/builds/2428/steps/make%20doxygen/logs/stdio).

stereomatchingkiss:
When I try to develop these algorithms, I am quite missed the good things offer by newer c++ standard, like(c++11/14), any plan to update from c++98 to c++11?

modules/img_hash/include/opencv2/img_hash/img_hash_base.hpp
```diff
@@ -0,0 +1,69 @@
...
+// indirect, incidental, special, exemplary, or consequential damages
+// (including, but not limited to, procurement of substitute goods or services;
+// loss of use, data, or profits; or business interruption) however caused
+// and on any theory of liability, whether in contract, strict liability,
+// or tort (including negligence or otherwise) arising in any way out of
+// the use of this software, even if advised of the possibility of such damage.
+//
+//M*/
+
+#ifndef __OPENCV_FUZZY_F0_MATH_H__
```
berak: you probably want to change the include guard

stereomatchingkiss: @berak Thanks for pointing this defect out

stereomatchingkiss:
Implement radial variance hash on this week

stereomatchingkiss:
The hash algorithm--radial variance hash is done(include example)

I found another bug of PHash and do some optimization when I port the algorithms to opencv.

1 : The paper said PHash apply gamma correction on the image, but in fact, they do not do that but wasting computation power.

```
//They want to apply gamma correction, but this line do not apply gamma correction 
//on th graysc image
(graysc/graysc.max()).pow(gamma);

```
2 : The function ph_crosscorr allocate memory to create an array, but we could just use a double variable to replace it.
3 : Their correlation function calculate same values again and again, these could be avoid

    ```
double meanx = sumx/N;
    double meany = sumy/N;
    double max = 0;
    for (int d=0;d<N;d++){
        double num = 0.0;
        double denx = 0.0;
        double deny = 0.0;
        for (int i=0;i<N;i++){
            //store the value of x_coeffs and y_coeffs in another array
           //this way we do not need to recalculate value - mean
            num  += (x_coeffs[i]-meanx)*(y_coeffs[(N+i-d)%N]-meany);
            //we could find out the standard deviation once and for all
            denx += pow((x_coeffs[i]-meanx),2);
            deny += pow((y_coeffs[(N+i-d)%N]-meany),2);
        }
       //we could save the result of sqrt(denx*deny) since they always same
        r[d] = num/sqrt(denx*deny);
        if (r[d] > max)
            max = r[d];
    }
```

ph_crosscorr slip through the n from 0~40(the formula of discrete cross-correlation could be [found at here](https://en.wikipedia.org/wiki/Cross-correlation)) and pick the largest value but I only calculate the value at n == 0. I do not know why PHash intent to find the maximum value of cross-correlation, any idea(This is not the issue I mentioned in 3)?Thanks

**Comment by stereomatchingkiss** [comment]
At last, I finished the block mean hash algorithm of PHash, for those who curious, I use [this file](https://gist.github.com/stereomatchingkiss/6b9034f72850b518f63631852d7b636f) to extract the golden data from PHash.

**Comment by stereomatchingkiss** [comment]
Write a small tool to remove trailing white space, I post it [on my blog](http://qtandopencv.blogspot.my/2016/06/remove-annoying-trailing-white-space-by.html), maybe this tool can help someone like me.

**Comment by sturkmen72** [comment]
@stereomatchingkiss you have a nice blog. could you post its link at http://answers.opencv.org/question/69691/informative-websites-related-to-opencv/

**Comment by stereomatchingkiss** [comment]
@sturkmen72 glad you like it, I think it is already listed in your post :)

**Comment by stereomatchingkiss** [comment]
Find out the reasons why ph_crosscorr slip through the n from [0~40)

pcc(peak cross-correlation I guess) is finding the value of the signals "align" the most. Problem is I do not know how would this affect the final results. Cannot download the paper since I do not have the account, I guess I would find it out by experiments.

**Comment by stereomatchingkiss** [comment]
I implement mode 2 of block mean hash, but it do not work well, maybe my implementations have some flaws, I put the codes in different branch([img_hash_rotation](https://github.com/stereomatchingkiss/opencv_contrib/tree/img_hash_rotation/modules/img_hash/src))

**Comment by stereomatchingkiss** [comment]
I write down a thorough introduction, do some performance tests(only measure speed) about this module [in my blog](http://qtandopencv.blogspot.my/2016/06/introduction-to-image-hash-module-of.html). 

The next plans are 

1 : implement faster resize function or
2 : use k-d tree  to reduce the complexity of the search problem from linear to sub-linear

**Comment by stereomatchingkiss** [comment]
If there are other good image hash algorithms worth to implement, please give me the link of the paper/websites, thanks

**Comment by stereomatchingkiss** [comment]
Implement [color moment hash](http://www.naturalspublishing.com/files/published/54515x71g3omq1.pdf), works quite good under rotation attack and others(ex : contrast, resize, noise)

**Comment by stereomatchingkiss** [comment]
I add the macro needed for python binding as [this post](http://docs.opencv.org/master/da/d49/tutorial_py_bindings_basics.html) mention, the build system of opencv should be able to generate the python binding. I try to generate python 2 wrapper on my laptop(Y410P, windows 10 64 bits) with following setting(as suggestion in [this post](http://docs.opencv.org/3.1.0/d5/de5/tutorial_py_setup_in_windows.html)).

[Pics for the setting](https://cdn-standard.discourse.org/uploads/pyimagesearch/original/1X/ea3b6611937299ea3c8454b02e4a9c7583c788f6.JPG)

**Comment by stereomatchingkiss** [comment]
Write a [new post](http://qtandopencv.blogspot.my/2016/06/speed-up-image-hashing-of-opencvimghash.html) to introduce img_hash module, in this post I 

1 : Show you how to improve speed of img_hash(outperform PHash library)
2 : Introduce new color moment hash which provide good resistance for rotation attack
3 : Provide better, more sophiticated test results under different attacks and show them by beautiful graph

**Comment by mshabunin** [comment]
@stereomatchingkiss , I've refactored interfaces and simplified the documentation. Please take a look: https://github.com/opencv/opencv_contrib/compare/master...mshabunin:img_hash

If you like it you can pull this commit to your branch:
```.sh
opencv_contrib> git checkout img_hash
opencv_contrib> git pull https://github.com/mshabunin/opencv_contrib img_hash
```

stereomatchingkiss:
Thanks for your contribution :), please give me some times to study your
codes.

2017-05-25 16:21 GMT+08:00 Maksim Shabunin <notifications@github.com>:

stereomatchingkiss:
Sorry for my late reply, I study your codes and I have one request

Instead of using void* to implement pimpl, could you instead use type safe
solution to it?
I think this is safer and also easier to understand, I do not find much
benefits to implement pimpl
by void*.

class something
{
  private:
    struct pimpl;
    cv::Ptr<pimpl> impl_; //I like unique_ptr better, but opencv still
using c++98
};

C++ pimpl avoiding void*
<https://stackoverflow.com/questions/35307934/c-pimpl-avoiding-void>

2017-05-26 7:57 GMT+08:00 ngap wei Tham <thamngapwei@gmail.com>:

mshabunin:
@stereomatchingkiss , I've updated my branch.

stereomatchingkiss:
Sorry for my late reply and thanks for your contribution :), I will try to
finish codes review on this weekend.


2017-06-05 17:46 GMT+08:00 Maksim Shabunin <notifications@github.com>:

stereomatchingkiss:
@mshabunin Thanks for your contribute, I merge the pull request already. I think this branch is ready to merge into opencv_contrib, anything need to be take care?

modules/img_hash/include/opencv2/img_hash/img_hash_base.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+//! @addtogroup img_hash
+//! @{
+
+/** @brief The base class for image hash algorithms
+ */
+class CV_EXPORTS_W ImgHashBase : public Algorithm
+{
+public:
+    class ImgHashImpl;
```
mshabunin: This declaration should be moved back to the top level, on line 15. Otherwise the compilation would be broken.

modules/img_hash/include/opencv2/img_hash.hpp
```diff
@@ -0,0 +1,78 @@
...
+- "Implementation and benchmarking of perceptual image hash functions" @cite zauner2010implementation
+- "Looks Like It" @cite lookslikeit
+
+### Code Example
+
+@include samples/hash_samples.cpp
+
+### Performance under different attacks
+
+![Performance chart](https://3.bp.blogspot.com/-Li-zoGXC6-I/V3Wnp5tbFwI/AAAAAAAAA1Y/iVQkZmI6wWQcpxynuzW4FngJYVdXw3AtgCLcB/s1600/overall_result.JPG)
```
mshabunin: Could you please add this image and two in the next section to the _img_hash/doc_ folder.

modules/img_hash/include/opencv2/img_hash/img_hash_base.hpp
stereomatchingkiss: This problem is solved, I forgot to commit one file, that is why it cannot compiled

modules/img_hash/include/opencv2/img_hash.hpp
stereomatchingkiss: Sure, will upload it on tomorrow

modules/img_hash/include/opencv2/img_hash.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+### Speed comparison with PHash library (100 images from ukbench)
+
+![Hash Computation chart](https://3.bp.blogspot.com/-XIs-olyuK9Q/V3NKRDRzUiI/AAAAAAAAAwU/k99xuDGlCBYwO3ZDZNHcLweuaAt_cpHtwCLcB/s1600/Capture.JPG)
+![Hash comparison chart](https://1.bp.blogspot.com/-anqfh2Awky4/V3NOOKvrQKI/AAAAAAAAAwo/pZjGDDnAPKooOZCCVnzGO4lJjKo7-KjlACLcB/s1600/Capture.JPG)
```
mshabunin: Please add these two images too.

modules/img_hash/include/opencv2/img_hash.hpp
mshabunin: Link should be changed to use local image:
```
![Performance chart](img_hash/doc/attack_performance.JPG)
```

mshabunin:
:+1:

PR #689, opened 2016-05-29
GSoC 2016: DIS optical flow algorithm and a benchmark script

sbokov:
Added a preliminary interface and a partial unoptimized implementation for the Dense Inverse Search (DIS) optical flow algorithm. The results so far on the full MPI-Sintel and Middlebury datasets: https://www.dropbox.com/sh/0qbyi0ds6sj6f1b/AADfLftQ_nSAwf13sRKLgIBKa?dl=0 . Also added a python benchmark script that can evaluate different optical flow implementations and build overall comparative charts like the ones provided above. 

@mshabunin Please, take a look. I wonder if such format for the benchmark script would be OK.


mshabunin:
Looks good! Please squash all commits into one.

mshabunin:
:+1:

PR #707, opened 2016-06-18
Dnn module enhancement [GSoC]

ludv1x:
This update contains:
- BLAS-based optimizations that speedup processing convolution and FC layers (up to 40x with Intel MKL; [perf tests](https://dl.dropboxusercontent.com/u/62894726/GSoC/perf_summary.zip))
- new LSTM and RNN layers
- more complete and clear module docs
- some API changes (backward-compatibility is broken int BlobShape(int, int) constructor, but API is more consistent now)


vpisarev:
looks cool! let's merge it in :+1:
please, submit further improvements (if any) as subsequent PR(s)

ludv1x:
@vpisarev Ok.

Please note, that there is an other [useful PR](https://github.com/opencv/opencv_contrib/pull/738) for dnn module that has conflicts after the merge.
I was working to merege this PR into my branch.
So, it seems that I can resolve conflicts quickly and make new PR with resolved conflicts.

@AnnaPetrovicheva @vbystricky In order to avoid extra effort, let me know if you plan to resolve the conflicts.

AnnaPetrovicheva:
@ludv1x, @vpisarevsky I'm planning to rebase my PR today/tomorrow. Great new module - good to see it merged:)

vbystricky:
@ludv1x 
Sorry if I don't clearly understand code. But in convolution_layer.cpp::Ln137 we try to get `blobs[1]` in any case, but if layer doesn't have a bias (it is possible) will we have a problem?

ludv1x:
@vbystricky Yes, you are right. It is a bug.
A correct solution shoul look like [this](https://github.com/opencv/opencv_contrib/blob/master/modules/dnn/src/layers/fully_connected_layer.cpp#L104) (example from InnerProduct layer).

vbystricky:
@ludv1x , @vpisarev, I think someone should fix this? And same problem with deconvolution layers.

ludv1x:
@vbystricky Yes, of course. I can fix it in next PR with upcoming module updates.
Or if it is critical for your changes I can fix it now in separate PR.

PR #710, opened 2016-06-21
[GSoC] PCAFlow implementation

VladX:
This pull request contains the basic implementation of the PCAFlow algorithm.

Results from the new benchmarking script:
evaluation_results_middlebury
Unfortunately, I wasn't able to compare more than 5 algorithms.


modules/optflow/src/pcaflow.cpp
```diff
@@ -0,0 +1,528 @@
...
+
+namespace cv
+{
+namespace optflow
+{
+namespace
+{
+
+#ifndef M_PI
+const float M_PI = 3.14159265358979323846;
```
sergregory: It seems you can use `CV_PI` from the `opencv2/core/cvdef.h`

modules/optflow/src/pcaflow.cpp
```diff
@@ -0,0 +1,528 @@
...
+  else if ( a == 0 )
+  {
+    c = 0;
+    s = mathSign( b );
+    r = std::abs( b );
+  }
+  else if ( std::abs( b ) > std::abs( a ) )
+  {
+    const double tau = a / b;
+    s = mathSign( b ) / sqrt( 1 + tau * tau );
```
sergregory: I'm not sure, but it seems to be more consistent to use `std::abs` with `std::sqrt` together

modules/optflow/include/opencv2/optflow/pcaflow.hpp
```diff
@@ -0,0 +1,127 @@
...
+  const float dampingFactor;
+  const float claheClip;
+
+public:
+  OpticalFlowPCAFlow( Ptr<const PCAPrior> _prior = Ptr<const PCAPrior>(), const Size _basisSize = Size( 18, 14 ),
+                      float _sparseRate = 0.024, float _retainedCornersFraction = 0.2,
+                      float _occlusionsThreshold = 0.0003, float _dampingFactor = 0.00002, float _claheClip = 14 );
+
+  void calc( InputArray I0, InputArray I1, InputOutputArray flow );
+  void collectGarbage();
```
sergregory: This method seems to be redundant.

sergregory:
@VladX, am I understand correctly that this implementation differs from the one described in a paper mostly by a following points:
- DCT instead of obtained via PCA of the learned with GPUFlow basis;
- `goodFeaturesToTrack` instead of libviso2/A-KAZE/... features;
- LSQR instead of IRLS?

modules/optflow/src/learn_prior.py
```diff
@@ -0,0 +1,166 @@
...
+            f.append(p)
+        else:
+            for root, subdirs, files in os.walk(p):
+                f += map(lambda x: os.path.join(root, x),
+                         filter(lambda x: x.split('.')[-1] == 'flo', files))
+    return list(set(f))
+
+
+def load_flo(flo):
+    with open(flo) as f:
```
sbokov: Please use open(flo, 'rb') to read binary files. It doesn't work on Windows otherwise.

modules/optflow/src/pcaflow.cpp
```diff
@@ -0,0 +1,528 @@
...
+  CV_Assert( occlusionsThreshold > 0 );
+}
+
+void OpticalFlowPCAFlow::collectGarbage() {}
+
+Ptr<DenseOpticalFlow> createOptFlow_PCAFlow() { return makePtr<OpticalFlowPCAFlow>(); }
+
+PCAPrior::PCAPrior( const char *pathToPrior )
+{
+  FILE *f = fopen( pathToPrior, "r" );
```
sbokov: The same thing here, i.e. fopen( pathToPrior, "rb" )

modules/optflow/src/pcaflow.cpp
```diff
@@ -0,0 +1,528 @@
...
+
+ocl::ProgramSource _ocl_fillDCTSampledPointsSource(
+  "__kernel void fillDCTSampledPoints(__global const uchar* features, int fstep, int foff, __global "
+  "uchar* A, int Astep, int Aoff, int fs, int bsw, int bsh, int sw, int sh) {"
+  "const int i = get_global_id(0);"
+  "const int n1 = get_global_id(1);"
+  "const int n2 = get_global_id(2);"
+  "if (i >= fs || n1 >= bsw || n2 >= bsh) return;"
+  "__global const float2* f = features + (fstep * i + foff);"
+  "__global float* a = A + (Astep * i + Aoff + (n1 * bsh + n2) * 4);"
```
sbokov: It seems that explicit casts should be here (I wasn't able to compile this kernel otherwise), i.e.:
"__global const float2* f = (__global const float2*)(features + (fstep * i + foff));"
"__global float* a = (__global float*)(A + (Astep * i + Aoff + (n1 * bsh + n2) * 4));"

modules/optflow/src/pcaflow.cpp
```diff
@@ -0,0 +1,528 @@
...
+  "const float2 p = f[0];"
+  "a[0] = cos((n1 * M_PI / sw) * (p.x + 0.5)) * cos((n2 * M_PI / sh) * (p.y + 0.5));"
```
sbokov: Please, define M_PI here

modules/optflow/src/pcaflow.cpp
```diff
@@ -0,0 +1,528 @@
...
+                                    const Size size )
+{
+  CV_Assert( prior->getBasisSize() == basisSize.area() );
+
+  A1Out.create( features.size() + prior->getPadding(), basisSize.area(), CV_32F );
+  A2Out.create( features.size() + prior->getPadding(), basisSize.area(), CV_32F );
+  b1Out.create( features.size() + prior->getPadding(), 1, CV_32F );
+  b2Out.create( features.size() + prior->getPadding(), 1, CV_32F );
+
+  if ( ocl::useOpenCL() )
```
sbokov: As far as I understand, OpenCL should be used only when a UMat is passed to the calc function. So there should be an additional check like here: https://github.com/opencv/opencv/blob/master/modules/video/src/tvl1flow.cpp#L382

sbokov:
I compared PCAFlow with other algorithms on both MPI-Sintel and Middlebury datasets:
pcaflow_comparison

Also, here is a qualitative comparison of PCAFlow and DISflow: 
[Middlebury](https://www.dropbox.com/sh/g9p8iwvmhfv5mt2/AACF5V5YitlFPt-XNkL7MKGHa?dl=0)
[MPI-Sintel](https://www.dropbox.com/sh/2fv3zy3y5bdcjz4/AAC-m1Krd313XfcL0VJbkUmJa?dl=0)

VladX:
@sergregory Yes, you are right. I tested both DCT basis and original learned basis provided by authors. I found that DCT has comparable performance though it doesn't bring complications like additional storage space. Also, there is no need to check that prior is learned and used with the same basis (prior is bound to the basis) since we know in advance that it always will be DCT.

sergregory:
@VladX, it would be great to place this information in the `pcaflow.hpp` header to make it available for future users :-)

modules/optflow/src/pcaflow.cpp
```diff
@@ -0,0 +1,525 @@
...
+  "__kernel void fillDCTSampledPoints(__global const uchar* features, int fstep, int foff, __global "
+  "uchar* A, int Astep, int Aoff, int fs, int bsw, int bsh, int sw, int sh) {"
+  "const int i = get_global_id(0);"
+  "const int n1 = get_global_id(1);"
+  "const int n2 = get_global_id(2);"
+  "if (i >= fs || n1 >= bsw || n2 >= bsh) return;"
+  "__global const float2* f = (__global const float2*)(features + (fstep * i + foff));"
+  "__global float* a = (__global float*)(A + (Astep * i + Aoff + (n1 * bsh + n2) * sizeof(float)));"
+  "const float2 p = f[0];"
+  "const float pi = 3.14159265358979323846;"
```
sergregory: It seems to me that it would be better to use stringification of the `CV_PI` macro, like that:
```cpp
#define STR_EXPAND_(arg) #arg
#define STR_EXPAND(arg) STR_EXPAND_(arg)
...
  "const float pi = " STR_EXPAND(CV_PI) ";"
...
#undef STR_EXPAND_
#undef STR_EXPAND

VladX: Not sure about this one :) Macro manipulation looks like unnecessary complication here. It's misleading and unclear.
But lets wait for another opinion, maybe I'm wrong.
Also, I thought that definition of M_PI is a part of the OpenCL standard. I could compile this kernel with M_PI.

modules/optflow/src/pcaflow.cpp
sergregory: @VladX, I agree with @sbokov, are you going to fix it?

VladX: I always pass Mat in the arguments so this check won't work in my case. But I'm not sure if this is correct to pass Mat and then get UMat from Mat using .getUMat() method. This seemed to work on my system.

sergregory: As far as I understand, the ideology behind the OpenCL usage is as following: if the user of the function do not mind the OpenCL to be used, he should inform the function by using the UMat as an argument. In this case if the OpenCL implementation is available, it will be used. But if the argument is of type Mat, the OpenCL implementation should not be used, since the user can not be informed in this case which of the implementations (vanilla/OpenCL) was used.

So, in your case, if you want to use OpenCL implementation, you should always pass UMat as the output arguments.

alalek: "Best known methods" for this condition are based on performance of controlled OpenCL/CPU code paths.

VladX: @sergregory I fixed behavior of this check so that the decision is now based on both conditions: whether or not UMat has been passed by a user into calc() output argument and the ocl::useOpenCL(). Now the behavior is controllable by a user as it should be. Am I supposed to do something like this?

sergregory: @VladX, yes, I think it is enough, thank you!

sergregory:
👍

vpisarev:
:+1:

PR #716, opened 2016-06-24
fix RGB channel order in the rgb2luv conversion

chammoru:
In the current rgb2luv() function, rgb[0], rgb[1], and rgb[2] are used as B,G, and R channel, respectively. This is not right. For the RGB to LUV color space conversion in the code, rgb[0], rgb[2] is should be R, B, respectively. This change simply fixes the problem by exchanging the position of R and B (recall that OpenCV's RGB channel order is BGR).

### DIFF
--- a/modules/ximgproc/src/structured_edge_detection.cpp
+++ b/modules/ximgproc/src/structured_edge_detection.cpp
@@ -152,7 +152,7 @@ static cv::Mat rgb2luv(const cv::Mat &src)
 
         for (int j = 0; j < src.cols*nchannels; j += nchannels)
         {
-            const float rgb[] = {pSrc[j + 0], pSrc[j + 1], pSrc[j + 2]};
+            const float rgb[] = {pSrc[j + 2], pSrc[j + 1], pSrc[j + 0]};
 
             const float xyz[] = {mX[0]*rgb[0] + mX[1]*rgb[1] + mX[2]*rgb[2],
                                  mY[0]*rgb[0] + mY[1]*rgb[1] + mY[2]*rgb[2],


alalek:
> This change simply fixes the problem by exchanging the position of R and B (recall that OpenCV's RGB channel order is BGR).

If you call function called "rgb" then you should pass "rgb" data. Don't try to pass "bgr" images, which [were received](https://www.learnopencv.com/why-does-opencv-use-bgr-color-format/) from "imread()" calls.

chammoru:
Thanks for the valuable comments, Alekhin.

From the following tutorial:
http://docs.opencv.org/3.1.0/d0/da5/tutorial_ximgproc_prediction.html
we can see the current code has the similar problem that you mentioned
(It is passing "bgr" images to the "rgb2luv()").

What would be a better choice?
1. Change the function name from 'rgb2luv' to 'bgr2luv' and apply my
previous patch,
2. Change the the tutorial so that it can call 'cvtColor(...,
CV_BGR2RGB)' before the detectEdges() function.

alalek:
This functionality with tutorial is introduced in this PR: #49 by @Bellaktris 

[Documentation](http://docs.opencv.org/3.1.0/d8/d54/classcv_1_1ximgproc_1_1StructuredEdgeDetection.html#a6582e80f4790e4f321dc1404a4adfb60) says that src is "source image (RGB, float, in [0;1]) to detect edges".
Also there is some [pre-trained model (random forest)](https://github.com/Itseez/opencv_extra/blob/master/testdata/cv/ximgproc/model.yml.gz), so it is not easy to add fixes if it has been created on BGR image data (but test passes with your changes for some reason).

chammoru:
It seems that my changes get more similar edge detection results with the
Matlab code outputs of Structured Forests for Fast Edge Detection
<https://www.microsoft.com/en-us/research/publication/structured-forests-for-fast-edge-detection/>
.
But I agree with you that this issue cannot be easily fixed by me, but can
be best handled by @Bellaktris <https://github.com/Bellaktris> (maybe with
his training code).

PR #720, opened 2016-06-27
GSOC'16 - DNN modern

edgarriba:
This pull request creates a new module for deep learning.
- it requires tiny-dnn (check README for installation)
- the module provides a caffe model loader
- a quick sample on how to use caffe loader is also provided


modules/dnn_modern/CMakeLists.txt
```diff
@@ -0,0 +1,111 @@
+set(the_description "Modern Deep Learning module")
+
+# Using cmake scripts and modules
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
+
+# ----------------------------------------------------------------------------
+# MODULE REQUIREMENTS
+# ----------------------------------------------------------------------------
+
+find_package(TinyDNN)
```
alalek: please use `QUIET` mode

modules/dnn_modern/cmake/FindTBB.cmake
```diff
@@ -0,0 +1,283 @@
+# Locate Intel Threading Building Blocks include paths and libraries
```
alalek: OpenCV has TBB support: https://github.com/opencv/opencv/blob/master/cmake/OpenCVDetectTBB.cmake
Are there any problems with reusing of mentioned code?

modules/dnn_modern/cmake/FindTinyCNN.cmake
```diff
@@ -0,0 +1,46 @@
+# Locate the tiny-cnn library.
```
alalek: Replace everywhere:
```
tiny-cnn -> tiny-dnn
FindTinyCNN -> FindTinyDNN
TinyCNN -> TinyDNN
etc
```

modules/dnn_modern/README.md
```diff
@@ -0,0 +1,43 @@
...
+ - System under Unix or Windows
+ - C++11 compiler
+ - tiny-cnn headers
+
+**How to install tiny-cnn?**
+
+  Download tiny-cnn project somewhere in your system
+
+    cd /opt
+    git clone https://github.com/nyanp/tiny-cnn.git
```
alalek: Project has new link:
```
https://github.com/tiny-dnn/tiny-dnn
```

edgarriba:
@vpisarev @garybradski Are we going to merge this module as it is or do we wait for a clear roadmap for the dnn module? /cc @bhack

bhack:
@edgarriba We are on this topic. I hope that we can have news soon

sovrasov:
@edgarriba this PR will be merged but first you need to resolve infrastructure problems mentioned by @alalek

modules/dnn_modern/cmake/FindTBB.cmake
edgarriba: the motivation for this is that I want the user to choose if he want to build the module with TBB totally independent of the rest of the opencv modules. So, if opencv was built with TBB support maybe for X reason and I don't to use it in this module. /cc @bhack

edgarriba:
@alalek @sovrasov  ready to merge

modules/dnn_modern/CMakeLists.txt
```diff
@@ -0,0 +1,111 @@
...
+if(TINYDNN_USE_NNPACK)
+   find_package(NNPACK REQUIRED)
+   add_definitions(-DCNN_USE_NNPACK)
+   include_directories(SYSTEM ${NNPACK_INCLUDE_DIR})
+   include_directories(SYSTEM ${NNPACK_INCLUDE_DIR}/../third-party/pthreadpool/include)
+   list(APPEND REQUIRED_LIBRARIES ${NNPACK_LIB})
+endif()
+
+find_package(Protobuf)
+find_package(Caffe)
```
sovrasov: Is Caffe required by this module?

modules/dnn_modern/CMakeLists.txt
```diff
@@ -0,0 +1,111 @@
...
+set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${EXTRA_EXE_LINKER_FLAGS}")
+set(CMAKE_EXE_LINKER_FLAGS_RELEASE "${CMAKE_EXE_LINKER_FLAGS_RELEASE} ${EXTRA_EXE_LINKER_FLAGS_RELEASE}")
+set(CMAKE_EXE_LINKER_FLAGS_DEBUG "${CMAKE_EXE_LINKER_FLAGS_DEBUG} ${EXTRA_EXE_LINKER_FLAGS_DEBUG}")
+
+# ----------------------------------------------------------------------------
+# DNN-MODERN MODULE
+# ----------------------------------------------------------------------------
+
+ocv_define_module(dnn_modern opencv_core opencv_imgproc opencv_imgcodecs WRAP python)
+ocv_target_link_libraries(${the_module} TinyDNN::tiny_dnn ${REQUIRED_LIBRARIES})
```
sovrasov: I followed the instructions form README, but tiny-dnn headers are missing during compilation,
so it's needed to add implicit include of sources.
Also in case if headers are included compilation fails due to -Werror=non-virtual-dtor warning.

modules/dnn_modern/src/caffe_converter.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+    vector<tiny_dnn::float_t> vec(inputvec.begin(), inputvec.end());
+
+    auto result = net_->predict(vec);
+
+    // allocate outpute
+    results->clear();
+    results->reserve(result.size());
+
+    for (cnn_size_t i = 0; i < result.size(); i++) {
```
sovrasov: `cnn_size_t` doesn't exist anymore and should be replaced with `size_t`.

modules/dnn_modern/CMakeLists.txt
```diff
@@ -0,0 +1,111 @@
+set(the_description "Modern Deep Learning module")
+
+# Using cmake scripts and modules
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
+
+# ----------------------------------------------------------------------------
+# MODULE REQUIREMENTS
+# ----------------------------------------------------------------------------
+
```
edgarriba: fixed

modules/dnn_modern/CMakeLists.txt
edgarriba: Did you download the last version and installed via cmake commands?

modules/dnn_modern/CMakeLists.txt
edgarriba: let me check

modules/dnn_modern/CMakeLists.txt
sovrasov: Install target in tiny-dnn's CMake is broken at the moment (headers of 3rd party libs are not installed), so I'm just using instructions from readme. CMake command is `cmake . -DTINYDNN_ROOT=~/repositories/tiny-dnn/`

edgarriba:
@sovrasov @alalek could you check again? We fixed some issues in [master](f8585abe6063a518259d42aaef3b5935e5db6eba) that allows to compile without external dependencies. I tried in local and it works for me.

modules/dnn_modern/CMakeLists.txt
```diff
@@ -0,0 +1,114 @@
+set(the_description "Modern Deep Learning module")
+
```
sovrasov: Please add CMake version check here:
```
if(${CMAKE_VERSION} VERSION_LESS 3.2)
  message(STATUS "Module opencv_dnn_modern disabled beacuse CMake version is less than 3.2")
  ocv_module_disable(dnn_modern)
  return()
endif()
```

modules/dnn_modern/CMakeLists.txt
```diff
@@ -0,0 +1,114 @@
...
+set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${EXTRA_EXE_LINKER_FLAGS}")
+set(CMAKE_EXE_LINKER_FLAGS_RELEASE "${CMAKE_EXE_LINKER_FLAGS_RELEASE} ${EXTRA_EXE_LINKER_FLAGS_RELEASE}")
+set(CMAKE_EXE_LINKER_FLAGS_DEBUG "${CMAKE_EXE_LINKER_FLAGS_DEBUG} ${EXTRA_EXE_LINKER_FLAGS_DEBUG}")
+
+# ----------------------------------------------------------------------------
+# DNN-MODERN MODULE
+# ----------------------------------------------------------------------------
+
+ocv_define_module(dnn_modern opencv_core opencv_imgproc opencv_imgcodecs WRAP python)
+ocv_target_link_libraries(${the_module} ${REQUIRED_LIBRARIES})
```
sovrasov: For me it doesn't work unless I add the below two lines:
```
ocv_target_include_directories(${the_module} ${TINYDNN_INCLUDE_DIRS})
target_compile_options(${the_module} PRIVATE "-Wno-error=non-virtual-dtor")
```

modules/dnn_modern/src/caffe_converter.cpp
```diff
@@ -0,0 +1,194 @@
...
+  and on any theory of liability, whether in contract, strict liability,
+  or tort (including negligence or otherwise) arising in any way out of
+  the use of this software, even if advised of the possibility of such damage.
+ */
+
+#include "precomp.hpp"
+
+#define CNN_USE_CAFFE_CONVERTER
+#include <tiny_dnn/tiny_dnn.h>
+
```
sovrasov: Build fails with tons of linker errors related to caffe, but it can be fixed by adding here
`#include <tiny_dnn/io/caffe/caffe.pb.cc>`

edgarriba: @sovrasov ok. However, not sure what's the proper way to handle this since in the library those proto files are generated during the cmake configuration
https://github.com/tiny-dnn/tiny-dnn/blob/master/examples/caffe_converter/CMakeLists.txt

Should we include this piece of code here?

sovrasov: I think it's better to handle this in CMake and just didn't know how it done in tiny-dnn.

In case if tiny-dnn was configured with it's own CMake the most part of dnn_modern CMake becomes useless and can be removed, but current implementation can work with raw unconfigured sources. Why did you choose to this way instead of relying on CMake from tiny-dnn?

edgarriba: well the point is that tiny-dnn is header-only and very configurable so you have to setup the configuration at the moment you "link" it to the target project. However, if you have any better idea it will be welcomed.

sovrasov: I see your point, just interested why tinydnn-config.cmake is not used.
Devs of tiny-dnn recommends it as far as I can see:
```
# Example to find TinyDNN headers only::
#
#   find_package(TinyDNN 0.1.0)
#   if(TinyDNN_FOUND)
#     add_executable(foo foo.cc)
#     target_link_libraries(foo TinyDNN::tiny_cnn)
#   endif()
#
# Example to find TinyDNN headers and some *static* libraries::
#
#   set(TinyDNN_USE_TBB              ON) # only find static libs
#   set(TInyCNN_USE_AVX2             ON)
#   find_package(TinyDNN 0.1.0)
#   if(TinyDNN_FOUND)
#     add_executable(foo foo.cc)
#     target_link_libraries(foo TinyDNN::tiny_cnn ${TinyDNN_LIBRARIES})
#   endif()
```

edgarriba: well yes, I'm the one who wrote those codes. But to use you need to install tiny-dnn headers via cmake in order to create all the targets. I didn't test at all the config file, so this can be a good starting point.

sovrasov: Let's leave all as is, because now rewriting dnn_modern CMake and testing tiny-dnn config will take too much effort.

modules/dnn_modern/cmake/FindTBB.cmake
alalek: This is to avoid potential conflicts, which are hard to debug.

modules/dnn_modern/README.md
alalek: Problem is still here.

modules/dnn_modern/cmake/FindTBB.cmake
edgarriba: my solution or your solution?

modules/dnn_modern/CMakeLists.txt
```diff
@@ -0,0 +1,124 @@
...
+    add_definitions(-DCNN_USE_TBB)
+    include_directories(${TBB_INCLUDE_DIRS})
+    link_directories(${TBB_LIBRARY_DIRS})
+    listt(APPEND REQUIRED_LIBRARIES ${TBB_LIBRARIES})
+elseif(TINYDNN_USE_OMP)
+    find_package(OpenMP REQUIRED)
+    add_definitions(-DCNN_USE_OMP)
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
+endif()
```
alalek: OpenCV already defines TBB and OpenMP dependencies.

edgarriba:
@alalek ping

alalek:
@edgarriba 
There are still several places with legacy name "tiny-cnn". Could you please replace it to "tiny-dnn" to prevent confusing of developers/users?

edgarriba:
@alalek now should be fine

alalek:
Thanks! :+1:

PR #708, opened 2016-06-20
GSoC 2016: Complete DIS optical flow implementation

sbokov:
Added a full optimized implementation of the Dense Inverse Search (DIS) optical flow algorithm with accuracy and performance tests. Some of the older tests in the optflow module were also refactored to decrease the amount of code duplication. Here is how it currently performs in comparison to other optical-flow algorithms in OpenCV:
objective_comparison_low_res

The variational refinement part of the DIS algorithm essentially duplicates the same component of the current DeepFlow implementation, but is noticeably faster. I could replace variational refinement in DeepFlow with my implementation which would yield a significant speedup with practically no quality loss (accelerated DeepFlow on the chart above).

PR #722, opened 2016-07-04
DeepFlow acceleration

sbokov:
Replaced DeepFlow's internal implementation of variational refinement with a new optimized one, which was introduced in #708.


mshabunin:
Thank you!
:+1:

PR #723, opened 2016-07-06
GSOC 2016 Holistic word spotter

anguelos:
Implemented a  classifier class that wraps caffe
Implemented a  word-spotting  class that employs a friendly interface
Simple example for using the word spotter in c++
An interactive example for using the word spotter in python


mshabunin:
Reintegrated this PR in #1348

PR #724, opened 2016-07-07
Add semantic segmentation to the module DNN.

vbystricky:
Add Crop and Eltwise layers. 
Add sample for using FCN32s or FCN8s neural network for image semantic segmentation.


vpisarev:
@vbystricky, if you have time, could you please re-base your PR?

vbystricky:
@vpisarev I will make re-base, but I think we have to fix problems with convolution and deconvolution layers code first. Because segmentation code fails on current implementation.

vbystricky:
@vpisarev rebased. But sample fcn_semsegm crashes because a problem in deconvolution layer.

vpisarev:
ok, let's put it in and then make it work! :+1:

PR #729, opened 2016-07-15
Add methods for get/set weights of variational refinement part of DISOpticalFlow

vbystricky:
Change DISOpticalFlow in module "optflow".

Add methods for get/set weights of variational refinement part of DISOpticalFlow.


alalek:
:+1:

PR #733, opened 2016-07-19
BM3D noise reduction algorithm added

bpawlik:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/301

### This pullrequest adds

- BM3D denoising algorithm (both first and second step):
   http://www.cs.tut.fi/~foi/GCF-BM3D/BM3D_TIP_2007.pdf
- For performance reasons, only Haar transform is used.

TODO list:
- [x] Unlimited size of the 3D group
- [x] Add sliding step size
- [x] Accelerate block-matching by moving average
- [x] Add Kaiser window
- [x] Unlimited size of the template window
- [x] Add enablers for different transforms

### DIFF
--- a/modules/xphoto/src/bm3d_image_denoising.cpp
+++ b/modules/xphoto/src/bm3d_image_denoising.cpp
@@ -1,0 +1,347 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
+//
+//  IMPORTANT: READ BEFORE DOWNLOADING, COPYING, INSTALLING OR USING.
+//
+//  By downloading, copying, installing or using the software you agree to this license.
+//  If you do not agree to this license, do not download, install,
+//  copy or use the software.
+//
+//
+//                        Intel License Agreement
+//                For Open Source Computer Vision Library
+//
+// Copyright (C) 2000, Intel Corporation, all rights reserved.
+// Third party copyrights are property of their respective icvers.
+//
+// Redistribution and use in source and binary forms, with or without modification,
+// are permitted provided that the following conditions are met:
+//
+//   * Redistribution's of source code must retain the above copyright notice,
+//     this list of conditions and the following disclaimer.
+//
+//   * Redistribution's in binary form must reproduce the above copyright notice,
+//     this list of conditions and the following disclaimer in the documentation
+//     and/or other materials provided with the distribution.
+//
+//   * The name of Intel Corporation may not be used to endorse or promote products
+//     derived from this software without specific prior written permission.
+//
+// This software is provided by the copyright holders and contributors "as is" and
+// any express or implied warranties, including, but not limited to, the implied
+// warranties of merchantability and fitness for a particular purpose are disclaimed.
+// In no event shall the Intel Corporation or contributors be liable for any direct,
+// indirect, incidental, special, exemplary, or consequential damages
+// (including, but not limited to, procurement of substitute goods or services;
+// loss of use, data, or profits; or business interruption) however caused
+// and on any theory of liability, whether in contract, strict liability,
+// or tort (including negligence or otherwise) arising in any way out of
+// the use of this software, even if advised of the possibility of such damage.
+//
+//M*/
+
+#include "opencv2/xphoto.hpp"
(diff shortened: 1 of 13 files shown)


bpawlik:
@vpisarev I have contacted one of the original authors of BM3D, prof. Alessandro Foi. It seems to be absolutely fine to include BM3D implementation in OpenCV. 
Prof. Foi said that: "There are no legal obstacles if you implement BM3D based on the description in the paper and for non-profit purposes."

bpawlik:
I believe this pull request is ready for reviewing. Looking forward to your comments!

vpisarev:
@bpawlik, thank you for the contribution and sorry for late follow-up!

I like the code! But we need to sort out legal issues to let it in. These few words "for non-profit purposes" actually mean a lot.

Here is the plan. @alalek will submit soon a pullrequest that adds top-level "ENABLE_NONFREE" option to CMake, which will be off by default. When user turns it on, people get special warning from CMake and at the build stage there will be HAVE_NONFREE macro. So you will need to include cvconfig.h in internal .hpp and .cpp files of your implementation, embrace implementation of all the code into

<pre>
#if HAVE_NONFREE
...
// real implementation
#else
// stub implementation, e.g. 
CV_EXPORTS_W void bm3dDenoising(
            InputArray,
            InputOutputArray,
            OutputArray,
            float,
            int ,
            int ,
            int ,
            int ,
            int ,
            int ,
            float ,
            int ,
            int ,
            int )
{
     CV_Error(CV_StsNotImplemented, "The algorithm is patented and is excluded in this configuration; Set ENABLE_NONFREE CMake option and rebuild the library");
}
#endif
</pre>

you can make this change now, because it does not depend from the patch from @alalek.

alalek:
https://github.com/opencv/opencv/pull/7043

Both `ENABLE_NONFREE` and `HAVE_NONFREE` are replaced to `OPENCV_ENABLE_NONFREE` for consistency and preventing name conflicts.
Any extra include of cvconfig.h is not needed.

bpawlik:
Awesome guys, thanks! Should I then document every function definition with `OPENCV_ENABLE_NONFREE` as @vpisarev mentioned? What about tests then? They'll start failing on empty implementation, right? Thanks again!

alalek:
These tests should be wrapped also (but without`#else` branch):
```
#ifdef OPENCV_ENABLE_NONFREE 
TEST(BM3DTestSuite, BM3DTestCase.)
{
...
}

TEST(...)
{
...
}
#endif // OPENCV_ENABLE_NONFREE
```

bpawlik:
Just pushed those changes that we were talking about.

modules/xphoto/src/bm3d_image_denoising.cpp
```diff
@@ -264,38 +268,38 @@ void bm3dDenoising(
...
+    int /* blockMatchingStep1 */,
+    int /* blockMatchingStep2 */,
+    int /* groupSize */,
+    int /* slidingStep */,
+    float /* beta */,
+    int /* normType */,
+    int /* step */,
+    int /* transformType */)
 {
     // Empty implementation
```
alalek: To prevent user confusing please raise an error in this case:
```
CV_Error(CV_StsNotImplemented, "The algorithm is patented and is excluded in this configuration; Set OPENCV_ENABLE_NONFREE CMake option and rebuild the library");
```

P.S. There is macro CV_UNUSED: https://github.com/opencv/opencv/blob/5f30a0a076e57c412509becd1fb618170cbfa179/modules/core/src/stat.cpp#L1745

bpawlik:
Thanks for the tip with CV_UNUSED, good to know! I've changed that plus added CV_Error. Now waiting for a green build and ready to merge :)

bpawlik:
Guys, anything else I should improve in this pull request?

alalek:
@bpawlik Looks good to me. Please squash all commits into one.

bpawlik:
@alalek Thanks, commits squashed into one.

alalek:
Thank you! :+1:

bpawlik:
Guys, don't forget to merge :)

PR #739, opened 2016-07-22
Changing filter so it treats every type() of disparity Matrix.

LeoBras:
I noted this filter only treats CV_16S, while it converts everything to float and back to CV_16S to get the answer.
### This pullrequest changes

Now it accepts both float and CV_16S (and probably future types).
I got really good results with float Disp_Matrices, and I felt no speed change.

If the disparity inputs are float, it only filters,  if it's of other type, it converts to float before doing the math. It already did most of that, I just changed the place it does, so it cant treat other kinds of disp matrix.


terfendail:
Transferred to #1404

PR #746, opened 2016-08-05
Add support for java in aruco module

smigielski:
### This pullrequest changes

This pull request is adding support for java wrapper generation. It is dependent on https://github.com/opencv/opencv/pull/7050 pull request from opencv.


vpisarev:
ok, the newer version of this patch has been merged

PR #747, opened 2016-08-11
A unified interface for white balance algorithms

sbokov:
Fixes some inconsistencies between interfaces provided by different white balance algorithms. Also adds an option of loading a model during runtime in the learning-based white balance algorithm.


modules/xphoto/test/simple_color_balance.cpp
```diff
@@ -4,9 +4,10 @@ namespace cvtest
 {
     TEST(xphoto_simplecolorbalance, regression)
     {
-        cv::String dir = cvtest::TS::ptr()->get_data_path() + "cv/xphoto/simple_white_balance/";
+        cv::String dir = cvtest::TS::ptr()->get_data_path() + "cv/xphoto/simple_white_balance/";;
```
mshabunin: Extra semicolon here

modules/xphoto/src/learning_based_color_balance.cpp
```diff
@@ -64,56 +64,115 @@ struct hist_elem
...
+  public:
+    LearningBasedWBImpl(String path_to_model)
+    {
+        range_max_val = 255;
+        saturation_thresh = 0.98f;
+        hist_bin_num = 64;
+        palette_size = 300;
+        palette_bandwidth = 0.1f;
+        prediction_thresh = 0.025f;
+        if (path_to_model.size() < 1)
```
mshabunin: `path_to_model.empty()` will be more readble I think.

mshabunin:
I suggest moving `learn_color_balance.py` from `/src/` to `xphoto/samples` folder and joining all `*_color_balance.cpp` samples into one.

modules/xphoto/include/opencv2/xphoto/white_balance.hpp
```diff
@@ -58,129 +58,172 @@ namespace xphoto
...
+    CV_WRAP virtual int getHistBinNum() const = 0;
+    /** @copybrief getHistBinNum @see getHistBinNum */
+    CV_WRAP virtual void setHistBinNum(int val) = 0;
+};
+
+/** @brief Creates an instance of LearningBasedWB
+
+@param path_to_model Path to a .yml file with the model. If not specified, the default model is used
+ */
+CV_EXPORTS_W Ptr<LearningBasedWB> createLearningBasedWB(String path_to_model = String());
```
mshabunin: It will be better to use `const String &` here and in implementation.

mshabunin:
In learning script (lines 228-231):
```.py
...
     if "groundtruth_illuminants" in gt.keys() and "darkness_level" in gt.keys():
         #NUS 8-camera dataset format
         base_gt_illuminants = gt["groundtruth_illuminants"]
         black_levels = len(gt_illuminants) * [gt["darkness_level"][0][0]]
...
```
`gt_illuminants` should be `base_gt_illuminants`

mshabunin:
@sbokov , I have no other comments. Thank you!

However two more ideas come into mind:
- extract common dataset-reading part from both python scripts and put it into either separate file (`color_balance_dataset.py`) either into one of original files
- add model filepath parameter for learning based algorithm to the benchmark script and allow users to compare different variants of same algorithm (something like `script.py -a grayworld,learning_based:model1.yml,learning_based:model2.yml` ...)

mshabunin:
Great! Thank you! :+1:

mshabunin:
Please squash all commits into one and we will merge it.

PR #748, opened 2016-08-14
add initial guess for cv::aruco::estimatePoseBoard()

### LINKED ISSUE #744: cv::aruco::estimatePoseBoard returns nonsensical values
I'm using cv::aruco to do global camera position tracking in a video. Most of the time it works as expected. However, on some occasions (seems to be happening only if all visible markers are coplanar, and mainly if there are only two of them within the field of view), rvec/tvec values returned by estimatePoseBoard jump from correct values to a completely different corner of the parameter space. Upon close examination, it appears that returned values are technically correct, but correspond to a transformation where all visible markers on the board are behind the camera (if I apply specified rotation and translation to elements of Board::objPoints, all resulting Point3f's have negative z-components). I sincerely doubt that this is the intended behavior.

This could be ameliorated to a degree by passing known good recent rvec/tvec to solvePnP as hints inside estimatePoseBoard (since solvePnP iterates to the correct transformation most of the time when given a good hint), but current API does not have this functionality.

I can provide sample input values that trigger this behavior. I've observed this in official 3.1.0 but I don't see any changes in git that would have fixed this. (estimatePoseBoard in particular is very short, and there weren't many changes to the implementation of solvePnP.)

savuor:
resolves #744 

### This pullrequest changes

If `rvec` and `tvec` arguments are not empty then use them as initial guess for pose estimation.

### Motivation
In case of small amount of object-image point correspondences (as it was described in issue #744) it should help with correct pose estimation.

### DIFF
--- a/modules/aruco/include/opencv2/aruco.hpp
+++ b/modules/aruco/include/opencv2/aruco.hpp
@@ -322,8 +322,9 @@ class CV_EXPORTS_W GridBoard : public Board {
  * @param distCoeffs vector of distortion coefficients
  * \f$(k_1, k_2, p_1, p_2[, k_3[, k_4, k_5, k_6],[s_1, s_2, s_3, s_4]])\f$ of 4, 5, 8 or 12 elements
  * @param rvec Output vector (e.g. cv::Mat) corresponding to the rotation vector of the board
- * (@sa Rodrigues).
+ * (@sa Rodrigues). Used as initial guess if not empty.
  * @param tvec Output vector (e.g. cv::Mat) corresponding to the translation vector of the board.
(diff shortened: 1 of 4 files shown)


terfendail:
I think it make sense to update [tutorial](https://github.com/savuor/opencv_contrib/blob/50f9b3adfd2fbd61f46994ddb66fa27bdcd9c269/modules/aruco/tutorials/aruco_board_detection/aruco_board_detection.markdown) as well

terfendail:
:+1:

alalek:
@savuor Please squash all commits into one. And please add used email into your GitHub account.

PR #751, opened 2016-08-17
add thinning

StevenPuttemans:
New functionality to thin out blobs if desired. Used it to perform skeletization on segmented blobs in fingerprints, to find minutae (intersections). Example of application can be seen below:

### DIFF
--- a/modules/ximgproc/include/opencv2/ximgproc.hpp
+++ b/modules/ximgproc/include/opencv2/ximgproc.hpp
@@ -104,6 +104,16 @@ CV_EXPORTS_W void niBlackThreshold( InputArray _src, OutputArray _dst,
                                     double maxValue, int type,
                                     int blockSize, double delta );
 
+/** @brief Applies a binary blob thinning operation, to achieve a skeletization of the input image.
+
+The function transforms a binary blob image into a skeletized form using the technique of Zhang-Suen.
+
+@param src Source 8-bit single-channel image, containing binary blobs, with blobs having 255 pixel values.
+@param dst Destination image of the same size and the same type as src. The function can work in-place.
+ */
+CV_EXPORTS_W void thinning( InputArray src, OutputArray dst);
+
+
 //! @}
 
 }
--- a/modules/ximgproc/src/thinning.cpp
+++ b/modules/ximgproc/src/thinning.cpp
@@ -1,0 +1,63 @@
+#include "precomp.hpp"
+
+using namespace std;
+
+namespace cv {
+namespace ximgproc {
+
+// Applies a thinning iteration to a binary image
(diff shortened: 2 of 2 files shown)


modules/ximgproc/src/thinning.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+namespace cv {
+namespace ximgproc {
+
+// Function prototypes to suppress -Wmissing-declarations warnings
+void thinningIteration(Mat img, int iter);
+void thinning(InputArray _input, OutputArray _output);
+
+// Applies a thinning iteration to a binary image
+void thinningIteration(Mat img, int iter){
```
alalek: Use "static" here.
Remove all these function prototypes.
Add include "opencv2/ximgproc.hpp"

StevenPuttemans: Okay will do so! Could you explain why the Ni Black hasn't got the ximgproc include but doesnt trigger an issue with the -Wmissing declarations?

alalek: Well, additional include is not required. It is done in [precomp.hpp file](https://github.com/opencv/opencv_contrib/blob/master/modules/ximgproc/src/precomp.hpp)

modules/ximgproc/include/opencv2/ximgproc.hpp
```diff
@@ -104,6 +104,16 @@ CV_EXPORTS_W void niBlackThreshold( InputArray _src, OutputArray _dst,
...
                                     int blockSize, double delta );
 
+/** @brief Applies a binary blob thinning operation, to achieve a skeletization of the input image.
+
+The function transforms a binary blob image into a skeletized form using the technique of Zhang-Suen.
+
+@param _src Source 8-bit single-channel image, containing binary blobs, with blobs having 255 pixel values.
+@param _dst Destination image of the same size and the same type as src. The function can work in-place.
+ */
+CV_EXPORTS_W void thinning( InputArray _src, OutputArray _dst);
```
alalek: Could you please remove underscore "_" from parameters here and from documentation comment.

modules/ximgproc/src/thinning.cpp
StevenPuttemans: Well but if i do not add the prototypes then I get this error when building locally. Will see what buildbot says.

modules/ximgproc/include/opencv2/ximgproc.hpp
StevenPuttemans: Will do so!

alalek:
Thank you! :+1:

StevenPuttemans:
Thank you!

PR #752, opened 2016-08-18
[GSoC] Implementation of the Global Patch Collider and demo for PCAFlow

VladX:
This pull request contains implementation of the sparse matching algorithm from the following paper -- http://research.microsoft.com/en-us/um/people/pkohli/papers/wfrik_cvpr2016.pdf
called the Global Patch Collider.

Also, added some demonstration code for PCAFlow.

OpenCV Global Patch Collider


VladX:
Here is the short video demonstration of the Global Patch Collider I made to the end of gsoc
https://drive.google.com/file/d/0B7Hb8cfuzrIIUHcweXRhdk1SdVU/view?usp=sharing

modules/optflow/samples/gpc_evaluate.cpp
```diff
@@ -0,0 +1,163 @@
...
+                 "here:\nhttps://drive.google.com/open?id=0B7Hb8cfuzrIIZDFscXVYd0NBNFU"
+              << std::endl;
+    return 1;
+  }
+
+  ocl::setUseOpenCL( useOpenCL );
+
+  Ptr< optflow::GPCForest< nTrees > > forest = Algorithm::load< optflow::GPCForest< nTrees > >( forestDumpPath );
+
+  Mat from = imread( argv[1] );
```
ethanrublee: Suggest using the parsed strings from above, rather than argv.

modules/optflow/samples/gpc_evaluate.cpp
```diff
@@ -0,0 +1,163 @@
...
+                    "{@image1      |<none>     | image1}"
+                    "{@image2      |<none>     | image2}"
+                    "{@groundtruth |<none>     | path to the .flo file}"
+                    "{@output      |           | output to a file instead of displaying, output image path}"
+                    "{f forest     |forest.dump| path to the forest.dump}";
+
+const int nTrees = 5;
+const bool useOpenCL = true;
+
+static double normL2( const Point2f &v ) { return sqrt( v.x * v.x + v.y * v.y ); }
```
ethanrublee: Isn't there an opencv function for norm?

modules/optflow/src/sparse_matching_gpc.cpp
```diff
@@ -187,9 +202,104 @@ GPCPatchDescriptor::GPCPatchDescriptor( const Mat *imgCh, int i, int j )
...
+    Mat cpuOut = out.getMat( 0 );
+    for ( int i = 0; i + 2 * patchRadius < sz.height; ++i )
+      for ( int j = 0; j + 2 * patchRadius < sz.width; ++j )
+        descr.push_back( *cpuOut.ptr< GPCPatchDescriptor >( i * globSize[1] + j ) );
+    return;
+  }
+
+  for ( int i = patchRadius; i + patchRadius < sz.height; ++i )
+    for ( int j = patchRadius; j + patchRadius < sz.width; ++j )
+      descr.push_back( GPCPatchDescriptor( imgCh, i, j ) );
```
ethanrublee: Would be nice to eventually add other descriptors as described in the paper, but this looks reasonably easy to factorize later.

ethanrublee:
What's the expected training time? I've been training for 1 hour so far on the first image from each of the "final" images in in sintel.

VladX:
I trained 5-tree forest on the full sintel dataset for around 4 days, so I guess it's ok in your case. You should expect several hours. Actually, the time linearly depends on the number of images (assuming all images have the same dimensions). Firstly, try to learn on one image and measure time, it should take 5-10 minutes. Then you can estimate how long it takes to learn on multiple images.

VladX:
The reason why it's so slow -- currently it collects quite a big number of samples from a single image 1024x436 (mpi sintel), around 30000 samples. I should have added some control parameter for tradeoff between speed and quality. But I think it doesn't matter that much since you only need to do it once for each dataset.

ethanrublee:
So I'm seeing around ~7 seconds per correspondence query on 800x600 images, with opencl.  Is this about the expected runtime?

It might be nice to be able to cache the descriptors for each frame, especially for analysis of video where its likely that the same image will be matched multiple times.

Also, I'm keen to experiment with different descriptors other than dct.

Have you done much profiling? I assume computing the descriptors is what takes time during the correspondence search.

VladX:
> Have you done much profiling? I assume computing the descriptors is what takes time during the correspondence search.
Yes, exactly. This is what takes most of the time.
Also, loading large *.yml files (this is the case with our GPC forest) takes noticeable time, but if you use my tool for evaluation, it should measure the right "matching" time, without initialization.

> Also, I'm keen to experiment with different descriptors other than dct.
It should be very easy to incorporate more descriptors to the GPC.

ethanrublee:
So I've tried training, and the gpc evaulation demo on the forests you trained.  Looks great!

ethanrublee:
Training finished on first image from each of sintel, and resulting forest worked in the evaluation.

modules/optflow/samples/gpc_evaluate.cpp
VladX: There is, but apparently it doesn't work for Point* types. Had to write my own.

VladX: Btw, I've seen some pieces of code where people get around this by writing like
`sqrt(pt.dot(pt))`

modules/optflow/src/sparse_matching_gpc.cpp
VladX: One of the authors of Global Patch Collider reports that DAISY descriptors also work well. OpenCV has DAISY. I will try to experiment with it.
Do you have some particular descriptors in mind?

cbalint13:
@VladX, @ethanrublee ,

Some hints:

  *  Could You let users pass ```cv::Ptr<cv::DescriptorExtractor>``` to your API ? This would allow to pass what desctriptor they want from OpenCV. You can then invoke defaultNorm() and descriptorSize() to choose your L1/L2/Hamming distance function. You might add this feature as seconday overloaded function and leave what you have now as primary option.

  * As alternative (perhaps as overloaded optional function) You could use [HDF5](http://docs.opencv.org/trunk/d3/d23/classcv_1_1hdf_1_1HDF5.html) which is very fast and highly compatible with outside world.

VladX:
@cbalint13 Thank you for the suggestions! I will add support for more descriptors supported by OpenCV. What I'm afraid, however, is that I need to compute descriptors for every point in the image: it would probably take eternity for most descriptors to be computed for every possible KeyPoint. Detecting small subset of useful keypoints with detect() doesn't make sense for Global Patch Collider algorithm.

tneumann:
@VladX To me, the training procedure seems to be different than the one described in the Global Patch Collider paper (Section 3.2). The paper uses triplets of patches (for each triplet: reference, correct, and false patch) and then optimizes according to the harmonic mean between recall and precision, Eq. (2). The training procedure implemented by you seems to just use precision as the objective score and uses only pairs consisting of the reference and correct patch. Is there a specific reason for implementing it in this way? Does this training approach perform better than the version in the paper?

VladX:
@tneumann the training procedure is different indeed. In the early testing I tried the triplet approach against simple pairs of reference/positive patches and found no useful gain from triplets. So I decided to stick with the simple pairs to not overcomplicate implementation. A little bit later I had a conversation with the author of the algorithm and he explained the original training procedure in details. Actually, I misunderstood the concept of precision/recall in my early implementation.
Right now I'm trying to experiment with the triplets again. I will rewrite the training procedure in case I find it useful.

tneumann:
@VladX excellent! Let me know what you find out. 

In the paper at the end of Sec. 3.2 there are some additional hints on how to improve training efficiency. For example, they "share features across nodes within the same layer", although I'm not completely sure what that means - maybe you can just train a random fern (series of hyperplane tests) instead of a full decision tree. The authors also mentions a method to search for the best threshold at each node (you use the median right now). There are also other possible collision detection strategies for finding the correspondences, e.g. the voting scheme proposed in [1]. 

[1] [Dou et al.: "Fusion4D: Real-time Performance Capture of Challenging Scenes" (SIGGRAPH 2016)](http://www.samehkhamis.com/dou-siggraph2016.pdf)

VladX:
@tneumann You can check [triplets branch](https://github.com/VladX/opencv_contrib/tree/triplets) of my repository where I replaced simple pairs with triplets. Taking random patch nearby the ground truth patch doesn't  seem to work just as before -- results are the same and sometimes (for bad precision/recall scoring choice) even worse. However, a series of sampling show slight improvement. I scan for a negative patch on a circle with the increasing radius around the ground truth location and take the one with the minimal L2 distance to the reference patch.

The particular choice of threshold (bias term) shouldn't have a visible impact. I've chosen median to keep the tree balanced so all the patches will have roughly the same path length falling down the tree and correspondence finding will be slightly faster which is important.

Voting method is interesting, I'll look into it.
Thanks for the suggestions!

VladX:
Seems like the test `Features2d_DescriptorExtractor_LUCID.regression` is broken.
Pushed some commits and the test failed.
I reverted my previous commits until the commit which got green checkmark several days ago and guess what: it is failed again.
After a quick look over this test I figured out that I have nothing to do with it. I was able to run it locally and got all test passed.

alalek:
@VladX Don't worry. This test failure is not related to you patch. We will fix it soon.

mshabunin:
@VladX @ethanrublee is it ready for integration?

VladX:
@mshabunin I guess it is.

But it would be nice if someone look over the code and recheck everything.

modules/optflow/src/sparse_matching_gpc.cpp
```diff
@@ -94,20 +117,321 @@ struct PartitionPredicate2
...
+  feature[16] = sumInt( imgCh[1], i, j, k, k );
+  feature[17] = sumInt( imgCh[2], i, j, k, k );
+
+  patchDescr.feature /= patchRadius;
+}
+
+#define STR_EXPAND_( arg ) #arg
+#define STR_EXPAND( arg ) STR_EXPAND_( arg )
+
+ocl::ProgramSource _ocl_getDCTPatchDescriptorSource(
```
mshabunin: It will be better to extract OpenCL kernel to separate file in `src/opencl` subfolder: [example](https://github.com/opencv/opencv/tree/master/modules/imgproc/src/opencl).

modules/optflow/src/sparse_matching_gpc.cpp
```diff
@@ -94,20 +117,321 @@ struct PartitionPredicate2
...
+    }
+  }
+};
+
+void getAllDCTDescriptorsForImage( const Mat *imgCh, std::vector< GPCPatchDescriptor > &descr, const GPCMatchingParams &mp )
+{
+  const Size sz = imgCh[0].size();
+  descr.reserve( ( sz.height - 2 * patchRadius ) * ( sz.width - 2 * patchRadius ) );
+
+  if ( mp.useOpenCL && ocl::useOpenCL() )
```
mshabunin: It is better to extract OpenCL specific code to separate function and use `CV_OCL_RUN` macro and `#ifdef HAVE_OPENCL` guards. Please take a look at the [example](https://github.com/opencv/opencv/blob/master/modules/imgproc/src/blend.cpp#L94-L151).

Also I think it is better to remove `GPCMatchingParams::useOpenCL` field and use only global `ocl::useOpenCL()` function (+ check whether the input is UMat).

modules/optflow/include/opencv2/optflow/sparse_matching_gpc.hpp
```diff
@@ -37,68 +37,159 @@ or tort (including negligence or otherwise) arising in any way out of
...
+
 struct CV_EXPORTS_W GPCPatchDescriptor
 {
-  static const unsigned nFeatures = 18; // number of features in a patch descriptor
+  static const unsigned nFeatures = 18; //!< number of features in a patch descriptor
   Vec< double, nFeatures > feature;
 
-  GPCPatchDescriptor( const Mat *imgCh, int i, int j );
+  double dot( const Vec< double, nFeatures > &coef ) const
+  {
```
mshabunin: It is better to move all implementation to source files (where possible).

mshabunin:
Could you please add basic test to `opencv_contrib/modules/optflow/test/test_OF_accuracy.cpp`?

modules/optflow/src/sparse_matching_gpc.cpp
VladX: Fixed.

Regarding `GPCMatchingParams::useOpenCL`: sometimes CPU implementation of this particular function works faster. With this option it is possible to benefit from faster OpenCL versions of core functions (like `split()` and `cvtColor()`) while leaving descriptor computation to CPU.

mshabunin:
:+1:

PR #758, opened 2016-08-22
TLD with parallel_for_ (targeting TBB enabled builds)

TheDZhon:

### DIFF
--- a/modules/tracking/src/tldDetector.cpp
+++ b/modules/tracking/src/tldDetector.cpp
@@ -41,6 +41,8 @@
 
 #include "tldDetector.hpp"
 
+#include <opencv2/core/utility.hpp>
+
 namespace cv
 {
 	namespace tld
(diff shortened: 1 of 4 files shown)


StevenPuttemans:
Do squash your commits into 1! That being said :+1: for the update!

TheDZhon:
Thank you for commenting!
It's done.

sovrasov:
:+1:

sovrasov:
FYI: [benchmark sample output](https://gist.github.com/sovrasov/9e92a24c45bd0690890d6a40fd0a9c83)

TheDZhon:
Ok, I think that problem is solved: standardPatches std::vector used copy-cted Mat_ instances with resize call. The instances is reference counted and uses same memory after all.

New version explicitly clones Mat_ instances with standardPatches growth. Parallel section body now enlarged with resample calls.

sovrasov:
With the last update performance has decreased comparing to previous parallel code, but results are more stable. :+1:  again

PR #759, opened 2016-08-23
fix CommandLineParser usage in samples

berak:
resolves https://github.com/opencv/opencv/issues/7151

### This pullrequest changes
remove "false" as default value for args used with CommandLineParser::has()


alalek:
:+1:

PR #761, opened 2016-08-23
This commit contains all significant contributions until 23/8/2016

anguelos:
### This pullrequest changes

Implemented ImagePreprocessor a genealogy of functors for preprocessing
samples before they are fed to a CNN.

Implemented TextImageClassifer as an abstraction of any object that
provides a vector of probabilities given an image. A gluing class
providing compatibility to callbacks for OCRBeamsearchDecoder allows
to plug any such object to the existing algorithms.

Implemented DeepCNN a class that is supposed to load pre-trained Caffe
models. Although several back-ends have been planed the only operational 
one is the one employing Caffe its self. GPU is fully supported but can only
be enabled and disabled globally.

Implemented a class using a DeepCNN that implements word-spotting

Implemented a python and a C++ demo demonstrating the Word-spotting
Class

Implemented a class that generates synthetic text samples at high speeds
and with a full pipeline of realistic distortions for training
Word-spotting CNN's who need millions of samples.

Implemented a python demo demonstrating the usage of the text synthesizer


modules/text/include/opencv2/text/ocr.hpp
```diff
@@ -61,82 +65,126 @@ enum
     OCR_LEVEL_TEXTLINE
 };
 
-//base class BaseOCR declares a common API that would be used in a typical text recognition scenario
+//base class BaseOCR declares a common API that would be used in a typical text
```
mshabunin: Please, restore the formatting of all documentation comments and code.

modules/text/src/ocr_beamsearch_decoder.cpp
```diff
@@ -43,6 +43,7 @@
 #include "precomp.hpp"
 #include "opencv2/imgproc.hpp"
 #include "opencv2/ml.hpp"
+#include "opencv2/highgui.hpp"
```
mshabunin: Is it needed?

modules/text/CMakeLists.txt
```diff
@@ -1,25 +1,72 @@
...
+endif() #HAVE_CAFFE
+
+message(STATUS "TEXT CAFFE SEARCH")
+if()
+  message(STATUS "TEXT NO CAFFE CONFLICT")
+else()
+  message(STATUS "TEXT CAFFE CONFLICT")
+endif()
+
+find_package(Qt5Gui)
```
mshabunin: It will be better to use `HAVE_QT5` from the main repository. ([examples](https://github.com/opencv/opencv/search?utf8=%E2%9C%93&q=HAVE_QT5))

mshabunin:
Ping @anguelos .

PR #768, opened 2016-08-31
Fix GCC 5.x warnings

mshabunin:
Companion for opencv/opencv#7210

### DIFF
--- a/modules/cvv/test/test_main.cpp
+++ b/modules/cvv/test/test_main.cpp
@@ -1,3 +1,3 @@
 #include "test_precomp.hpp"
 
-CV_TEST_MAIN(".")
+CV_TEST_MAIN(".", /*Empty VA_ARGS for C++11*/)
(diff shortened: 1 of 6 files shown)


alalek:
:+1:

modules/ximgproc/src/l0_smooth.cpp
```diff
@@ -212,14 +212,6 @@ namespace
         D.copyTo(dst);
     }
 
-    void addComplex(InputArray aSrc, int bSrc, OutputArray dst)
```
mshabunin: I got warning about unused method here.

modules/cvv/test/test_main.cpp
```diff
@@ -1,3 +1,3 @@
 #include "test_precomp.hpp"
 
-CV_TEST_MAIN(".")
+CV_TEST_MAIN(".", /*Empty VA_ARGS for C++11*/)
```
mshabunin: `cvv` module is compiled with C++11 flags, and gcc does not like empty VA_ARGS in variadic macros. I hope it will not break anything.

Issue #230, opened 2015-05-13
updateMotionHistory, calcMotionGradient, calcGlobalOrientation not exposed to python/java


berak:
they're tagged as CV_EXPORTS_W in optflow/include/opencv2/optflow/motempl.hpp, but don't end up in the python or java wrappers.

this is probably due to a namespace problem, all other code is in `namespace cv::optflow`, the code from motempl.cpp in `namespace cv::motempl`

changing the namespace to `cv::optflow`, too will fix it, but feels like a gross change, hopefully someone has a better idea here.

DuinoDu:
Yes, I come across the same problem when running the python sample "motempl.py" in opencv_contrib/optflow/sample. I need to change "cv2.updateMotionHistory" to "cv2.motempl.updateMotionHistory". I guess this is because this algorithm was once in opencv lib. But I don't know why it is moved to contrib lib. Can anyone give some reasons?

berak:
@DuinoDu , originally, there was a problem with the script generator, it could not handle nested c++ namespaces correctly, so those functions were not available from python at all.

but now this has changed (and this issue can be closed).

you are right, you need to call `cv2.motempl.updateMotionHistory` , thanks for pointing out a flaw in the motempl.py sample, i'll go and fix it later !

PR #773, opened 2016-09-09
optflow: updated motempl.py sample

berak:
vaguely related to #230

* removed video.py and common.py dependancies
* added motempl namespace to functions


berak:
question: since this is just a (python) sample, and won't get tested by buildbots anyway, should i add a **WIP** tag or similar ?

modules/optflow/samples/motempl.py
```diff
@@ -43,10 +47,10 @@ def draw_motion_comp(vis, (x, y, w, h), angle, color):
         gray_diff = cv2.cvtColor(frame_diff, cv2.COLOR_BGR2GRAY)
```
sovrasov: To prevent error in `absdiff` in case of empty frame was retrieved by `cam.read()` block
```
if frame is None:
            break
```
is needed after `cam.read()`

sovrasov:
@berak we have no special tags for similar cases. All commits are tested on buildbot anyway.

berak:
numpy >=1.10 seems to be picky comparing a (supposed) numpy array against None, so i'll opt for `ret==False`

> we have no special tags for similar cases. 

ok. just asking.

sovrasov:
:+1:

PR #778, opened 2016-09-14
Fix rounding in FREAK descriptor

### LINKED ISSUE #477: Incorrect rounding
https://github.com/Itseez/opencv_contrib/blob/master/modules/xfeatures2d/src/freak.cpp#L541

Rounding towards positive infinity is performed here. If the initial angle is negative, the thetaIdx could be off by one. This will introduce random-looking orientation errors of about ~1.4 degrees.

### LINKED ISSUE #478: Missing rounding
The division performed here does not account for round correction:
https://github.com/Itseez/opencv_contrib/blob/master/modules/xfeatures2d/src/freak.cpp#L675

This can introduce noise in the patch values, which can actually cause instability in the descriptor bits, since patch values tend to be similar for a given keypoint, and off-by-one differences will cause the pair comparisons to change.

sovrasov:
**Merge with opencv_extra**: https://github.com/opencv/opencv_extra/pull/313

resolves #478 
resolves #477


terfendail:
:+1:

PR #750, opened 2016-08-17
Python bindings for dnn module

ludv1x:
Python bindings for dnn module

Also merge of PR #738 is added

PR #781, opened 2016-09-18
fixed some compile-time and runtime problems in semantical segmentation

vpisarev:
thanks to Vladimir Bystritsky for the "null bias" fix in convolution_layer.cpp.
made gemmCPU() parallel (probably need to do that in core too) - it noticeably improved speed of the sample


alalek:
Mentioned fix waiting in this PR #750. I believe it should be merged first.

vpisarev:
will rebase it w.r.t. the new master

PR #787, opened 2016-09-23
Plot module update

### LINKED ISSUE #581: Add an option to show points only for plot module

sovrasov:
Python bindings for plot module are enabled.
resolves #581


mshabunin:
:+1:

PR #788, opened 2016-09-25
Fix Board draw function in aruco module

sergarrido:
For instance:
./bin/example_aruco_create_board --bb=1 -d=4 -l=120 -m=0 -s=24 -w=4 -h=2 image.png
creates a board with a (incorrect) white border on bottom and right.

Also, marker border (`borderBits` variable) is not used correctly.

### DIFF
--- a/modules/aruco/src/aruco.cpp
+++ b/modules/aruco/src/aruco.cpp
@@ -1542,22 +1542,22 @@ void _drawPlanarBoardImpl(Board *_board, Size outSize, OutputArray _img, int mar
             // remove negativity
             p1.x = p0.x - minX;
             p1.y = p0.y - minY;
-            pf.x = p1.x * float(markerZone.cols - 1) / sizeX;
-            pf.y = float(markerZone.rows - 1) - p1.y * float(markerZone.rows - 1) / sizeY;
+            pf.x = p1.x * float(markerZone.cols) / sizeX;
+            pf.y = float(markerZone.rows - 1) - p1.y * float(markerZone.rows) / sizeY;
             outCorners[j] = pf;
         }
 
         // get tiny marker
-        int tinyMarkerSize = 10 * dictionary.markerSize + 2;
+        int tinyMarkerSize = 10 * (dictionary.markerSize + 2 * borderBits);
         Mat tinyMarker;
         dictionary.drawMarker(_board->ids[m], tinyMarkerSize, tinyMarker, borderBits);
 
         // interpolate tiny marker to marker position in markerZone
         Mat inCorners(4, 1, CV_32FC2);
         inCorners.ptr< Point2f >(0)[0] = Point2f(0, 0);
-        inCorners.ptr< Point2f >(0)[1] = Point2f((float)tinyMarker.cols, 0);
-        inCorners.ptr< Point2f >(0)[2] = Point2f((float)tinyMarker.cols, (float)tinyMarker.rows);
-        inCorners.ptr< Point2f >(0)[3] = Point2f(0, (float)tinyMarker.rows);
+        inCorners.ptr< Point2f >(0)[1] = Point2f((float)tinyMarker.cols-1, 0);
+        inCorners.ptr< Point2f >(0)[2] = Point2f((float)tinyMarker.cols-1, (float)tinyMarker.rows-1);
+        inCorners.ptr< Point2f >(0)[3] = Point2f(0, (float)tinyMarker.rows-1);
 
         // remove perspective
         Mat transformation = getPerspectiveTransform(inCorners, outCorners);


modules/aruco/src/aruco.cpp
```diff
@@ -1542,22 +1542,22 @@ void _drawPlanarBoardImpl(Board *_board, Size outSize, OutputArray _img, int mar
             // remove negativity
             p1.x = p0.x - minX;
             p1.y = p0.y - minY;
-            pf.x = p1.x * float(markerZone.cols - 1) / sizeX;
-            pf.y = float(markerZone.rows - 1) - p1.y * float(markerZone.rows - 1) / sizeY;
+            pf.x = p1.x * float(markerZone.cols) / sizeX;
```
paroj: I think this should be `p1.x * float(markerZone.cols) / sizeX - 1`

modules/aruco/src/aruco.cpp
```diff
@@ -1542,22 +1542,22 @@ void _drawPlanarBoardImpl(Board *_board, Size outSize, OutputArray _img, int mar
...
+            pf.y = float(markerZone.rows - 1) - p1.y * float(markerZone.rows) / sizeY;
```
paroj: and this should be rewritten as `markerZone.rows - p1.y * float(markerZone.rows) / sizeY - 1` for readability

modules/aruco/src/aruco.cpp
sergarrido: Why?
- `p0` is the coordinate in the Board ref system
- `p1` is the coordinate in the Board ref system after removing negativity
- `pf` is the coordinate in the image ref system (in pixels)
- `sizeX` is the total range in Board coordinates (for x axis)
- `markerZone.cols` is the total range in image coordinates (for x axis in pixels)

In this line we are just converting from `p1` to `pf`. This is scaling with the rate `markerZone.cols/sizeX`

modules/aruco/src/aruco.cpp
sergarrido: I think it is this way just for uniformity with the previous line `pf.x = ...`

modules/aruco/src/aruco.cpp
paroj: yeah, you are right. `p1.x in [0;sizeX)` so `pf.x != markerZone.cols`

modules/aruco/src/aruco.cpp
paroj: but then you neither need the -1 here

modules/aruco/src/aruco.cpp
paroj: `p1.x in [0;sizeX]` so for `p1.x == sizeX` you get `markerZone.cols` and are out of bounds

modules/aruco/src/aruco.cpp
paroj: `(1.0f - p1.y / sizeY) * float(markerZone.rows) - 1`

modules/aruco/src/aruco.cpp
paroj: which means previous code was correct

modules/aruco/src/aruco.cpp
paroj: and this should be `(1.0f - p1.y / sizeY) * float(markerZone.rows - 1)`

modules/aruco/src/aruco.cpp
sergarrido: But `pf` is not used to access the image, but to create the transformation later in `getPerspectiveTransform`.
So we have to use the `markerZone.cols` (without -1).

modules/aruco/src/aruco.cpp
sergarrido: Indeed there was another error here.
This should be:
`pf.y = float(markerZone.rows) - p1.y * float(markerZone.rows) / sizeY;`
or:
`pf.y = (1.0f - p1.y / sizeY) * float(markerZone.rows)`

modules/aruco/src/aruco.cpp
paroj: then you compute the wrong transform.
the maximum valid value for `pf.x` is  `markerZone.cols - 1`.

paroj: (we want the transform to be `(maxX, maxY) -> max(markerZone)`)

sergarrido: Sorry, I dont agree.
In `getPerspectiveTransform` the maximum size is `markerZone.cols` since you have to count the length of the last pixel.

For instance, if we are in a hypothetical case where `markerZone` is 1x1 pixel, the transformation should be:
`(minX, minY) -> (0,0)`
`(maxX, maxY) -> (1,1)`

If we do what you are proposing we would get:
`(minX, minY) -> (0,0)`
`(maxX, maxY) -> (0,0)`

Which I think it is clearly wrong.

paroj: according to [this thread](http://answers.opencv.org/question/87923/sub-pixel-coordinate-origin/) the location `(0,0)` refers to the center of the top left pixel.

This would then mean we have to write
`pf.x = p1.x * float(markerZone.cols - 1) / sizeX - 0.5;`
still need to verify this however..

paroj:
I took a step back: computing a perspectiveTransform for a pure scaling operation of fixed point 2D coordinates seemed over the top for me. See the linked PR for an alternative take.

sergarrido:
I see. I implemented it this way to allow planar boards with markers not necessarily parallel to the image axis.

paroj:
> I see. I implemented it this way to allow planar boards with markers not necessarily parallel to the image axis.

did you try this at some point? I guess this does not go well with `INTER_NEAREST`..

paroj:
I was curious on how to use the `warp*` functions correctly and came up with the alternative PR.

I still prefer the solution just using `resize` as probably nobody uses non aligned markers and if they do the results will be bad due to INTER_NEAREST

sovrasov:
Fix already done in #792

PR #789, opened 2016-09-26
Fix for SFM module build on Windows

terfendail:
### This pullrequest changes
This PR fix the build of SFM module on Windows with VisualStudio


PhilLab:
Thanks! Is the installation process similar to the [linux doc](http://docs.opencv.org/master/db/db8/tutorial_sfm_installation.html)?

terfendail:
There are a few special moments related to prerequisite installation(since there is no package manager for that). i.e. glog and gflags should be built as dynamic libraries, while SuiteSparse should be static. I'm going to repeat build process from scratch and prepare update for installation tutorial.

vpisarev:
:+1: If I'm not mistaken, we use gflags and glog in a few OpenCV packages already; probably it makes sense to unify the scripts; maybe put some version of the libraries into 3rdparty, similar to gtest that we put to OpenCV

LaurentBerger:
So something must be wrong in my installed :

Caffe:   NO
Protobuf:   NO
Glog:   NO
Found required Ceres dependency: Eigen version 3.3.0 in F:/lib/eigen-eigen
No preference for use of exported glog CMake configuration set, and no hints for include/library directories provided. Defaulting to preferring an installed/exported glog CMake configuration if available.
Found installed version of glog: F:/lib/glog
Detected glog version: 0.3.4
Found required Ceres dependency: Glog in 
Found Ceres version: 1.12.0 installed in: C:/Program Files/Ceres with components: [SchurSpecializations, OpenMP]
Checking SFM deps... TRUE
Module opencv_sfm disabled because the following dependencies are not found: Eigen

LaurentBerger:
@terfendail Hi,
I have build sfm on windows using ceres glog gflags and protobuf. I have found some errors in CMake. Do you want to help me to solve or understand what is wrong in cmake or in my install ?

terfendail:
@LaurentBerger Hi,
I'm not a CMake guru but will try to do my best.

LaurentBerger:
Thanks for your answer.
I think I solve some problem on my configuration (windows 10 VS2015). First I have build ceres-solver and ask to ceres mailing list and many problems are solved now (see https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/ceres-solver/IaWVXbNasqM/ZuDqzw9bBAAJ).

In multiwiew cmakelists some lines must be add :
ADD_LIBRARY(multiview STATIC ${MULTIVIEW_SRC} ${MULTIVIEW_HDRS})
**if (MSVC)
    set_target_properties(multiview PROPERTIES  COMPILE_DEFINITIONS  "_USE_MATH_DEFINES")
endif()**

Now I have made a PR in OpenCV to detect glog and gflags using CERES cmake. You don't need those lines if pr is accepted

if(NOT Ceres_FOUND)  # Looks like Ceres find glog on the own, so separate search isn't necessary
  find_package(Glog QUIET)
endif()

PR #792, opened 2016-09-29
aruco: correct marker coordinates in drawPlanarBoard

### LINKED ISSUE #592: Aruco: Charuco board markers are imprecise
The Charuco board code uses several image scaling and warping steps to create the final board image. Along the way, pixel precision gets lost. Creating a Charuco board with just as many pixels as are needed to place all markers will show that some markers are one row/column too small. Also with really large image sizes, the differences between the markers' square sizes are noticeable.

paroj:
see #788. fixes #592.

### DIFF
--- a/modules/aruco/include/opencv2/aruco.hpp
+++ b/modules/aruco/include/opencv2/aruco.hpp
@@ -235,15 +235,15 @@ class CV_EXPORTS_W Board {
     *
     */
     CV_WRAP static Ptr<Board> create(InputArrayOfArrays objPoints, const Ptr<Dictionary> &dictionary, InputArray ids);
-    // array of object points of all the marker corners in the board
-    // each marker include its 4 corners, i.e. for M markers, the size is Mx4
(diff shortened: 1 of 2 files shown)


sovrasov:
:+1:

sovrasov:
I think this approach is preferable because it covers more usecases even if they seems to be non-realistic.

sergarrido:
Maybe you could check if marker is parallel to image axis or not, and depending on that use INTER_NEAREST or a different interpolation approach.

paroj:
> Maybe you could check if marker is parallel to image axis or not, and depending on that use INTER_NEAREST or a different interpolation approach.

this will break the `borderValue` hack used for masking. I still think that supporting rotated markers is more trouble than its worth..

paroj:
done. `BORDER_TRANSPARENT` did the trick

sovrasov:
@paroj thanks for patch! :+1: again

PR #795, opened 2016-09-30
Fix TLD fail on gray images

### LINKED ISSUE #625: Tracking TLD: Greyscale Image bug
I am using the Tracking Module with the TLD variant. When I pass a greyscale image (CV_8U) it gives an assertion error:

`openCV Error: Assertion failed (scn == 3 || scn == 4)`

That means it is trying to convert the image from RGB (CV_8UC3) to greyscale...

I think that it should be able to detect that is CV_8U and avoid this step.

Or does the algorithm needs to have RGB images for another reason?

sovrasov:
resolves #625


savuor:
:+1:

PR #462, opened 2015-12-04
Exporting saliency classes to Python

Ignotus:
resolves #681
The following proposed patch exports StaticSaliencySpectralResidual and MotionSaliencyBinWangApr2014 algorithms to Python.

```python
import cv2
img = cv2.imread('DSCF4494.JPG')
saliency = cv2.saliency.StaticSaliencySpectralResidual()
saliency.computeSaliency(img)
```

PR #796, opened 2016-10-04
Exporting saliency classes to Python

### LINKED ISSUE #681: Dose module "saliency" work in python?
It does not seem like the "saliency" has python bindings.

I success build and import cv2 in python.
But "saliency" not work in python, though it work in c++.

This is very useful if it can use.

sovrasov:
resolves #681
replaces #462

### DIFF
--- a/modules/saliency/CMakeLists.txt
+++ b/modules/saliency/CMakeLists.txt
@@ -1,3 +1,11 @@
 set(the_description "Saliency API")
 set(OPENCV_MODULE_IS_PART_OF_WORLD OFF)
+add_definitions(-DOBJECTNESS_MODELS_PATH="${CMAKE_INSTALL_PREFIX}/${OPENCV_OTHER_INSTALL_PATH}/Objectness")
+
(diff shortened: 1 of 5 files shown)


modules/saliency/CMakeLists.txt
```diff
@@ -1,3 +1,11 @@
 set(the_description "Saliency API")
 set(OPENCV_MODULE_IS_PART_OF_WORLD OFF)
+add_definitions(-DOBJECTNESS_MODELS_PATH="${CMAKE_INSTALL_PREFIX}/${OPENCV_OTHER_INSTALL_PATH}/Objectness")
```
mshabunin: I don't think it's a good approach.

vpisarev:
thanks! :+1:

saeed68gm:
What this change reverted at the end? so still no saliency for python?

sovrasov:
@saeed68gm saliency works, from CMake was removed a couple of redundant definitions.

PR #797, opened 2016-10-04
Add threshold setter for FaceRecognizer class

furushchev:
### Add virtual method to set threshold for FaceRecognizer class

This pull request adds virtual method `setThreshold` to `FaceRecognizer` class.
This method is already implemented in its child classes: `BasicFaceRecognizer` and `LBPHFaceRecognizer`.
Currently `FaceRecognizer` has only getter `getThreshold`.


sovrasov:
:+1:

furushchev:
Thanks for merging!

PR #801, opened 2016-10-05
Fix line descriptor module build on Win64

### LINKED ISSUE #799: ERROR result of 32-bit shift implicitly converted to 64 bits: When building opencv_line_descriptor on 64 bit RELEASE

- OpenCV => 3.1
- Operating System / Platform => Windows 7 (ultimate N), 64 bit

##### Detailed description

When building RELEASE version of line_descriptor project the following lines produce a build fail https://github.com/opencv/opencv_contrib/blob/6cd8e9f556c8c55c05178dec05d5277ae00020d9/modules/line_descriptor/src/bitops.hpp#L146-L147

> Severity Code Description Project File Line Suppression State Warning C4334 '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?) opencv_line_descriptor e:\git\opencv_contrib\modules\line_descriptor\src\bitops.hpp 146

The problem seems to be that the number '1' is interpreted as a 32 bit number and then involved in a 64 bit shift.

Solution is to replace the '1's with a static cast:
```
printf( "%llu", (long long int) tmp / (UINT64) (static_cast<UINT64>(1) << j ) );
tmp = tmp - ( tmp / (UINT64) (static_cast<UINT64>(1) << j ) ) * (UINT64) (static_cast<UINT64>(1) << j );
```

##### Steps to reproduce

Simply build the contrib lib in 64 bit RELEASE.

sovrasov:
resolves #799

### DIFF
--- a/modules/line_descriptor/src/bitops.hpp
+++ b/modules/line_descriptor/src/bitops.hpp
@@ -143,8 +143,8 @@ inline void print_code( UINT64 tmp, int b )
 {
   for ( long long int j = ( b - 1 ); j >= 0; j-- )
   {
-    printf( "%llu", (long long int) tmp / (UINT64) ( 1 << j ) );
-    tmp = tmp - ( tmp / (UINT64) ( 1 << j ) ) * (UINT64) ( 1 << j );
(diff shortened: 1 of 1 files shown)


mshabunin:
:+1:

PR #808, opened 2016-10-08
text: fix memleak in ocr_hmm_decoder

### LINKED ISSUE #529: Memory leak in OCRHMMClassifierCNN
In file ocr_hmm_decoder.cpp, the function OCRHMMClassifierCNN has a memory leak. The predicted labels aren't de-allocated after copying to output vectors. This could be added as the last line:
```.cpp
delete[] p;
```
OCRBeamSearchClassifierCNN handles this issue by creating the vector as a pointer to already allocated "p" array.

berak:
resolves #529

added a delete[] call in ocr_hmm_decoder.cpp

### DIFF
--- a/modules/text/src/ocr_hmm_decoder.cpp
+++ b/modules/text/src/ocr_hmm_decoder.cpp
@@ -935,7 +935,7 @@ public:
 
 protected:
     void normalizeAndZCA(Mat& patches);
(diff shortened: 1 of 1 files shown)


sovrasov:
:+1:

paroj:
you still leak at the CV_Error call some lines above. Change to std::vector?

berak:
good idea, i'll do that.

PR #811, opened 2016-10-10
bug fix, parameter adjustment, and some enhancements

### LINKED ISSUE #170: Surface Matching and x64 :  heap corruption
Hi,

In surface_matching\src\ppf_match_3d.cpp a function "murmurHash" is used like the following :
  KeyType hashKey=0;
  murmurHash(key, 4*sizeof(int), 42, &hashKey);

In case of a 64bit system, "murmurHash" is defined via a macro as "hashMurmurx64" in surface_matching\src\hash_murmur.hpp
However this function takes a void\* as a parameter which should at least of length (2*sizeof(unsigned int)). Here &hashKey is given, and hashKey is an unsigned int, thus there is a heap corruption.

A workaround consists in forcing "murmurHash" to be defined as hashMurmurx86 even on 32 bit system, but in this case I don't understand the purpose of specific code for x64 systems.

Regards,
Romain

tolgabirdal:
### This pullrequest changes
some parameters in the samples, and includes a tiny performance enhancement when searching for the model.

### DIFF
--- a/modules/surface_matching/include/opencv2/surface_matching/t_hash_int.hpp
+++ b/modules/surface_matching/include/opencv2/surface_matching/t_hash_int.hpp
@@ -55,7 +55,12 @@ namespace ppf_match_3d
 //! @addtogroup surface_matching
 //! @{
 
+#if (defined x86_64 || defined _M_X64)
+#include 
+typedef uint64_t KeyType;
+#else
 typedef unsigned int KeyType;
+#endif
 
 typedef struct hashnode_i
 {
--- a/modules/surface_matching/samples/ppf_load_match.cpp
+++ b/modules/surface_matching/samples/ppf_load_match.cpp
@@ -106,7 +106,7 @@ int main(int argc, char** argv)
     cout << endl << "Starting matching..." << endl;
     vector<Pose3DPtr> results;
     tick1 = cv::getTickCount();
-    detector.match(pcTest, results, 1.0/40.0, 0.05);
+    detector.match(pcTest, results, 1.0/4.0, 0.05);
     tick2 = cv::getTickCount();
     cout << endl << "PPF Elapsed Time " <<
          (tick2-tick1)/cv::getTickFrequency() << " sec" << endl;
--- a/modules/surface_matching/src/ppf_match_3d.cpp
+++ b/modules/surface_matching/src/ppf_match_3d.cpp
@@ -241,7 +241,8 @@ void PPF3DDetector::trainModel(const Mat &PC)
   float dy = yRange[1] - yRange[0];
   float dz = zRange[1] - zRange[0];
   float diameter = sqrt ( dx * dx + dy * dy + dz * dz );
-
+  m_diameter = (double)diameter;
+  
   float distanceStep = (float)(diameter * sampling_step_relative);
 
   Mat sampled = samplePCByQuantization(PC, xRange, yRange, zRange, (float)sampling_step_relative,0);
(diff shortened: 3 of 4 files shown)


modules/surface_matching/include/opencv2/surface_matching/t_hash_int.hpp
```diff
@@ -55,7 +55,12 @@ namespace ppf_match_3d
 //! @addtogroup surface_matching
 //! @{
 
+#if (defined x86_64 || defined _M_X64)
+#include
```
terfendail: Looks like this line could brake compilation

modules/surface_matching/src/ppf_match_3d.cpp
```diff
@@ -539,51 +540,56 @@ void PPF3DDetector::match(const Mat& pc, std::vector<Pose3DPtr>& results, const
...
+          KeyType hashValue = hashPPF(f, angle_step, distanceStep);
+  
+          // we don't need to call this here, as we already estimate the tsg from scene reference point
+          // double alpha = computeAlpha(p1, n1, p2);
+          p2t[1] = tsg[1] + row2[0] * p2[0] + row2[1] * p2[1] + row2[2] * p2[2];
+          p2t[2] = tsg[2] + row3[0] * p2[0] + row3[1] * p2[1] + row3[2] * p2[2];
+
+          alpha_scene=atan2(-p2t[2], p2t[1]);
+
+          if ( alpha_scene != alpha_scene)
```
terfendail: Looks like impossible case. I think it make sense to remove the check(even if compiler optimize it out)

modules/surface_matching/src/ppf_match_3d.cpp
```diff
@@ -539,51 +540,56 @@ void PPF3DDetector::match(const Mat& pc, std::vector<Pose3DPtr>& results, const
...
+          {
+            continue;
+          }
+
+          if (sin(alpha_scene)*p2t[2]<0.0)
```
terfendail: IMHO it make sense to invert condition here and completely remove line 564

modules/surface_matching/src/ppf_match_3d.cpp
```diff
@@ -539,51 +540,56 @@ void PPF3DDetector::match(const Mat& pc, std::vector<Pose3DPtr>& results, const
...
-          unsigned int accIndex = corrI * numAngles + alpha_index;
+            unsigned int accIndex = corrI * numAngles + alpha_index;
 
-          accumulator[accIndex]++;
-          node = node->next;
-        }
+            accumulator[accIndex]++;
+            node = node->next;
+          }
+		}
```
terfendail: Could you please fix indentation

modules/surface_matching/samples/ppf_load_match.cpp
```diff
@@ -106,7 +106,7 @@ int main(int argc, char** argv)
     cout << endl << "Starting matching..." << endl;
     vector<Pose3DPtr> results;
     tick1 = cv::getTickCount();
-    detector.match(pcTest, results, 1.0/40.0, 0.05);
+    detector.match(pcTest, results, 1.0/4.0, 0.05);
```
bmagyar: Now we know this may be related to #1079 although here it's 0.25 not 0.5. Still the same ballpark.

bmagyar:
"bug fix, parameter adjustment, and some enhancements" -- this is addressing at least 3 things at once, changes like this doesn't allow for tracking individual issues or improvements.

Please split that commit into several smaller ones where each one includes changes related to one of the issues.

The written description on the PR is a tad bit too vague. First step is a clear description as to why this should be considered for a merge.

modules/surface_matching/samples/ppf_load_match.cpp
tolgabirdal: Well, for this data both values should work. Some tougher scenarios would require higher values.

alalek:
As a bugfix this should go onto 3.4 branch: https://github.com/opencv/opencv/wiki/Branches


> Merge branch 'master'

Do NOT use merges in PRs.

modules/surface_matching/src/ppf_match_3d.cpp
terfendail: Get the check idea. However it would be better to use `cvIsNaN(alpha_scene)` to make the code cleaner

PR #822, opened 2016-10-17
aruco: correctly round pixels when computing marker size & fix assertion

paroj:

### DIFF
--- a/modules/aruco/src/aruco.cpp
+++ b/modules/aruco/src/aruco.cpp
@@ -1505,7 +1505,7 @@ void _drawPlanarBoardImpl(Board *_board, Size outSize, OutputArray _img, int mar
             pf -= Point2f(minX, minY);
             pf.x = pf.x / sizeX * float(out.cols);
             pf.y = (1.0f - pf.y / sizeY) * float(out.rows);
-            outCorners[j] = Point(pf);
+            outCorners[j] = pf;
         }
 
         // get marker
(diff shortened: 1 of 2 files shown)


modules/aruco/src/aruco.cpp
```diff
@@ -1505,7 +1505,7 @@ void _drawPlanarBoardImpl(Board *_board, Size outSize, OutputArray _img, int mar
             pf -= Point2f(minX, minY);
             pf.x = pf.x / sizeX * float(out.cols);
             pf.y = (1.0f - pf.y / sizeY) * float(out.rows);
-            outCorners[j] = Point(pf);
+            outCorners[j] = pf;
```
sovrasov: Why not use `Point(pf.x + 0.5, pf.y + 0.5)` here?
Will your changes break float-point equality check at L1515?

paroj: this would force a `ceil()` behaviour which is wrong. the problem at hand is that any rounding should only happen after L1512 or we compute the wrong size. This is what now happens implicitly at the Size and Rect constructors. (Point2f -> Point conversion uses a `saturate_cast` at some point)
The L1515 checks whether the points are aligned *exactly*, so two exact input values should also map to the same floating point outputs. If the check fails, we try to apply rotation which is correct.

sovrasov: :+1:

sovrasov:
:+

PR #824, opened 2016-10-19
Fix for SFM module configuration when no gflags have been found

terfendail:
### This pullrequest changes
SFM shouldn't fail overall cmake configuration when no gflags have been found


alalek:
:+1:

PR #825, opened 2016-10-19
aruco: interpolateCornersCharuco - make minMarkers property configurable

paroj:
allows trading corner precision for corner quantity


sovrasov:
:+1:

PR #855, opened 2016-11-16
Fix crash when no lines found in BinaryDescriptor

### LINKED ISSUE #647: Line segment detector crashes when no keylines found!
This bug report follows [This question at OpenCV Answers](http://answers.opencv.org/question/93765/line-segment-detector-crashes-when-no-keylines-found/). In short, the  line detector (both of them) crash when no keylines are found. Either a graceful error (throw etc.) or a normal behavior (e.g. set the keylines detector to a special value that will produce a descriptor of all-zeroes), would be fine.

The code fails at line 2475 of binary_descriptor.cpp, with an `Access violation writing location 0x0000000000000000` message.

The image, a minimal reproducing example, and possibly some more info can be found at [the above mentioned Opencv Answers question.](http://answers.opencv.org/question/93765/line-segment-detector-crashes-when-no-keylines-found/)
**UPDATE:** Since answers.opencv.org seems to be down, i will repost all necessary info here:

The image i refer to is 
06-scale_1_im_1_col
 The code that crashes is `cv::Mat Image= cv::imread("pathToImage", CV_LOAD_IMAGE_COLOR);  Ptr<cv::line_descriptor::BinaryDescriptor> LINES = cv::line_descriptor::BinaryDescriptor::createBinaryDescriptor();
std::vector<cv::line_descriptor::KeyLine> keylines; 
LINES->detect(Image, keylines);`

sovrasov:
Resolves #647

### DIFF
--- a/modules/line_descriptor/src/LSDDetector.cpp
+++ b/modules/line_descriptor/src/LSDDetector.cpp
@@ -104,7 +104,7 @@ inline void checkLineExtremes( cv::Vec4f& extremes, cv::Size imageSize )
 void LSDDetector::detect( const Mat& image, CV_OUT std::vector<KeyLine>& keylines, int scale, int numOctaves, const Mat& mask )
 {
   if( mask.data != NULL && ( mask.size() != image.size() || mask.type() != CV_8UC1 ) )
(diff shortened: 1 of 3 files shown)


mshabunin:
:+1:

PR #857, opened 2016-11-18
Replace dummy pointer with smart in ERStat

### LINKED ISSUE #637: Possible memory leak in ERFilter (text module)
When calling ERFilter's `run()` method or the `MSERsToERStats()` method, I experienced a memory issue. Memory is not completely deallocated after leaving the caller's scope.
I suspected the `crossings = new deque<int>();` line in ERStat's constructor.
Memory might be freed correctly internally, however the ERStat objects returned by `virtual void run( InputArray image, std::vector<ERStat>& regions )` need to be handled, too.

I therefore switched from simple pointers to OpenCV Ptr in the ERStat struct definition: 
`Ptr<std::deque<int> > crossings; //!< horizontal crossings`

I then commented out the delete and NULL assignments to crossings and the leak was gone.

sovrasov:
Resolves #637


alalek:
:+1:

PR #865, opened 2016-11-22
tracking: make datasets dependancy optional

### LINKED ISSUE #864: tracking depends on datasets
the opencv_tracking module depends on opencv_datasets, which again depends on opencv_face.

actually the dependancy is only for a [single example prog](https://github.com/opencv/opencv_contrib/blob/master/modules/tracking/samples/tracker_dataset.cpp)  so it better should be OPTIONAL.

berak:
resolves #864

### DIFF
--- a/modules/tracking/samples/multiTracker_dataset.cpp
+++ b/modules/tracking/samples/multiTracker_dataset.cpp
@@ -39,6 +39,10 @@
 //
 //M*/
 
+#include "opencv2/opencv_modules.hpp"
+#include "opencv2/core.hpp"
+#ifdef HAVE_OPENCV_DATASETS
+
 #include "opencv2/datasets/track_vot.hpp"
 #include <opencv2/core/utility.hpp>
 #include <opencv2/tracking.hpp>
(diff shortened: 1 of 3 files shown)


mshabunin:
@berak , it will break _tracking_ module build with cmake options `-DBUILD_EXAMPLES=ON -DBUILD_opencv_datasets=OFF`:
```
opencv_contrib/modules/tracking/samples/multiTracker_dataset.cpp:42:42: fatal error: opencv2/datasets/track_vot.hpp: No such file or directory
```
Since we don't have mechanism to specify dependencies for each sample app, you have to add guards to `multiTracker_dataset.cpp` itself:
```.cpp
#include "opencv2/opencv_modules.hpp"
#ifdef HAVE_OPENCV_DATASETS
...
```

berak:
ah right, will do !

modules/tracking/samples/multiTracker_dataset.cpp
```diff
@@ -227,4 +234,11 @@ int main(int argc, char *argv[])
 	waitKey(0);
 
 	return 0;
-}
\ No newline at end of file
+}
+
+#else // ! HAVE_OPENCV_DATASETS
+int main() {
+	// !!! this sample requires opencv_datasets !!!
```
sovrasov: It's better to place CV_Error() call here because not everyone looks at sources before running samples.

mshabunin:
:+1:

PR #866, opened 2016-11-22
Fix memory leak in MIL tracker

### LINKED ISSUE #430: MIL Tracker Implementation in 3.0.0 Has Memory Leaking
I noticed obvious memory leaking when I was testing the MIL algorithm (https://github.com/Itseez/opencv_contrib/blob/master/modules/tracking/samples/tracker.cpp) with opencv and opencv_contrib both in 3.0.0. I probably won't have time to look into the source code. Hopefully some one who are familiar with this implementation can figure out where is the problem.

sovrasov:
resolves #430


mshabunin:
:+1:

PR #867, opened 2016-11-22
Remove duplicated code in UnscentedKalmanFilter

sovrasov:
Local copy of Colesky decomposition function was replaced with the similar function from `hal` namespace.

### DIFF
--- a/modules/tracking/include/opencv2/tracking/kalman_filters.hpp
+++ b/modules/tracking/include/opencv2/tracking/kalman_filters.hpp
@@ -62,13 +62,13 @@ public:
     * @param control - the current control vector,
     * @return the predicted estimate of the state.
     */
-    virtual Mat predict( const Mat& control = Mat() ) = 0;
+    virtual Mat predict( InputArray control = noArray() ) = 0;
 
     /** The function performs correction step of the algorithm
     * @param measurement - the current measurement vector,
     * @return the corrected estimate of the state.
     */
-    virtual Mat correct( const Mat& measurement ) = 0;
+    virtual Mat correct( InputArray measurement ) = 0;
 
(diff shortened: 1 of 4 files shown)


modules/tracking/include/opencv2/tracking/kalman_filters.hpp
```diff
@@ -222,6 +224,47 @@ CV_EXPORTS Ptr<UnscentedKalmanFilter> createUnscentedKalmanFilter( const Unscent
 */
 CV_EXPORTS Ptr<UnscentedKalmanFilter> createAugmentedUnscentedKalmanFilter( const AugmentedUnscentedKalmanFilterParams &params );
 
+
+namespace internal {
```
mshabunin: I think it will be better to move it to `tracking/src/<some-hpp-file>`, for example `precomp.hpp`, or some new file.

modules/tracking/include/opencv2/tracking/kalman_filters.hpp
```diff
@@ -222,6 +224,47 @@ CV_EXPORTS Ptr<UnscentedKalmanFilter> createUnscentedKalmanFilter( const Unscent
...
+
+    astep /= sizeof(_Tp);
+    lstep /= sizeof(_Tp);
+
+    for(int i = 0; i < asize; i++)
+        for(int j = 0; j <= i; j++)
+            L[i*lstep + j] = A[i*astep + j];
+
+
+    if(typeid(_Tp) == typeid(float))
```
paroj: you have this information already at compile time. use template specialization instead of RTTI.

vpisarev:
nice improvement! :+1:

PR #869, opened 2016-11-22
cnn_3dobj: add call to ocv_module_disable so documentation is generated

paroj:
also fix some minor doxygen issues


sovrasov:
:+1:

PR #883, opened 2016-12-02
Fix rounding in StructuredEdgeDetection

### LINKED ISSUE #881: Memory leak and bug in Structured Edge Detector
- OpenCV => 3.1

##### Detailed description

We found a bug in Structured Edge Detector which also creates a memory leak.
in structured_edge_detection.cpp line 246 the code is:
      `pHist[cvRound((j/pSize + pPhase[j])*nBins)] += pMagn[j] / CV_SQR(pSize);`

this creates an unintentional writing to memory outside the histogram Mat as well as introducing an error in the histogram itself.
when pPhase[j] > 0.875 the cvRound causes the pointer to point to the adjacent pixel.

our solution was to change the code:
      `pHist[cvRound((j / pSize + pPhase[j] * 0.9999 - 0.124999)*nBins)] += pMagn[j] / CV_SQR(pSize);`

##### Steps to reproduce

reproducing the bug is straight forward. 
to reproduce the memory leak, one must use an image where the value of one of the four lower right pixels of the phase Mat has a value > 0.875.

sovrasov:
Resolves #881

### DIFF
--- a/modules/ximgproc/src/structured_edge_detection.cpp
+++ b/modules/ximgproc/src/structured_edge_detection.cpp
@@ -235,6 +235,7 @@ static void gradientHist(const cv::Mat &src, cv::Mat &magnitude, cv::Mat &histog
     magnitude /= imsmooth( magnitude, gnrmRad )
         + 0.01*cv::Mat::ones( magnitude.size(), magnitude.type() );
 
+    int pHistSize = histogram.cols*histogram.channels() - 1;
     for (int i = 0; i < phase.rows; ++i)
     {
         const float *pPhase = phase.ptr<float>(i);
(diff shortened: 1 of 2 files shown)


modules/ximgproc/src/structured_edge_detection.cpp
```diff
@@ -243,7 +245,7 @@ static void gradientHist(const cv::Mat &src, cv::Mat &magnitude, cv::Mat &histog
         float *pHist = histogram.ptr<float>(i/pSize);
 
         for (int j = 0; j < phase.cols; ++j)
-            pHist[cvRound((j/pSize + pPhase[j])*nBins)] += pMagn[j] / CV_SQR(pSize);
+            pHist[cvRound((j/pSize + pPhase[j]*phaseRoundingCoeff - phaseRoundingShift)*nBins)] += pMagn[j] / CV_SQR(pSize);
```
alalek: We can use something like this in case of occasional out of range results in floating point calculations:
```
size_t index  = cvRound(... accurate formula without magic coefficients ...)
index = max(0, min(index, <sizeof pHist> - 1));
pHist[index] = ...
```

alalek:
:+1:

PR #884, opened 2016-12-03
Fuzzy module optimization

tucna:
**Merge with extra:**: https://github.com/opencv/opencv_extra/pull/318

Update of the **fuzzy** module.

- Python support added.
- Everything changed to **InputArray/OutputArray**.
- Tests updated and fixed.

Related **opencv_extra** commit exists.


alalek:
Please remove unnecessary merge commits.

tucna:
@alalek Sure, I wanted to check the build first. Thank you.

tucna:
@alalek I have read your suggestions in: https://github.com/opencv/opencv_extra/pull/318

Anyway, I have used **SourceTree** GUI for commit of this PR. I have tried to squash commits using `rebase` but it still does not work properly (because of _merge commits_ I guess).

Can I kindly ask you for a help with squashing of this PR as well? I can not find any guide how to do this in this particular case.

alalek:
Merge commits were removed from patch.

Also I added whitespace cleanup commit - please review and squash it into your one commit.

tucna:
Done, thank you @alalek. For the reference, I have used

```
git pull --rebase
git rebase -i upstream/master
git push origin fuzzy_optimization --force
```

The _whitespace cleanup_ you did - you have some script for this?

catree:
For whitespace commit, you can look here: [Q3. I was asked to remove whitespace issues, how can I do that?](https://github.com/opencv/opencv/wiki/How_to_contribute).

You can also configure your editor to remove trailing whitespace when saving.

tucna:
Thank you @catree I have seen this link before, but I do not have any pre-commit hooks. You know, I am asking because build machine told me I have some trailing whitespace issues. So I have fixed it, build was clear, but on top of that @alalek commited his _whitespace cleanup commit_.

Well, I thought maybe I did it wrong and he used some special script to do something (like replace tabs by spaces).

alalek:
Builder doesn't check for `tabs` in indentation, so I have [changed](https://github.com/alalek/opencv_contrib/commit/31e7f98f44f12e3dbb6b36f8c12723747b66efae) them to 4 spaces to make used code style consistent.

tucna:
Yes, I understood. My question was exactly about this - you have some script for that? Or you just use git hook to fix it how it is written in the link above?

I would like to know for future reference, to make in right next time :)

alalek:
"git hook" just warns you about whitespace issues.
In this case I just use [sed](http://stackoverflow.com/questions/11094383/how-can-i-convert-tabs-to-spaces-in-every-file-of-a-directory) for module source files(!don't use this for image and other binary files) to replace `\t` to `    ` (4 spaces).

alalek:
BTW, There is failed test in Debug build (on Linux).

tucna:
@alalek Thank you, great! That is exactly it.

I guess that the tests fail because of the missing pictures included in https://github.com/opencv/opencv_extra/pull/318. Can you merge the data if it will help? Everything is fine on my machine.

alalek:
Mentioned opencv_extra PR is [used](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_linux64_no_opt/builds/62/steps/Merge%20opencv_extra%20with%20test%20branch/logs/stdio) during patch testing, so all pictures are included. Problem is somewhere else...

tucna:
It is suspicious. Success on Windows, fail on Linux. Moreover exactly these two tests which use the new images is failing (but only on Linux).

tucna:
Ok, now I do not know what to do. I have updated the tests (according to another ones in **photo** module) but linux build is `not_queued` so I do not know if it helped.

Well, should I squash the commit and let it be?

alalek:
Linux debug build doesn't start automatically. I have scheduled it.

> should I squash the commit

Yes, It is better to keep 1 commit in GitHub PR.

tucna:
Same errors as before.  Are there any known differences between tests implementations for Windows and Linux, to care about?

alalek:
There is no precommit Windows debug builder.
BTW, What system do your have? Windows/Linux?

It is possible that `floating-point` computation results are different between debug and release builds due aggressive compiler optimization. Computation results are usually slightly different, but it is enough to provide completelly different total results after calls like cvRound (for example for 1.4999999 and 1.5) or `if` condition checks.
There is the same problem between platforms too (x86 -> arm).
How did your get reference image?

tucna:
Different precision came to my mind as well. Is there any bullet proof way how to compare if output and expected image is the same then?

I have Windows (VS 2015). I just tried several combinations **Debug/Release x86/x64** and tests passed all the time. I have obtained the reference images by calling `imwrite` just after the each tested operation in the test methods.

alalek:
I will try to extract results from Linux debug build tomorrow.

tucna:
Thank you. I have a feeling that maybe `imread/imwrite` works differently in **Debug/Release** or between compilers. Because when I use pure `Mat` manually filled, tests passed without a problem (basically all other tests in my fuzzy set).

alalek:
I have extracted results via this code:
```
    double n1 = cvtest::norm(exp1, res1, NORM_INF);
    double n2 = cvtest::norm(exp2, res2, NORM_INF);
    double n3 = cvtest::norm(exp3, res3, NORM_INF);

#if 1
    imwrite("res1.png", res1);
    imwrite("res2.png", res2);
    imwrite("res3.png", res3);

    Mat diff1; cv::absdiff(exp1, res1, diff1); imwrite("diff1.png", diff1*255);
    Mat diff2; cv::absdiff(exp2, res2, diff2); imwrite("diff2.png", diff2*255);
    Mat diff3; cv::absdiff(exp3, res3, diff3); imwrite("diff3.png", diff3*255);
#endif

    EXPECT_LE(n1, 1);
    EXPECT_LE(n2, 1);
    EXPECT_LE(n3, 1);
```
Diff images (multiplied by 255):
diff1
diff2
diff3
diff2 is just zero.

Results are different by 1, which is result of floating-point precision (at least you have `convertTo(8UC3)` call, where 1.4999999 is converted to 1 and 1.5 -> 2).
So my suggestion is to change checks and allow small tolerance (see code above).

You can try to reproduce problem with these additional CMake flags:
```
-DCMAKE_BUILD_TYPE=Debug -DENABLE_SSE3=OFF -DENABLE_SSE2=OFF -DWITH_IPP=OFF
```

Other failed test probably has similar issue.

tucna:
Thank you for great analysis! With your help I was able to reproduce the behaviour in my environment.

I have updated the tests. Can I kindly ask you for Linux build schedule?

tucna:
Great, finally green. I will follow this patern of tests for future PRs.

alalek:
:+1:

PR #887, opened 2016-12-05
Added a forced include of limits to fix compilation on ubuntu 16.04.

hibes:
### This pullrequest adds

A forced include on the <limits> dependency so that it will compile on ubuntu 16.04.  This probably fixes builds on other platforms as well.


alalek:
`<limits>` are already [included](https://github.com/opencv/opencv/blob/37cbcf024c2b4160648299887a7c28d7cf28b1d3/modules/core/include/opencv2/core/types.hpp#L54) from OpenCV basic header file (master branch, after OpenCV 3.1 release).

hibes:
Oh, gotcha, that would fix it upstream, then.  Will close the merge request.

Thanks for looking into it for me!

PR #888, opened 2016-12-07
Add default value for LUCID feature.

catree:
### This pullrequest adds

the default values for LUCID feature and also a note about a color image required as an input.


StevenPuttemans:
Could you add an extra check/assert [here](https://github.com/opencv/opencv_contrib/blob/master/modules/xfeatures2d/src/lucid.cpp#L103) that explicitly checks if the input image is of type CV_8UC3, before continuing?

catree:
I added: `CV_Assert(_src.depth() == CV_8U && _src.channels() == 3);`.

sovrasov:
:+1:

PR #891, opened 2016-12-09
Fast line detector

spacetrain:
A fast line detector. This detector runs up to two times faster than LSD while achieving competitive results.
Please evaluate it with ximgproc/samples/fld.lines.cpp.
A test code is also included to ximgproc/text/text_fld.cpp.


modules/ximgproc/include/opencv2/ximgproc/fast_line_detector.hpp
```diff
@@ -0,0 +1,109 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ * Copyright (c) 2014, 2015
+ * Zhengqin Li <li-zq12 at mails dot tsinghua dot edu dot cn>
+ * Jiansheng Chen <jschenthu at mail dot tsinghua dot edu dot cn>
+ * Tsinghua University
```
alalek: Is this copyright right?

modules/ximgproc/include/opencv2/ximgproc/fast_line_detector.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+/** @example fld_lines.cpp
+	An example using the FastLineDetector
+*/
+
+class CV_EXPORTS_W FastLineDetector : public Algorithm
+{
+		public:
+
+		/** @brief Finds lines in the input image.
```
alalek: Please use 4 spaces instead of `tabs`

modules/ximgproc/src/fld.cpp
```diff
@@ -0,0 +1,794 @@
...
+//  By downloading, copying, installing or using the software you agree to this license.
+//  If you do not agree to this license, do not download, install,
+//  copy or use the software.
+//
+//
+//                           License Agreement
+//                For Open Source Computer Vision Library
+//
+// Copyright (C) 2013, OpenCV Foundation, all rights reserved.
+// Third party copyrights are property of their respective owners.
```
alalek: You may use short version of license header instead:
```
```

modules/ximgproc/include/opencv2/ximgproc/fast_line_detector.hpp
spacetrain: Oops... Because I wrote the code based on LSD,, I made such a mistake. I will change it right away.

modules/ximgproc/src/fld.cpp
```diff
@@ -0,0 +1,794 @@
...
+
+void FastLineDetectorImpl::lineDetection(const Mat& src, std::vector<SEGMENT>& segments_all)
+{
+    int r, c;
+    imageheight=src.rows; imagewidth=src.cols;
+
+    std::vector<Point2i> points;
+    std::vector<SEGMENT> segments, segments_tmp;
+    Mat canny;
+    Canny(src, canny, 50, 50, 3);
```
alalek: Canny result depends on image quality (contrast). There is no universal parameters to get good edge maps in all cases.

Probably this method may accept "edge" map directly instead of original image.
Or Canny parameters may be exposed via public interface.

modules/ximgproc/src/fld.cpp
```diff
@@ -0,0 +1,794 @@
...
+    else if(dx > 0.0f && dy < 0.0f)
+        ang = 2*CV_PI + atan( dy/dx );
+    else if(dx < 0.0f && dy < 0.0f)
+        ang = CV_PI + atan( dy/dx );
+    else
+        ang = atan( dy/dx );
+
+    if(ang > 2.0 * CV_PI)
+        ang -= 2.0 * CV_PI;
+    seg.angle = (float)ang;
```
alalek: Probably you need "atan2" / "cv::fastAtan2" function here.

doc/opencv.bib
```diff
@@ -0,0 +1,907 @@
+@comment{Bib-it,
```
alalek: You should use module file instead: https://github.com/opencv/opencv_contrib/blob/master/modules/ximgproc/doc/ximgproc.bib

modules/ximgproc/include/opencv2/ximgproc/fast_line_detector.hpp
```diff
@@ -0,0 +1,73 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#ifndef __OPENCV_FAST_LINE_DETECTOR_HPP__
+#define __OPENCV_FAST_LINE_DETECTOR_HPP__
+#ifdef __cplusplus
```
alalek: `__cplusplus` check is not need for "hpp" header files.

doc/opencv.bib
spacetrain: Changed as you pointed out

modules/ximgproc/include/opencv2/ximgproc/fast_line_detector.hpp
spacetrain: Removed the line

modules/ximgproc/src/fld.cpp
spacetrain: Okay. I will modify the code to have public interface for Canny parameters.

modules/ximgproc/src/fld.cpp
spacetrain: Thanks for smart suggestion. I modified the lines to use cv::fastAtan2. =)

PR #905, opened 2016-12-17
Bioinspired module: fix some typos, add verbose option.

catree:
### This pullrequest adds

verbose option to avoid unnecessary prints. Some typos are also fixed.

@amroamroamro Can you test to see if the modifications did not break anything and solve your issue?

Related: [unnecessary print statements #7781](https://github.com/opencv/opencv/issues/7781)

### DIFF
--- a/modules/bioinspired/include/opencv2/bioinspired/retina.hpp
+++ b/modules/bioinspired/include/opencv2/bioinspired/retina.hpp
@@ -226,8 +226,9 @@ public:
     - warning, Exceptions are thrown if read XML file is not valid
     @param retinaParameterFile the parameters filename
     @param applyDefaultSetupOnFailure set to true if an error must be thrown on error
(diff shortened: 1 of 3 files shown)


vpisarev:
@catree, thanks for the patch! I wonder if we can avoid adding verbose, and instead call printSetup() from the application after setup() is called?

catree:
@vpisarev Indeed, your solution is better. I will revert the verbose changes and add a note in the documentation.

amroamroamro:
@catree thanks for submitting a fix.
As vpisarev suggested, I don't think there is a need for "verbose" flag in the setup() functions. The API already provides public methods for that, so users are free to print the params themselves if they wish..

catree:
I have reverted my changes and removed the printing. I have fixed some typos also.

vpisarev:
thanks! :+1:

PR #906, opened 2016-12-18
build(freetype): move 3rdparty include files from precomp.hpp

alalek:
resolves https://github.com/opencv/opencv/issues/7886


alalek:
:+1:

PR #910, opened 2016-12-21
Python wrapper for both RGBD and Surface Matching modules

Sahloul:
### This pullrequest changes

- Adds a Python wrapper for both RGBD and Surface Matching modules
  - Replaces `(*T)(&mat.data[i * mat.step])` into `mat.ptr<T>(i)` for wrappers compatibility.
  - Replaces two-words types with one-word (e.g. `unsigned int` into `uint`) for wrappers compatibility.
  - Removes some `enum` names for wrappers compatibility.
  - Replaces some `Mat` types with appropriate `InputArray ` or `OutputArray` types.
  - Converts some pointers into vectors.
  - Adds `const` modifier where necessary.
- Provides sample python scripts for testing the wrappers.
- Supports OpenNI2 in the original `pyopencv_rgbd` sample code.

### DIFF
--- a/modules/rgbd/include/opencv2/rgbd.hpp
+++ b/modules/rgbd/include/opencv2/rgbd.hpp
@@ -109,7 +109,7 @@ namespace rgbd
   public:
     CV_WRAP enum
     {
-      RGBD_NORMALS_METHOD_FALS = 1, RGBD_NORMALS_METHOD_LINEMOD = 2, RGBD_NORMALS_METHOD_SRI = 3,
+      RGBD_NORMALS_METHOD_FALS = 0, RGBD_NORMALS_METHOD_LINEMOD = 1, RGBD_NORMALS_METHOD_SRI = 2,
     };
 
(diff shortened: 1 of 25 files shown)


vpisarev:
@Sahloul, thank you! Can you provide any sample or test to check that the wrappers work?

Sahloul:
@vpisarev I have provided two samples (ppf_load_match.py and ppf_normal_computation.py) for the surface matching modules.
As you can see, I really needed to compact the original implementation to enable the wrapper functionality. And since the RGBD module might need much more efforts in its code compaction, I would prefer to dedicate a new PR for it once I have time.

Sahloul:
@vpisarev I have provided an example for RGBD module as well.

Sahloul:
@vpisarev Mind to advice why is the delay with this pull request?

Sahloul:
Maybe it is too huge! I will close this one push separated ones in my free time.

PR #913, opened 2016-12-22
moved BLAS/LAPACK detection scripts to the main repository. Try to use external gemm in dnn

vpisarev:
It should improve performance of dnn module by a lot, even comparing to dnn's own parallel gemm implementation. Merge after https://github.com/opencv/opencv/pull/7918


vpisarev:
:+1:

PR #915, opened 2016-12-22
Update sift.cpp

sturkmen72:
i think no need to the line
`img.copyTo(gray);`


terfendail:
@sturkmen72 
Could you please squash commits into one?

terfendail:
👍

PR #916, opened 2016-12-23
Added Torch ENet support

arrybn:
### This pullrequest changes

Added loading Torch ENet model: changed Torch models reader and added all necessary layers

### DIFF
--- a/modules/dnn/samples/torch_enet.cpp
+++ b/modules/dnn/samples/torch_enet.cpp
@@ -1,0 +1,176 @@
+/*
+Sample of using OpenCV dnn module with Torch ENet model.
+*/
+
+#include <opencv2/dnn.hpp>
+#include <opencv2/imgproc.hpp>
+#include <opencv2/highgui.hpp>
+using namespace cv;
+using namespace cv::dnn;
+
+#include <fstream>
+#include <iostream>
+#include <cstdlib>
+#include <sstream>
+using namespace std;
+
+const String keys =
+        "{help h    || Sample app for loading ENet Torch model. "
+                       "The model and class names list can be downloaded here: "
+                       "https://www.dropbox.com/sh/dywzk3gyb12hpe5/AAD5YkUa8XgMpHs2gCRgmCVCa }"
+        "{model m   || path to Torch .net model file (model_best.net) }"
+        "{image i   || path to image file }"
+        "{i_blob    | .0 | input blob name) }"
+        "{o_blob    || output blob name) }"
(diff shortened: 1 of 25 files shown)


modules/dnn/samples/torch_enet.cpp
```diff
@@ -0,0 +1,185 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
```
alalek: Usually we don't put license header into sample files.

modules/dnn/samples/torch_enet.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+    if (inputImgSize != img.size())
+        resize(img, img, inputImgSize);       //Resize image to input size
+
+    if(img.channels() == 3)
+        cv::cvtColor(img, img, cv::COLOR_BGR2RGB);
+
+    img.convertTo(img, CV_32F);
+
+    img /= 255;
```
alalek: Merge into single line:
```
img.convertTo(img, CV_32F, 1/255.0);
```

modules/dnn/src/layers/batch_norm_layer.hpp
```diff
@@ -0,0 +1,38 @@
...
+// Copyright (C) 2016, Intel Corporation, all rights reserved.
+// Third party copyrights are property of their respective owners.
+
+/*
+Declaration of Batch Normalization layer.
+*/
+
+#ifndef __OPENCV_DNN_LAYERS_BATCH_NORM_LAYER_HPP__
+#define __OPENCV_DNN_LAYERS_BATCH_NORM_LAYER_HPP__
+#include "../precomp.hpp"
```
alalek: `precomp.hpp` should be included from `.cpp` files only.

modules/dnn/samples/torch_enet.cpp
arrybn: Removed, thanks

modules/dnn/samples/torch_enet.cpp
arrybn: fixed, thanks

modules/dnn/src/layers/batch_norm_layer.hpp
arrybn: Removed, thanks

vpisarev:
cool! let's put it in! :+1:

PR #918, opened 2016-12-25
Improve paillou filter

LaurentBerger:
### This pullrequest changes

Improve paillou filter
add license
InputArray OutputArray use as parameter
template for IIR filter
use Ptr<Mat> for example


alalek:
:+1:

PR #921, opened 2016-12-27
Bug in Text module when set Optimization Level to "-Ofast" in xcode

logic1988:
Bug when set Optimization Level to "-Ofast" (Fastest, Aggressive Optimizations) in xcode, read the wrong patch_size parameters to 7 (should be 8). 
Take the following code for example:
`patch_size  = (int)sqrt((float)kernels.cols);`
if `kernels.cols == 64.0`,  `patch_size` may assigned by `(int)sqrt(64.0f) = (int)(7.99999999) = 7`, It was dangerous and crashed in "-Ofast" setting. Therefore, the code needs to be changed to:
`patch_size  = (int)round(sqrt((float)kernels.cols));`

This bug will occured in `loadOCRHMMClassifierCNN` function called by `example_text_character_recognition` and `example_text_segmented_word_recognition`.

I also optimized a little bit of code by the way.

In addition, I also found a potential risk, but did **not** modify them, I hope the authors can improve them.
In `example_text_cropped_word_recognition`, the result is also different depending on the degree of optimization. The OCR output was empty when  in "-Ofast" setting. It caused by: 
In [ocr_beamsearch_decoder.cpp#L413](https://github.com/opencv/opencv_contrib/blob/246ea8f3bdf174a2aad6216c2601e3a93bf75c29/modules/text/src/ocr_beamsearch_decoder.cpp#L413), The value `step_size` is always **zero** because of **getStepSize() function always return 0** [ocr_beamsearch_decoder.cpp#L170](https://github.com/opencv/opencv_contrib/blob/246ea8f3bdf174a2aad6216c2601e3a93bf75c29/modules/text/src/ocr_beamsearch_decoder.cpp#L170) and [ocr.hpp#L368](https://github.com/opencv/opencv_contrib/blob/246ea8f3bdf174a2aad6216c2601e3a93bf75c29/modules/text/include/opencv2/text/ocr.hpp#L368), so `interdist.at<float>((int)i,0)` was always 0, `if ((float)interdist.at<float>((int)i,0)/win_size > 2.25)` and `if ((float)interdist.at<float>((int)i,0)/win_size < 0.15)` was equal to if `Nan > 2.25`  or if `Nan < 0.15`.
While is not "-Ofast" setting, the above judgment always returns false. But in "-Ofast" setting, it may return true, this causes the function to return prematurely, resulting in incorrect results.
The division-by-zero operation and the nan value are both dangerous and can lead to undesirable results.

### DIFF
--- a/modules/text/src/ocr_beamsearch_decoder.cpp
+++ b/modules/text/src/ocr_beamsearch_decoder.cpp
@@ -566,7 +566,7 @@ OCRBeamSearchClassifierCNN::OCRBeamSearchClassifierCNN (const string& filename)
 
     nr_feature = weights.rows;
     nr_class   = weights.cols;
-    patch_size  = (int)sqrt((float)kernels.cols);
+    patch_size  = (int)round(sqrt((float)kernels.cols));
     window_size = 4*patch_size;
     step_size   = 4;
     quad_size   = 12;
@@ -602,8 +602,12 @@ void OCRBeamSearchClassifierCNN::eval( InputArray _src, vector< vector<double> >
     Mat tmp;
     Mat img;
 
+    int sz = src.cols - window_size;
+    int sz_window_quad = window_size - quad_size;
+    int sz_half_quad = (int)(quad_size/2-1);
+    int sz_quad_patch = quad_size - patch_size;
     // begin sliding window loop foreach detection window
-    for (int x_c=0; x_c<=src.cols-window_size; x_c=x_c+step_size)
+    for (int x_c = 0; x_c <= sz; x_c += step_size)
     {
 
         img = src(Rect(Point(x_c,0),Size(window_size,window_size)));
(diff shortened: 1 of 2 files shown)


modules/text/src/ocr_beamsearch_decoder.cpp
```diff
@@ -566,7 +566,7 @@ OCRBeamSearchClassifierCNN::OCRBeamSearchClassifierCNN (const string& filename)
 
     nr_feature = weights.rows;
     nr_class   = weights.cols;
-    patch_size  = (int)sqrt((float)kernels.cols);
+    patch_size  = (int)round(sqrt((float)kernels.cols));
```
alalek: `cvRound()` is preferred for `float->int` conversion. `round()` is `float->float`.

alalek:
:+1:

PR #922, opened 2016-12-27
BugFix SLIC, SLICO & LSC.

cbalint13:
### BugFix SLIC, SLICO & LSC

  * This PR fix an initialization bug for the parallel_reduce() structure. The structure won't get re-updated inside loops with the newly computed arrays and the quality of results may be degraded.


alalek:
:+1:

PR #924, opened 2016-12-29
Overhaul bioinspired opencl implementation

pengli:
This patch update bioinspired opencl code to make it working
with latest opencv, major changes include:

1. data structure update, e.g. oclMat to UMat.
2. opencl kernel optimization to get it running faster
3. accuracy and perf test update.

Signed-off-by: Li Peng <peng.li@intel.com>


alalek:
@pengli Thank you!
There are some [build warnings](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_opencl/builds/23). Could you please take a look on them?

pengli:
@alalek，new updates are pushed，should fix the build warnings，pls check

modules/bioinspired/src/retina_ocl.cpp
```diff
@@ -472,6 +439,13 @@ void RetinaOCLImpl::activateContoursProcessing(const bool activate)
     _retinaFilter->activateContoursProcessing(activate);
 }
 
+// unimplemented interfaces:
+void RetinaOCLImpl::applyFastToneMapping(InputArray /*inputImage*/, OutputArray /*outputToneMappedImage*/) { NOT_IMPLEMENTED; }
+void RetinaOCLImpl::getParvoRAW(OutputArray /*retinaOutput_parvo*/) { NOT_IMPLEMENTED; }
+void RetinaOCLImpl::getMagnoRAW(OutputArray /*retinaOutput_magno*/) { NOT_IMPLEMENTED; }
+const Mat RetinaOCLImpl::getMagnoRAW() const { NOT_IMPLEMENTED; return Mat(); }
+const Mat RetinaOCLImpl::getParvoRAW() const { NOT_IMPLEMENTED; return Mat(); }
```
alalek: Transparent API must avoid these failures and automatically drop to CPU implementation.

modules/bioinspired/src/retina_ocl.cpp
```diff
@@ -48,70 +48,37 @@
...
+    inline void ensureSizeIsEnough(int rows, int cols, int type, cv::UMat &m)
+    {
+        m.create(rows, cols, type, m.usageFlags);
+    }
+}
+
 namespace cv
 {
 static ocl::ProgramEntry retina_kernel = ocl::bioinspired::retina_kernel;
+static ocl::ProgramSource retina_kernel_source(retina_kernel.programStr);
```
alalek: These 2 lines should be defined in `opencl_kernels_bioinspired.hpp` and `opencl_kernels_bioinspired.cpp` (located in `<build_dir>/modules/bioinspired`). No need to re-declare them.

modules/bioinspired/src/retina.cpp
```diff
@@ -295,8 +296,21 @@ class RetinaImpl : public Retina
...
 // smart pointers allocation :
-Ptr<Retina> createRetina(Size inputSize){ return makePtr<RetinaImpl>(inputSize); }
-Ptr<Retina> createRetina(Size inputSize, const bool colorMode, int colorSamplingMethod, const bool useRetinaLogSampling, const float reductionFactor, const float samplingStrenght){
+Ptr<Retina> createRetina(Size inputSize)
+{
+#ifdef HAVE_OPENCL
+    if (cv::ocl::useOpenCL() && (inputSize.width % 4 == 0))
+        return makePtr<cv::bioinspired::ocl::RetinaOCLImpl>(inputSize);
+#endif
+    return makePtr<RetinaImpl>(inputSize);
```
alalek: I believe there are should be one implementation that is able to handle OCL and CPU code paths (via `OCL_RUN` hooks).
But there is main problem with data storage: Mat <-> UMat conversion is not cheap now for CPU data storage (there is forced OpenCL buffer creation).

modules/bioinspired/src/retina.cpp
```diff
@@ -649,7 +663,7 @@ void RetinaImpl::_convertValarrayBuffer2cvMat(const std::valarray<float> &grayMa
             for (unsigned int j=0;j<nbColumns;++j)
             {
                 cv::Point2d pixel(j,i);
-                outMat.at<unsigned char>(pixel)=(unsigned char)*(valarrayPTR++);
+                outMat.at<unsigned char>(pixel)=(unsigned char)round(*(valarrayPTR++));
```
alalek: What kind of problem is fixed here (and below)?

BTW, it is better to use `cvRound()` instead. Also saturation may be added (if necessary).

modules/bioinspired/test/test_retina_ocl.cpp
```diff
@@ -134,31 +94,35 @@ TEST_P(Retina_OCL, Accuracy)
 
     Mat gold_parvo;
     Mat gold_magno;
-    ocl::oclMat ocl_parvo;
-    ocl::oclMat ocl_magno;
+    UMat ocl_parvo;
+    UMat ocl_magno;
 
-    for(int i = 0; i < RETINA_ITERATIONS; i ++)
+    for(int i = 0; i < test_loop_times; i ++)
```
alalek: `test_loop_times` is not suitable here (it is used as counter of random image generation - sizes/ROI)

modules/bioinspired/test/test_retina_ocl.cpp
```diff
@@ -82,48 +66,24 @@ PARAM_TEST_CASE(Retina_OCL, bool, int, bool, double, double)
...
         colorMode,
         colorSamplingMethod,
         useLogSampling,
         reductionFactor,
         samplingStrength);
 
+    OCL_OFF();
     Ptr<bioinspired::Retina> gold_retina = bioinspired::createRetina(
         input.size(),
         colorMode,
```
alalek: Only one creation scheme should be used and technically `ocl_retina` and `gold_retina` should be the same object or  similarly constructed (preferred). Please remove `OCL_ON`/`OCL_OFF`.

modules/bioinspired/src/retina_ocl.cpp
pengli: with new update, unimplemented API will fallback to CPU implementation. so application won't see these failures

modules/bioinspired/src/retina_ocl.cpp
pengli: fixed in new update

modules/bioinspired/perf/opencl/perf_retina.ocl.cpp
```diff
@@ -0,0 +1,56 @@
...
+    float samplingStrength = static_cast<float>(get<3>(params));
+
+    Mat input = imread(getDataPath("cv/shared/lena.png"), colorMode);
+    ASSERT_FALSE(input.empty());
+
+    UMat ocl_parvo, ocl_magno;
+
+    {
+        OCL_ON();
+        Ptr<cv::bioinspired::Retina> ocl_retina = cv::bioinspired::createRetina(
```
pengli: in new update, only one retina ptr is created, ocl_retina is removed.

modules/bioinspired/perf/opencl/perf_retina.ocl.cpp
```diff
@@ -0,0 +1,56 @@
...
+    float reductionFactor = static_cast<float>(get<2>(params));
+    float samplingStrength = static_cast<float>(get<3>(params));
+
+    Mat input = imread(getDataPath("cv/shared/lena.png"), colorMode);
+    ASSERT_FALSE(input.empty());
+
+    UMat ocl_parvo, ocl_magno;
+
+    {
+        OCL_ON();
```
pengli: fixed

modules/bioinspired/perf/opencl/perf_retina.ocl.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+typedef tuple<bool, int, double, double> RetinaParams;
+typedef TestBaseWithParam<RetinaParams> RetinaFixture;
+
+OCL_PERF_TEST_P(RetinaFixture, Retina,
+                ::testing::Combine(testing::Bool(), testing::Values((int)cv::bioinspired::RETINA_COLOR_BAYER),
+                                   testing::Values(1.0, 0.5), testing::Values(10.0, 5.0)))
+{
+    if (!cv::ocl::haveOpenCL())
+        throw TestBase::PerfSkipTestException();
```
pengli: fixed

modules/bioinspired/perf/opencl/perf_retina.ocl.cpp
```diff
@@ -0,0 +1,56 @@
+#include "../perf_precomp.hpp"
+#include "opencv2/ts/ocl_perf.hpp"
+
+#ifdef HAVE_OPENCL
```
pengli: This comment is addressed in the new update.

modules/bioinspired/src/retina.cpp
pengli: OCL_RUN hooks are used.

modules/bioinspired/src/retina.cpp
pengli: This is to fix accuracy test, because in cpu implementation, final pixel is converted from float to int with truncating conversion,  but in ocl implementation, float to int conversion is rounding conversion.  this change is to make sure we use same conversion method

modules/bioinspired/test/test_retina_ocl.cpp
pengli: only one retina ptr is created. OCL_ON and OCL_OFF is used as same as other accuracy test

modules/bioinspired/test/test_retina_ocl.cpp
pengli: fixed in the new update.

pengli:
@alalek , new update is pushed to address your comments, pls check and give your feedback. thanks

modules/bioinspired/src/retina.cpp
```diff
@@ -729,7 +814,20 @@ bool RetinaImpl::_convertCvMat2ValarrayBuffer(InputArray inputMat, std::valarray
...
+    _ocl_retina->clearBuffers();
+    return true;
+}
+#endif
+
+void RetinaImpl::clearBuffers()
+{
+    CV_OCL_RUN((_ocl_retina != 0), ocl_clearBuffers());
+
+    _retinaFilter->clearAllBuffers();
```
pengli: both patches are applied, new update is pushed. Thanks for your review.

alalek:
Thank you! :+1:

Issue #584, opened 2016-03-16
SURF_CUDA produces different number of keypoints on the same image


drtimes:
For some images at each run SURF_CUDA() produces different number of detected keypoints on the same image, no matter if it is the same instance or another instance. I'm using OpenCV 3.1.0, CUDA 7.5

![seaman_surf-cuda.jpg](https://cloud.githubusercontent.com/assets/13232153/13806850/c44caa56-eb66-11e5-9023-3e018b0651de.jpg)

Sample image: [seaman.jpg](https://cloud.githubusercontent.com/assets/13232153/13806851/c44ce9ee-eb66-11e5-906b-ef91a08e44fd.jpg)

```
#include <iostream>
#include "opencv2/core/core.hpp"
#include "opencv2/features2d/features2d.hpp"
#include "opencv2/highgui/highgui.hpp"
#include "opencv2/cudafeatures2d.hpp"
#include "opencv2/xfeatures2d/cuda.hpp"
#include "opencv2/opencv_modules.hpp"

using namespace std;
using namespace cv;
using namespace cv::cuda;

int main(int argc, char *argv[])
{
    setDevice(0);

    GpuMat img1, img2;
    img1.upload(imread("seaman.jpg", IMREAD_GRAYSCALE));
    img2.upload(imread("seaman.jpg", IMREAD_GRAYSCALE));

    SURF_CUDA surf;
    GpuMat keypoints1GPU, keypoints2GPU;
    GpuMat descriptors1GPU, descriptors2GPU;

    surf(img1, GpuMat(), keypoints1GPU, descriptors1GPU);
    surf(img2, GpuMat(), keypoints2GPU, descriptors2GPU);

    cout << "FOUND " << keypoints1GPU.cols << " keypoints on first image" << endl;
    cout << "FOUND " << keypoints2GPU.cols << " keypoints on second image" << endl;

    Ptr<DescriptorMatcher> matcher = DescriptorMatcher::createBFMatcher(surf.defaultNorm());
    vector<DMatch> matches;
    matcher->match(descriptors1GPU, descriptors2GPU, matches);

    vector<KeyPoint> keypoints1, keypoints2;
    vector<float> descriptors1, descriptors2;
    surf.downloadKeypoints(keypoints1GPU, keypoints1);
    surf.downloadKeypoints(keypoints2GPU, keypoints2);
    surf.downloadDescriptors(descriptors1GPU, descriptors1);
[...]

PR #931, opened 2017-01-04
Fix SURF OCL

acyen:
Enable SURF OCL. Fix incorrect calling sequence for two kernels.
Fix compiler warnings in surf.cl.
Enable SURF OCL test. Test runs with no failures.


alalek:
Please take a look on existed OpenCV tests. They are failed on SURF checks.

acyen:
What is the right way to fix this test - DescriptorExtractor_SURF? The test is comparing the CPU descriptor calculation with OCL.  I fixed the OCL code to match CUDA, but the CPU code is different from both and gives a different result. Choices are:
1. Create a separate test for OCL and do not compare CPU with OCL.
2. Change CPU code to match OCL.
3. Change OCL code to match CPU.

catree:
My comments, as just an OpenCV user. The best would be in my opinion:

- for the descriptor part: have a numerically accurate code + without decrease in term of precision / recall

- match the CPU and GPU code if possible

Looks like it is not so simple (only bullet 3 should concern this PR):

- detection part of OCL SURF seems to be of less quality: [OpenCL SURF matching quality decreased #4685](https://github.com/opencv/opencv/issues/4685)

- Cuda SURF seems to suffer of the same issue (at least for the detection part): [SURF_CUDA produces different number of keypoints on the same image #584](https://github.com/opencv/opencv_contrib/issues/584)

- one user reports NaN values with Cuda SURF in some cases: [SURF_CUDA produces nan values](http://answers.opencv.org/question/130616/surf_cuda-produces-nan-values/)

- even if the OCL / Cuda descriptor part matches the CPU descriptor code, the Cuda detection part seems to be [not deterministic](https://github.com/opencv/opencv_contrib/issues/584). It is not necessary an issue though, as the GPU architecture is of course different of a CPU architecture.

To recap, as a user the best would be option 2 if there is no decrease in term of performance and quality (precision / recall) for CPU SURF.

PS: I found an old [SURF benchmark](http://boofcv.org/index.php?title=Performance:SURF).

acyen:
Changing CPU SURF would break tests unless the test data were changed at the same time.  Is there a procedure to do this?

catree:
I think that you will have to submit a PR with the new test data [here](https://github.com/opencv/opencv_extra).

acyen:
The testdata PR has to be coordinated with this one, otherwise tests will break.  How does that work?

alalek:
You need to use the same branch name for patches in both forks (opencv_contrib/opencv_extra).
But this doesn't work for "master" source branch(see above: "from acyen:master"). You need to recreate PR with another branch name (for example, surf_ocl).

Issue #851, opened 2016-11-12
Add Harris-Laplace / Hessian-Laplace detectors?


Yorwba:
I need the Harris-Laplace / Hessian-Laplace affine feature detectors for a project and found [an old set of patches](http://code.opencv.org/issues/1053) implementing the Harris-Laplace detector, that was apparently never merged.
I plan to massage the patches into shape for use with current versions of OpenCV but I'm unsure whether I should aim for inclusion in (presumably the xfeatures2d module of) OpenCV or whether I should choose a more standalone approach.
Since I do not know why the patches weren't merged when they were submitted, I would like to know whether you would accept the contribution if I made a PR, or whether these feature detectors won't be included in OpenCV for some reason.

StevenPuttemans:
I do not think that if you get a stable working release, that inclusion is not possible. Like suggested, go for inclusion inside xfeatures2d, which is best approach for new feature detectors.

Good luck with the PR!

alalek:
Thank you! #933

PR #933, opened 2017-01-04
Add Harris-Laplace affine detector

Yorwba:
### This pullrequest changes

I took the old patches found at http://code.opencv.org/issues/1053 and rewrote the interface to (hopefully) conform to the OpenCV coding style. While doing this I noticed that the code for affine adaptation was actually mostly independent from the key-point detector, so I pulled it out into a separate class AffineFeature2D that can wrap any feature detector and descriptor.

What I did not get to work were the Python/Java wrappers, simply putting CV_EXPORTS_W on the functions lead to issues with the generated code.


Yorwba:
The failed checks taught me a lesson about the difficulty of cross-platform C++ development. Hopefully I didn't miss anything.

As a note: the warning about overloaded virtuals that clang helpfully shows can also be enabled in GCC using -Woverloaded-virtual, but then I get a number of additional warnings across the OpenCV code base. I'm not sure whether that is indicative of low precision by GCC or whether those warnings ought to be fixed.

modules/xfeatures2d/src/affine_feature2d.cpp
```diff
@@ -0,0 +1,714 @@
+/*IMPORTANT: READ BEFORE DOWNLOADING, COPYING, INSTALLING OR USING.
+
+ By downloading, copying, installing or using the software you agree to this license.
+ If you do not agree to this license, do not download, install,
+ copy or use the software.
+
+
+ License Agreement
+ For Open Source Computer Vision Library
```
alalek: Could you please use "unmodified" license header or just use the shorter variant:
```
// of this distribution and at http://opencv.org/license.html
```

modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
```diff
@@ -834,6 +834,100 @@ class CV_EXPORTS_W PCTSignaturesSQFD : public Algorithm
...
+* @brief Elliptic region around an interest point.
+*/
+class CV_EXPORTS Elliptic_KeyPoint : public KeyPoint
+{
+public:
+    Point centre;
+    Size_<float> axes; //!< the lengths of the major and minor ellipse axes
+    double phi; //!< the rotation angle
+    float size;
+    float si;  //!< the integration scale at which the parameters were estimated
```
alalek: KeyPoint already have similar [fields](https://github.com/opencv/opencv/blob/3.2.0/modules/core/include/opencv2/core/types.hpp#L701-L708):
- centre -> pt
- phi -> angle
- size -> size
- si (float) -> octave (int)

modules/xfeatures2d/src/affine_feature2d.cpp
```diff
@@ -0,0 +1,714 @@
...
+        size(0, 0) = size(1, 0) = it->size;
+
+        //U matrix
+        Mat transf = it->transf;
+        Mat_<float> U(2, 2);
+        U.setTo(0);
+        Mat col0 = U.col(0);
+        transf.col(0).copyTo(col0);
+        Mat col1 = U.col(1);
+        transf.col(1).copyTo(col1);
```
alalek: For fixed size matrices it is better to use `Matx`: Matx22f and similar.

Yorwba:
@alalek, I hope my changes address your concerns.

However, the build bot seems to have hit a bug: It has been "building" for 3 hours now and the build IDs link to builds for an unrelated pull request.

alalek:
:+1:

modules/xfeatures2d/src/affine_feature2d.cpp
```diff
@@ -0,0 +1,681 @@
...
+            p = p + Matx22f(Matx22d(U).inv()) * c;
+            px = int(p(0, 0));
+            py = int(p(1, 0));
+
+            q = calcSecondMomentSqrt(Lxm2smooth, Lxmysmooth, Lym2smooth, Point(cx, cy), Mk);
+
+            float ratio = 1 - q;
+
+            //if ratio == 1 means q == 0 and one axes equals to 0
+            if (!isnan(ratio) && ratio != 1)
```
rversaw: Should this be using `std::isnan()` similar to [other places](https://github.com/opencv/opencv_contrib/blob/246ea8f3bdf174a2aad6216c2601e3a93bf75c29/modules/sfm/src/libmv_light/libmv/numeric/numeric.h#L461) in the codebase? I ran into an error locally until I switched to that, though I'm still hitting other issues with the `master` branch, so I can't confirm that it solved everything.

lovaj:
@Yorwba I'm a little bit confused about how to use this, could you please give an example code?

Yorwba:
@lovaj Say you have an `image` and want to extract affine covariant elliptic regions for keypoints detected by the Harris-Laplace detector. Then you would write
```c++
    Ptr<AffineFeature2D> detector = AffineFeature2D::create(
        HarrisLaplaceFeatureDetector::create(/* adjust parameters as necessary */),
        yourFeatureDescriptor
    );
    vector<Elliptic_KeyPoint> keypoints;
    Mat descriptors;
    detector->detectAndCompute(image, noArray() /* no mask */, keypoints, descriptors);
```
and then you can use the `keypoints` together with their `descriptors` for whatever it is you want to do.

oqtvs:
Is AffineFeature2D equivalent to Harris-affine(when using harris detector)?

oqtvs:
Ok only to make it clear, this PR added Harris-Laplace and Harris-Affine ( If i am correct harris-affine can be achieved by using Harris-Laplace detector with `AffineFeature2D`) by bringing back patches from Harris Laplace and Harris Affine code [(Patch #1053)](http://code.opencv.org/issues/1053). That said the issue #851 is closed but hessian-laplace/hessian-affine isn't yet implemented. So on #1116 maybe implementing only Hessian-laplace one can achieve Hessian-Affine only by using `AffineFeature2D` (must check if things match).

cyberbeat:
I would like to add support for elliptic keypoints to the mser detector. How would I do that? The problem is, that mser is in the features2d module, and elliptic keypoints in the xfeatures2d module.

PR #936, opened 2017-01-05
Support getTextSize() to resolve issue #927

### LINKED ISSUE #927: Freetype2: returns the maximum advanced point

- OpenCV => 3.2
- Operating System / Platform => Linux
- Compiler => ❔ 

##### Detailed description
Currently, the `putText` has no other output than the printed text. It could be a great feature if the `mOrg` point is returned, either in the input parameter or in the return value, since it will be useful to print more text, either on the same line or on another line.

Kumataro:
This getTextSize() funtion helps to print more text, either on the same line or on another line.
To add getTextSize() function, some member variables are removed which are corresponding at putText() function's arguments. And a negative fontHeight must be asserted.

resolves #927 

### This pullrequest changes

Add getTextSize() function at freetype2 module similar to imgproc module.
In order to avoid confusion, refactoring is now required for a member variable which are hold the argument of putText (). 
Also, CV_Assert is required for cases where fontHeight is negative.(it is an invalid parameter).

### DIFF
--- a/modules/freetype/include/opencv2/freetype.hpp
+++ b/modules/freetype/include/opencv2/freetype.hpp
@@ -115,7 +115,67 @@ The function putText renders the specified text string in the image. Symbols tha
         int thickness, int line_type, bool bottomLeftOrigin
     ) = 0;
 
+/** @brief Calculates the width and height of a text string.
+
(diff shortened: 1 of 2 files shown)


alalek:
:+1:

Issue #853, opened 2016-11-15
Sfm module on windows using VS 2015


LaurentBerger:
Hi,

Building SFM module with VS2015
For multiview project M_PI is undefined. Preprocessor definition _USE_MATH_DEFINES can be add using VS 2015 (windows 10)

terfendail:
@LaurentBerger 
Could you please provide your CMake and build output and describe used prerequisites and libraries?
I've successfully build SFM with VS2015(Win 10). Additional investigation shows that M_PI constant was defined in `Eigen/Geometry` file in my case.

LaurentBerger:
Sorry for delay : 
Cmake output in GUI : 
FP16: Compiler support is available
A library with BLAS API not found. Please specify library location.
LAPACK requires BLAS
A library with LAPACK API not found. Please specify library location.
found IPP (ICV version): 9.0.1 [9.0.1]
at: F:/lib/opencv/3rdparty/ippicv/unpack/ippicv_win
CUDA detected: 7.5
CUDA NVCC target flags: -gencode;arch=compute_20,code=sm_20;-gencode;arch=compute_30,code=sm_30;-gencode;arch=compute_35,code=sm_35;-gencode;arch=compute_37,code=sm_37;-gencode;arch=compute_50,code=sm_50;-gencode;arch=compute_52,code=sm_52;-D_FORCE_INLINES
Could NOT find JNI (missing:  JAVA_AWT_LIBRARY JAVA_JVM_LIBRARY JAVA_INCLUDE_PATH JAVA_INCLUDE_PATH2 JAVA_AWT_INCLUDE_PATH) 
Could NOT find Matlab (missing:  MATLAB_MEX_SCRIPT MATLAB_INCLUDE_DIRS MATLAB_ROOT_DIR MATLAB_LIBRARIES MATLAB_LIBRARY_DIRS MATLAB_MEXEXT MATLAB_ARCH MATLAB_BIN) 
VTK is not found. Please set -DVTK_DIR in CMake to VTK build directory, or to VTK install subdirectory with VTKConfig.cmake file
Caffe:   NO
Protobuf:   NO
Glog:   NO
Found required Ceres dependency: Eigen version 3.3.0 in F:/lib/eigen-eigen
No preference for use of exported glog CMake configuration set, and no hints for include/library directories provided. Defaulting to preferring an installed/exported glog CMake configuration if available.
Found installed version of glog: F:/lib/glog
Detected glog version: 0.3.4
[...]

PR #937, opened 2017-01-08
solve issue #853

LaurentBerger:
Using eigen 3.3 on windows with vs 2015 constant M_Pi is undefined now . Add compilation directive _USE_MATH_DEFINES solves this problem
I have modified multiview cmakelists to check eigen version


sovrasov:
Strange, I have no problems with the latest Eigen master on Ubuntu 16.04.

LaurentBerger:
@sovrasov  I don't use Ubuntu. I use vs 2015 with opencv 3.3

Origin is  this commit https://bitbucket.org/eigen/eigen/commits/0dfb2d5d3d33d9c44b5cfbd6c5fb68768c05c88a#general-comments

alalek:
Probably it is better to add this define globally. @LaurentBerger Could you check https://github.com/opencv/opencv/pull/8001 please?

terfendail:
Agree with @alalek 
Since M_PI issue on VS2015 is generic it could later raise in another module so I also consider it better option to fix it globally

LaurentBerger:
I think that @alalek solution is  better than mine so we can close this issue

PR #938, opened 2017-01-08
Add readNetFromTensorflow

UnaNancyOwen:
### This pullrequest changes

Add function that reads a network stored in TensorFlow graph files.
It provides a simple way to load a network like <code>cv::dnn::readNetFromCaffe()</code>.

```cpp
// cv::Ptr<cv::dnn::Importer> importer = cv::dnn::createTensorflowImporter( "graph.pb" );
// cv::dnn::Net net;
// importer->populateNet( net );
cv::Ptr<cv::dnn::Net> net = cv::dnn::readNetFromTensorflow( "graph.pb" );
```

### DIFF
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -306,7 +306,7 @@ namespace dnn //! This namespace is used for dnn module functionlaity.
     CV_EXPORTS Ptr<Importer> createTensorflowImporter(const String &model);
 
     /** @brief Reads a network model stored in TensorFlow model files.
-      * @details This is shortcut consisting from createTensorflowImporter and Net::populateNet calls.
+      * @details This is shortcut consisting from createTensorflowImporter and Importer::populateNet calls.
       */
     CV_EXPORTS_W Net readNetFromTensorflow(const String &model);
 
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -308,7 +308,7 @@ namespace dnn //! This namespace is used for dnn module functionlaity.
     /** @brief Reads a network model stored in TensorFlow model files.
       * @details This is shortcut consisting from createTensorflowImporter and Importer::populateNet calls.
       */
-    CV_EXPORTS_W Net readNetFromTensorflow(const String &model);
+    CV_EXPORTS_W Ptr<Net> readNetFromTensorflow(const String &model);
 
     /** @brief Creates the importer of <a href="http://torch.ch">Torch7</a> framework network.
      *  @param filename path to the file, dumped from Torch by using torch.save() function.
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -305,6 +305,11 @@ namespace dnn //! This namespace is used for dnn module functionlaity.
      */
     CV_EXPORTS Ptr<Importer> createTensorflowImporter(const String &model);
 
+    /** @brief Reads a network model stored in TensorFlow model files.
+      * @details This is shortcut consisting from createTensorflowImporter and Net::populateNet calls.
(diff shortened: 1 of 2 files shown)


modules/dnn/src/tensorflow/tf_importer.cpp
```diff
@@ -747,3 +747,20 @@ Ptr<Importer> cv::dnn::createTensorflowImporter(const String&)
...
+Net cv::dnn::readNetFromTensorflow(const String &model)
+{
+    Ptr<Importer> tensorflowImporter;
+    try
+    {
+        tensorflowImporter = createTensorflowImporter(model);
+    }
+    catch(...)
+    {
+    }
```
alalek: Empty catch blocks must be avoided.

UnaNancyOwen: @alalek Thank you for your advice. 
I fixed it to returns an empty cv::dnn::Net().

UnaNancyOwen:
I squashed fix commits.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -305,6 +305,11 @@ namespace dnn //! This namespace is used for dnn module functionlaity.
      */
     CV_EXPORTS Ptr<Importer> createTensorflowImporter(const String &model);
 
+    /** @brief Reads a network model stored in TensorFlow model files.
+      * @details This is shortcut consisting from createTensorflowImporter and Net::populateNet calls.
```
alalek: > Net::populateNet calls

It is not true (looking on code).

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -305,6 +305,11 @@ namespace dnn //! This namespace is used for dnn module functionlaity.
...
+      */
+    CV_EXPORTS_W Net readNetFromTensorflow(const String &model);
```
alalek: We usually return Ptr<> for non scalar types.

modules/dnn/src/tensorflow/tf_importer.cpp
```diff
@@ -747,3 +747,22 @@ Ptr<Importer> cv::dnn::createTensorflowImporter(const String&)
...
+    Ptr<Importer> tensorflowImporter;
+    try
+    {
+        tensorflowImporter = createTensorflowImporter(model);
+    }
+    catch (cv::Exception& e)
+    {
+        std::cerr << "Exception: " << e.what() << std::endl;
+        return Net();
+    }
```
alalek: I don't like this design.
How user application can handle a error and detect "not implemented" or "I/O error" error types?
Why do not pass exception directly to user code?

modules/dnn/include/opencv2/dnn/dnn.hpp
UnaNancyOwen: Oh, That was wrong.
It is "Importer::populateNet calls".

modules/dnn/include/opencv2/dnn/dnn.hpp
UnaNancyOwen: I will change to return <code>Ptr\<Net\></code>.

modules/dnn/src/tensorflow/tf_importer.cpp
UnaNancyOwen: I will change to re-throw exception.

alalek: It is better to remove try-catch at all.

UnaNancyOwen: OK, I will remove try-catch.

modules/dnn/include/opencv2/dnn/dnn.hpp
UnaNancyOwen: @alalek Tests doesn't pass because occur the builds errors in python binding. Could you advise me about it?

sovrasov:
Replaced with #1050

PR #940, opened 2017-01-10
Correcting wrong type check in Aruco Board creation

### LINKED ISSUE #939: Wrong type checking in Aruco Board_create
The board_create python function implemented following #771 throws a type error

To reproduce the bug:

```.python
import cv2.aruco as aruco
import numpy as np

points = [np.array([(0,0,0),(1,0,0),(1,-1,0),(0,-1,0)],('float32'))]

arucoDict = aruco.Dictionary_get(aruco.DICT_4X4_50)

ids = np.array([[0]])

board = aruco.Board_create(points,arucoDict,ids)
```

will throw the following error:
```.bash
    _board = aruco.Board_create(_points,_arucoDict,_ids)
cv2.error: /home/forssea/opencv_contrib/modules/aruco/src/aruco.cpp:1281: error: (-215) objPoints.type() == CV_32FC3 in function create
```

Shouldn't the type to check be CV_32F instead of CV_32FC3?

nicolaje:
resolves #939

### This pullrequest changes
The type checked when creating a board with the board_create python helper function is now CV_32F instead of CV_32FC3

### DIFF
--- a/modules/aruco/src/aruco.cpp
+++ b/modules/aruco/src/aruco.cpp
@@ -1278,7 +1278,7 @@ void GridBoard::draw(Size outSize, OutputArray _img, int marginSize, int borderB
 Ptr<Board> Board::create(InputArrayOfArrays objPoints, const Ptr<Dictionary> &dictionary, InputArray ids) {
 
     CV_Assert(objPoints.total() == ids.total());
-    CV_Assert(objPoints.type() == CV_32FC3);
+    CV_Assert(objPoints.type() == CV_32F);
 
     std::vector< std::vector< Point3f > > obj_points_vector;
     for (unsigned int i = 0; i < objPoints.total(); i++) {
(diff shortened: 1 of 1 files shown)


modules/aruco/src/aruco.cpp
```diff
@@ -1278,7 +1278,7 @@ void GridBoard::draw(Size outSize, OutputArray _img, int marginSize, int borderB
 Ptr<Board> Board::create(InputArrayOfArrays objPoints, const Ptr<Dictionary> &dictionary, InputArray ids) {
 
     CV_Assert(objPoints.total() == ids.total());
-    CV_Assert(objPoints.type() == CV_32FC3);
+    CV_Assert(objPoints.type() == CV_32F);
```
sovrasov: We can support both formats, so here can be `CV_Assert(objPoints.type() == CV_32FC3 || objPoints.type() == CV_32FC1);`

nicolaje:
Thanks for the review, checking both cases works for me, see updated pull request.

```
For me points = [np.array([[0,0],[1,0],[1,-1],[0,-1]],('float32'))] works without any exceptions.
```

Weird. Could it be related to Python's bitness? Mine is 64 bits.

sovrasov:
In this case index overflow occurs in `corners_mat.at<Point3f>(j);`, it's not related to python. In release mode we have no any index checks inside `.at()` and results of out-of-bound access can be different on different machines.

sovrasov:
👍

PR #945, opened 2017-01-12
Added im2row, tiny optimiziations

arrybn:
### This pullrequest changes

Made some optimizations: added im2row, rewrite some frequently called loops
Rewrited Torch ENet sample to make it more useful

### DIFF
--- a/modules/dnn/src/layers/convolution_layer.cpp
+++ b/modules/dnn/src/layers/convolution_layer.cpp
@@ -58,8 +58,7 @@ BaseConvolutionLayerImpl::BaseConvolutionLayerImpl():
     inpH(0), inpW(0), inpCn(0),
     outH(0), outW(0), outCn(0),
     inpGroupCn(0), outGroupCn(0),
-    ksize(0), colBlobCols(0),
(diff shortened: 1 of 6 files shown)


modules/dnn/src/layers/convolution_layer.cpp
```diff
@@ -238,6 +238,31 @@ void ConvolutionLayerImpl::im2col(const Mat &srcImg, Mat &dstCol)
...
+        im2row_CpuPBody<double>::run(srcImg.ptr<double>(), inpGroupCn, inpH, inpW, kernel.height,
+                                     kernel.width, pad.height, pad.width, stride.height, stride.width,
+                                     dilation.height, dilation.width, outW, outH, colMat.ptr<double>());
+
+    dstRow = colMat;
+}
+
+void ConvolutionLayerImpl::im2row(const UMat &srcImg, UMat &dstCol)
+{
+}
```
alalek: "not implemented" exception?

arrybn: Added, thanks

alalek:
:+1:

PR #947, opened 2017-01-13
cmake: add global defines to enable useful macros

alalek:
**Merge with**: https://github.com/opencv/opencv/pull/8001


alalek:
:+1:

PR #953, opened 2017-01-19
Disable all Torch importer source when have no Torch

sovrasov:


modules/dnn/src/torch/torch_importer.cpp
```diff
@@ -964,6 +964,13 @@ Blob readTorchBlob(const String &filename, bool isBinary)
 
     return importer->tensors.begin()->second;
 }
+#else
+
+Ptr<Importer> createTorchImporter(const String &filename, bool isBinary)
+{
+    return Ptr<Importer>();
```
alalek: Probably exception with description is better.

alalek:
:+1:

PR #968, opened 2017-01-24
tracking:Disable all tests except kalman filter

sovrasov:
Tests was disabled because they are partially broken and take long time.


mshabunin:
:+1:

PR #973, opened 2017-01-27
new corner refinement method :: using the contour-lines

szk1509:
### This pullrequest changes:

The pullrequest is trying to do an extension of the "_doCornerRefinement_" parameter.
The idea is to use the contour-points to calculate the corners, in some cases more accurately.

#### Steps (roughly):

1. Marker was detected.
2. Group and classify the contour-points according to their position.
3. Evaluates the direction it follows (clockwise or not). This is for saving the corners at the right position.
4. Fit a line who passes through the points (for each group).
5. Find the Points where the edge lines crosses.
6. Save the updated corners.

### DIFF
--- a/modules/aruco/src/aruco.cpp
+++ b/modules/aruco/src/aruco.cpp
@@ -788,7 +788,7 @@ class MarkerSubpixelParallel : public ParallelLoopBody {
  * Line fitting  A * B = C :: Called from function refineCandidateLines
  * @param nContours, contour-container
  */
-static Point3f interpolate2Dline(const std::vector<cv::Point2f>& nContours){
+static Point3f _interpolate2Dline(const std::vector<cv::Point2f>& nContours){
 	float minX, minY, maxX, maxY;
 	minX = maxX = nContours[0].x;
 	minY = maxY = nContours[0].y;
@@ -833,13 +833,13 @@ static Point3f interpolate2Dline(const std::vector<cv::Point2f>& nContours){
  * @param nLine2
  * @return Crossed Point
  */
-static Point2f getCrossPoint(Point3f nLine1, Point3f nLine2){
+static Point2f _getCrossPoint(Point3f nLine1, Point3f nLine2){
 	Matx22f A(nLine1.x, nLine1.y, nLine2.x, nLine2.y);
 	Vec2f B(-nLine1.z, -nLine2.z);
 	return Vec2f(A.solve(B).val);
 }
 
-static void distortPoints(vector<cv::Point2f>& in, const Mat& camMatrix, const Mat& distCoeff) {
+static void _distortPoints(vector<cv::Point2f>& in, const Mat& camMatrix, const Mat& distCoeff) {
     // trivial extrinsics
     Matx31f Rvec(0,0,0);
     Matx31f Tvec(0,0,0);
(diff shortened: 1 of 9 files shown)


modules/aruco/src/aruco.cpp
```diff
@@ -771,13 +784,153 @@ class MarkerSubpixelParallel : public ParallelLoopBody {
...
+}
 
+/**
+ * Refine Corners using the contour vector :: Called from function detectMarkers
+ * @param nContours, contour-container
+ * @param nCorners, candidate Corners
+ * @param camMatrix, cameraMatrix input 3x3 floating-point camera matrix
+ * @param distCoeff, distCoeffs vector of distortion coefficient
+ */
+static void refineCandidateLines(std::vector<Point>& nContours, std::vector<Point2f>& nCorners, const Mat& camMatrix, const Mat& distCoeff){
```
szk1509: done, i hope i didn't forget anything :)
thank you so much for your help

modules/aruco/src/aruco.cpp
```diff
@@ -771,13 +784,153 @@ class MarkerSubpixelParallel : public ParallelLoopBody {
...
+/**
+ * Find the Point where the lines crosses :: Called from function refineCandidateLines
+ * @param nLine1
+ * @param nLine2
+ * @return Crossed Point
+ */
+static Point2f _getCrossPoint(Point3f nLine1, Point3f nLine2){
+	Matx22f A(nLine1.x, nLine1.y, nLine2.x, nLine2.y);
+	Vec2f B(-nLine1.z, -nLine2.z);
+	return Vec2f(A.solve(B).val);
```
paroj: last warning is here, but does not come from this change but from `determinant()` usage in `Matx_FastSolveOp`

paroj: see: https://github.com/opencv/opencv/pull/8206

paroj:
@alalek can you trigger a rebuild for this one as the required changes are in opencv now?

modules/aruco/src/aruco.cpp
```diff
@@ -819,6 +972,17 @@ void detectMarkers(InputArray _image, const Ptr<Dictionary> &_dictionary, Output
         parallel_for_(Range(0, _corners.cols()),
                       MarkerSubpixelParallel(&grey, _corners, _params));
     }
+
+    /// STEP 4, Optional : Corner refinement :: use contour container
+    if( _params->cornerRefinementMethod == CORNER_REFINE_CONTOUR){
+
+        if(! _ids.empty()){
+            _refineCandidateLines(contours[0], candidates[0], camMatrix.getMat(), distCoeff.getMat());
```
sovrasov: Why do we have refinement only for `candidates[0]` here?

szk1509: D: sorry... i was a little bit confused... my bad
Thank you all! :)

sovrasov:
:+1:

szk1509:
Hi @alalek :)
After the Thumbs up I've changed a tiny little comment on the code and i regret it, because it is taking so long to get into the queue (and with it the pull request) so i was wondering... should i just undo the last commit? or maybe should i just keep waiting (because it is normal that it takes so long)

sorry about bothering you (this is my first time) and thank you so much for your help :)

alalek:
@szk1509 
This patch is not queued automatically because the source branch has reserved name "master" (this should be avoided to eliminate conflicts with other OpenCV repositories: opencv/opencv_extra).
Please select appropriate name in the next time. Thank you for contribution!

PR #978, opened 2017-01-31
tracking: refactoring in tests

sovrasov:
- all testing methods are moved to one class (they were very similar)
- amount and length of testcases were reduced


vpisarev:
looks great! :+1:

PR #997, opened 2017-02-13
Update omnidir_tutorial.markdown

np-csu:
fix the order of parameter while using cv::omnidir::stereoReconstruct


alalek:
Thanks! :+1:

PR #1003, opened 2017-02-15
Improved the textdetection.py sample to remove overlapping rects

chrizandr:
### This pullrequest changes

Modified the python text detection sample in the samples folder of text module.
Previously, the script would detect text regions multiple times, the rects would overlap and detect same text multiple times.
**Old detection:**
old
Added a new grouping algorithm to group these overlapping rects together.
The algorithm groups the rects together to create one larger rect.
Also made some modifications to the previous script by adding comments and changing some parts.
**New detection:**
new

### DIFF
--- a/modules/text/samples/textdetection.py
+++ b/modules/text/samples/textdetection.py
@@ -15,42 +15,98 @@ if (len(sys.argv) < 2):
   print(' (ERROR) You must call this script with an argument (path_to_image_to_be_processed)\n')
   quit()
 
+# Grouping overlapping rects into one larger rect
+def rectGrouping(rects):
+    # Create buckets equal to number of rect
+    buckets = [[] for i in range(len(rects))]
+    # Initialising first bucket
+    buckets[0].append(0)
+    # Implemented an O(n^3) algorithm to merge rects, possible to do it more efficiently
+    for i in range(1,len(rects)):
+        flag = False
+        for bucket in buckets:
+            for j in range(len(bucket)):
+                if rectOverlap(rects[bucket[j]], rects[i]):
+                    bucket.append(i)
+                    flag = True
+                    break
+            if flag:
+                break
+        if not flag:
+            buckets[i].append(i)
+    # Creating bigger rects from the already grouped smaller rects
+    new_rects = list()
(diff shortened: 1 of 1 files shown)


alalek:
Sample is a simple demo for OpenCV API.
Implementing algorithms in samples is not a good idea.

chrizandr:
@alalek : I see, should I close then?
I figured recognition would be easier if we had one rect rather than multiple ones.

StevenPuttemans:
Actually I am not convince your result is better. At the top two lines, you get now a giant overlapping box, while at the original code, you have boxes for each line, which could easily be filtered then.. 

Can you explain again why you think your solutions is better?

I see it works nice for the other lines, but it doesn't for the ones at the top ...

chrizandr:
@StevenPuttemans : I see what you mean. Basically I am working on a sample of end_to_end_recognition using Tesseract. The problem with the ER algorithm was that it would detect the same text multiple times, which we need to filter to get a single bounding box for one text region. 
I observed that in most cases, the whole region is detected along with sub-regions that overlap with the bigger parent. To avoid this, I tried to extract only the larger parent bounding box. I gave this larger bounding box to Tesseract, which then recognized it without any problems.
I could add a small threshold to avoid the case you mentioned above. But I want to know if the actual outcome is useful or not.

StevenPuttemans:
I understand why you do this, but in my opinion, this is something case specific, where you want this output, and can generate it from the output existing now. Therefore I would not update the sample or create a completely new one to show the difference.

chrizandr:
Alright. I understand, closing request then :)

StevenPuttemans:
Ah well, maybe you should consider waiting on the response of @alalek because OpenCV core devs might have another idea on this matter?

alalek:
I mean that implementation of algorithms should be done in the library modules itself, not in the samples. Ideally samples should contain just OpenCV API calls (keep code simple as possible).

chrizandr:
@alalek : Yes, but like I said, I saw something similar in end_to_end_recognition.cpp, so I thought it would be better to have something like that in the python version.

PR #1004, opened 2017-02-15
Missed layer params

dkurt:
### This pullrequest changes

* Added missed initializations of layer names and types.
* Parameters initializations of elementwise layers created with functors. Inherited parameter fields are not initialized.


arrybn:
:+1:

PR #1006, opened 2017-02-16
Getting layer inputs

dkurt:
### This pullrequest changes

Added function for getting input layers of specific layer from Net by layerId.

### DIFF
--- a/modules/dnn/src/dnn.cpp
+++ b/modules/dnn/src/dnn.cpp
@@ -568,7 +568,7 @@ Ptr<Layer> Net::getLayer(LayerId layerId)
 {
     LayerData &ld = impl->getLayerData(layerId);
     if (!ld.layerInstance)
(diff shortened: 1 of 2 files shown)


modules/dnn/src/dnn.cpp
```diff
@@ -572,6 +572,21 @@ Ptr<Layer> Net::getLayer(LayerId layerId)
     return ld.layerInstance;
 }
 
+std::vector<Ptr<Layer> > Net::getLayerInputs(LayerId layerId)
+{
+    LayerData &ld = impl->getLayerData(layerId);
+    if (!ld.layerInstance)
+        CV_Error(Error::StsNullPtr, format("Requseted layer \"%s\" was not initialized", ld.name.c_str()));
```
alalek: Please fix typo here

dkurt: Thanks! Fixed.

alalek:
:+1:

PR #1011, opened 2017-02-20
sfm: fix missing module include in cmake

### LINKED ISSUE #993: Build bot does not build sfm module
When checking the build logs, it seems that the sfm module is not build and thus not incorporated in the CI.

E.g. 
http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_windows64/builds/10031/steps/cmake/logs/stdio
> Module opencv_sfm disabled because the following dependencies are not found: Eigen Glog/Gflags

and the corresponding linux build (where at least Eigen is found):
http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_linux64/builds/10032/steps/cmake/logs/stdio
> Module opencv_sfm disabled because the following dependencies are not found: Glog/Gflags

This also corresponds with the finding that we are currently unavailable to build SFM because the Cmake does not find the required dependencies and aborts. We use MSVC2013 x64 which should work since PR #789

sovrasov:
Resolves #993

### DIFF
--- a/modules/sfm/CMakeLists.txt
+++ b/modules/sfm/CMakeLists.txt
@@ -2,7 +2,7 @@ set(the_description "SFM algorithms")
 
 
 ### LIBMV LIGHT EXTERNAL DEPENDENCIES ###
-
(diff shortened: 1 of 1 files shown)


alalek:
:+1:

PR #1017, opened 2017-02-22
Changed parameters of unpooling layer

arrybn:
### This pullrequest changes
list of parameters, which are passed to unpooling layer. Now this layer computes output size by itself.


alalek:
:+1:

PR #1020, opened 2017-02-22
Create a float-type cloud when depth map is not double-type

### LINKED ISSUE #1021: cv::rgbd::warpFrame fails when camera matrix is double
- OpenCV => 3.0+
- Operating System / Platform => ALL
- Compiler => ALL
##### Detailed description

Camera matrix can be either float or double, but when a `double` matrix is used, the generated cloud would be of double type as well. This is problematic for the `warpFrame` function which expects float instead, causing a crash with a some misleading error...

##### Steps to reproduce
```.cpp
#include <opencv2/rgbd.hpp>

void main()
{
	cv::Size size(640, 480);
	cv::Mat image(size, CV_8UC3),
		depth(size, CV_16U),
		mask(size, CV_8U);
	cv::Mat Rt = cv::Mat(cv::Matx44d::eye()),
		K = cv::Mat(cv::Matx33d::eye()), // cv::Matx33f::eye() would work just fine!
		D = cv::Mat(cv::Vec4d::all(0));
	cv::Mat warpedImage, warpedDepth, warpedMask;

	cv::rgbd::warpFrame(image, depth, mask, Rt, K, D, warpedImage, warpedDepth, warpedMask);
}
```

This can be fixed by merging the PR #1020

Sahloul:
resolves #1021 

### This pullrequest changes
The current implementation of `cv::rgbd::depthTo3d` would return CV_64F cloud if camera matrix argument is of type CV_64F and the depth map is non-float type.

However, since `cv::rgbd::warpFrameImpl` uses `Point2f` type for its points internally, it always expects CV_32F clouds. This results in a crash when camera matrix passed to `cv::rgbd::warpFrame` is of type CV_64F (the default data type in numpy).

This PR resolves the above mentioned issue...

### DIFF
--- a/modules/rgbd/src/depth_to_3d.cpp
+++ b/modules/rgbd/src/depth_to_3d.cpp
@@ -232,14 +232,8 @@ namespace rgbd
         depth.type() == CV_64FC1 || depth.type() == CV_32FC1 || depth.type() == CV_16UC1 || depth.type() == CV_16SC1);
     CV_Assert(mask.empty() || mask.channels() == 1);
 
-    // TODO figure out what to do when types are different: convert or reject ?
     cv::Mat K_new;
-    if ((depth.depth() == CV_32F || depth.depth() == CV_64F) && depth.depth() != K.depth())
-    {
-      K.convertTo(K_new, depth.depth());
-    }
-    else
-      K_new = K;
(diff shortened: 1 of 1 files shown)


Sahloul:
@alalek Mind to review?

alalek:
:+1:

PR #1022, opened 2017-02-23
Compacted `(type)&var.data[i*step]` code into `var.ptr<type>(i)`

Sahloul:
### This pullrequest changes
Seems like some variable is being casted at some point, but its step is used at another place without considering the cast. While C++ did not complain, Numpy had some issues with it. Perhaps around this line:
```.cpp
unsigned int ppfInd = corrInd*ppfStep;
```
Anyhow, it is always great to compact the code! And both Numpy and C++ are happy with this PR..


Sahloul:
@alalek Mind to check why the test failed?

Sahloul:
@alalek Thanks! Now it worked. Mind to merge?

alalek:
:+1:

Issue #941, opened 2017-01-11
GOTURN Tracker results in errors


bassamarshad:
##### System information (version)

- OpenCV => 3.2.0 & opencv_contrib 3.2 
- Operating System / Platform => Windows 10 64 Bit
- Compiler => Visual Studio 2015

##### Detailed description

I have been trying to run the GOTURN tracker OpenCV implementation. I was firstly not able to find the goturn.prototxt and goturn.caffemodel (pretrained versions) on the opencv_extra github site, but was able to pull them from other github repos.

The program fails on execution with the following error, as soon as you select a ROI  : 

OpenCV Error: Assertion failed (input.dims() == 4 && (input.type() == CV_32F || input.type() == CV_64F)) in cv::dnn::ConvolutionLayerImpl::allocate, file E:\opencv3.2.0_sources\opencv_contrib-master\modules\dnn\src\layers\convolution_layer.cpp, line 90
OpenCV Error: Assertion failed (The following error occured while making allocate() for layer "conv11": input.dims() == 4 && (input.type() == CV_32F || input.type() == CV_64F)) in cv::dnn::ConvolutionLayerImpl::allocate, file E:\opencv3.2.0_sources\opencv_contrib-master\modules\dnn\src\layers\convolution_layer.cpp, line 90 

I tried fixing this by updating the following lines of code in gtrtacker.cpp as follows :
 
dnn::Blob targetBlob = dnn::Blob(targetPatch);
 dnn::Blob searchBlob = dnn::Blob(searchPatch);

TO

dnn::Blob targetBlob = dnn::Blob::fromImages(targetPatch);
    dnn::Blob searchBlob = dnn::Blob::fromImages(searchPatch);

NOW,

I get a different error altogether that seems to have something to do with OpenCL, i get this error in the allocate() function in dnn.cpp ::

OpenCV Error: Assertion failed (dims <= 3) in cv::ocl::OpenCLAllocator::checkContinuous, file E:\opencv3.2.0_sources\3.2.0\opencv-3.2.0\modules\core\src\ocl.cpp, line 4899
OpenCV Error: Assertion failed (The following error occured while making forward() for layer "concat1": dims <= 3) in cv::ocl::OpenCLAllocator::checkContinuous, file E:\opencv3.2.0_sources\3.2.0\opencv-3.2.0\modules\core\src\ocl.cpp, line 4899
[...]

PR #1026, opened 2017-02-27
Partial fix to GOTURN tracker

LorenaGdL:
Added fix to GOTURN tracker, when creating dnn::Blob objects

Related issue #941 
OpenCV Q&A related question [here](http://answers.opencv.org/question/129140/goturn-tracker-error/)


mshabunin:
:+1:

PR #1031, opened 2017-03-02
Optimize SIFT with AVX2

woodychow:
Using AVX2  to optimize SIFT descriptor /  histogram calculation.

There are 2 things that I am not too sure about the change
- Should I copy the definition of USE_AVX2 to modules/xfeatures2d/src/precomp.hpp? I see that in the "opencv" repo, every #if CV_AVX2 is followed by if ( USE_AVX2 )
- "#if 0 //CV_AVX2". Enabling that section will break opencv_test_xfeatures2d test. Descriptors are off by at most 1 because nrm2 is summed up in a different order because of AVX2.

Thanks,
Woody

### DIFF
--- a/modules/xfeatures2d/src/precomp.hpp
+++ b/modules/xfeatures2d/src/precomp.hpp
@@ -61,4 +61,6 @@
 
 #include "opencv2/core/private.hpp"
 
+#define USE_AVX2  (cv::checkHardwareSupport(CV_CPU_AVX2))
+
 #endif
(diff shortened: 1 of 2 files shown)


sovrasov:
@woodychow please re-submit this PR using different name of source branch than `master`. Now our buildbot can't automatically fetch your changes and run the builds.

woodychow:
@sovrasov thanks very much.

Here is the new one https://github.com/opencv/opencv_contrib/pull/1033. I am closing this one.

PR #1033, opened 2017-03-03
Optimize SIFT with AVX2

woodychow:
Using AVX2 to optimize SIFT descriptor / histogram calculation.

There are 2 things that I am not too sure about the change

* Should I copy the definition of USE_AVX2 to modules/xfeatures2d/src/precomp.hpp? I see that in the "opencv" repo, every #if CV_AVX2 is followed by if ( USE_AVX2 )
*  "#if 0 //CV_AVX2". Enabling that section will break opencv_test_xfeatures2d test. Descriptors are off by at most 1 because nrm2 is summed up in a different order because of AVX2.

Thanks,
Woody

Recreating PR because https://github.com/opencv/opencv_contrib/pull/1031 was bad.

### DIFF
--- a/modules/xfeatures2d/src/sift.cpp
+++ b/modules/xfeatures2d/src/sift.cpp
@@ -344,7 +344,40 @@ static float calcOrientationHist( const Mat& img, Point pt, int radius,
     cv::hal::fastAtan2(Y, X, Ori, len, true);
     cv::hal::magnitude32f(X, Y, Mag, len);
 
-    for( k = 0; k < len; k++ )
+    k = 0;
+#if CV_AVX2
+    if( USE_AVX2 )
+    {
+        __m256 __nd360 = _mm256_set1_ps(n/360.f);
+        __m256i __n = _mm256_set1_epi32(n);
+        int CV_DECL_ALIGNED(32) bin_buf[8];
+        float CV_DECL_ALIGNED(32) w_mul_mag_buf[8];
+        for ( ; k <= len - 8; k+=8 )
+        {
+            __m256i __bin = _mm256_cvtps_epi32(_mm256_round_ps(_mm256_mul_ps(__nd360, _mm256_loadu_ps(&Ori[k])), _MM_FROUND_TO_NEAREST_INT |_MM_FROUND_NO_EXC));
+
+            __bin = _mm256_sub_epi32(__bin,
+                _mm256_and_si256(__n, _mm256_or_si256(_mm256_cmpeq_epi32(__bin, __n), _mm256_cmpgt_epi32(__bin, __n))));
+            __bin = _mm256_add_epi32(__bin,
+                _mm256_and_si256(__n, _mm256_cmpgt_epi32(_mm256_setzero_si256(), __bin)));
+
+            __m256 __w_mul_mag = _mm256_mul_ps(_mm256_loadu_ps(&W[k]), _mm256_loadu_ps(&Mag[k]));
+
+            _mm256_store_si256((__m256i *) bin_buf, __bin);
+            _mm256_store_ps(w_mul_mag_buf, __w_mul_mag);
+
+            temphist[bin_buf[0]] += w_mul_mag_buf[0];
+            temphist[bin_buf[1]] += w_mul_mag_buf[1];
+            temphist[bin_buf[2]] += w_mul_mag_buf[2];
+            temphist[bin_buf[3]] += w_mul_mag_buf[3];
+            temphist[bin_buf[4]] += w_mul_mag_buf[4];
+            temphist[bin_buf[5]] += w_mul_mag_buf[5];
+            temphist[bin_buf[6]] += w_mul_mag_buf[6];
+            temphist[bin_buf[7]] += w_mul_mag_buf[7];
+        }
+    }
+#endif
+    for( ; k < len; k++ )
     {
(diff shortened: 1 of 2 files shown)


modules/xfeatures2d/src/sift.cpp
```diff
@@ -627,7 +693,141 @@ static void calcSIFTDescriptor( const Mat& img, Point2f ptf, float ori, float sc
...
+                _mm256_mullo_epi32(
+                    _mm256_add_epi32(
+                        _mm256_mullo_epi32(_mm256_add_epi32(_mm256_cvtps_epi32(__r0), __one), _mm256_set1_epi32(d + 2)),
+                        _mm256_add_epi32(_mm256_cvtps_epi32(__c0), __one)),
+                    _mm256_set1_epi32(n + 2)),
+                __o0i);
+
+            _mm256_store_si256((__m256i *)idx_buf, __idx);
+
+            hist[idx_buf[0]] +=               __v_rco000[0];
```
alalek: MSVC compiler stops in this line (same problem with lines below too):
```
error C2676: binary '[': '__m256' does not define this operator or a conversion to a type acceptable to the predefined operator [C:\build\precommit-contrib_windows64\build\modules\xfeatures2d\opencv_xfeatures2d.vcxproj]
```
http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_windows64/builds/10121

woodychow:
Thanks for looking into this. I thought all the tests passed.

I will replace the code with below. When I profiled the code on my machine (g++), the code that doesn't compile on MSVC is slightly faster than below. But I think the below code is easier to read.

```cpp
 #define SUM_HELPER(offset, __v)                \
            _mm256_store_ps(val_buf, __v);        \
            hist[idx_buf[0] + (offset)] += val_buf[0]; \
            hist[idx_buf[1] + (offset)] += val_buf[1]; \
            hist[idx_buf[2] + (offset)] += val_buf[2]; \
            hist[idx_buf[3] + (offset)] += val_buf[3]; \
            hist[idx_buf[4] + (offset)] += val_buf[4]; \
            hist[idx_buf[5] + (offset)] += val_buf[5]; \
            hist[idx_buf[6] + (offset)] += val_buf[6]; \
            hist[idx_buf[7] + (offset)] += val_buf[7]; \
        
        SUM_HELPER(0, __v_rco000);
        SUM_HELPER(1, __v_rco001);
        SUM_HELPER(n+2, __v_rco010);
        SUM_HELPER(n+3, __v_rco011);
        SUM_HELPER((d+2)*(n+2), __v_rco100);
        SUM_HELPER((d+2)*(n+2)+1, __v_rco101);
        SUM_HELPER((d+3)*(n+2), __v_rco110);
        SUM_HELPER((d+3)*(n+2)+1, __v_rco111);
#undef SUM_HELPER

**Comment by woodychow** [comment]
Actually what I posted above doesn't pass the tests because it alters the sum order. Also, it is less efficient.

I changed the code to use _mm256_extract_epi32 to extract the floats every line. However, I think the most efficient way would be to allocate an aligned array which holds 64 floats, and store __v_rco to them. The problem with that is, the aligned array should be allocated only once and used for all iterations, but that would require adding an extra argument to calcSIFTDescriptor only when AVX2 is enabled. I prefer not using #ifdefs on function arguments.

By the way, I also have code for multi threading SIFT which I will create a PR soon.

Thanks,
Woody

**Comment by alalek** [comment]
You can try to allocate aligned data on the stack, no need to allocate on heap and propagate this pointer.

**Comment by alalek** [comment]
BTW, `_mm256_extract_epi32` has similar compatibility [issues](https://github.com/opencv/opencv/blob/6fe1898ab6d2a8a81ff2026e381b692cffe00266/modules/core/src/stat.cpp#L4229-L4235) with MSVS.

**Comment by woodychow** [comment]
> You can try to allocate aligned data on the stack, no need to allocate on heap and propagate this pointer.

Yes that is what I meant.

```cpp
float CV_DECL_ALIGNED(32) val_buf[64];
```
I want to put this in calcDescriptors not calcSIFTDescriptor. That would require adding #ifdef to the function arguments of calcSIFTDescriptor. Should I just put that in calcSIFTDescriptor? If we do that, _mm256_extract_epi32 can be removed.

alalek:
> _mm256_extract_epi32 can be removed

[Yes](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_windows64/builds/10128/steps/compile%20release/logs/stdio):
```
C:\build\precommit-contrib_windows64\opencv_contrib\modules\xfeatures2d\src\sift.cpp(779): error C3861: '_mm256_extract_epi32': identifier not found
```

> I want to put this in calcDescriptors not calcSIFTDescriptor. 

Why? It is better to put this temporary buffer locally, because we should not block compiler optimizations (compiler could check that this buffer is not used really).

woodychow:
OK I moved the temporary buffer to calcSIFTDescriptor. I just thought it would be better to have the buffer declared outside, and pass it into calcSIFTDescriptor as it is a bit big (64 floats). I am not too familiar with the compiler optimizations, anyways.

woodychow:
@alalek, is there something that should to be fixed / improved for this branch?

alalek:
Looks good to me. Thank you!
:+1:

woodychow:
Thanks a lot!

PR #1037, opened 2017-03-04
datasets: remove face dependancy

berak:
changed fr_lfw_benchmark, so it no more needs classes from opencv_face, so we can remove the dependancy (tracking module also depends on datasets).

(lfw also is a verification task, same / not-same, and should be trained without looking at labels)


sovrasov:
:+1:

PR #1040, opened 2017-03-04
Updated xobjdetect module CMakeLists.txt

sergboec:
Currently if you are building OpenCV with contrib under ios, not as framework, that can leads to errors in config station of cmake.
I suppose it would be better if xobjdetect module tool will be optional with the BUILD_opencv_apps just like others tools in OpenCV

### This pullrequest changes

CMakeLists.txt in xobjdetect module. Add option not to build tool at xobjdetect module


alalek:
Looks good to me! :+1:

PR #1045, opened 2017-03-06
Update thinning.cpp

sturkmen72:
Update thinning.cpp

### DIFF
--- a/modules/ximgproc/samples/thinning.cpp
+++ b/modules/ximgproc/samples/thinning.cpp
@@ -11,6 +11,7 @@ using namespace cv;
 int main()
 {
     Mat img = imread("opencv-logo.png", IMREAD_COLOR);
+    resize(img, img, Size(), 0.5, 0.5);
 
     /// Threshold the input image
     Mat img_grayscale, img_binary;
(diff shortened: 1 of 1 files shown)


sturkmen72:
@StevenPuttemans could you check this

modules/ximgproc/samples/thinning.cpp
```diff
@@ -10,7 +10,7 @@ using namespace cv;
 
 int main()
 {
-    Mat img = imread("opencv-logo.png", IMREAD_COLOR);
+    Mat img = imread("../data/opencv-logo.png", IMREAD_REDUCED_COLOR_2);
```
StevenPuttemans: I am not convinced that changes like this work. If the samples get build, the data will not be one folder up. It is only like this in the case of the source files. Thus I suggest to undo this change.

modules/ximgproc/samples/thinning.cpp
```diff
@@ -19,8 +19,8 @@ int main()
 
     /// Apply thinning to get a skeleton
     Mat img_thinning_ZS, img_thinning_GH;
-    //ximgproc::thinning(img_binary, img_thinning_ZS, THINNING_ZHANGSUEN);
-    //ximgproc::thinning(img_binary, img_thinning_GH, THINNING_GUOHALL);
+    ximgproc::thinning(img_binary, img_thinning_ZS, ximgproc::THINNING_ZHANGSUEN);
+    ximgproc::thinning(img_binary, img_thinning_GH, ximgproc::THINNING_GUOHALL);
```
StevenPuttemans: This is wrong, the constants are not under te ximgproc namespace for the moment, and I am still awaiting response on why they are not. So this would break the sample.

StevenPuttemans: Forget this last remark, it should indeed be there or at least include `using namespace cv::ximgproc` in the beginning of the file!

modules/ximgproc/samples/thinning.cpp
sturkmen72: hi @StevenPuttemans 
"../data/" used in many sample 
for example https://github.com/opencv/opencv_contrib/blob/master/modules/ximgproc/samples/disparity_filtering.cpp#L18-L20

indeed i don't know another way and i always move the opencv/samples/data folder to a proper place according to where the compiled exe files placed.

so still i see this change will be useful.

also IMREAD_REDUCED_COLOR_2 will load the image resized. it is also better to see result image.

StevenPuttemans: Well it is a matter of flavor. 
About the reduced flag, I would not do this, it is additional functionality that is NOT needed in this case and could confuse people.

sovrasov:
:+1:

PR #1038, opened 2017-03-04
tracking: make opencv_dnn dependancy optional

berak:
resolves #1036
resolves #1030
resolves #1046

### This pullrequest changes

add a few HAVE_OPENCV_DNN conditionals, based on opencv_modules.hpp

PR #1046, opened 2017-03-06
one of the tracking files will not compile without an include from da…

dtmoodie:
Without this fix compilation of opencv_tracking will fail with:

In file included from /code/opencv_contrib/modules/tracking/src/gtrTracker.hpp:49:0,
                 from /code/opencv_contrib/modules/tracking/src/gtrTracker.cpp:42:
/code/opencv_contrib/modules/tracking/src/gtrUtils.hpp:7:43: fatal error: opencv2/datasets/track_alov.hpp: No such file or directory
compilation terminated.


alalek:
Should be fixed here: #1038

PR #1054, opened 2017-03-09
Fix reallocation issue of raw poses vector

Sahloul:
### This pullrequest changes
Fixes a reallocation issue that causes crashes rarely in Python wrapper (and possibly in C++ apps as well).

Edit: Seems the crashes were caused by OpenMP as in #1055, however, fixing the reallocation is still recommended.


alalek:
:+1:

PR #1056, opened 2017-03-09
Computation reduction of angle between two normalized vectors

Sahloul:
When it comes to the angle between two vectors:
- angle in range [-pi/2, pi/2] can be given by: `asin(|a x b|)`
- angle in range [0, pi] can be given by either: `atan2(|a x b|, a dot b)` or `acos((a dot b) / (|a| |b|))`

However, knowing that the two vectors are either:
- Two normalized surface Normals, or
- One normalized surface vector and the normalized distance vector

then, expression could be simplified to just `acos(a dot b)`
In fact, the above expression is largely employed in many software.

### This pullrequest changes
Reduces the angle computation between two normalized vectors

### DIFF
--- a/modules/surface_matching/src/c_utils.hpp
+++ b/modules/surface_matching/src/c_utils.hpp
@@ -83,11 +83,15 @@ static inline void TCross(const double a[], const double b[], double c[])
   c[2] = (a[0])*(b[1])-(a[1])*(b[0]);
 }
 
-static inline double TAngle3(const double a[3], const double b[3])
+static inline double TAngle3Normalized(const double a[3], const double b[3])
 {
-  double c[3];
-  TCross(a,b,c);
-  return (atan2(TNorm3(c), TDot3(a, b)));
+  /*
+   angle = atan2(a dot b, |a x b|) # Bertram (accidental mistake)
+   angle = atan2(|a x b|, a dot b) # Tolga Birdal (correction)
+   angle = acos(a dot b)           # Hamdi Sahloul (simplification, a & b are normalized)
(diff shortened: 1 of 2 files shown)


modules/surface_matching/src/c_utils.hpp
```diff
@@ -85,9 +85,13 @@ static inline void TCross(const double a[], const double b[], double c[])
 
 static inline double TAngle3(const double a[3], const double b[3])
```
sovrasov: I think the name of this function should be changed accordingly. Something like `TAngle3Normalized` is more suitable.

Sahloul: @sovrasov Thank you very much! I revised accordingly.

sovrasov:
:+1:

tolgabirdal:
The idea using this code is not to assert a range but numerical stability. Please check Section 12 of the Mindless:
https://people.eecs.berkeley.edu/~wkahan/Mindless.pdf

Sahloul:
@tolgabirdal The real problem is if `TDot3(a, b) ` results in `π+ε`, in which `acos` become unstable.
But is this case possible? How?

Sahloul:
@sovrasov @alalek Please judge the situation. Personally, I am not biased to this PR as much as to the efficiency. Feel free to revert this PR if there is a possibility for instability issue (e.g. `π+ε` might occur).

alalek:
I believe the current code is fine until there is

> an example that fails with this formula but pass with the other one

that can be used as a regression test.

PR #1058, opened 2017-03-10
Stop ICP iterations before solver crashes due to underdetermined system

Sahloul:
The `minimizePointToPlaneMetric()` calls `cv::solve()` to solve a linear system consisting of six compoenents:
- Euler angles (3 components: roll, pitch, yaw) 
- Translation (three components: X, Y, Z)

In order to solve such linear system, there should be at least six points corresponding between the model and scene clouds. However, current implementation constrain minimal limit to only one point, which may cause crashes.

### This pullrequest changes
Sets a minimum limit of six-points correspondance in order to iterate ICP further


Sahloul:
@sovrasov Mind to review this as well?

sovrasov:
:+

PR #1059, opened 2017-03-10
Parallelize `ICP::registerModelToScene`

Sahloul:
### This pullrequest changes
Speedup processing poses passed to `registerModelToScene()`

### DIFF
--- a/modules/surface_matching/src/icp.cpp
+++ b/modules/surface_matching/src/icp.cpp
@@ -540,7 +540,10 @@ int ICP::registerModelToScene(const Mat& srcPC, const Mat& dstPC, double& residu
 // source point clouds are assumed to contain their normals
 int ICP::registerModelToScene(const Mat& srcPC, const Mat& dstPC, std::vector<Pose3DPtr>& poses)
 {
-  for (size_t i=0; i<poses.size(); i++)
+  #if defined _OPENMP
+  #pragma omp parallel for
+  #endif
+  for (int i=0; i<(int)poses.size(); i++)
   {
     Matx44d poseICP = Matx44d::eye();
     Mat srcTemp = transformPCPose(srcPC, poses[i]->pose);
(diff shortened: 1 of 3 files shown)


Sahloul:
@alalek @sovrasov Kindly be asked to review this PR!

modules/surface_matching/src/icp.cpp
```diff
@@ -540,12 +540,22 @@ int ICP::registerModelToScene(const Mat& srcPC, const Mat& dstPC, double& residu
 // source point clouds are assumed to contain their normals
 int ICP::registerModelToScene(const Mat& srcPC, const Mat& dstPC, std::vector<Pose3DPtr>& poses)
 {
-  for (size_t i=0; i<poses.size(); i++)
+  #if defined _OPENMP
```
sovrasov: In OpenCV there is own [parallell for](http://docs.opencv.org/trunk/d7/dff/tutorial_how_to_use_OpenCV_parallel_for_.html), which is more universal than implicit OpenMP usage, but it hasn't critical sections unfortunately. Processing of every `poses[i]` seems to be independent from others. Why do you need `omp critical` here?

Sahloul: @sovrasov I agree that using `cv::parallel_for` seems more wise, and I would consider that in my rewrite version. However, the rest of the module is using OpenMP for now, so I prefer to remain consistent.

As for the `omp critical` part, although it does not seem to be needed, it does not harm the execution as well. Heavy processes are already done before the critical part, and it is just a kind of protective practice. Anyway, I am removing it since you do not like such practice.

sovrasov:
:+1:

Sahloul:
@alalek Please merge this PR.

Sahloul:
@sovrasov @alalek Any chance of merge?

PR #1063, opened 2017-03-13
Reduce modules dependencies

mshabunin:
### This pullrequest changes

Linked PR: opencv/opencv#8376

- Many modules do not need _highgui_ for build - it has been removed or replaced with _imgcodecs_ (_ccalib_ still have more or less complex function which calls `imshow`).
- Samples always have _highgui_, _imgcodecs_ and _videoio_ as a dependency, so they can be removed from the module itself. When a dependency is only required by a sample application, it can be made optional by adding `#ifdef HAVE_OPENCV_<module>` guard to the app and moving module to the `OPTIONAL` part in `ocv_define_module`.
- `roiSelector` has been moved from the _tracking_ module public interface to the samples folder and transformed to the single header file.

**Notes:**
- It is often needed to debug existing functions by inserting `imshow` calls to the algorithm implementations, but without explicit _highgui_ dependency this method will not work. I suggest adding logging functions to the _core_ (or some other) module which will be able to store or display images when _highgui_ and _imgcodecs_ modules are available and dump data as _xml_ otherwise.
- `roiSelector` can be moved to the _highgui_ module in future.
- Several modules (_cvv_, _viz_, _cuda*_, _sfm_, ?) has not been tested by the "disable-one-module-and-build" test. Windows and Android builds has not been tested too. Java and Python tests has not been run.
- Instead of guarding samples with `ifdef`, we can implement cmake mechanism for adding dependencies to single file: `ocv_add_sample_dependency(compute_descriptors.cpp features2d)` before `ocv_define_module`


alalek:
:+1:

PR #1075, opened 2017-03-21
dnn_modern: update docs, fix build problems

sovrasov:
Several updates in `dnn_modern` module:
- enable Python bindings
- update docs
- fix build instruction
- code cleanup
- disable warnings which come form tiny-dnn
- add tiny-dnn sources download procedure to CMake


vpisarev:
:+1:

PR #1082, opened 2017-03-24
Parallelize SIFT

woodychow:
- Parallelize calcDescriptors and buildDoGPyramid
- Simplify 2 lines of AVX2 instructions

I have also parallelized findScaleSpaceExtrema in my local branch. The thing is, that will change the order of the keypoints, and I think the tests will break. To get stable result, the keypoints have to be sorted. As that is more involved, I guess I will add that to a separate PR.


alalek:
:+1:

woodychow:
Thanks!

PR #1084, opened 2017-03-24
tracking: hide getModel() method in Tracker API

sovrasov:
`TrackerModel` is private stuff, so it's better to hide it.
Related: https://github.com/opencv/opencv/issues/8451


mshabunin:
:+1:

PR #1093, opened 2017-03-28
Added CV_OUT in process() of selective search segmentation module.

lhelontra:
process method of createSelectiveSearchSegmentation don't returns bounding boxes in python bindings.
This PR fixes this. I also added a sample in python.


sovrasov:
:+1:

PR #1094, opened 2017-03-29
Add implementation of read/write methods to PPF3DDetector

### LINKED ISSUE #1091: PPF3DDetector::read & ::write methods missing

- OpenCV => 3.2-dev/master
- Operating System / Platform => Ubuntu 16.04
- Compiler => gcc 5.4

##### Detailed description

This is kind-of a feature request, because it looks like PPF3DDetector::read and ::write were somewhat intended and defined in the ppf_match_3d.hpp header file but never implemented in the ppf_match_3d.cpp file.

##### Steps to reproduce

1) Build / install OpenCV with opencv_contrib modules.
2) Write any short program trying to use the cv::ppf_match_3d::PPF3DDetector::read or ~::write methods.
The above results in an (obvious) linker error of the form ```undefined reference to `cv::ppf_match_3d::PPF3DDetector::read(cv::FileNode const&)'```

sovrasov:
Resolves #1091
**WIP**

### DIFF
--- a/modules/surface_matching/src/ppf_match_3d.cpp
+++ b/modules/surface_matching/src/ppf_match_3d.cpp
@@ -300,6 +300,34 @@ void PPF3DDetector::trainModel(const Mat &PC)
   trained = true;
 }
 
+void PPF3DDetector::read(const FileNode& fn)
+{
+    CV_Assert(!fn["sampling_step_relative"].empty() &&
+            !fn["sampling_step_relative"].empty() &&
+            !fn["distance_step_relative"].empty() &&
+            !fn["num_angles"].empty() &&
+            !fn["position_threshold"].empty() &&
+            !fn["use_weighted_clustering"].empty());
+
+    fn["sampling_step_relative"] >> sampling_step_relative;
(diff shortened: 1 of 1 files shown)


Algomorph:
@sovrasov , this alone does not really resolve #1091, perhaps because I was not explicit enough in my description. I don't know if this is what you meant by "**WIP**". We need a way to save and load a pre-trained model, meaning all those parameters that you're saving, yes, but also the hash itself, the ```sampled_pc``` matrix, and the ```ppf``` matrix (and possibly other things).

I want to call to your attention that the hashtableWrite method, which may be possible to adapt for this purpose, currently has a bug on line 295.
Instead of
```C++
            fwrite(&node->data, dataSize, 1, f);
```
it should be
```C++
            fwrite(node->data, dataSize, 1, f);
```
This way, the actual data, not just it's address in memory, gets written to the binary file.

MartinPernus:
Are there any plans to actually be able to save and load trained model? This commit only implements writing and reading parameters instead of saving trained data.

PR #1098, opened 2017-03-29
ximgproc: fix weightedMedianFilter crash on empty mask

### LINKED ISSUE #1086: cv::ximgproc::weightedMedianFilter fails with mask containing 0's

- OpenCV => 3.2.0
- Operating System / Platform => Mac OS X 64 Bit
- Compiler => Apple Clang

##### Detailed description

The function cv::ximgproc::weightedMedianFilter is happy only when the mask contains all non-zeros. I believe this is inherent to the implementation and has nothing to do with system parameters. Also note that this function might be better off as a class - fewer memory allocations, de-allocations. 

##### Steps to reproduce

 Use this function with any mask containing zeros.

sovrasov:
Resolves #1086


vpisarev:
:+1:

PR #1101, opened 2017-03-30
issue_1100 dnn and xfeatures2d cmake ocv_download needs PACKAGE instead of FILENAME

mevanecek:
Update the dnn and xfeatures2d cmake files to use PACKAGE instead of
FILENAME when invoking opencv’s ocv_download function in
opencv-3.2.0/cmake/OpenCVUtils.cmake


alalek:
Just keep up-to-date main opencv repository too.
There is no compatibility support for CMake scripts between versions.

PR #1102, opened 2017-03-30
refactored DNN

vpisarev:
**WIP**

[done] 1. public API is converted to Blobs.
[tbd] 2. rebase other parts of the patch - get rid of OpenCL, get rid of Blob's, optimize internals


vpisarev:
:+1:

PR #1105, opened 2017-04-02
Reuse DNN outputs

dkurt:
### This pullrequest changes

Proposed manager for reusing allocated blobs for layers' outputs and internals.
Enabled FCN-8s test for Win64.
Now layers just tell us that they could work in-place. But their implementation should be independent of it.
Split, batch normalization and scale layers now could work in-place. 

Note: in-place layers are only layers with single input blob (number of output blobs might be more than one).

**Measurements:**
Load model, allocate and do single forward pass.

| Model | Maximum resident memory (based on /usr/bin/time --verbore) |
|--------|------|
| ENet, 512x512x3 input image | from 828MB to 190MB (x4.36) |
|ResNet-50, 224x224x3 | from 343MB to 212MB (x1.62) |
|FCN-8s, 500x500x3 | from 4.5GB to 2.78GB (x1.62) |

### DIFF
--- a/modules/dnn/src/dnn.cpp
+++ b/modules/dnn/src/dnn.cpp
@@ -154,6 +154,11 @@ struct LayerPin
     {
         return (lid == r.lid && oid == r.oid);
     }
+
+    bool operator<(const LayerPin &r) const
+    {
+        return lid < r.lid || lid == r.lid && oid < r.oid;
+    }
 };
 
 struct LayerData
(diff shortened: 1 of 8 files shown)


dkurt:
WIP. After resolving conflicts I'll do memory allocation outside layers.

dkurt:
WIP: reuse data layer.

dkurt:
WIP: extend blobs manager to making decisions about in-place computations (for layers where it supported). Pins tracker works well and it could be used safely in cases like:
Split (forward input references instead copying the data)
Element-wise layers that currently are not in-place (scale, batch normalization)
Element-wise addition, multiplication, maximum (use one of inputs for accumulation).

arrybn:
:+1:

arrybn:
There is some necessary change to do for this PR: rework process of getting blobs from net after forward

arrybn:
So please don't merge it until I make this

dkurt:
Merged with #1205

PR #1110, opened 2017-04-06
MSLIC: fix uninitialized iteration variable

### LINKED ISSUE #1109: MSLIC has undeterministic behaviour
- OpenCV => 3.x (master branch)
- Operating System / Platform => Ubuntu 16.04 LTS
- Compiler => gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609

##### Detailed description
When rerunning the MSLIC implementation multiple times on the same image, it sometimes results in a very bad segmentation. To me the algorithm looks like it should be deterministic. But the implementation sometimes yields unreasonable bad results. I assume there is a problem with accessing invalid memory.

Expected and sometimes actual outcome:
mslic_normal

Bad result:
mslic_bug

##### Steps to reproduce

Take any image and run MSLIC on it.

jexner:
resolves #1109 

### This pullrequest changes

The variable `m_iterations` is never initialized. This probably leads to the undeterministic behaviour as described in #1109. Maybe @cbalint13 can confirm that `itrnum` is the correct value for `m_iterations`.

### DIFF
--- a/modules/ximgproc/src/slic.cpp
+++ b/modules/ximgproc/src/slic.cpp
@@ -278,6 +278,9 @@ void SuperpixelSLICImpl::initialize()
 
 void SuperpixelSLICImpl::iterate( int num_iterations )
 {
+    // store total iterations
+    m_iterations = num_iterations;
+
     if( m_algorithm == SLICO )
       PerformSLICO( num_iterations );
     else if( m_algorithm == SLIC )
--- a/modules/ximgproc/src/slic.cpp
+++ b/modules/ximgproc/src/slic.cpp
@@ -1375,10 +1375,6 @@ inline void SuperpixelSLICImpl::PerformMSLIC( const int&  itrnum )
     for( int b = 0; b < m_nr_channels; b++ )
       sigma[b].resize(m_numlabels, 0);
 
-    vector<float> sigmax(m_numlabels, 0);
-    vector<float> sigmay(m_numlabels, 0);
-    vector<int> clustersize(m_numlabels, 0);
-
     Mat distvec( m_height, m_width, CV_32F );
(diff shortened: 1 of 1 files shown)


cbalint13:
@jexner 

(1) m_iterations is a global thing that store desired iterations, itrnum is a different thing that track other thing.

* May i see your test code ? You cannot call interate() on SLIC calss only once.

jexner:
@cbalint13 Ok, so maybe `itrnum` isn't the right value for `m_iterations`. But `m_iterations` is never set, but read.

cbalint13:
@jexner ,

* Can try this patch: http://openrisc.rdsor.ro/slic.patch

jexner:
@cbalint13 Yes, that will work. `num_iterations` and `itrnum` are the same variables. `num_iterations` is passed to `PerformMSLIC` [here](https://github.com/jexner/opencv_contrib/blob/698b6852a7fe5d9e852a47efa14edc93b014df70/modules/ximgproc/src/slic.cpp#L286) and then becomes `itrnum` [here](https://github.com/jexner/opencv_contrib/blob/698b6852a7fe5d9e852a47efa14edc93b014df70/modules/ximgproc/src/slic.cpp#L1369).

cbalint13:
@jexner ,

Can rebase this PR ? (Can delete those three vector containers too).

Will work in weekend for a reusable iteration(), also planning a cuda version too plus testcases.

jexner:
@cbalint13 I integrated your patch and rebased the PR. For my needs multiple calls to `iterate()` are not necessary. But probably a handy feature.

sovrasov:
:+1:

PR #1112, opened 2017-04-07
Fixed gflags search in Ubuntu

mshabunin:
resolves https://github.com/opencv/opencv/issues/8536

`set(... PARENT_SCOPE)` block at the end of the function *__find_exported_gflags* should not be executed in case of failure. Otherwise, following calls to `find_path(GFLAGS_INCLUDE_DIR ...)` will not work.


alalek:
:+1:

PR #1115, opened 2017-04-11
Tracking API update

sovrasov:
- Delete `create(String)` function.
- Use `InputArray` in public methods of the `Tracker` class.
- Change samples to be consistent with the new API.
Merge after https://github.com/opencv/opencv/pull/8598


vpisarev:
looks great! :+1:

Issue #1118, opened 2017-04-12
Matlab bindings failed to build on latest master + contrib


StevenPuttemans:
When building latest master of both repos and enabling `BUILD_opencv_matlab`, I get the following error in both release and debug.

```CMake Error at /home/spu/Github/opencv_contrib/modules/matlab/compile.cmake:47 (message):
  Failed to compile wrapperEMD:
  /home/spu/Github/opencv/build_release/modules/matlab/src/wrapperEMD.cpp: In
  function ‘void mexFunction(int, mxArray**, int, const mxArray**)’:

  
  /home/spu/Github/opencv/build_release/modules/matlab/src/wrapperEMD.cpp:47:3:
  error: ‘Ptr_float’ was not declared in this scope

     Ptr_float lowerBound = inputs[4].empty() ? (Ptr_float) Ptr<float>() : inputs[4].toPtrFloat();
     ^

  
  /home/spu/Github/opencv/build_release/modules/matlab/src/wrapperEMD.cpp:54:69:
  error: ‘lowerBound’ was not declared in this scope

       retval = cv::wrapperEMD(signature1, signature2, distType, cost, lowerBound, flow);
                                                                       ^

  
  /home/spu/Github/opencv/build_release/modules/matlab/src/wrapperEMD.cpp:65:16:
  error: ‘lowerBound’ was not declared in this scope

     outputs[1] = lowerBound;
```

For now just disabling the matlab bindings work, but since it is generated code that goes wrong, I am unsure how I can fix this.

alalek:
Related: https://github.com/opencv/opencv/pull/8496

Try to add new entry for `Ptr_float` [here](https://github.com/opencv/opencv_contrib/blob/40f1ac514b27d2c4cd4fac81ce54f3b9b68ff41b/modules/matlab/include/opencv2/matlab/bridge.hpp#L97).

StevenPuttemans:
Okay first issue fixed, however as expected now `inputs[4].toPtrFloat();` raises an issue because it does not exist. Going to take a look at that!

PR #1119, opened 2017-04-12
fixing matlab wrapping

StevenPuttemans:
This fix resolves issue #1118

### DIFF
--- a/modules/matlab/include/opencv2/matlab/bridge.hpp
+++ b/modules/matlab/include/opencv2/matlab/bridge.hpp
@@ -95,7 +95,7 @@ typedef cv::Ptr<TonemapDrago> Ptr_TonemapDrago;
 typedef cv::Ptr<TonemapDurand> Ptr_TonemapDurand;
 typedef cv::Ptr<TonemapMantiuk> Ptr_TonemapMantiuk;
 typedef cv::Ptr<TonemapReinhard> Ptr_TonemapReinhard;
-
+typedef cv::Ptr<float> Ptr_float;
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


modules/matlab/include/opencv2/matlab/bridge.hpp
```diff
@@ -522,6 +522,11 @@ class Bridge {
   Bridge& operator=(const Ptr_TonemapReinhard& ) { return *this; }
   Ptr_TonemapReinhard toPtrTonemapReinhard() { return Ptr_TonemapReinhard(); }
   operator Ptr_TonemapReinhard() { return toPtrTonemapReinhard(); }
+
+  // ---------------------------   Ptr_float   ----------------------
+  Bridge& operator=(const Ptr_float& ) { return *this; }
+  Ptr_float toPtrFloat() { return Ptr_float(); }
```
alalek: BTW, Looks like this method always create new empty "Ptr", so statement below will not use "input value" at all:
```
Ptr_float lowerBound = inputs[4].empty() ? (Ptr_float) Ptr<float>() : inputs[4].toPtrFloat();
```

alalek:
:+1:

modules/matlab/include/opencv2/matlab/bridge.hpp
StevenPuttemans: This is just to fix the build process. If they want to make sure the code is doing something useful, than that is for the guy of the other PR that wrapped this wrongfully xD

StevenPuttemans: But I get what you saying!

PR #1123, opened 2017-04-14
Add confidence threshold to TrackerKCF

### LINKED ISSUE #1121: Tracker::update() always returns true
- OpenCV => 3.2 (latest master)
- Operating System / Platform => any
- Compiler => any

[Tracker::update](http://docs.opencv.org/master/d0/d0a/classcv_1_1Tracker.html#a6b2bc1d8bf400424a883367edc94dbcf) should return false, if it lost the object, but both the KCF and the GOTURN implementation always return true, so we can never find out, if it's nessecary to re-initialize it.

sovrasov:
Resolves #1121

### DIFF
--- a/modules/tracking/include/opencv2/tracking/tracker.hpp
+++ b/modules/tracking/include/opencv2/tracking/tracker.hpp
@@ -1236,6 +1236,7 @@ public:
     */
     void write(FileStorage& /*fs*/) const;
 
+    double detect_thresh;         //!<  detection confidence threshold
     double sigma;                 //!<  gaussian kernel bandwidth
     double lambda;                //!<  regularization
     double interp_factor;         //!<  linear interpolation factor for adaptation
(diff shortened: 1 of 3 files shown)


vpisarev:
:+1:

anandcu3:
Which version is this available from? Is it available in the cv2 Python bindings? If so, how do I use it in Python?

galdalali:
I am trying to understand the logic for determinate lost of tracking on KCF tracker.
Can anyone explain to me superficially what is the logic?

Especially , what is "response"?

`
calcResponse(alphaf,kf,response, spec);
`

And whay doed the maxVal is between 0-1?

```
// extract the maximum response
minMaxLoc( response, &minVal, &maxVal, &minLoc, &maxLoc );
 if (maxVal < params.detect_thresh)
```

PR #1127, opened 2017-04-17
Accuracy scripts

arrybn:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/328

Added python scripts to measuring and comparing accuracy of dnn module with different frameworks in different tasks: classification (AlexNet in Caffe, GoogleNet in Tensorflow, Caffe), semantic segmentation (FCN in Caffe, ENet in torch)
Also made some changes in the code: added functions of loading models, added logic of reinitialize network after setting blob with another dimension


vpisarev:
:+1:

PR #1134, opened 2017-04-24
Visibility of layer options

dkurt:
### This pullrequest changes

BatchNorm, Scale, ReLU, Power and MaxUnpool layers parameters and flags are public. Motivated by another layers (Pooling, Conv, etc.).
Added initialization of names of Blank layers (Dropout, Identity).


arrybn:
:+1:

PR #1137, opened 2017-04-26
linemod: add rudimentary python bindings

paroj:
requires https://github.com/opencv/opencv/pull/8654


alalek:
These build failures are not related to your patch, looking on these tests

alalek:
Thank you! :+1:

PR #1144, opened 2017-05-01
[bugfix] Rounding error in cv::ximgproc::StructuredEdgeDetection

### LINKED ISSUE #1143: [bugfix] Rounding error in cv::ximgproc::StructuredEdgeDetection
There is an error in the implementation of the [Piotr's toolbox gradQuantize() function](https://github.com/pdollar/toolbox/blob/1a3c9869033548abb0c7a3c2aa6a7902c36f39c2/channels/private/gradientMex.cpp).

The original implementation in /modules/ximgproc/src/structured_edge_detection.cpp::248
```
int index  = cvRound((j/pSize + pPhase[j])*nBins);
index = std::max(0, std::min(index, pHistSize));
```

Fix:
```
int angle = cvRound(pPhase[j]*nBins);
if(angle >= nBins)
{
    angle = 0;
}
const int index = (j/pSize)*nBins + angle;
```

To illustrate the bug, here is a representation of one channel of the gradient histogram:

bug

And after the fix:

fix

olivierpascal:
resolves #1143

### DIFF
--- a/modules/ximgproc/src/structured_edge_detection.cpp
+++ b/modules/ximgproc/src/structured_edge_detection.cpp
@@ -245,8 +245,12 @@ static void gradientHist(const cv::Mat &src, cv::Mat &magnitude, cv::Mat &histog
 
         for (int j = 0; j < phase.cols; ++j)
         {
-            int index  = cvRound((j/pSize + pPhase[j])*nBins);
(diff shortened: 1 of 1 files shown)


alalek:
:+1:

PR #1146, opened 2017-05-02
Framework for a viable way to add some more binarization algorithms, …

### LINKED ISSUE #1138: Extend current niBlackThreshold to handle general Local Adaptive Threshold algorithms, Sauvola at the least
I'd like to extend our current implementation of [Niblack's](https://github.com/opencv/opencv_contrib/blob/master/modules/ximgproc/src/niblack_thresholding.cpp) to implement Sauvola's binarization technique.

Current niBlackThreshold signature:

`void niBlackThreshold( InputArray _src, OutputArray _dst, double maxValue, int type, int blockSize, double delta )`

Desired signature that would minimally change niBlackThreshold and maintain backwards compatibility:

`void niBlackThreshold( InputArray _src, OutputArray _dst, double maxValue, int type, int blockSize, double delta, bool useSauola = false )`

Extremely desirable signature that would be slightly more substantial but would allow for newer Niblack-inspired binarization algos to be implemented in the future without much refactoring and moving around:

`void niBlackThreshold( InputArray _src, OutputArray _dst, double maxValue, int type, int blockSize, double delta, int method = NIBLACK )`, where `method` is an enum that specifies a couple of popular NIblack-inspired binarization techniques, like Wolf's and NICK.

The reason I focus on modifying niBlackThreshold is that the formula most people associate with Sauvola document binarization is very similar to Niblack's and another source just for Sauvola's seems unwarranted.

**Desired outcome:** At the least, I'd like to submit a PR for the first signature, but before doing that, I want to leave this Issue open to foster a little community discussion on such a change, take in other's ideas as well.

saskatchewancatch:
…by extending the current niBlackThreshold implementation

### This pullrequest changes

Please see #1138 for context, where I proposed adding some more binarization algos.

This PR is a _draft_ of my proposed way to add some more Niblack inspired binarization algorithms to opencv_contrib. It is only a draft, I am looking for a first review.

Could I kindly get someone to review it and get an opinion on whether this is a way to go to add some more popular binarization algos? If this is the way to go, then I will gladly finish it off: tidying up source, update docs, etc...

3 major changes here to give the idea of what I propose:
1. Added enum to denote common NIblack inspired binarization algos.

2. Changed the current Niblack implementation (cv::ximgproc::niBlackThreshold)to handle multiple Niblack inspired algos including the original Niblack. I'm using the "delta" parameter of existing niBlackThreshold to pass values to algos' user-defined parameters. 

3. Changed the NIblack example to show that it works. Did some basic testing with DIBCO2009 and [DIBCO2016 dataset](https://vc.ee.duth.gr/h-dibco2016/benchmark/DIBCO2016_dataset-original.zip) and results seem reasonable. Use the "method" slider to switch between algos.

### DIFF
--- a/modules/ximgproc/include/opencv2/ximgproc.hpp
+++ b/modules/ximgproc/include/opencv2/ximgproc.hpp
@@ -79,10 +79,21 @@ enum ThinningTypes{
     THINNING_GUOHALL      = 1  // Thinning technique of Guo-Hall
 };
 
+/**
+* @brief Specifies the binarization method to use in cv::ximgproc::niBlackThreshold
+*/
+enum LocalBinarizationMethods{
+	BINARIZATION_NIBLACK = 0, //!< Classic Niblack binarization. See @cite Niblack1985 .
+	BINARIZATION_SAUVOLA = 1, //!< Sauvola's technique. See @cite Sauvola1997 .
+	BINARIZATION_WOLF = 2,    //!< Wolf's technique. See @cite Wolf2004 .
+	BINARIZATION_NICK = 3     //!< NICK technique. See @cite Khurshid2009 .
+};
+
 //! @addtogroup ximgproc
 //! @{
 
-/** @brief Applies Niblack thresholding to input image.
+/** @brief Performs thresholding on input images using Niblack's technique or some of the
+popular variations it inspired.
 
 The function transforms a grayscale image to a binary image according to the formulae:
 -   **THRESH_BINARY**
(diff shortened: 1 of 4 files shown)


modules/ximgproc/include/opencv2/ximgproc.hpp
```diff
@@ -103,14 +115,17 @@ used with the THRESH_BINARY and THRESH_BINARY_INV thresholding types.
...
+the mean.
+@param binarizationMethod Binarization method to use. By default, Niblack's technique is used.
+Other techniques can be specified, see cv::ximgproc::LocalBinarizationMethods.
 
 @sa  threshold, adaptiveThreshold
  */
 CV_EXPORTS_W void niBlackThreshold( InputArray _src, OutputArray _dst,
                                     double maxValue, int type,
-                                    int blockSize, double delta );
+                                    int blockSize, double k, int binarizationMethod = 0 );
```
sovrasov: Please, replace 0 with `BINARIZATION_NIBLACK`.

sovrasov:
@saskatchewancatch good job! Please, fix warnings on [Windows platform](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_windows64/builds/10320/steps/compile%20release/logs/warnings%20%283%29) and squash your commits into one before merge.

saskatchewancatch:
Squashed and pushed. I think that should be it. 

Thanks @sovrasov for the feedback.

sovrasov:
:+1:

PR #1142, opened 2017-04-28
Fixed bugs after dnn refactoring. Added new tests for layers

arrybn:
### This pullrequest changes
Fixed some bugs after refactoring dnn. Added new tests, testdata is here: https://github.com/opencv/opencv_extra/pull/334

PR #1150, opened 2017-05-04
Fix incorrect memset in im2row

pengli:
should add out_row_offset when memset data_col,
otherwise it will clear wrong data in data_col and
cause running "example_dnn_caffe_googlenet" have
different result.

for example, without this fix

"example_dnn_caffe_googlenet" will return Probability: 99.6437%

with this fix

"example_dnn_caffe_googlenet" will return Probability: 99.6378%

and 99.6378% is the good result.

Signed-off-by: Li Peng <peng.li@intel.com>

### DIFF
--- a/modules/dnn/src/layers/op_im2col.cpp
+++ b/modules/dnn/src/layers/op_im2col.cpp
@@ -203,7 +203,7 @@ public:
             }
             else
             {
-                memset(data_col_, 0, kw*kh*channels*sizeof(data_col_[0]));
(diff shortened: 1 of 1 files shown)


mshabunin:
@pengli , in recently merged #1142 this `memset` line has been removed. I tried to run the latest version of the example and got 99.6378%. Can you please check whether this line is needed or not?

cc @arrybn

pengli:
@mshabunin, yes, remove "memset" could also fix this problem, I will close this PR.

PR #1151, opened 2017-05-04
Names of blank layers

dkurt:
### This pullrequest changes

Setting up names and types for BlankLayers (Dropout and Identity).


dkurt:
#1134

PR #1152, opened 2017-05-05
Enabled parallel computation of ReLU activation layer

dkurt:
### This pullrequest changes

Improve efficiency of networks with ReLU layers by doing it in parallel.

Motivated by another activation layers. With PR changes, efficiency of AlexNet improves on ~9.4%, SqueezeNet v1.1 on ~1.7%, ResNet-50 on ~2.5%.


vpisarev:
@dkurt, this is too low improvement. Let's fuse conv layer with PReLU, and then speedup will be much more sabstantial

PR #1156, opened 2017-05-11
Made separate functions for computing output shapes for all layers. R…

arrybn:
…emoved output blobs allocation from layers

### This pullrequest changes
Added functions for computing layers shapes without inference. Moved  creating blobs from layers to Net


arrybn:
@vpisarev please take a look. I've changed Net interface by adding new functions

vpisarev:
getOutShapes() has been introduced, but it's not used in the corresponding allocate() methods. Also, as we discussed, getOutShapes() should also return size of a temporary buffer. And also it should report whether inputs can be processed in-place.

arrybn:
@vpisarev I've made all changes, which you've mentioned in your comment above. Please take a look

vpisarev:
well, it's not quite what I'd like to see, but let's merge it in and then refine :+1:

PR #1162, opened 2017-05-15
dnn: extra testdata files from OPENCV_DNN_TEST_DATA_PATH env

alalek:


alalek:
@arrybn Please take a look on failures.

PR #1163, opened 2017-05-15
Fast F-transform added

tucna:
**Merge with extra:**

Update of the **fuzzy** module. New methods optimised for speed were added for roughly 10 times faster processing. 

- New methods in fuzzy module.
- Tests updated.
- Full Python support.

Related **opencv_extra** commit exists.


modules/fuzzy/src/fuzzy_F0_math.cpp
```diff
@@ -43,6 +43,262 @@
...
+    uchar *im_b = channel[0].data;
+
+    int width = imagePadded.cols;
+    int height = imagePadded.rows;
+    int n_width = width / radius + 1;
+    int n_height = height / radius + 1;
+
+    float *c_r = new float[n_width * n_height];
+    float *c_g = new float[n_width * n_height];
+    float *c_b = new float[n_width * n_height];
```
alalek: It is better to avoid manual memory management via "new/delete" or "malloc/free" calls. I didn't see "delete" calls here, so probably there is memory leak.
Please use std::vector<float> instead.

tucna: Oh, you are right about memory leak, thanks. Anyway, I am going to use `vector` and check a change in a performance.

tucna:
I have made several fixes. As @alalek proposed, I used `vector` (no negative impact on performance found). I left pure arrays only on places where they are used for conversion to `Mat`.

alalek:
:+1:

PR #1166, opened 2017-05-17
Removed an extra reallocation of networks every setBlob call.

dkurt:
### This pullrequest changes

Before PR changes, every `setBlob` call reallocated network. Despite the same shapes.


vpisarev:
:+1:

PR #1169, opened 2017-05-18
Sfm fixes

PhilLab:
Fixes compilation error with MSVC < 2015 and enables refinement of intrinsics

### DIFF
--- a/modules/sfm/src/libmv_light/libmv/logging/logging.h
+++ b/modules/sfm/src/libmv_light/libmv/logging/logging.h
@@ -23,6 +23,11 @@
 
 #include <glog/logging.h>
 
+
+#if defined _MSC_VER && _MSC_VER < 1900
+#  define snprintf _snprintf
+#endif
+
 #define LG LOG(INFO)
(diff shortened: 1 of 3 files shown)


sovrasov:
:+1:

PhilLab:
@edgarriba

LaurentBerger:
@PhilLab  ( and @edgarriba) with your PR I have got some problem to run example given [in doc ](http://docs.opencv.org/master/d4/d18/tutorial_sfm_scene_reconstruction.html#sagrada) because of [this line](https://github.com/opencv/opencv_contrib/commit/4f8a400d5d17bf80dc4f25d6f2699a9a93829cc1)
Without optimisation example is aable to reconstruct scene but reconstruction when parameters are refined.

I understand that you PR is usefull but I think a function must be added to sfm classes which can process parameters given in libmv_ReconstructionOptions libmv_CameraIntrinsicsOptions
My questions are : can you reproduce problem with example? What do you think about a new method?

Thanks for your answer

PR #1174, opened 2017-05-21
F1-transform added.

tucna:
New functionality for **fuzzy** module. F-transform of first degree (F1-transform) added for advanced computation with gradients. 

- Documentation updated.
- Full Python support.

**Related paper**

Vlašánek, Pavel, and Perfilieva, Irina. "The F-transform in Terms of Image Processing Tools." _Journal of Fuzzy Set Valued Analysis_ 2016.1 (2016): 54-62.

https://www.researchgate.net/publication/304708729_The_F-transform_in_Terms_of_Image_Processing_Tools

_Edit_: minor fixes in F0 added in the same manner as in F1.


modules/fuzzy/include/opencv2/fuzzy/fuzzy_F1_math.hpp
```diff
@@ -0,0 +1,152 @@
...
+    */
+    CV_EXPORTS_W void FT12D_components(InputArray matrix, InputArray kernel, OutputArray components);
+
+    /** @brief Computes elements of F1-transform components.
+    @param matrix Input array.
+    @param kernel Kernel used for processing. Function **createKernel** can be used.
+    @param c00 Elements represent average color.
+    @param c10 Elements represent average vertical gradient.
+    @param c01 Elements represent average horizontal gradient.
+    @param components Output 32-bit array for the components.
```
sovrasov: It's unclear what type of 32-bit array is assumed? It can be 32bit integer or 32bit float.

modules/fuzzy/include/opencv2/fuzzy/fuzzy_F1_math.hpp
```diff
@@ -0,0 +1,152 @@
...
+    /** @brief Computes F1-transfrom and inverse F1-transfrom at once.
+    @param matrix Input matrix.
+    @param kernel Kernel used for processing. Function **createKernel** can be used.
+    @param output Output 32-bit array.
+
+    This function computes F1-transfrom and inverse F1-transfotm in one step. It is fully sufficient and optimized for **Mat**.
+    */
+    CV_EXPORTS_W void FT12D_process(InputArray matrix, InputArray kernel, OutputArray output);
+
+    /** @brief Computes F1-transfrom and inverse F1-transfrom at once.
```
sovrasov: How it's possible to compute direct and inverse transform at once, should it have two output parameters for both transforms? Or type of transform depends on passed kernel?

modules/fuzzy/include/opencv2/fuzzy/fuzzy_F1_math.hpp
tucna: Thank you for the code review @sovrasov!

I will fix it to be more concrete.

modules/fuzzy/include/opencv2/fuzzy/fuzzy_F1_math.hpp
```diff
@@ -0,0 +1,152 @@
...
+    @param c10 Elements represent average vertical gradient.
+    @param c01 Elements represent average horizontal gradient.
+    @param components Output 32-bit array for the components.
+
+    The function computes components and its elements using predefined kernel.
+
+    @note
+        F-transform technique of first degreee is described in paper @cite Vlas:FT.
+    */
+    CV_EXPORTS_W void FT12D_polynomial(InputArray matrix, InputArray kernel, OutputArray c00, OutputArray c10, OutputArray c01, OutputArray components);
```
tucna: This is interesting. I did this because of problems with Python wrapper. Anyway, I tried now and everything works. I will fix it in F0 transform as well.

modules/fuzzy/include/opencv2/fuzzy/fuzzy_F1_math.hpp
tucna: This method is aimed to situations where you have an image and you want to get the output image only. It is faster than usage of the methods for direct FT and inverse FT after that.

So, if you are not interested in fuzzy components but just in the result, you can use this method.

sovrasov: I've understood, but description is a bit unclear.

tucna: I see. I would suggest to make some bigger changes in the documentation in separated PR. I would also like to add some tutorial.

sovrasov:
:+1:

PR #1176, opened 2017-05-23
added optimized anisotropic diffusion filter

vpisarev:
This is rewrite of https://github.com/opencv/opencv/pull/8362. Thanks to sav-chris for the reference implementation

### DIFF
--- a/modules/ximgproc/include/opencv2/ximgproc.hpp
+++ b/modules/ximgproc/include/opencv2/ximgproc.hpp
@@ -123,19 +123,19 @@ The function transforms a binary blob image into a skeletized form using the tec
 CV_EXPORTS_W void thinning( InputArray src, OutputArray dst, int thinningType = THINNING_ZHANGSUEN);
 
 /** @brief Performs anisotropic diffusian on an image.
- 
+
  The function applies Perona-Malik anisotropic diffusion to an image. This is the solution to the partial differential equation:
- 
+
  \f[{\frac  {\partial I}{\partial t}}={\mathrm  {div}}\left(c(x,y,t)\nabla I\right)=\nabla c\cdot \nabla I+c(x,y,t)\Delta I\f]
- 
+
  Suggested functions for c(x,y,t) are:
- 
+
  \f[c\left(\|\nabla I\|\right)=e^{{-\left(\|\nabla I\|/K\right)^{2}}}\f]
- 
+
  or
- 
+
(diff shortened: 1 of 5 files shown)


vpisarev:
:+1:

pmoulon:
Why the initial author owner/author @alalek  of the code is not in the git history?

alalek:
Because original patch was not adapted to OpenCV contribution guidelines. See here: https://github.com/opencv/opencv/pull/8362

Anyway author is [mentioned](https://github.com/opencv/opencv_contrib/pull/1176/files#diff-90c11f8b94f2cb2e61adeebe1434daa8R42) in the source code and this PR description.

simmplecoder:
Hi,

I've been trying to implement a strategy for Boost.GIL's anisotropic diffusion that would reproduce behavior of opencv, but I found some weird differences between source implementation and this PR. I know it has been a while, but it seems like this implementation diverges in many different ways.

1. The diagonal sums are not halved ([source](https://github.com/opencv/opencv/pull/8362/files#diff-6d8b3a1cb04f8b3e18c8e911b3a9fabaR77), [this PR](https://github.com/opencv/opencv_contrib/pull/1176/files#diff-90c11f8b94f2cb2e61adeebe1434daa8R177))

The halving is due to it being an approximation of Laplacian. If you have a look at section "Implementation via operator discretization" [here](https://en.wikipedia.org/wiki/Discrete_Laplace_operator), you will see the matrix

```
[ 0.25 0.5 0.25
  0.5  -3  0.5
  0.25 0.5 0.25]

which multiplied by two is

[ 0.5 1 0.5
  1  -6  1
  0.5 1 0.5]

2. The diffusivity function is different ([source](https://github.com/opencv/opencv/pull/8362/files#diff-6d8b3a1cb04f8b3e18c8e911b3a9fabaR54), [this PR](https://github.com/opencv/opencv_contrib/pull/1176/files#diff-90c11f8b94f2cb2e61adeebe1434daa8R262))

The new function makes the distribution lie in close neighborhood to 1, which makes most differences apply quickly even with low alpha.

3. Brightness function is different ([source](https://github.com/sav-chris/opencv/blob/67db4ba5ce56c51b96280f274d491c33e491bd16/modules/imgproc/src/peronamalik.cpp#L67), [this PR](https://github.com/opencv/opencv_contrib/pull/1176/files#diff-90c11f8b94f2cb2e61adeebe1434daa8R183))

The original treated channels separately, while this one seems to just sum the channels together.

----

I wanted to ask why those changes were made? Are they more correct? 

By the way, the exponent table is a great optimization. I've never seen the brightness function and diffusion function in use in any matlab versions I've seen so far and no book or material suggested such a function either.
```

PR #1179, opened 2017-05-24
Added statistics functions

arrybn:
### This pullrequest changes
Added statistics functions for models: counting FLOPS, getting all used layer types, getting count of layers with certain type, computing memory consumption


vpisarev:
perhaps, some examples and/or tests should be extended to check or demonstrate this functionality, e.g. print the memory consumption and gflops for the popular networks :+1:

PR #1183, opened 2017-05-24
Fix changes with allocating memory outside layers.

dkurt:
### This pullrequest changes

* Removed `allocate` method from Blank layer.
* Correct global average pooling output shape computation (in case of global pooling kernel size initializes properly in `finalize` that calls after all allocations).
* Removed a temporal Mat allocation from Reshape layer.

### DIFF
--- a/modules/dnn/src/layers/reshape_layer.cpp
+++ b/modules/dnn/src/layers/reshape_layer.cpp
@@ -149,6 +149,7 @@ public:
             outputs.push_back(MatShape());
             computeShapeByReshapeMask(inputs[i], newShapeDesc, newShapeRange, outputs.back());
         }
+        internals = outputs;
 
         return true;
     }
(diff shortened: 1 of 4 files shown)


vpisarev:
@arrybn, please, review

modules/dnn/src/layers/reshape_layer.cpp
```diff
@@ -169,14 +169,13 @@ class ReshapeLayerImpl : public ReshapeLayer
...
+            Mat& srcBlob = *inputs[i];
+            Mat& dstBlob = outputs[i];
+            MatShape inputShape = shape(srcBlob);
 
             if (performReordering)
             {
-                Mat reordered_blob(inputShape, srcBlob.type());
-
-                float *dstData = reordered_blob.ptr<float>();
+                float *dstData = dstBlob.ptr<float>();
```
arrybn: In some cases this line will lead to dstData and srcData point on the same memory (because layer is in-place). In can cause wrong results because scrData will be erased by writing into dstData and then will be read again. There are few ways to correct this: 
1. Reorder weights in fully-connected layer in case when there is a Reshape->FC consequence in tensorflow model; Reordering was implemented to handle such cases (Reshape->FC)
2. Check if reordering is needed in getMemoryShapes() function and return shape of input as shape for internal necessary blob; also it's possible to always request internal blob

dkurt: Chosen the second way of implementation. Thanks!
Also, modified reordering conditions.

arrybn:
:+1:

chrisrn:
Is this fix related to this error? :

The following error occurred while making allocate() for layer "Reshape": -dims() <= axis && axis < dims() in canonicalAxis, file /HOME/opencv_contrib/modules/dnn/include/opencv2/dnn/blob.inl.hpp, line 155

arrybn:
@chrisrn do you have this error now? if so, please give details

chrisrn:
I am using the dnn module to load a tensorflow model using tf_inception.cpp script. When the graph contains the reshape operation on the input image I have this error. If I remove this tensorflow operation and perform reshaping with numpy I don't have that problem. So if I export the tensorflow graph without using
```
tf.reshape(image, shape)
```
it works fine.

arrybn:
I think it's a tensorflow-specific issue ant probably it still exists in the latest master. I'll take a look

chrisrn:
During the forward pass the reshape operations work fine.

dkurt:
@chrisrn , I was wondering if you could apply changes from #1194 . If it doesn't working, please tell us about input and output shapes.

PR #1185, opened 2017-05-25
Changed blobFromImages() function, so it can handle 1-channel images

### LINKED ISSUE #1184: DNN module now can not work with 1-channel input images
- OpenCV => 3.2 (snaphot at 24.05.2017)

##### Detailed description

@arrybn, thanks to the new [dnn::blobFromImage](https://github.com/opencv/opencv_contrib/blob/master/modules/dnn/src/dnn.cpp#L71) implementation all networks with the single-channel input (data blob) can not work now! Can you bring back this possibility?

P.S. It is also starnge that if you pass 3-channel image into the dnn::blobFromImage then last 4-th layer of the output blob could contain arbitrary noise not related with the input image. So, why this 4-layers blobs are used?

##### Steps to reproduce
You can reproduce issue by means of this networks [protos](https://github.com/AlfredXiangWu/face_verification_experiment/blob/master/proto) and [caffemodels](https://github.com/AlfredXiangWu/face_verification_experiment/tree/master/model)

arrybn:
Resolves #1184

### DIFF
--- a/modules/dnn/src/dnn.cpp
+++ b/modules/dnn/src/dnn.cpp
@@ -70,25 +70,8 @@ static String toString(const T &v)
 
 Mat blobFromImage(const Mat& image_, double scalefactor, bool swapRB)
 {
-    Mat image;
-    if(image_.depth() == CV_8U)
-    {
-        image_.convertTo(image, CV_32F, scalefactor);
-    }
-    else
-        image = image_;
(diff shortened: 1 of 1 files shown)


arrybn:
@dkurt @vpisarev please review

dkurt:
:+1:

PR #1186, opened 2017-05-25
Added Halide backend support for deep learning layers

dkurt:
#### Implemented some layers using Halide language backend:

Layer | Covered all configurations | Test
---------------------------------|----------------------------|-----
Concat                           |          by channels (axis = 1)        | x
Convolution                      |                          x | x
Deconvolution                    |    no group factor    | x
(attachable) ReLU, TanH, Sigmoid, Abs, BNLL, Power, PReLU, Batch normalization, Scale | x | x
Element-wise addition, multiplication, maximum | x | x
Fully-connected | axis = 1 | x
LRN (across and within channels) | x | x
Maximum unpooling with indices support | x | x
Maximum pooling | x | x
Average pooling | only whole kernels, no padding | x
SoftMax | axis == 1, no spatial size more than 1x1 | x

Supported CPU only, float32 computations. No ahead-of-time compilation.
Added tutorial with installation guide and sample, tests for some layers and networks, performance tests.

#### Generalized manual scheduling of 5 networks into the layer-wise automatic scheduling.
Intel&reg; Core&trade; i7-6700K CPU @ 4.00GHz x 8

Architecture | Manual scheduling | Automatic scheduling | Ratio
------|------|------|------
AlexNet |            22.38ms |  25.5ms | x0.87
ResNet-50 | 73.91ms | 84.26ms | x0.87
SqueezeNet v1.1 | 8.21ms | 6.44ms | **x1.27**
Inception-5h | 37.06ms |  32.01ms | **x1.15**
ENet @ 3x512x256 | 41.21ms | 41.07ms| x1.00

For testing purpose have taken GoogLeNet network. L1 diff is 1.26789e-09. Efficiency is 33ms using Halide (automatic scheduling) against 34.89ms using MKL (x1.05).

TODO list:
- [x] Reproduce Halide installation by tutorial (Linux) https://github.com/halide/Halide/commit/abef1d9bf6cb3f866393fa4c5f48726f728285ee
- [ ] Reproduce Halide installation by tutorial (Windows): installation reproduced but SegFault appears (Some layers works with replacing vectorization by `x` variable to factor 4 (from 8 or 16). All right in case of no scheduling).
- [x] Cover more layers and configurations in tests
- [x] Write scheduling tutorial
- [x] Remove local paths
- [x] Figure out about `supportBackend` and layer configurations (axis, etc.)
- [x] Automatic scheduling with no more than 20% efficiency loss.

### DIFF
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -55,6 +55,23 @@ namespace dnn //! This namespace is used for dnn module functionlaity.
 
     typedef std::vector<int> MatShape;
 
+    /**
+     * @brief Enum of computation backends supported by layers.
+     */
+    enum Backend
+    {
+        DNN_BACKEND_DEFAULT,
+        DNN_BACKEND_HALIDE
+    };
+
+    /**
+     * @brief Enum of target devices for computations.
+     */
+    enum Target
+    {
+        DNN_TARGET_CPU
+    };
+
     /** @brief Initialize dnn module and built-in layers.
      *
      * This function automatically called on most of OpenCV builds,
@@ -75,6 +92,54 @@ namespace dnn //! This namespace is used for dnn module functionlaity.
 
         String name; //!< Name of the layer instance (optional, can be used internal purposes).
         String type; //!< Type name which was used for creating layer by layer factory (optional).
+    };
+
+   /**
+    * @brief Derivatives of this class encapsulates functions of certain backends.
+    */
+    class BackendNode
+    {
+    public:
+        BackendNode(int backendId);
+
+        virtual ~BackendNode(); //!< Virtual destructor to make polymorphism.
+
+        int backendId; //!< Backend identifier.
+    };
+
+    /**
+     * @brief Derivatives of this class wraps cv::Mat for different backends and targets.
+     */
+    class BackendWrapper
+    {
+    public:
+        BackendWrapper(int backendId, int targetId);
+
+        /**
+         * @brief Wrap cv::Mat for specific backend and target.
+         * @param[in] targetId Target identifier.
+         * @param[in] m cv::Mat for wrapping.
+         *
+         * Make CPU->GPU data transfer if it's require for the target.
+         */
+        BackendWrapper(int targetId, const cv::Mat& m);
+
+        /**
+         * @brief Make wrapper for reused cv::Mat.
+         * @param[in] base Wrapper of cv::Mat that will be reused.
+         * @param[in] shape Specific shape.
+         *
+         * Initialize wrapper from another one. It'll wrap the same host CPU
+         * memory and mustn't allocate memory on device(i.e. GPU). It might
+         * has different shape. Use in case of CPU memory reusing for reuse
+         * associented memory on device too.
+         */
+        BackendWrapper(const Ptr<BackendWrapper>& base, const MatShape& shape);
+
+        virtual ~BackendWrapper(); //!< Virtual destructor to make polymorphism.
+
+        int backendId;  //!< Backend identifier.
+        int targetId;   //!< Target identifier.
     };
 
     /** @brief This interface class allows to build new Layers - are building blocks of networks.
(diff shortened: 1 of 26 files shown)


modules/dnn/tutorials/tutorial_dnn_halide.markdown
```diff
@@ -0,0 +1,151 @@
...
+write image processing algorithms in well-readable format, schedule computations
+according to specific device and evaluate it with a quite good efficiency.
+
+An official website of the Halide project: http://halide-lang.org/.
+
+## Efficiency comparison
+Measured on Intel&reg; Core&trade; i7-6700K CPU @ 4.00GHz x 8.
+
+<html>
+  <table>
```
alalek: Please use markdown tables. [Example](https://github.com/opencv/opencv/blob/ec49eb813cbde55927d269a2e89a4dbb4a6b3262/modules/core/include/opencv2/core/hal/intrin_cpp.hpp#L183-L238). [Tutorial](https://help.github.com/articles/organizing-information-with-tables/)

modules/dnn/test/test_halide_nets.cpp
```diff
@@ -0,0 +1,120 @@
...
+// Copyright (C) 2017, Intel Corporation, all rights reserved.
+// Third party copyrights are property of their respective owners.
+
+namespace cvtest
+{
+
+using namespace cv;
+using namespace dnn;
+
+void loadNet(const std::string& weights, const std::string& proto,
```
alalek: please use "static" or "anonymous namespace" for internal functions.

BTW, Where is `#include` directive in this file?

modules/dnn/perf/perf_halide_net.cpp
```diff
@@ -0,0 +1,93 @@
...
+    net.setPreferableBackend(HALIDE); \
+    net.allocate(); \
+    net.applyHalideScheduler(scheduler); \
+    net.forward(net.getLayerId(outputLayer)); \
+    \
+    TEST_CYCLE_N(10) \
+    { \
+        net.forward(net.getLayerId(outputLayer)); \
+    } \
+    SANITY_CHECK_NOTHING();
```
dkurt: Thanks! Replaced to static function.

modules/dnn/perf/perf_halide_net.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+inline std::string getOpenCVExtraDir()
+{
+    return "/home/dkurtaev/opencv_extra/testdata/";
```
dkurt: Modified modules/dnn/perf/perf_main.cpp for it.

modules/dnn/CMakeLists.txt
```diff
@@ -54,11 +54,34 @@ if(NOT PROTOBUF_FOUND)
...
+# ----------------------------------------------------------------------------
+if(HALIDE_FOUND)
+  set(DNN_INCLUDE_DIRS
+    ${DNN_INCLUDE_DIRS}
+    ${HALIDE_INCLUDE_DIRS}
+  )
+  set(DNN_LIBRARIES
+    ${DNN_LIBRARIES}
+    ${HALIDE_LIBRARIES}
+  )
```
dkurt: Eliminated CMakeLists changes at all.

modules/dnn/tutorials/tutorial_dnn_halide.markdown
dkurt: Fixed, thanks!

modules/dnn/test/test_halide_nets.cpp
dkurt: It's working somehow. Should I add it for prevent fails in different build cases?

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -55,6 +55,23 @@ namespace dnn //! This namespace is used for dnn module functionlaity.
 
     typedef std::vector<int> MatShape;
 
+    /**
+     * @brief Enum of computation backends supported by layers.
+     */
+    enum Backend
+    {
+        DEFAULT,
```
vpisarev: I would rename them to DNN_BACKEND_DEFAULT, DNN_BACKEND_HALIDE, to make it more self-explanatory

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -55,6 +55,23 @@ namespace dnn //! This namespace is used for dnn module functionlaity.
...
+        HALIDE
+    };
+
+    /**
+     * @brief Compilation targets of Halide code.
+     */
+    enum HalideTarget
+    {
+        HALIDE_CPU
```
vpisarev: again, let's use DNN_TARGET_CPU, DNN_TARGET_GPU etc. I would not use the word "Halide" here, because other backends, like OpenVX or even the default one, can be potentially customized to use CPU, GPU or other device to run the networks

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -131,6 +148,47 @@ namespace dnn //! This namespace is used for dnn module functionlaity.
...
+        /**
+         * @brief Returns pointers to Halide functions.
+         * @param[in] inputs Pointers to input Halide buffers.
+         *
+         * Input buffers should be exactly the same that will be used in forward invocations.
+         * Despite we can use Halide::ImageParam based on input shape only,
+         * it helps prevent some memory management issues (if something wrong,
+         * Halide tests will be failed).
+         */
+        virtual std::vector<void*> initHalide(const std::vector<void*> &inputs);
```
vpisarev: instead of very low-level void* I would use some abstract class HalideFunc (that can encapsulate void*) with a smart pointer to it. It's comparatively low overhead, but it would be much more safe from memory management point of view.

dkurt: Introduced `BackendNode` and `BackendWrapper` whose derivatives store explicit `Halide::Func` and `Halide::Buffer` correspondingly. Moreover, it helps remove `Halide::Func` members from layers as it was before.

vpisarev:
ok, let's merge it in while it's fresh! :+1:

alalek:
This PR is rely on https://github.com/opencv/opencv/pull/8794 . Need to review/merge them too.

PR #1194, opened 2017-05-29
Fix reshape layer

dkurt:
### This pullrequest changes

Added more complicated tests for Torch reshape layer in https://github.com/opencv/opencv_extra/pull/340 .
Fixed layer more carefully process `batch_size = 1` and cases with reshape without batching.


vpisarev:
:+1:

PR #1196, opened 2017-05-29
Fuzzy module - Tutorials and Documentation

tucna:
Update of the **fuzzy** module. I have added three tutorials with basics and usage explanation. I have also updated documentation and made one minor fix. Content:

- tutorial with basic theory,
- tutorial about inpainting,
- tutorial about filtering,
- documentation updated,
- minor fix about multiplication order.

Guys, one question - I do not know the process of `contrib -> root` delivery, but I would like to extend existing OpenCV `inpaint` method to be able to use also **FT** algorithms (not only `TELEA` and `NS`). I guess it is not good idea to do it now because of  `root -> contrib` dependency, am I right?


tucna:
There is a warning in **Docs** build:

__stat" returned: (33206, 6030656, 64771, 1, 100000, 100000, 1753926, 1496054846, 1496054846, 1496054846)
File size: 1753926 bytes = 1712 KiB = 1 MiB__

Unfortunately, I have no clue how to prevent it. Can you give me any advice?

alalek:
@tucna Please convert some images into ".jpg" format (like fuzzy_filt_output.png)

sovrasov:
@tucna thanks for the great job!
Is the quality of the source images for tutorials very important? May be convert them into jpeg and leave `fuzzy_BF_view.png` and `fuzzy_pixel_view.png` as they are?

tucna:
Thank you @sovrasov and @alalek. I changed all images to jpeg, their quality is not so important.

tucna:
@sovrasov, @alalek 

Guys, my question is maybe not so visible in the first post:

>  I do not know the process of contrib -> root delivery, but I would like to extend existing OpenCV inpaint method to be able to use also FT algorithms (not only TELEA and NS). I guess it is not good idea to do it now because of root -> contrib dependency, am I right?

Should I add FT methods to the current OpenCV inpainting method, or better not yet?

sovrasov:
Such dependency is not a good idea. It's better to add a new `inpaint' in xphoto module. But fuzzy dependency should be optional for xphoto.

tucna:
I see. Maybe all `fuzzy_image` can be kicked out to `xphoto` and **fuzzy module** can be used just for pure computations of related methods.

I will think about that and prepare a new PR hypothetically.

modules/fuzzy/include/opencv2/fuzzy/fuzzy_F0_math.hpp
```diff
@@ -53,68 +53,64 @@ namespace ft
     //! @addtogroup f0_math
     //! @{
 
-    /** @brief Computes components of the array using direct F0-transform.
+    /** @brief Computes components of the array using direct \f$F^0\f$-transform.
     @param matrix Input array.
-    @param kernel Kernel used for processing. Function **createKernel** can be used.
+    @param kernel Kernel used for processing. Function `createKernel` can be used.
```
tucna: Thank you, fixed.

sovrasov:
:+1:

PR #1207, opened 2017-06-02
Fixed a buffer overflow which occurs when setImage is called with dif…

natsuki14:
…ferent images and default image_id.

### This pullrequest changes

Modified a conflict between comments ("If the image_id is not equal to -1 and the same as the previous call for setImage, computations are used again") and codes.
I believe the comment is correct because of a buffer overflow problem.


sovrasov:
:+1:

PR #1212, opened 2017-06-05
Import SoftMax & LogSoftMax layers from Torch

dkurt:
**Merge with**: https://github.com/opencv/opencv_extra/pull/341

### This pullrequest changes
Classification networks trained in Torch use criterion that expected output from `SoftMax` and espetially from `LogSoftMax`: https://github.com/torch/nn/blob/master/doc/criterion.md.
From my point of view, the most popular criterion for multi-class training is a [ClassNLLCriterion](https://github.com/torch/nn/blob/master/doc/criterion.md#classnllcriterion).
SoftMax and LogSoftMax are monotonic functions that means predicted class (maximum of output) is the same if we exclude these computations. Thanks for Torch framework it's so simple doing by [replace layer to nn.Identity](https://github.com/torch/nn/blob/master/doc/module.md#replacefunction) or [remove it at all](https://github.com/torch/nn/blob/65f42723f51d13683a431329452e5f385988e984/doc/containers.md#removeindex).
Currently DNN's torch importer doesn't import SoftMax and LogSoftMax layers. Despite it's not so critical for model deployment, we'd better to cover it.

### DIFF
--- a/modules/dnn/include/opencv2/dnn/all_layers.hpp
+++ b/modules/dnn/include/opencv2/dnn/all_layers.hpp
@@ -251,6 +251,8 @@ namespace dnn
     class CV_EXPORTS SoftmaxLayer : public Layer
     {
     public:
+        bool logSoftMax;
+
         static Ptr<SoftmaxLayer> create(const LayerParams& params);
     };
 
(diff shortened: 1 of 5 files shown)


dkurt:
@arrybn , @vpisarev , could you review it please?

arrybn:
:+1:

PR #1214, opened 2017-06-06
Fixed SSD example, added test

arrybn:
### This pullrequest changes
Made SSD sample work again!
Updated caffe.proto file to support the latest ssd models
Fixed all things which cause differences between dnn and caffe outputs for SSD
Added test for SSD


vpisarev:
excellent, thank you! :+1:

vpisarev:
@arrybn, looks like you forgot to create pull request for opencv_extra

PR #1222, opened 2017-06-12
Fixed while loop in buildPyramid in deepflow.cpp

### LINKED ISSUE #1218: Minor Bug in Deep Flow
Hi

In the member function [buildPyramid](https://github.com/opencv/opencv_contrib/blob/master/modules/optflow/src/deepflow.cpp#L95) of deepflow,  the variable "i" is never incremented in the while loop hence the while loop never terminates as a result of the termination condition instead it terminates as a result of the minSize breaking condition.

Best Regards!

ceroytres:
resolves #1218 

### This pullrequest changes:
The while loop in member function buildPyramid in deepflow.cpp terminated as a result of the break statement and never as a result of the termination criteria because the "i" variable in the while loop was never incremented.  The proposed change is to simply add a line that increments "i". Also, this is the first time I submit a pull request :).

### DIFF
--- a/modules/optflow/src/deepflow.cpp
+++ b/modules/optflow/src/deepflow.cpp
@@ -97,8 +97,7 @@ std::vector<Mat> OpticalFlowDeepFlow::buildPyramid( const Mat& src )
     std::vector<Mat> pyramid;
     pyramid.push_back(src);
     Mat prev = pyramid[0];
-    int i = 0;
-    while ( i < this->maxLayers )
+    for( int i = 0; i < this->maxLayers; ++i)
     {
         Mat next; //TODO: filtering at each level?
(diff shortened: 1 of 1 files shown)


StevenPuttemans:
I am just wondering if it then would not be better to replace the while loop with an internal increment, into a limited for loop that increments `i`? Certainly if `i` is never used.

ceroytres:
I agree with you completely.

sovrasov:
@ceroytres please fix trailing whitespaces [here](https://github.com/opencv/opencv_contrib/pull/1222/commits/ecf5a1cf3775a9c33ac5bc86d593deac2fe64e07#diff-f525c17eb61dbdd971cf42ef2f5136bfR100) and squash your commits into one before merge.

sovrasov:
:+1:

PR #1224, opened 2017-06-13
added parallel convolution layer that includes parallel im2row and bias term addition

vpisarev:
**WIP**
merge after https://github.com/opencv/opencv/pull/8907


modules/dnn/src/layers/layers_common_opt_avx.cpp
```diff
@@ -0,0 +1,199 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
```
alalek: Please use this name to enable AVX compilation flags: layers_common.avx.cpp

BTW, we have short version of license header.

modules/dnn/src/layers/layers_common.hpp
```diff
@@ -65,12 +65,12 @@ void getConvPoolPaddings(const Size& inp, const Size& out,
                          const Size &kernel, const Size &stride,
                          const String &padMode, Size &pad);
 
-#if defined __clang__ && (defined __i386__ || defined __x86_64__)
-#define CV_DNN_TRY_AVX 1
+#if CV_SSE2
```
alalek: Current guard to access AVX2 optimized functions is:
```
#if !defined CV_DISABLE_OPTIMIZATION && defined CV_ENABLE_INTRINSICS && \
    (defined CV_CPU_COMPILE_AVX2 || defined CV_CPU_DISPATCH_COMPILE_AVX2)
```
(see failure of Linux Debug builder - `.avx2.cpp` is not processed at all)

Currently dispatching calls are done via `CV_CPU_CALL_AVX2` macro (function should be under "opt_AVX2" namespace). But this approach doesn't fit well into large functions (this method contains 200+ lines).
[Example with single dispatched code path](https://github.com/opencv/opencv/blob/f49f056d29e71ce0a48cfa4d8ea82681da52990c/modules/core/src/convert.cpp#L4600).
[Example with multiple dispatched paths](https://github.com/opencv/opencv/blob/master/modules/core/src/mathfuncs_core.dispatch.cpp) ([PR](https://github.com/opencv/opencv/pull/8441))

Some updates are prepared here: https://github.com/opencv/opencv/pull/8899/files

PR #1229, opened 2017-06-16
Added possibility of getting any intermediate blobs

arrybn:
### This pullrequest changes
Added possibility of getting correct intermediate blobs. Added test. Fixed few bugs with several sequent forwards.


modules/dnn/src/dnn.cpp
```diff
@@ -1670,26 +1831,5 @@ Ptr<Layer> LayerFactory::createLayerInstance(const String &_type, LayerParams& p
     }
 }
 
-BackendNode::BackendNode(int backendId) : backendId(backendId) {}
```
dkurt: These classes are base classes for `HalideBackendNode` and `HalideBackendWrapper`. If Halide backend enabled, implementations undefined.

modules/dnn/src/dnn.cpp
```diff
@@ -598,27 +605,72 @@ struct Net::Impl
...
+                                   DNN_TARGET_CPU);
+            }
+        }
+    }
+
+    void setUpNet(const std::vector<LayerPin>& blobsToKeep = std::vector<LayerPin>())
+    {
+        if (!netWasAllocated || this->blobsToKeep != blobsToKeep)
+        {
+            if (!netWasAllocated )
```
dkurt: I think it should be after `initBackend()`.

arrybn: Changed order

modules/dnn/src/dnn.cpp
arrybn: Returned, it was deleted due to bad merge

arrybn:
All tests have been passed, only check of files size to commit to opencv_extra is failed

mshabunin:
:+1:

PR #1230, opened 2017-06-16
aruco: fix non-square marker size in _drawPlanarBoardImpl

### LINKED ISSUE #1226: aruco::CharucoBoard::draw() crashes with specific image size

- OpenCV => 3.2.0-dev
- Operating System / Platform => Linux Ubuntu 16.04 - 64 bit
- Compiler => GCC 5.4.0

##### Detailed description

When calling `cv::aruco::CharucoBoard::draw()` using specific board and image sizes, code crashes with `OpenCV Error: Assertion failed (!fixedSize() || ((Mat*)obj)->size.operator()() == Size(_cols, _rows)) in create`.
Specifically, I've been able to reproduce the problem when creating a 7x5 board and then drawing it at size 1600x1200 pixels.

##### Steps to reproduce

```
#include <opencv2/opencv.hpp>
#include <opencv2/aruco.hpp>
#include <opencv2/aruco/charuco.hpp>
int main(int argc, char **argv)
{
  int squares_x = 7,
      squares_y = 5;

  int bwidth = 1600,
      bheight = 1200;

  cv::Ptr<cv::aruco::Dictionary> dict = cv::aruco::getPredefinedDictionary(cv::aruco::DICT_4X4_50);
  cv::Ptr<cv::aruco::CharucoBoard> board = cv::aruco::CharucoBoard::create(squares_x, squares_y, 1.0, 0.75, dict);
  cv::Size sz(bwidth, bheight);
  cv::Mat mat;
  board->draw(sz, mat, 0, 1);
  cv::imshow("test", mat);
  cv::waitKey(-1);
  return 0;
}
```
This code triggers the bug.
Changing the size of the board to, e.g. 9x7, or changing the size of the image to, eg. 1600x1201, does not trigger the bug.

##### Possible fix

I've managed to trace the bug to `cv::aruco::_drawPlanarBoardImpl()` at line 1683 (https://github.com/opencv/opencv_contrib/blob/master/modules/aruco/src/aruco.cpp#L1683):
```
if((outCorners[0].y == outCorners[1].y) && (outCorners[1].x == outCorners[2].x)) {
            // marker is aligned to image axes
            marker.copyTo(out(Rect(outCorners[0], dst_sz)));
            continue;
}
```
When using the parameters above, `dst_sz == Size(172, 171)` whereas `marker.size() == Size(172,172)`, which causes the crash.
I managed to fix it simply by adding the condition that `marker.size() == dst_sz`. It seems to work correctly, but I'm not expert enough on the matter to know whether it's the right fix.

sovrasov:
resolves #1226

### DIFF
--- a/modules/aruco/src/aruco.cpp
+++ b/modules/aruco/src/aruco.cpp
@@ -1676,6 +1676,7 @@ void _drawPlanarBoardImpl(Board *_board, Size outSize, OutputArray _img, int mar
 
         // get marker
         Size dst_sz(outCorners[2] - outCorners[0]); // assuming CCW order
+        dst_sz.width = dst_sz.height = std::min(dst_sz.width, dst_sz.height); //marker should be square
         dictionary.drawMarker(_board->ids[m], dst_sz.width, marker, borderBits);
 
         if((outCorners[0].y == outCorners[1].y) && (outCorners[1].x == outCorners[2].x)) {
--- a/modules/aruco/test/test_misc.cpp
+++ b/modules/aruco/test/test_misc.cpp
@@ -1,0 +1,24 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#include "test_precomp.hpp"
+#include <opencv2/aruco/charuco.hpp>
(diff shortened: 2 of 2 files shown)


vpisarev:
thanks! :+1:

PR #1234, opened 2017-06-19
Fix implicit mask case in reshape layer.

dkurt:
### This pullrequest changes
Added one more test for reshape layer. Made changes require for correct output shape of layer. Before PR `{1, 128, 4, 4}` with mask `{-1, 2048}` reshaped to `{1, 1, 2048}`.


arrybn:
:+1:

PR #1236, opened 2017-06-19
Added another overloads of blobFromImage functions 

arrybn:
### This pullrequest changes
Added blobFromImage functions which can crop input from the center and subtract mean

### DIFF
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -598,9 +598,37 @@ namespace dnn //! This namespace is used for dnn module functionlaity.
      *  @warning This function has the same limitations as createTorchImporter().
      */
     CV_EXPORTS_W Mat readTorchBlob(const String &filename, bool isBinary = true);
-
-    CV_EXPORTS Mat blobFromImage(const Mat& image, double scalefactor=1.0, bool swapRB=true);
-    CV_EXPORTS Mat blobFromImages(const std::vector<Mat>& image, double scalefactor=1.0, bool swapRB=true);
+    /** @brief Creates 4-dimensional blob from image. Optionally resizes and crops @p image from center,
+     *  subtract @p mean values, scales values by @p scalefactor, swap Blue and Red channels.
+     *  @param image input image (with 1- or 3-channels).
+     *  @param size spatial size for output image
+     *  @param mean scalar with mean values which are subtracted from channels. Values are intended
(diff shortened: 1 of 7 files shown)


modules/dnn/src/dnn.cpp
```diff
@@ -154,6 +155,38 @@ Mat blobFromImages(const std::vector<Mat>& images, double scalefactor, bool swap
...
+            size = imgSize;
+        float resizeFactor = std::max(size.width / (float)imgSize.width,
+                                      size.height / (float)imgSize.height);
+        images[i].convertTo(images[i], CV_32F);
+        resize(images[i], images[i], Size(), resizeFactor, resizeFactor);
+        Rect crop(Point(0.5 * (images[i].cols - size.width),
+                        0.5 * (images[i].rows - size.height)),
+                  size);
+        Scalar mean = mean_;
+        if (swapRB)
```
dkurt: It seems to me that with `swapRB == true` we'll do it twice. In this function and in origin `blobFromImages`.

arrybn: here swap is only for mean values Scalar, not for input image. there's no double swapping

dkurt: Yeah, thanks, I see.

dkurt:
:+1:

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -598,9 +598,55 @@ namespace dnn //! This namespace is used for dnn module functionlaity.
...
+     *  @param mean scalar with mean values which are subtracted from channels. Values are intended
+     *  to be in (mean-R, mean-G, mean-B) order if @p image has BGR ordering and @p swapRB is true.
+     *  @param scalefactor multiplier for @p image values.
+     *  @param swapRB flag which indicates that swap first and last channels
+     *  in 3-channel image is necessary.
+     *  @details input image is resized so one side after resize is equal to corresponing
+     *  dimension in @p size and another one is equal or larger. Then, crop from the center is performed.
+     *  @returns 4-dimansional Mat with NCHW dimensions order.
+     */
+    CV_EXPORTS_W Mat blobFromImage(const Mat& image, const Size& size,
```
vpisarev: can we just make the `size` parameter optional too? (`const Size& size=Size()`) and reduce the number of overloaded methods?

arrybn: Removed 2 overloads

vpisarev:
:+1:

PR #1237, opened 2017-06-19
some further optimizations and cleanups in dnn

vpisarev:
**WIP**
+ got rid of dnn::gemm; it's not perf critical anymore (perhaps)
+ embedded col2im functionality into convolution_layer.cpp, since it's not used anywhere else
+ parallel max pooling. even better performance can be achieved if we knew that max indices are not needed (and they are not needed in most networks)
+ somewhat optimized deconvolution layer: optimized bias addition (merged it with col2im), optimized col2im slightly.
+ hopefully fixed incorrect memory access in fully-connected layer; restored aligned memory reads (they should work fine now)


dkurt:
Best observed efficiency (no Lapack/MKL):
Intel® Core™ i7-6700K CPU @ 4.00GHz x 8

|Model| before PR | with PR |
|---------|--------------|------------|
|GoogLeNet| 34.91ms|25.73ms|
|AlexNet| 16.55ms| 15.51ms|
|ResNet-50|62.72ms|57.11ms|
|SqueezeNet v1.1| 10.32ms|8.11ms|
|Inception-5h| 36.79ms|25.63ms|
|ENet @ 3x512x256| 115ms (82ms with MKL)|109.05ms (both with and without MKL)|

vpisarev:
@dkurt, thank you for the analysis! I will try to fix the regressions in ENet performance

PR #1243, opened 2017-06-21
Get rid of exit() calls in algorithms and tests

### LINKED ISSUE #1238: throw exception instead of calling exit()
### Version

- opencv: master
- opencv_contrib: master

### Description

I noticed a couple of places in library code with hard process termination instead of throwing exceptions. Those `exit()` calls should be replaced with `cv::error`, probably using one of the convenient macros `CV_Assert` or `CV_Error`:

- [`cv::plot::Plot2dImpl::Plot2dImpl`][1] constructors in `modules/plot/src/plot.cpp`
- [`cv::ppf_match_3d::writePLY`][2] and [`cv::ppf_match_3d::writePLYVisibleNormals`][3] in `modules/surface_matching/src/ppf_helpers.cpp`
- [`EDLineDetector::nfa`][4] (inside `cv::line_descriptor::BinaryDescriptor`) in `modules/line_descriptor/include/opencv2/line_descriptor/descriptor.hpp`

[1]: https://github.com/opencv/opencv_contrib/blob/master/modules/plot/src/plot.cpp#L67
[2]: https://github.com/opencv/opencv_contrib/blob/master/modules/surface_matching/src/ppf_helpers.cpp#L121
[3]: https://github.com/opencv/opencv_contrib/blob/master/modules/surface_matching/src/ppf_helpers.cpp#L174
[4]: https://github.com/opencv/opencv_contrib/blob/master/modules/line_descriptor/include/opencv2/line_descriptor/descriptor.hpp#L752

sovrasov:
resolves #1238

### DIFF
--- a/modules/line_descriptor/include/opencv2/line_descriptor/descriptor.hpp
+++ b/modules/line_descriptor/include/opencv2/line_descriptor/descriptor.hpp
@@ -747,10 +747,7 @@ class CV_EXPORTS BinaryDescriptor : public Algorithm
 
       /* check parameters */
       if( n < 0 || k < 0 || k > n || p <= 0.0 || p >= 1.0 )
-      {
-        std::cout << "nfa: wrong n, k or p values." << std::endl;
-        exit( 0 );
-      }
+        CV_Error(Error::StsBadArg, "nfa: wrong n, k or p values.\n");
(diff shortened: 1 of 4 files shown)


vpisarev:
thank you! :+1:

PR #1246, opened 2017-06-21
Halide OpenCL target for deep learning networks

dkurt:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/345

### This pullrequest changes
Added OpenCL target for Halide.
One more interface function: `setPreferableTarget`. Default target is CPU.

| Model | Manual scheduling | Auto scheduling |
|---:|---:|---:|
| GoogLeNet | | 93.66ms|
| AlexNet | 49ms | 63ms (x0.77)|
|ResNet-50| 189ms| 190.32ms (x0.99) |
| SqueezeNet v1.1| 15.61ms | 20.05ms (x0.77)|
| Inception-5h | 89.52ms | 98.43ms (x0.90) |
| ENet @ 512x256 | 39.67ms | 42.12ms (x0.94)|

TODO list:
- [ ] Select GPU device by id

### DIFF
--- a/modules/dnn/perf/perf_halide_net.cpp
+++ b/modules/dnn/perf/perf_halide_net.cpp
@@ -41,114 +41,131 @@ static void loadNet(std::string weights, std::string proto, std::string schedule
 
     net->setInput(blobFromImage(input, 1.0, false));
     net->setPreferableBackend(DNN_BACKEND_HALIDE);
+    net->setPreferableTarget(targetId);
     net->setHalideScheduler(scheduler);
     net->forward(outputLayer);
 }
 
+////////////////////////////////////////////////////////////////////////////////
+// CPU target
+////////////////////////////////////////////////////////////////////////////////
 PERF_TEST(GoogLeNet, HalidePerfTest)
 {
-    try {
-        Net net;
-        loadNet("dnn/bvlc_googlenet.caffemodel", "dnn/bvlc_googlenet.prototxt",
-                "", 227, 227, "prob", "caffe", DNN_TARGET_CPU, &net);
-
-        TEST_CYCLE_N(10)
-        {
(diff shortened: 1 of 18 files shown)


modules/dnn/perf/perf_halide_net.cpp
```diff
@@ -150,6 +154,115 @@ PERF_TEST(ENet, HalidePerfTest)
...
+// OpenCL target
+////////////////////////////////////////////////////////////////////////////////
+PERF_TEST(GoogLeNet_opencl, HalidePerfTest)
+{
+    try {
+        Net net;
+        loadNet("dnn/bvlc_googlenet.caffemodel", "dnn/bvlc_googlenet.prototxt",
+                "", 227, 227, "prob", "caffe", DNN_TARGET_OPENCL, &net);
+
+        TEST_CYCLE_N(10)
```
alalek: What is about `TEST_CYCLE()`?

modules/dnn/perf/perf_halide_net.cpp
```diff
@@ -150,6 +154,115 @@ PERF_TEST(ENet, HalidePerfTest)
...
+        {
+            net.forward();
+        }
+        SANITY_CHECK_NOTHING();
+    } catch (SkipTestException& e) {
+        throw PerfSkipTestException();
```
alalek: Perhaps we should unify behavior for these exceptions (via patch into opencv repository).

modules/dnn/perf/perf_halide_net.cpp
dkurt: Thanks! Replaced. It runs the same 10 iterations by default.

vpisarev:
thank you! :+1:

PR #1249, opened 2017-06-22
Updated Halide pipelines

dkurt:
### This pullrequest changes
* Replaced `blobFromImage` in Halide tests.
* Removed Halide buffer allocation in Batch norm layer.
* Create fused weights in convolutional layer separately from origin weights. In example, deconvolution's `weightsMat` creates by `transpose` with data copying. So, in this PR added `clone` call in convolution layer in the same way.


vpisarev:
:+1:

PR #1239, opened 2017-06-20
face: refactoring of factory methods

sovrasov:
resolves #1235

PR #1251, opened 2017-06-24
face: minor cleanup after #1239

berak:
### This pullrequest changes
* change some leftover createXXXRecognizer mentioning in docs / examples / header comments
* to make Algorithm::load() work without any loss, the threshold member variable should be serialized, too.


sovrasov:
:+1:

berak:
as an afterthought,  implementation of Algorithm::empty() was added, too, since currently there is no way to check, if loading succeeded, or the model was trained successfully before

PR #1199, opened 2017-05-30
Face alignment

sukhad-app:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/346
**Merge with 3rdparty**: https://github.com/opencv/opencv_3rdparty/pull/24

## Overview

- Proposal: [Face alignment using OpenCV.](https://summerofcode.withgoogle.com/projects/#6093777644027904)

- Blog:[ A complete report of work done.](https://sukhad007.wordpress.com/)

- Tutorial blog: [A tutorial to use the described functionality](https://sukhad-app.github.io/final/) 

- Mentor : Steven Puttemans @StevenPuttemans 

- Student: Sukhad Anand @sukhad-app
 

## Contribution of this pull request

- This pull request will add a new functionality of face_alignment to OpenCV.
- Face alignment is a computer vision technology for identifying the geometric structure of human faces in digital images. Given the location and size of a face, it automatically determines the shape of the face components such as eyes and nose.
- The main aim of the project is an implementation of an excellent paper from 2014 Computer Vision and Pattern Recognition Conference: [One Millisecond Face Alignment with an Ensemble of Regression Trees by Vahid Kazemi and Josephine Sullivan ](http://www.cv-foundation.org/openaccess/content_cvpr_2014/papers/Kazemi_One_Millisecond_Face_2014_CVPR_paper.pdf)
- The pull request adds one-millisecond face alignment to the [face](https://github.com/sukhad-app/opencv_contrib/tree/face_alignment/modules/face) module of OpenCV library.

### Applications added
- [Training face landmark detector](https://github.com/sukhad-app/opencv_contrib/blob/face_alignment/modules/face/samples/sample_train_landmark_detector.cpp)

- [Face landmark detection in an image](https://github.com/sukhad-app/opencv_contrib/blob/face_alignment/modules/face/samples/sampleDetectLandmarks.cpp)

- [ Face landmark detection in a video](https://github.com/sukhad-app/opencv_contrib/blob/face_alignment/modules/face/samples/sampleDetectLandmarksvideo.cpp)

- [Face swapping](https://github.com/sukhad-app/opencv_contrib/blob/face_alignment/modules/face/samples/sample_face_swapping.cpp)

### Results

- [
image
](https://www.youtube.com/embed/ZtaV07T90D8)
- [Results](https://sukhad007.wordpress.com/)
- [Tutorial](https://sukhad-app.github.io/final/)

 All discussion regarding the project can be found at OpenCV face alignment slack channel:
[...]

PR #1257, opened 2017-06-27
GSOC17 - Facemark API

kurnianggoro:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/365

Project: API for Facial Landmark Detector
Mentor: Delia Passalacqua

This pull request contains the implementation of Facemark API as well as two facemark algorithm, AAM and LBF.

video: https://www.youtube.com/watch?v=B7WGyhl2zm8
report: https://gist.github.com/kurnianggoro/74de9121e122ad0bd825176751d47ecc

Currently, there are several interfaces provided in the Facemark base class including:
- load training data
- set custom face detector
- face detection
- training the model
- saving a trained model
- loading a trained model
- fitting

For the AAM algorithm, all of its training functionalities are already coded.
Several related functionalities are listed as follows:
 - Procrustes analysis
 - PCA projection
 - customized Delaunay triangulation
 - warp image
 - Gram-Schmidth orthonormalization
 - feature extraction


sergboec:
@kurnianggoro  Do you have plans to add implementation of algorithm from dlib ?

alalek:
@kurnianggoro What is the license of images from opencv_extra patch?

kurnianggoro:
hi @alalek , I just found them on google, will it be problem?

alalek:
test_stereo has sporadic failures - build restarted.

All images should have known proper usage rights. It is better to reuse existed OpenCV images, or use Google [filtering](https://support.google.com/websearch/answer/29508?hl=en) (but anyway double check is required).

kurnianggoro:
@alalek hi, i have changed the file and provide the source links (the license is available,you can check), they are from wikipedia, so i hope it will not be problem

vpisarev:
@kurnianggoro, let's not put faces of public people (or not public, it does not matter) to OpenCV. lena is the exception, 'cause it's considered a standard test image.

modules/face/src/facemarkLBF.cpp
```diff
@@ -0,0 +1,1294 @@
...
+        if(params.verbose) printf("loading data from : %s\n", s.c_str());
+        std::ifstream infile;
+        infile.open(s.c_str(), std::ios::in);
+        if (!infile) {
+           std::string error_message = "No valid input file was given, please check the given filename.";
+           CV_Error(CV_StsBadArg, error_message);
+        }
+
+        FILE *fd = fopen(s.c_str(), "rb");
+        regressor.read(fd, params);
```
alalek: OpenCV modules should not read/write files directly, especially via old "C-API" (there are a lot of missing error checks at least; there is no in-memory storage support; cross-platform issues).
Why FileStorage doesn't work here?

modules/face/src/facemarkLBF.cpp
```diff
@@ -0,0 +1,1294 @@
...
+        std::random_shuffle(bboxes.begin(), bboxes.end());
+        std::srand(seed);
+        std::random_shuffle(current_shapes.begin(), current_shapes.end());
+
+
+        regressor.initRegressor(params);
+        regressor.trainRegressor(imgs, gt_shapes, current_shapes, bboxes, mean_shape, 0, params);
+
+        FILE *fd = fopen(params.model_filename.c_str(), "wb");
+        assert(fd);
```
alalek: `assert()` should not be used in OpenCV, use `CV_Assert()` instead.

modules/face/src/facemarkLBF.cpp
```diff
@@ -0,0 +1,1294 @@
...
+                imgs[idx] = data_faces[i];
+                gt_shapes[idx] = data_shapes[i];
+                bboxes[idx] = data_boxes[i];
+                current_shapes[idx] = data_boxes[i].reproject(data_boxes[k].project(data_shapes[k]));
+            }
+        }
+
+        // random shuffle
+        unsigned int seed = (unsigned int)std::time(0);
+        std::srand(seed);
```
alalek: `rand()` is not thread-safe - can't provide reproducible results (for testing).
Consider to use cv::rand() , cv::theRNG() instead.

modules/face/src/liblinear.hpp
```diff
@@ -0,0 +1,3968 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifdef F2C_COMPAT
+
+void cdotc_(fcomplex *dotval, int *n, fcomplex *cx, int *incx,
+            fcomplex *cy, int *incy);
+
+void cdotu_(fcomplex *dotval, int *n, fcomplex *cx, int *incx,
```
alalek: Using of external libraries should be avoided. Especially in case of duplicated functionality.
Why does `cv::ml::LogisticRegression` not work for you?

modules/face/test/test_facemark_lbf.cpp
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    EXPECT_NO_THROW(facemark->training());
+}
+
+TEST(CV_Face_FacemarkLBF, can_load_model) {
+    string model_filename = "LBF.model";
+
+    Ptr<Facemark> facemark = FacemarkLBF::create();
+    EXPECT_NO_THROW(facemark->loadModel(model_filename.c_str()));
```
alalek: Each test should be standalone and don't rely on results of other tests.
Also tests should not write any data (models) if it is not required.

Probably we should put model somewhere or we can merge train/load tests into one.

kurnianggoro:
@vpisarev the test function needs at least 2 images, so i put mark zuckerberg and downey johnson photos from wikipedia (the license allows it, CCA 2.0 and CCA 3.0), will that be a problem? also regarding to the lena image, in other library it was removed due to copyright matter, i am not sure why opencv still have it. check it here https://github.com/torch/nn/issues/854 also in wikipedia it is stated that lena is non free image https://en.wikipedia.org/wiki/File:Lenna.png

modules/face/src/facemarkLBF.cpp
kurnianggoro: according to the documentation http://docs.opencv.org/3.2.0/d2/de8/group__core__array.html#ga75843061d150ad6564b5447e38e57722 theRNG function will use different random generator for each thread, but in this case the program only needs to randomize the order of training sample, which should produce same random order for each sample category (i.e. the randomized order or image sample should corresponds to the randomized order of the annotation). If reproducible results is expected, then i think we just need to set the seed value to a fixed number, or just remove this shuffling part (i don't think it will produce significantly different results)

modules/face/src/facemarkLBF.cpp
kurnianggoro: ok, i will try to change to filestorage.

modules/face/src/liblinear.hpp
kurnianggoro: I tried to remove the liblinear dependency but was not successful (using gradient descent regression and simple regression w^T x = y). The LogisticRegression result is in range 0~1, isn't it? I considered to use this function before but worried that the result will be different to the original version (i need to normalize the target values, i.e. the landmark points, into 0~1 range,maybe by dividing the values by width and height of input image, and not sure about the result) and the deadline was approaching, so i did not test LogisticRegression. But let me try now...

modules/face/test/test_facemark_lbf.cpp
kurnianggoro: merging the train-load test is easier, i can put a flag for not saving the model.

modules/face/src/facemarkLBF.cpp
```diff
@@ -0,0 +1,1294 @@
...
+
+        cascade_face = "../data/haarcascade_frontalface_alt.xml";
+        shape_offset = 0.0;
+        n_landmarks = 68;
+        initShape_n = 10;
+        stages_n=5;
+        tree_n=6;
+        tree_depth=5;
+        bagging_overlap = 0.4;
+        model_filename = "LBF.model";
```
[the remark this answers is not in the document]
kurnianggoro: this is just the default filename for the saving file, is it also problem? if it were for reading file like the cascade face then for sure it will be problem

modules/face/src/facemarkLBF.cpp
alalek: `cv::theRNG` is thread-safe (maintains own state for each thread).
Seed initialization from `time()` should be removed from algorithm implementations - seed should be controlled externally.

Regarding this exact algorithm, I believe it should work without shuffling of input data.

modules/face/src/facemarkLBF.cpp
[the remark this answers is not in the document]
alalek: This is minor problem (at least there is no relative path to subdirectory).

modules/face/src/facemarkLBF.cpp
alalek: Well, there is similar PR #1199 with same [problem](https://github.com/opencv/opencv_contrib/pull/1199/files#diff-bda6c2019eab834caf1e9453102e8025). Probably we should keep this as is if there is some problem with `cv::FileStorage`.
I believe it is fine for the first version. So consider this problem as minor too.

modules/face/src/liblinear.hpp
alalek: OK. Thank you for information.

alalek:
AAM tests are crashed in the "Debug" [build](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_linux64_no_opt/builds/10037). Please take a look.

modules/face/src/facemarkAAM.cpp
```diff
@@ -560,8 +560,8 @@ namespace face {
         reduce(Ys,sumYs, 0, CV_REDUCE_SUM);
 
         //calculate the normrnd
-        double normX = sqrt(sumXs.at<float>(0)+sumXs.at<float>(1));
-        double normY = sqrt(sumYs.at<float>(0)+sumYs.at<float>(1));
+        double normX = sqrt(Mat(sumXs.reshape(1)).at<float>(0)+Mat(sumXs.reshape(1)).at<float>(1));
```
kurnianggoro: @alalek the error in build debug is fixed here, but why it is not checked anymore?

alalek: This build is optional, it is not triggered automatically.
Debug build is passed, thanks!

modules/face/test/test_facemark_aam.cpp
```diff
@@ -0,0 +1,177 @@
...
+        image = imread(images_train[i].c_str());
+        EXPECT_TRUE(loadFacePoints(points_train[i].c_str(),landmarks));
+        EXPECT_TRUE(landmarks.size()>0);
+        EXPECT_TRUE(facemark->addTrainingSample(image, landmarks));
+    }
+
+    EXPECT_NO_THROW(facemark->training());
+}
+
+TEST(CV_Face_FacemarkAAM, can_load_model) {
```
alalek: The same note about standalone tests.

kurnianggoro:
Hi @alalek , I already fixed all the problems you've mentioned, but the building bot failed due to stereo module, can you please restart the building? also regarding the liblinear depency, I've tried to use opencv's logistric regression and neural network but both of them are not compatible. For the log-res, it seems like that the algorithm automatically assign categories to the training data and the number of resulted categories are inconsistent across different training data. As for the neural network, it do not support linear activation, I tried to train but the results are nan. So, I grab the support vector regression code from the liblinear, it just one function, not too long. I've tested it and the result is good. You can try to train using the Helen dataset if you want. The tutorial is here http://pullrequest.opencv.org/buildbot/export/pr_contrib/1257/docs/d7/dec/tutorial_facemark_usage.html but make sure your system has around 9GB of RAM and just wait 45in~1hour for the for the training. I've tried to train on smaller dataset but the results are not good.

alalek:
@kurnianggoro Thank you! Update looks good to me.

alalek:
The current issue is related to test images. We can't accept such risk.
Could you please replace these images? For example, from free datasets (but the most part of datasets is not free/restricted/missing license), or very old photos.

modules/face/src/facemarkLBF.cpp
```diff
@@ -1191,6 +1181,185 @@ namespace face {
...
+            }
+            index[i] = i;
+        }
+
+        while(iter < max_iter){
+            Gmax_new = 0;
+            Gnorm1_new = 0;
+
+            for(i=0; i<active_size; i++){
+                int j = i+rand()%(active_size-i);
```
alalek: Please avoid using of `rand()` - it doesn't provide reproducible results.

Like "stereo" test (it uses `rand()` too), which fails time-to-time.

kurnianggoro:
hi @alalek i have changed the test images using frames taken from the tracking module test video, I hope it will not be a problem now.

alalek:
Thank you! Looks good to me :+1:

Macbull:
Is there a python wrapper of this API in any release version?
Thanks

PR #1268, opened 2017-07-06
face: add a toplevel node to write(String) method

### LINKED ISSUE #1267: FaceRecognizer load() method is missing from Java wrapper

- OpenCV =>3.3.0
- Compiler =>Visual Studio 2017

##### Detailed description
The `load(String filename) `is missing from OpenCV Java wrapper, so unable to load the trained data from HD.

Note: I have tried the` read(String filename)`, but it is not working. Throws the following exception:

> OpenCV Error: Bad argument (This LBPH model is not computed yet. Did you call the train method?) in cv::face::LBPH::predict, file E:\opencv_contrib\modules\face\src\lbph_faces.cpp, line 396
> CvException [org.opencv.core.CvException: cv::Exception: E:\opencv_contrib\modules\face\src\lbph_faces.cpp:396: error: (-5) This LBPH model is not computed yet. Did you call the train method? in function cv::face::LBPH::predict
> ]

**Edit:** 
I have a test, which I trained the images with the new version of OpenCV (3.3.0-rc) and again load the trained data by the previous version (3.2.0) the problem still existed, I think the problem can be with the train() method also, because if the data was trained well, then the 3.2.0 version can successfully load the data. And also the load() method is missed totally two bugs till now I have detected....

berak:
resolves #1267

### This pullrequest changes

* add a toplevel node to write(String) method, so it does the same as the save(String) method.
(nessecary to preserve compatibility with pretrained models from older versions)
* add a unit test for this
* minor correction to loading the threshold value, make sure to overwrite the default only IF there's a valid entry in the file (older versions might not have this)

### DIFF
--- a/modules/face/src/face_basic.cpp
+++ b/modules/face/src/face_basic.cpp
@@ -53,7 +53,10 @@ cv::Mat BasicFaceRecognizer::getMean() const
 void BasicFaceRecognizer::read(const FileNode& fs)
 {
     //read matrices
-    fs["threshold"] >> _threshold;
+    double _t = 0;
+    fs["threshold"] >> _t; // older versions might not have "threshold"
+    if (_t !=0)
+        _threshold = _t;    // be careful, not to overwrite DBL_MAX with 0 !
     fs["num_components"] >> _num_components;
     fs["mean"] >> _mean;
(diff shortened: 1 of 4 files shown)


sovrasov:
:+1:

PR #1270, opened 2017-07-10
Update doc and method String getDefaultName()

LaurentBerger:
Just a pr to update doc see http://answers.opencv.org/question/26777/createlbphfacerecognizer-radius-parameter/ and method getDefaultName which is not defined in FaceRecognizer fisher, eigen or lbph


sovrasov:
:+1:

PR #1276, opened 2017-07-14
Example for plot module

LaurentBerger:
A sample for plot module : draw histogram

### DIFF
--- a/modules/plot/samples/plot_histo.cpp
+++ b/modules/plot/samples/plot_histo.cpp
@@ -1,0 +1,102 @@
+#include <opencv2/core.hpp>
+#include <opencv2/highgui.hpp>
+#include <opencv2/imgproc.hpp>
+#include <opencv2/plot.hpp>
+#include <iostream>
+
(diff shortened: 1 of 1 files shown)


sturkmen72:
it looks nice
but i did not understand
X=252
Y=12335 -> what does it mean ?
histogram_1

LaurentBerger:
I think you remember http://answers.opencv.org/question/73233/how-do-you-plot-graphs-in-opencv-projects/#80870. @berak started a pr but he closed it. I wrote an example how to use this module and now I think a lot of work is necessary to really use this module...

berak:
the numbers are [the coords of the last point of the curve](https://github.com/opencv/opencv_contrib/blob/master/modules/plot/src/plot.cpp#L330-L331) , whatever the author needed that for.

sovrasov:
:+1:

alalek:
@LaurentBerger Thank you! Could you please rebase your patch before merge? (to eliminate "merge" commit)

Issue #961, opened 2017-01-19
some issues with plotting module


darkdragon-001:
I identified multiple issues while trying to run the following code

	// generate data
	vector<double> x{0,1,2,3};
	vector<double> y{.0,.2,.5,1.5};

	// plot
	Mat plotMat;
	Ptr<plot::Plot2d> plot;
	plot = plot::createPlot2d(x, y);
	plot->render(plotMat);
	plot->setMaxY(1.);
	plot->setMinY(-1.);
	imshow("MTF", plotMat);
	imwrite("./plot.jpg", plotMat);

Issues
1. Only `double` is allowed, no automatic implicit conversion possible for easy plotting
2. `setMinY()` and `setMaxY()` do not have any effects
3. Grid lines should be distributed equally between full integers
4. Y axis seems to go down instead of going up as it is common for plotting

sovrasov:
@darkdragon-001 try to call `setMinX()` and `setMinY()` before `render` to fix issue 2.

alalek:
Feel free to contribute useful changes for OpenCV community via pull requests.

StevenPuttemans:
First of all, nice to see that people actually intend to use the plotting module. As said by @alalek fixes or suggested changes always welcome!

> Only double is allowed, no automatic implicit conversion possible for easy plotting

Agreed that floating point or integer should be valid also. Adjusting the pipeline for this should be a minimal effort. You up for it?

> Y axis seems to go down instead of going up as it is common for plotting

Well OpenCV has always used a downwards pointing y axis for all it's operations, and I remember the original discussion, where we decided to keep it like this to be uniform in the library. However, you could add an overload function to `plot::createPlot2d(x, y);`, something like  `plot::createPlot2d(x, y, plotOrientation);`. This might be some more work, but in my eyes the best approach possible.

> The display of the last value ("X = ", "Y = ") is useless and should be disabled by default.

Agreed, however, I am unsure if we need to keep binary compatibility in contrib modules, if so, we cannot simply change this behaviour I am afraid. @alalek can you get back on this?

> Support construction via Nx2 matrix

Agreed, again an overload method could do the trick here. I am happily to take a look at this and provide a PR if you want!

> Support construction via Nx1x2 matrix (two channel Mat - Mat with data type Vec2x)

Same here.

> Ability to add multiple axis to one plot (comparing multiple data sets)

[...]

PR #1278, opened 2017-07-16
plot: fix yaxis and grid

berak:
### This pullrequest changes
* make y axis point up
* fix the grid (snap to 1, 0.1, etc)
* change text rendering to show min/max values at resp. axes
* make it possible to reuse the drawing image for multiple curve plots

this hopefully fixes issues 3,4,5, and feature 4 in #961


StevenPuttemans:
Thanks but, if we revert axis directions, than for me, that kind of is unnatural. Also it will get users, who are already used to this, end up with broken code?

berak:
@StevenPuttemans , yes, that's  true, and i'm also a bit unsure here. (let's see, if we can get more comments)

StevenPuttemans:
We can just add thumbs up and down :+1:  :-1:

modules/plot/src/plot.cpp
```diff
@@ -62,9 +61,17 @@ namespace cv
                 Mat _plotData = plotData.getMat();
                 //if the matrix is not Nx1 or 1xN
                 if(_plotData.cols > 1 && _plotData.rows > 1)
-                    CV_Error(Error::StsBadArg, "ERROR: Plot data must be a 1xN or Nx1 matrix.\n");
+                {
+                    std::cout << "ERROR: Plot data must be a 1xN or Nx1 matrix." << std::endl;
+                    exit(0);
```
alalek: Please keep `CV_Error` here and below.

berak: ah, true. no real reason to change it

berak:
i'm somewhat not happy with this, it does not quite deliver:

* ppl will *still* get confused, what's up or down. 
* having a nice grid is nice, but if it "auto-snaps" to the data, it introduces a lot of jitter, if used real-time (like in @LaurentBerger 's histogram demo
* appending multiple plots to the same image still requires to decide on ranges beforehand, and setting them for each run, so not real usable, atm.

PR #1284, opened 2017-07-20
Added python bindings for rgbd module

Piggelinus:
### This pullrequest changes

Exposes RGBD Odometry classes and functions to Python

### DIFF
--- a/modules/rgbd/include/opencv2/rgbd.hpp
+++ b/modules/rgbd/include/opencv2/rgbd.hpp
@@ -56,13 +56,13 @@ namespace rgbd
    * a limit. For a float/double, we just check if it is a NaN
    * @param depth the depth to check for validity
    */
-  CV_EXPORTS
+  CV_EXPORTS_W
   inline bool
   isValidDepth(const float & depth)
   {
     return !cvIsNaN(depth);
   }
-  CV_EXPORTS
+  CV_EXPORTS_AS(isValidDepth2)
   inline bool
   isValidDepth(const double & depth)
   {
@@ -81,7 +81,7 @@ namespace rgbd
     return (depth != std::numeric_limits<unsigned short int>::min())
         && (depth != std::numeric_limits<unsigned short int>::max());
   }
-  CV_EXPORTS
+  CV_EXPORTS_AS(isValidDepth3)
   inline bool
   isValidDepth(const int & depth)
   {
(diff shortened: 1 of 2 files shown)


berak:
hmm, did you check, if this is producing working python bindings locally on your box ?
(i'd think, it would need `static Ptr<MyClass>::create()` functions here, not public class constructors)

the [whitespace issue](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_docs/builds/10771/steps/whitespace%20opencv_contrib/logs/stdio) should be easy to fix.

Piggelinus:
Yeah, I've got RgbdOdometry working on correctly on the TUM freiburg1_xyz dataset in Python 2.7 on Ubuntu 16.04.

Do you mean adding a static Ptr<MyClass>::create() method to all classes instead of wrapping the constructors directly? That does indeed seem like the way it's done in other modules.

berak:
> working on correctly on the TUM

oh, apologies for the noise, then, misread it.

Piggelinus:
Should I make the changes I mentioned?

modules/rgbd/include/opencv2/rgbd.hpp
```diff
@@ -56,13 +56,13 @@ namespace rgbd
...
    * @param depth the depth to check for validity
    */
-  CV_EXPORTS
+  CV_EXPORTS_W
   inline bool
   isValidDepth(const float & depth)
   {
     return !cvIsNaN(depth);
   }
-  CV_EXPORTS
```
sovrasov: Do we really need `isValidDepth*` functions to be exported? I think `nan` checks can be done in Python.

modules/rgbd/include/opencv2/rgbd.hpp
```diff
@@ -81,7 +81,7 @@ namespace rgbd
     return (depth != std::numeric_limits<unsigned short int>::min())
         && (depth != std::numeric_limits<unsigned short int>::max());
   }
-  CV_EXPORTS
```
sovrasov: `isValidDepth3` is a strange name, it's better to use something like `isValidDepthInt` to make code readable.

modules/rgbd/include/opencv2/rgbd.hpp
```diff
@@ -238,44 +243,46 @@ namespace rgbd
...
     ~DepthCleaner();
 
+    CV_WRAP static Ptr<DepthCleaner> create(int depth, int window_size = 5, int method = DepthCleaner::DEPTH_CLEANER_NIL);
+
     /** Given a set of 3d points in a depth image, compute the normals at each point.
      * @param points a rows x cols x 3 matrix of CV_32F/CV64F or a rows x cols x 1 CV_U16S
      * @param depth a rows x cols matrix of the cleaned up depth
      */
-    void
+    CV_PROP void
```
sovrasov: As far as I know `CV_PROP` allows to wrap simple public data members like
`CV_PROP int a;`
does this operator really works from python?

Piggelinus: No, I see what you mean. I couldn't get it to work with CV_WRAP either so I'll remove the bindings for operator().

sovrasov:
:+1:

PR #1291, opened 2017-07-24
Protobuf parser module

dkurt:


vpisarev:
let's integrate it finally! :+1:

dkurt:
I think we can close it. Perhaps we can implement a parser of text protobuf messages by FileStorage.

PR #1297, opened 2017-07-27
ximgproc: change M_PI to CV_PI in structured_edge_detection

### LINKED ISSUE #1273: 'M_PI' undefined Error, compiling ximgproc module for Visual Studio 2017 on Windows

- OpenCV => 3.2.0
- Operating System / Platform => Windows 7 64 bit

##### Detailed description

I am trying to compile `opencv_contrib 3.3.0 rc` on Vistual Studio 2017 for Windows 64-bit system. I am building for Win32 though. It's giving me an error while building module `ximgproc`. The error is `identifier "M_PI" is undefined` in `opencv_contrib/modules/ximgproc/src/structured_edge_detection.cpp`. I checked online if it can be fixed and yes, I found a way to fix it. As per [Visual Studio 2017 documentation ](https://docs.microsoft.com/en-us/cpp/c-runtime-library/math-constants) for Math constants, we need to add `#define _USE_MATH_DEFINES` before including `cmath` (for C++) or `math.h` (for C). So I edited the `structured_edge_detection.cpp` and added `#define _USE_MATH_DEFINES` before all include statements. And I was able to build complete `opencv` successfully. 

Earlier, a somewhat similar [issue](https://github.com/opencv/opencv_contrib/issues/853 ) was found and fixed in a different way for a different module. I don't know if the same solution can be applied here as well. If not, the solution I provided worked fine for me.

##### Steps to reproduce

One can reproduce the error following the general instructions for building opencv 3.2.0 with opencv_contrib 3.3.0 rc

berak:
resolves #1273

### DIFF
--- a/modules/ximgproc/src/structured_edge_detection.cpp
+++ b/modules/ximgproc/src/structured_edge_detection.cpp
@@ -581,8 +581,8 @@ public:
       {
           int xysign = -((oxy[i] > 0) - (oxy[i] < 0));
           o[i] = (atan((oyy[i] * xysign / (oxx[i] + 1e-5))) > 0) ? (float) fmod(
-                  atan((oyy[i] * xysign / (oxx[i] + 1e-5))), M_PI) : (float) fmod(
-                  atan((oyy[i] * xysign / (oxx[i] + 1e-5))) + M_PI, M_PI);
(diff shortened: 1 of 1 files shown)


sovrasov:
:+1:

PR #1298, opened 2017-07-27
Update line descriptor samples

sturkmen72:
see http://answers.opencv.org/question/169646

### DIFF
--- a/modules/xfeatures2d/samples/shape_transformation.cpp
+++ b/modules/xfeatures2d/samples/shape_transformation.cpp
@@ -2,6 +2,7 @@
  * shape_context.cpp -- Shape context demo for shape matching
  */
 #include <iostream>
+#include "opencv2/opencv_modules.hpp"
 
 #ifdef HAVE_OPENCV_SHAPE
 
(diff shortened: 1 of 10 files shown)


modules/xfeatures2d/samples/shape_transformation.cpp
```diff
@@ -2,10 +2,10 @@
  * shape_context.cpp -- Shape context demo for shape matching
  */
 #include <iostream>
+#include "opencv2/shape.hpp"
```
sovrasov: Why has it been moved outside of `#ifdef HAVE_OPENCV_SHAPE`?

sturkmen72: @sovrasov we need to determine `HAVE_OPENCV_SHAPE` is defined or not. in actual code `HAVE_OPENCV_SHAPE` is always undefined. by this change it works as it should be

alalek: Use `#include "opencv2/opencv_modules.hpp"` for proper `HAVE_OPENCV_SHAPE` definition.

sovrasov:
:+1:

sturkmen72:
thanks, i will rebase commits later asap

sturkmen72:
Commits are rebased

PR #1303, opened 2017-07-30
Method: Convert Image To Highly Customizable String

ramadawn:
See Readme.md for details of module.

Optimized with Cython


alalek:
Probably this PR would be rejected:
- It doesn't integrated with OpenCV (it is fully standalone code).
- Purpose of this code is not clear - we already have cv::FileStorage and support wide range of popular image codecs.
- OpenCV source code is licensed under BSD. GPL code is not applicable due license conflict
- Patch includes binary blobs: `itsl.cpython-36m-x86_64-linux-gnu.so` and huge parts of generated code (about 10k lines)
- missing tests

ramadawn:
In your opinion what would I need to do to get it accepted?

StevenPuttemans:
> In your opinion what would I need to do to get it accepted?

1. It should be integrated into OpenCV. It should prove to be more efficient than the current cv::FileStorage interface at least, to be considered a contribution.
2. It cannot contain a single framework, or make use of any, that is not under a BSD license. 
3. Avoid binary blobs that just cripple the repository.
4. At least at tests for this functionality.

PR #1307, opened 2017-08-01
plot: make the module more customizable

### LINKED ISSUE #1306: ideas about improving plot module
test code

```
#include <opencv2/highgui.hpp>
#include <opencv2/plot.hpp>
#include <iostream>

using namespace cv;

{
	Mat data_x(1, 11, CV_64F);
	Mat data_y(1, 11, CV_64F);

	for (int i = 0; i < 11; i++)
	{
		data_x.at<double>(0, i) = i - 5;
		data_y.at<double>(0, i) = i - 5;
	}

	std::cout << "data_x : " << data_x << std::endl;
	std::cout << "data_y : " << data_y << std::endl;

	Mat plot_result0, plot_result1, plot_result2;

	Ptr<plot::Plot2d> plot = plot::createPlot2d(data_x, data_y);
	plot->render(plot_result0);
	plot->setPlotSize(500, 500);
	plot->render(plot_result1);

	//plot->setPlotAxisColor( Scalar(0, 0, 0) );
	plot->setPlotGridColor( Scalar(0, 0, 0) );
	plot->setPlotTextColor( Scalar(0, 0, 0) );
	plot->render(plot_result2);

	imshow("plot0", plot_result0);
	imshow("plot1", plot_result1);
	imshow("plot2", plot_result2);
	waitKey();

	return 0;
}
```

# 1
output plot0 (PlotSize is 600x400 default)

plot0

output plot1 (PlotSize is 500x500)
plot1

two graphs with same data and different size. i think there is a problem to be solved.

# 2

actually by setting `PlotGridColor`,`PlotTextColor` etc. you can do some customization on display like
plot2

probably adding some more functions to customization like `setShowGrid()`, `setShowText()` etc.

probably adding showGrid showText

sovrasov:
related to  #1306

- fix wrong grid drawing when plot is not a square
- add `showGrid` and `showText` flags
- add `gridLinesNumber` parameter
- move `create` functions into `Plot2d` scope.

### DIFF
--- a/modules/plot/include/opencv2/plot.hpp
+++ b/modules/plot/include/opencv2/plot.hpp
@@ -84,23 +84,33 @@ namespace cv
             CV_WRAP virtual void setPlotGridColor(Scalar _plotGridColor) = 0;
             CV_WRAP virtual void setPlotTextColor(Scalar _plotTextColor) = 0;
             CV_WRAP virtual void setPlotSize(int _plotSizeWidth, int _plotSizeHeight) = 0;
+            CV_WRAP virtual void setShowGrid(bool needShowGrid) = 0;
+            CV_WRAP virtual void setShowText(bool needShowText) = 0;
+            CV_WRAP virtual void setGridLinesNumber(int gridLinesNumber) = 0;
+            /**
+             * @brief Sets the index of a point which coordinates will be printed on the top left corner of the plot (if ShowText flag is true).
+             *
+             * @param pointIdx index of the required point in data array.
+             */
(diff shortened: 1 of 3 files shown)


LaurentBerger:
Thanks for PR. Can you add a variable cursorPos to use at [line ](https://github.com/opencv/opencv_contrib/blob/master/modules/plot/src/plot.cpp#L192-L193) and replace NumVecElements-1 by cursorPos and add a method to set cursorPos

alalek:
:+1:

PR #1314, opened 2017-08-07
Load SED model from binary file.

csukuangfj:
Fix issue: https://github.com/opencv/opencv_contrib/issues/1313

### This pullrequest changes

merge with https://github.com/opencv/opencv_extra/pull/362

### DIFF
--- a/modules/ximgproc/src/structured_edge_detection.cpp
+++ b/modules/ximgproc/src/structured_edge_detection.cpp
@@ -69,7 +69,7 @@ static cv::Mat imresize(const cv::Mat &src, const cv::Size &nSize)
 }
 
 /*!
- * The function filters src with triangle filter with radius equal rad
+ * The function filters src with triangle filter with radius rad
  *
  * \param src : source image to filter
  * \param rad : radius of filtering kernel
@@ -256,7 +256,7 @@ static void gradientHist(const cv::Mat &src, cv::Mat &magnitude, cv::Mat &histog
 }
 
 /*!
- * The class parallelizing the edgenms algorithm.
+ * The class parallelises the edgenms algorithm.
  *
  * \param E : edge image
  * \param O : orientation image
@@ -403,15 +403,8 @@ Ptr<RFFeatureGetter> createRFFeatureGetter()
         return makePtr<RFFeatureGetterImpl>();
 }
 
-}
-}
 
 /********************* StructuredEdgeDetection class *********************/
-
-namespace cv
-{
-namespace ximgproc
-{
 
 class StructuredEdgeDetectionImpl : public StructuredEdgeDetection
 {
@@ -419,7 +412,9 @@ public:
     /*!
      * This constructor loads __rf model from filename
      *
-     * \param filename : name of the file where the model is stored
+     * \param filename : name of the file where the model is stored. It can have an extension
+     *                   `.bin`, `.yml` or `.yml.gz`. The `.bin` format is loaded much faster than
+     *                   the `.yml` format.
      */
     StructuredEdgeDetectionImpl(const cv::String &filename,
         Ptr<const RFFeatureGetter> _howToGetFeatures)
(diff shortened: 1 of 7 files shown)


modules/ximgproc/src/structured_edge_detection.cpp
```diff
@@ -810,6 +830,171 @@ class StructuredEdgeDetectionImpl : public StructuredEdgeDetection
...
+
+        fclose(f);
+    }
+
+    /**
+     * Save model to a file.
+     * \param filename  [in] Model file name.
+     * \param rf        [in] Model parameters.
+     */
+    void saveModelToBinaryFile(const String &filename, const struct RandomForest &rf)
```
sovrasov: This method doesn't seem to be used anywhere and accessible form the public API. Why do you need it?

modules/ximgproc/src/structured_edge_detection.cpp
```diff
@@ -419,7 +419,16 @@ class StructuredEdgeDetectionImpl : public StructuredEdgeDetection
     /*!
      * This constructor loads __rf model from filename
      *
-     * \param filename : name of the file where the model is stored
+     * Refer to <https://github.com/opencv/opencv_extra/blob/master/testdata/cv/ximgproc/model.yml.gz>
```
sovrasov: This description should be visible from the generated documentation, so it's better to place it 
before `createStructuredEdgeDetection`. And also `create` functions are usually satic methods of corresponding classes (template methods of Algorithm class supposes that). Could you please move `createStructuredEdgeDetection` inside `StructuredEdgeDetection` and rename it to `create`?

csukuangfj: Hi, thank you for the comments. I have added a static `create` method to the class instead of
moving the old one, as some code may depend on that function. Some usage code demos
are added to the comments.

modules/ximgproc/src/structured_edge_detection.cpp
csukuangfj: Yes, you are right. I used this method initially to convert the the existing text format model to binary.
Now a Matlab script is used to do this job. This method is removed in the new commit.

modules/ximgproc/src/structured_edge_detection.cpp
sovrasov: Ok, but please add before the declaration of the old create function `CV_DEPRECATED` specifier.

csukuangfj: Done

csukuangfj: I have changed all the code in  opencv_contrib to use the new `create` API, because
the `CV_DEPRECATED` results in compilation error in the buildbot for python wrappers.
The old API is now deleted.

sovrasov:
:+1:

modules/ximgproc/src/structured_edge_detection.cpp
```diff
@@ -810,6 +815,93 @@ class StructuredEdgeDetectionImpl : public StructuredEdgeDetection
...
+        uint64_t len_;
+        ret = fread(&len_, sizeof(uint64_t), 1, f);
+        len = (size_t)len_;
+        rf.childs.resize(len);
+        ret = fread(rf.childs.data(), sizeof(int), len, f);
+
+        ret = fread(&len_, sizeof(uint64_t), 1, f);
+        len = (size_t)len_;
+        rf.featureIds.resize(len);
+        ret = fread(rf.featureIds.data(), sizeof(int), len, f);
```
alalek: We definitely should avoid such code for custom formats.

Consider to use BASE64 encoding or HDF5 binary formats.

csukuangfj: @alalek 
I find that you tried to add HDF5 support to OpenCV (https://github.com/opencv/opencv/compare/master...alalek:hdf5)

But it seems that the work is stopped. Maybe I could reuse your work.

alalek: There is hdf module in opencv_contrib: https://github.com/opencv/opencv_contrib/tree/3.3.0/modules/hdf

csukuangfj:
Anyone who is interested in this work are welcome. I will continue with this issue after November if it is not fixed by then.

csukuangfj:
Finally, I come back :smile: 

I have compared the running time for using YML format and HDF format.
HDF is about 20x faster than YML, though the model file size
is increased from 6.1MB (zipped) to 34 MB.

Here is the gist file for the test
<https://gist.github.com/csukuangfj/2cbe9374a28df0f2d709c88325357a53>

To run the gist:
 - Use this pull request <https://github.com/opencv/opencv_contrib/pull/1500>
 - Download the model file from <https://github.com/opencv/opencv_extra/blob/master/testdata/cv/ximgproc/model.yml.gz>

 - Delete line <https://github.com/opencv/opencv_contrib/blob/master/modules/hdf/src/hdf5.cpp#L579>

 - Delete line <https://github.com/opencv/opencv_contrib/blob/master/modules/hdf/src/hdf5.cpp#L697>

Here is the output of the code

```
$ ./sed
Start to read yml file....
Finish reading yml file in 2.192 seconds
Start to generate hdf file....
Finish generating hdf file in 0.367 seconds
Start to read hdf file....
Finish reading hdf file in 0.108 seconds
$ ls -lh model.yml.gz model.h5 
-rw-r--r-- 1 xx xx  34M Jan  2 15:45 model.h5
-rw-r--r-- 1 xx xx 6.1M Jan  2 15:13 model.yml.gz
```

If you think this patch is useful, I will close this pull request and start another pull request
to work on this. Since the current master branch has changed a lot, I would like to work on
the latest commit and do not want to merge the commits and solve the conflicts.


@alalek

csukuangfj:
Some algorithms use their own binary format for models, such as
<https://github.com/opencv/opencv_contrib/blob/master/modules/face/src/getlandmarks.cpp#L54>.

Maybe there would be a lot of work to convert existing models if custom formats are not allowed.

The current YML format in OpenCV is only suitable for text files with only small amounts of data,
otherwise, it will spend a lot of time on parsing it. 

Introduce a binary data format for OpenCV would be a good idea.

PR #1318, opened 2017-08-11
bioinspired: improve samples

sovrasov:


alalek:
:+1:

amroamroamro:
@sovrasov, @alalek
The use of the new variable `_wasOCLRunCalled` from ca7683e is not quite right, as you can get failed assertions in certain cases.

Assume that opencv was built with OpenCL enabled (`HAVE_OPENCL` is defined), then call `Retina::run` with a regular `Mat` as input image (not `UMat`). According to the implementation, it will run the ocl version, and sets `_wasOCLRunCalled = true`:

```cpp
#ifdef HAVE_OPENCL
bool RetinaImpl::ocl_run(InputArray inputMatToConvert)
{
    _ocl_retina->run(inputMatToConvert);
    _wasOCLRunCalled = true;
    return true;
}
#endif

void RetinaImpl::run(InputArray inputMatToConvert)
{
    CV_OCL_RUN((_ocl_retina != 0), ocl_run(inputMatToConvert));
    _wasOCLRunCalled = false;

    //...
}
```

However when you next try to retrieve the retina outputs with `Retina::getParvo` and `Retina::getMagno` (again using a regular `Mat` as output), the assertion would fail:

```cpp
#ifdef HAVE_OPENCL
bool RetinaImpl::ocl_getParvo(OutputArray retinaOutput_parvo)
{
    CV_Assert(_wasOCLRunCalled);
    _ocl_retina->getParvo(retinaOutput_parvo);
    return true;
}
#endif

void RetinaImpl::getParvo(OutputArray retinaOutput_parvo)
{
    CV_OCL_RUN((_ocl_retina != 0) && retinaOutput_parvo.isUMat(), ocl_getParvo(retinaOutput_parvo));
    CV_Assert(!_wasOCLRunCalled);  // <-- this check fails

    //...
}
```

Perhaps you shouldn't execute the OpenCL path unless the input image was a `UMat` in the first place, i.e:

```cpp
CV_OCL_RUN((_ocl_retina != 0) && inputMatToConvert.isUMat(), ocl_run(inputMatToConvert));
```

Can you confirm this is a valid fix?

PR #1330, opened 2017-08-18
Replace OpenMP parallelization with OpenCV's general parallel_for_ to…

### LINKED ISSUE #1329: Structured Edge Detection only optimized for OpenMP
Examining the code of ximgproc's structured edge detection shows that parallelization in `predictEdges()` is only accomplished with OpenMP. 
This could be improved using OpenCV's own parallel_for which to my knowledge chooses the appropriate backend (TBB, OpenMP, GCD).
See also issue [#8179](https://github.com/opencv/opencv/issues/8179) of the main repo.

A PR will be submitted soon.

Update: PR #1330 is submitted, however it will fail until #1168 is resolved as it uses C++11's Lambdas.

Jaykob:
… cover other backends than OpenMP.

resolves #1329
depends on #1168

### This pullrequest changes

This replaces specific OpenMP parallelization in `predictEdges()` with the more general OpenCV parallel_for.
As C++11`s lambdas are used, this will depend on #1168.

### DIFF
--- a/modules/ximgproc/src/structured_edge_detection.cpp
+++ b/modules/ximgproc/src/structured_edge_detection.cpp
@@ -730,9 +730,10 @@ protected:
             }
             // lookup tables for mapping linear index to offset pairs
 
-        const cv::Range range(0, height);
         #ifdef CV_CXX11
-        parallel_for_(range, [&](const cv::Range& range)
+        parallel_for_(cv::Range(0, height), [&](const cv::Range& range)
+        #else
+        const cv::Range range(0, height);
         #endif
         {
             for(int i = range.start; i < range.end; ++i) {
@@ -788,7 +789,7 @@ protected:
 
         float step = 2.0f * CV_SQR(stride) / CV_SQR(ipSize) / nTreesEval;
         #ifdef CV_CXX11
-        parallel_for_(range, [&](const cv::Range& range)
+        parallel_for_(cv::Range(0, height), [&](const cv::Range& range)
         #endif
         {
             for(int i = range.start; i < range.end; ++i)
--- a/modules/ximgproc/src/structured_edge_detection.cpp
+++ b/modules/ximgproc/src/structured_edge_detection.cpp
@@ -729,65 +729,72 @@ protected:
                 offsetY[n] = x2*features.cols*nchannels + y2*nchannels + z;
             }
             // lookup tables for mapping linear index to offset pairs
-        #ifdef _OPENMP
-        #pragma omp parallel for
+
+        const cv::Range range(0, height);
+        #ifdef CV_CXX11
+        parallel_for_(range, [&](const cv::Range& range)
         #endif
-        for (int i = 0; i < height; ++i)
         {
-            float *regFeaturesPtr = regFeatures.ptr<float>(i*stride/shrink);
-            float  *ssFeaturesPtr = ssFeatures.ptr<float>(i*stride/shrink);
+            for(int i = range.start; i < range.end; ++i) {
+                float *regFeaturesPtr = regFeatures.ptr<float>(i*stride/shrink);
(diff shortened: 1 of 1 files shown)


sovrasov:
@Jaykob OpenCV will switch to cxx11 by default with the next major version i.e. 4.0., so if you want it to be merged this year, you should use old-style parallel for.

Jaykob:
For the moment I'm OK with waiting as it builds successfully with cxx11 here.
Should I find the time, I'll implement it using the old-style, however I'm not a big fan of it as I think readability suffers a lot...

csukuangfj:
I would like to change it to ParallelLoopBody if you do not want to.

Jaykob:
I found the time and started already, should be finished soon.

csukuangfj:
You could check whether the macro "HAVE_CXX11" is defined or not
to switch to C++11

alalek:
`HAVE_CXX11` is available in CMake scripts only. In C++ code please use [`CV_CXX11`](https://github.com/opencv/opencv/blob/3.3.0/modules/core/include/opencv2/core/cvdef.h#L385-L396).

Example:
```
#ifdef CV_CXX11
parallel_for_(cv::Range(0, height), [&](const cv::Range& range)
#else
const cv::Range range(0, height);
#endif
{
    ... regular parallel body code ...
}
#ifdef CV_CXX11
);
#endif
```

(there is no parallel loop execution in C++98 compilation mode)

Jaykob:
Good idea, thanks! I just commited it this way.

csukuangfj:
My suggestion was to implement both C++11 version and ParallelLoopBody version and
switch between them with the macro. Now there is no parallelization in case C++11 is unavailable.

modules/ximgproc/src/structured_edge_detection.cpp
```diff
@@ -730,7 +730,10 @@ class StructuredEdgeDetectionImpl : public StructuredEdgeDetection
             }
             // lookup tables for mapping linear index to offset pairs
```
csukuangfj: The buildbot says there are some spaces.

Jaykob:
Yes that's true. I think it should be OK to assume C++11 in the year 2017, though. As it will be activated by default with 4.0.0, it's only temporary anyway.
But feel free to submit a PR that implements both versions.

sovrasov:
:+1:

modules/ximgproc/src/structured_edge_detection.cpp
```diff
@@ -729,65 +729,72 @@ class StructuredEdgeDetectionImpl : public StructuredEdgeDetection
                 offsetY[n] = x2*features.cols*nchannels + y2*nchannels + z;
             }
             // lookup tables for mapping linear index to offset pairs
-        #ifdef _OPENMP
-        #pragma omp parallel for
+
+        const cv::Range range(0, height);
+        #ifdef CV_CXX11
+        parallel_for_(range, [&](const cv::Range& range)
         #endif
```
alalek: C++11 mode emit these warnings:
```
structured_edge_detection.cpp:735:56: warning: declaration of 'range' shadows a previous local [-Wshadow]
structured_edge_detection.cpp:733:25: warning: shadowed declaration is here [-Wshadow]
structured_edge_detection.cpp:791:56: warning: declaration of 'range' shadows a previous local [-Wshadow]
structured_edge_detection.cpp:733:25: warning: shadowed declaration is here [-Wshadow]
```

Please move top-level `range` declaration under `#else`.

Jaykob:
I just found out that the parallelization of the second loop causes a nondeterministic behaviour. 
Results differ from run-to-run using the same image.
Seems like this bug has been there ever since it was parallelized using OpenMP in my eyes. Will have a closer look at it soon.

PR #1332, opened 2017-08-18
making minRepDistance not quadratically dependent on the image size

### LINKED ISSUE #1331: CharUco Diamond detection detects too many false positives when the image resolution is high.

- OpenCV => 3.3
- Operating System / Platform => Any
- Compiler => Any

##### Detailed description

CharUco Diamond detection detects too many false positives when the image resolution is high.

The reason is that a variable is that  ```minRepDistance``` inside ```detectCharucoDiamond``` is growing in squared proportion to the image width, and then its value is squared again for the thresholding using ```closestCandidateDistance``` inside of ```refineDetectedMarkers```.

##### Steps to reproduce

Take for instance the image of original size 1920x1080. It recognizes (7, 9, 45, 2), although the marker is (2,7,45,9). And the pose is totally wrong. 

With the downsampled image of 960x540 (and of course a corresponding calibration file of half the resolution), the id is correctly recognized as (2,7,45,9) and the pose estimation is accurate.

You can try moving the window size parameters, but reality is that ```detectCharucoDiamond``` does not consider any detector parameters. 

Original image size (1920x1080):
example1085

Downsampled image (960x540):
example1085_resized

The results yout get are:

Original image size (1920x1080):
out_screenshot_04 08 2017

Downsampled image (960x540):
out_screenshot_04 08 2017_resized

You can use the very same code available in the ```modules/aruco/samples/detect_diamonds.cpp``` for testing. 
The calibration files are available as 

Camera calibration for original image:
[calibration_tango.txt](https://github.com/opencv/opencv_contrib/files/1234019/calibration_tango.txt)

Camera calibration for resized image: 
[calibration_tango_resized.txt](https://github.com/opencv/opencv_contrib/files/1234021/calibration_tango_resized.txt)

##### How to fix

There are much better ways to fix (between them to consider detector parameters), but the least intrusive would be to change ```minRepDistanceRate``` and ```minRepDistance```, inside of ```charuco.cpp```. I attach a diff :

```
diff -ruN OpenCV_contrib/src/modules/aruco/src/charuco.cpp OpenCV_contrib.1/src/modules/aruco/src/charuco.cpp
--- OpenCV_contrib/src/modules/aruco/src/charuco.cpp	2016-12-23 12:09:43.000000000 +0100
+++ OpenCV_contrib.1/src/modules/aruco/src/charuco.cpp	2017-08-09 16:15:38.003573042 +0200
@@ -738,7 +738,7 @@
 
     CV_Assert(_markerIds.total() > 0 && _markerIds.total() == _markerCorners.total());
 
-    const float minRepDistanceRate = 0.12f;
+    const float minRepDistanceRate = 1.3024550356922115f;
 
     // create Charuco board layout for diamond (3x3 layout)
     Ptr<Dictionary> dict = getPredefinedDictionary(PREDEFINED_DICTIONARY_NAME(0));
@@ -771,7 +771,7 @@
           perimeterSq += edge.x*edge.x + edge.y*edge.y;
         }
         // maximum reprojection error relative to perimeter
-        float minRepDistance = perimeterSq * minRepDistanceRate * minRepDistanceRate;
+        float minRepDistance = sqrt(perimeterSq) * minRepDistanceRate;
 
         int currentId = _markerIds.getMat().at< int >(i);
```

A pull request is on its way. Please support it! :+1:

cduguet:
### System information (version)

OpenCV => 3.3
Operating System / Platform => Any
Compiler => Any
Detailed description

CharUco Diamond detection detects too many false positives when the image resolution is high.

The reason is that a variable is that ```minRepDistance``` inside ```detectCharucoDiamond``` is growing in squared proportion to the image width, and then its value is squared again for the thresholding using ```closestCandidateDistance``` inside of ```refineDetectedMarkers```.

The images for testing this problem and the corresponding calibration files can be found in: 

https://github.com/opencv/opencv_contrib/issues/1331

### DIFF
--- a/modules/aruco/src/charuco.cpp
+++ b/modules/aruco/src/charuco.cpp
@@ -738,7 +738,7 @@ void detectCharucoDiamond(InputArray _image, InputArrayOfArrays _markerCorners,
 
     CV_Assert(_markerIds.total() > 0 && _markerIds.total() == _markerCorners.total());
 
-    const float minRepDistanceRate = 0.12f;
+    const float minRepDistanceRate = 1.302455f;
 
     // create Charuco board layout for diamond (3x3 layout)
     Ptr<Dictionary> dict = getPredefinedDictionary(PREDEFINED_DICTIONARY_NAME(0));
@@ -771,7 +771,7 @@ void detectCharucoDiamond(InputArray _image, InputArrayOfArrays _markerCorners,
           perimeterSq += edge.x*edge.x + edge.y*edge.y;
         }
         // maximum reprojection error relative to perimeter
-        float minRepDistance = perimeterSq * minRepDistanceRate * minRepDistanceRate;
+        float minRepDistance = sqrt(perimeterSq) * minRepDistanceRate;
 
         int currentId = _markerIds.getMat().at< int >(i);
 


modules/aruco/src/charuco.cpp
```diff
@@ -738,7 +738,7 @@ void detectCharucoDiamond(InputArray _image, InputArrayOfArrays _markerCorners,
 
     CV_Assert(_markerIds.total() > 0 && _markerIds.total() == _markerCorners.total());
 
-    const float minRepDistanceRate = 0.12f;
+    const float minRepDistanceRate = 1.302455f;
```
sovrasov: How did you calculate this value?

sovrasov:
:+1:

PR #1339, opened 2017-08-22
fix cmake configuration error

jasjuang:
WIthout this change, there will be the below cmake configuration error

``````
CMake Error at /home/ubuntu/ppa/opencv_contrib/modules/cnn_3dobj/CMakeLists.txt:39 (target_link_libraries):
  The keyword signature for target_link_libraries has already been used with
  the target "opencv_test_cnn_3dobj".  All uses of target_link_libraries with
  a target must be either all-keyword or all-plain.

  The uses of the keyword signature are here:

   * /home/ubuntu/ppa/opencv/cmake/OpenCVUtils.cmake:911 (target_link_libraries)
``````


sovrasov:
:+1:

PR #1341, opened 2017-08-23
Comments minor typo

f3rm4rf3r:

### DIFF
--- a/modules/tracking/src/trackerKCF.cpp
+++ b/modules/tracking/src/trackerKCF.cpp
@@ -47,7 +47,7 @@
 |---------------------------*/
 namespace cv{
    /**
-  * \brief Implementation of TrackerModel for MIL algorithm
+  * \brief Implementation of TrackerModel for KCF algorithm
   */
   class TrackerKCFModel : public TrackerModel{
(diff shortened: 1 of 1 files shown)


modules/tracking/src/trackerKCF.cpp
```diff
@@ -47,7 +47,7 @@
 |---------------------------*/
 namespace cv{
    /**
-  * \brief Implementation of TrackerModel for MIL algorithm
+  * \brief Implementation of TrackerModel for KCF algorithm
   */
   class TrackerKCFModel : public TrackerModel{
```
alalek: Similar problem is [here](https://github.com/opencv/opencv_contrib/blob/3.3.0/modules/tracking/doc/%5BTutorial%5D%20Adding%20new%20Tracker%20Method%20for%20dummies) too

sovrasov: BTW mentioned tutorial is wrong, tracking API was slightly changed.

alalek:
Lets put it in and remove outdated tutorial later.
Thank you! :+1:

PR #1343, opened 2017-08-24
omnidir: fix wrong memory access in estimateUncertainties

### LINKED ISSUE #1340: Possible bug in cv::omnidir::internal::estimateUncertainties
##### Detailed description
When some camera parameters are fixed (ex: omnidir::CALIB_FIX_CENTER) , in _cv :: omnidir :: calibrate_, the system crashes.

##### Steps to fix the problem
Comment in _cv :: omnidir :: internal :: estimateUncertainties_, the following line:
```
errors = 3 * s * _JTJ_inv.diag();
//checkFixed(errors, flags, n); <-----
rms = 0;
```

This line (checkFixed) is useless because it does not intervene in calculations and variable return.

The explanation of the crash is the matrix **errors** is not always of dimensions (6 * n + 10) x (6 * n + 10), because it comes from the Jacobian matrix **_JTJ_inv** which is a submatrix of the main jacobian matrix. 
This is due to the following matrix reduction operation in _cv::omnidir::internal::computeJacobian_ :
```
    std::vector<int> _idx(6*n+10, 1);
    flags2idx(flags, _idx, n);

    subMatrix(JTJ, JTJ, _idx, _idx);
    subMatrix(JTE, JTE, std::vector<int>(1, 1), _idx);
```

- OpenCV => 3.2.0
- Operating System / Platform =>Ubuntu 16.04.2 LTS
- Compiler => gcc 5.4.0 64bit

sovrasov:
resolves #1340

### DIFF
--- a/modules/ccalib/src/omnidir.cpp
+++ b/modules/ccalib/src/omnidir.cpp
@@ -1785,8 +1785,6 @@ void cv::omnidir::internal::estimateUncertainties(InputArrayOfArrays objectPoint
 
     errors = 3 * s * _JTJ_inv.diag();
 
-    checkFixed(errors, flags, n);
-
     rms = 0;
     const Vec2d* ptr_ex = reprojError.ptr<Vec2d>();
     for (int i = 0; i < (int)reprojError.total(); i++)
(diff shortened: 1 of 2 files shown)


modules/ccalib/src/omnidir.cpp
```diff
@@ -1995,52 +1993,6 @@ double cv::omnidir::internal::computeMeanReproErrStereo(InputArrayOfArrays objec
     return reProErr;
 }
 
-void cv::omnidir::internal::checkFixed(Mat& G, int flags, int n)
```
alalek: It should be dropped from `.hpp` file too.

sovrasov: @alalek fixed

alalek:
:+1:

PR #1347, opened 2017-08-28
Update webcam_demo.cpp

sturkmen72:
### This pullrequest changes
1. the program was working only with webcam. alternatively using a video file or using a  single image file as an option is available now.

2. Visual improvements

### DIFF
--- a/modules/text/samples/webcam_demo.cpp
+++ b/modules/text/samples/webcam_demo.cpp
@@ -1,20 +1,18 @@
 /*
  * webcam-demo.cpp
  *
- * A demo program of End-to-end Scene Text Detection and Recognition.
+ * A demo program of End-to-end Scene Text Detection and Recognition using webcam or video.
  *
  * Created on: Jul 31, 2014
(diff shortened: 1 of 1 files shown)


modules/text/samples/webcam_demo.cpp
```diff
@@ -140,120 +183,79 @@ int main(int argc, char* argv[])
...
 
         if (downsize)
             resize(frame,frame,Size(320,240));
 
         /*Text Detection*/
-
-        cvtColor(frame,grey,COLOR_RGB2GRAY);
-        grey.copyTo(orig_grey);
+        cvtColor(frame,gray,COLOR_BGR2GRAY);
+        gray.copyTo(orig_gray);
```
terfendail: It looks like `orig_gray` isn't used anywhere else.

sturkmen72: @terfendail  thanks for review. `orig_gray `removed.

terfendail:
👍

PR #1357, opened 2017-09-05
Updated default tracker parameters

f3rm4rf3r:
Updated the lambda coefficient to the one suggested in the original paper (http://www.robots.ox.ac.uk/~joao/publications/henriques_tpami2015.pdf, page 11, Table 2) which works (for me) better then the previous default one.


sovrasov:
@f3rm4rf3r in future, please put all of your small changes into one PR.

PR #1374, opened 2017-09-20
KCF speedup

sovrasov:
replaces #579

Results on dudek benchmark video:

|Data Type|FPS|
|----------|--------------|
| float | 174 |
| double | 124 |

Quality metrics remained the same.

OpenCL kernel doesn't give speedup on iGPU (tested on i7-6700K and i5-6300U), but for faster GPUs and large resolutions the results may change. For now OpenCL TMM kernel is disabled for small patches (original threshold was increased in two times).


vpisarev:
@sovrasov, very nice! Hope, you will finish it as well

sovrasov:
PR is ready for merge.

PR #1385, opened 2017-09-29
fixed typo in tutorial

flashspys:
### This pullrequest changes

rvecs and tvecs should be of type std::vector<cv::Vec3d> as used in the detect_markers.cpp example. Otherwise an exception is thrown.


savuor:
:+1:

PR #1393, opened 2017-10-04
surface_matching loadPLYSimple() now reports that only ascii format is supported

terfendail:
### Original PR description
in addition to ascii there is a binary .ply format, which this function does not support. I added code to report that unsupported data is being read.

### DIFF
--- a/modules/surface_matching/src/ppf_helpers.cpp
+++ b/modules/surface_matching/src/ppf_helpers.cpp
@@ -69,10 +69,13 @@ std::vector<std::string> split(const std::string &text, char sep) {
 
 
 
-Mat loadPLYSimple(const char* fileName, int withNormals)
+Mat loadPLYSimple(const char* fileName, int withNormals /* = 0 */)
 {
   Mat cloud;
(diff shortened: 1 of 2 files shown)


bmagyar:
Looks good to me on first read :+1:

vpisarev:
:+1:

modules/surface_matching/src/ppf_helpers.cpp
```diff
@@ -55,55 +55,92 @@ void getRandomRotation(double R[9]);
...
-    cloud=Mat(numVertices, 6, CV_32FC1);
-  else
-    cloud=Mat(numVertices, 3, CV_32FC1);
+  cloud = Mat(numVertices, withNormals ? 6 : 3, CV_32FC1);
 
   for (int i = 0; i < numVertices; i++)
   {
     float* data = cloud.ptr<float>(i);
+    int col = 0;
+    for (; col < withNormals ? 6 : 3; ++col)
```
alalek: infinite loop

PR #1399, opened 2017-10-09
CNN-based text detector

sovrasov:
Reintegrated detection part from #1348.


sovrasov:
@vpisarev could you take a look?

vpisarev:
looks great! :+1:

PR #1401, opened 2017-10-10
hfs segment module

Yujun-Shi:
An efficient algorithm for image segmentation

### This pullrequest changes

Adding a new module called hfs_segment to the original repository. This module is used to do efficient image segmentation.

### DIFF
--- a/modules/hfs_segment/CMakeLists.txt
+++ b/modules/hfs_segment/CMakeLists.txt
@@ -1,2 +1,6 @@
+if(NOT HAVE_CUDA)
+  ocv_module_disable(hfs_segment)
+endif()
+
 set(the_description "Hierarchical Feature Selection for Efficient Image Segmentation")
 ocv_define_module(hfs_segment opencv_core opencv_imgproc)


modules/hfs_segment/src/egb_segment/egb_segment.cpp
```diff
@@ -0,0 +1,98 @@
...
+// indirect, incidental, special, exemplary, or consequential damages
+// (including, but not limited to, procurement of substitute goods or services;
+// loss of use, data, or profits; or business interruption) however caused
+// and on any theory of liability, whether in contract, strict liability,
+// or tort(including negligence or otherwise) arising in any way out of
+// the use of this software, even if advised of the possibility of such damage.
+//
+//M*/
+/*
+Copyright (C) 2006 Pedro Felzenszwalb
```
vpisarev: sorry, but we do not allow GPL code in OpenCV

vpisarev:
@Yujun-Shi, thank you for the contribution! As I see, the patch includes some GPL code. We do not allow it in OpenCV

modules/hfs_segment/src/egb_segment/egb_segment.cpp
Yujun-Shi: Sorry about that. So, can I revise this piece of code and make a new pull request? (By revise, I mean implementing this algorithm again on my own)

Issue #1376, opened 2017-09-21
Matlab module build issue with new type traits


catree:
##### System information (version)

- OpenCV => cc547e8
- OpenCV contrib => 4bc49c0
- Operating System / Platform => Ubuntu 16.04 / 64 bits

##### Detailed description

Related: [core: fix type traits #9457](https://github.com/opencv/opencv/pull/9457)
Related: [fix type traits #1352](https://github.com/opencv/opencv_contrib/pull/1352)

I think the Matlab contrib module has not been updated to use the correct "type traits".
Errors:

> [ 80%] Compiling Matlab source files. This could take a while...
> CMake Error at /home/opencv_contrib/modules/matlab/compile.cmake:47 (message):
>   Failed to compile drawMatches: In file included from
>   /home/opencv/modules/core/include/opencv2/core/cvdef.h:91:0,
> 
> 
>                    from /home/opencv/modules/core/include/opencv2/core.hpp:52,
>                    from /home/opencv_contrib/modules/matlab/include/opencv2/matlab/bridge.hpp:52,
>                    from /home/opencv-build/modules/matlab/src/drawMatches.cpp:14:
> 
>   
>   /home/opencv_contrib/modules/matlab/include/opencv2/matlab/bridge.hpp:
>   In instantiation of ‘cv::Mat cv::bridge::Bridge::toMat() const [with
>   Scalar = long int]’:
> 
>   
>   /home/opencv_contrib/modules/matlab/include/opencv2/matlab/bridge.hpp:579:49:
>   required from here
> 
>   
>   /home/opencv/modules/core/include/opencv2/core/hal/interface.h:77:42:
>   error: ‘type’ is not a member of ‘cv::DataType<long int>’
> 
>    #define CV_MAT_DEPTH(flags)     ((flags) & CV_MAT_DEPTH_MASK)
>                                             ^
> 
>   
>   /home/opencv/modules/core/include/opencv2/core/hal/interface.h:79:32:
>   note: in expansion of macro ‘CV_MAT_DEPTH’
> 
>    #define CV_MAKETYPE(depth,cn) (CV_MAT_DEPTH(depth) + (((cn)-1) << CV_CN_SHIFT))
>                                   ^
> 
>   
>   /home/opencv_contrib/modules/matlab/include/opencv2/matlab/bridge.hpp:244:43:
>   note: in expansion of macro ‘CV_MAKETYPE’
> 
[...]

PR #1406, opened 2017-10-13
matlab: disable module by default

alalek:
- broken bindings generator
- broken conversion functions (uses non-supported cv::Mat types)
- other problems like missing/invalid conversion functions for Moments and other structs

related #1376


alalek:
:+1:

PR #1412, opened 2017-10-17
saliency: improve BING handling

### LINKED ISSUE #1411: saliency: ObjectnessBING::matRead always returns true
- OpenCV => 3.3 (master)
- Compiler => mingw64

##### Detailed description

ObjectnessBING needs to read pretrained model files, but its outcome is [never checked in matRead()](https://github.com/opencv/opencv_contrib/blob/master/modules/saliency/src/BING/objectnessBING.cpp#L437)

right now, trying to load any of those Filestorages will go unchecked, and will throw an [assertion here](https://github.com/opencv/opencv_contrib/blob/master/modules/saliency/src/BING/FilterTIG.cpp#L77) later, if unsuccsesful.

(an invalidMat is passed on)

the default trainingpath in the computeSaliency sample (needed for BING) also has the default set to "1" instead of "ObjectnessTrainedModel"

##### Steps to reproduce
start [the sample](https://github.com/opencv/opencv_contrib/blob/master/modules/saliency/samples/computeSaliency.cpp) using "BING", but without pointing to a valid "training_path" argument.

berak:
resolves #1411

### This pullrequest changes

* make the BING path fail better, if the pretrained models are not present or found
* fix some minor assumptions in the sample, like the default trainingpath
* add some visualization for BING to the sample. (bonus, prev. output went entirely unsused.)
* all result files are written to a local "Results" dir. let's assume, "trainingpath" is read-only

### DIFF
--- a/modules/saliency/samples/computeSaliency.cpp
+++ b/modules/saliency/samples/computeSaliency.cpp
@@ -52,7 +52,7 @@ static const char* keys =
 { "{@saliency_algorithm | | Saliency algorithm <saliencyAlgorithmType.[saliencyAlgorithmTypeSubType]> }"
     "{@video_name      | | video name            }"
     "{@start_frame     |1| Start frame           }"
-    "{@training_path   |1| Path of the folder containing the trained files}" };
+    "{@training_path   |ObjectnessTrainedModel| Path of the folder containing the trained files}" };
 
 static void help()
 {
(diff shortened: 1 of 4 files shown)


modules/saliency/src/BING/objectnessBING.cpp
```diff
@@ -104,7 +104,7 @@ int ObjectnessBING::loadTrainedModel( std::string modelName )  // Return -1, 0,
 
   if( !matRead( s1, filters1f ) || !matRead( sI, idx1i ) )
```
sovrasov: Could you use a constant reference as an argument of `loadTrainedModel`?

modules/saliency/samples/computeSaliency.cpp
```diff
@@ -150,11 +150,28 @@ int main( int argc, char** argv )
       saliencyAlgorithm = ObjectnessBING::create();
       vector<Vec4i> saliencyMap;
       saliencyAlgorithm.dynamicCast<ObjectnessBING>()->setTrainingPath( training_path );
-      saliencyAlgorithm.dynamicCast<ObjectnessBING>()->setBBResDir( training_path + "/Results" );
```
sovrasov: Unfortunately on Linux output file can't be created properly.
Please change zero [here](https://github.com/opencv/opencv_contrib/blob/933f53e47fe93b8318f17aa1878469923d62b1d6/modules/saliency/src/BING/CmFile.cpp#L75) and few lines below to `S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH` to fix it.

alalek: BTW, This fix could be postponed - some filesystem utility functions will be added into OpenCV's core module including recursive directories creation.

berak: sure. while it could be postponed, having the correct flags here won't do any harm in the meantime.

modules/saliency/src/BING/objectnessBING.cpp
berak: uh, found [the reason, it wasn't const](https://github.com/berak/opencv_contrib/blob/62331a6ce0f252b363b0db8adc3e4a7540e46348/modules/saliency/src/BING/objectnessBING.cpp#L101)

but since loadTrainedModel() is [only called once](https://github.com/berak/opencv_contrib/blob/62331a6ce0f252b363b0db8adc3e4a7540e46348/modules/saliency/src/BING/objectnessBING.cpp#L387) , and that without any args, i'd rather remove the whole arg.

sovrasov:
:+1:

PR #1416, opened 2017-10-19
Fix for infinite loop in loadPLYSimple()

terfendail:
Additional change for #1393
### This pullrequest changes
Fixed infinite loop in loadPLYSimple()


alalek:
Please update to latest master.
Fixed in #1413

PR #1419, opened 2017-10-20
GaussianBlur of the source image for LATCH descriptor made optional

terfendail:
This PR is reworked change of LATCH descriptor offered by @ofirbb in PR #545
GaussianBlur made optional as suggested by @cbalint13 in PR #545 discussion

### This pullrequest changes
Provide ability to disable GaussianBlur of the source image for LATCH descriptor evaluation

### DIFF
--- a/modules/xfeatures2d/src/latch.cpp
+++ b/modules/xfeatures2d/src/latch.cpp
@@ -63,7 +63,7 @@ namespace cv
         public:
             enum { PATCH_SIZE = 48 };
 
-            LATCHDescriptorExtractorImpl(int bytes = 32, bool rotationInvariance = true, int half_ssd_size = 3);
+            LATCHDescriptorExtractorImpl(int bytes = 32, bool rotationInvariance = true, int half_ssd_size = 3, double sigma = 2.0);
 
             virtual void read( const FileNode& );
             virtual void write( FileStorage& ) const;
@@ -81,14 +81,15 @@ namespace cv
             PixelTestFn test_fn_;
             bool rotationInvariance_;
             int half_ssd_size_;
+            double sigma_;
 
 
             std::vector<int> sampling_points_ ;
         };
 
-        Ptr<LATCH> LATCH::create(int bytes, bool rotationInvariance, int half_ssd_size)
+        Ptr<LATCH> LATCH::create(int bytes, bool rotationInvariance, int half_ssd_size, double sigma)
         {
-            return makePtr<LATCHDescriptorExtractorImpl>(bytes, rotationInvariance, half_ssd_size);
+            return makePtr<LATCHDescriptorExtractorImpl>(bytes, rotationInvariance, half_ssd_size, sigma);
         }
         void CalcuateSums(int count, const std::vector<int> &points, bool rotationInvariance, const Mat &grayImage, const KeyPoint &pt, int &suma, int &sumc, float cos_theta, float sin_theta, int half_ssd_size);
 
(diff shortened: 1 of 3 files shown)


terfendail:
@GilLevi , @ofirbb , @StevenPuttemans
Could you please provide you opinion on this change?

GilLevi:
In my experiments the Gaussian blur improves result.

terfendail:
@GilLevi
There still is an opportunity to enable Gaussian blur if it is considered useful for particular case. But existing tests use 8UC3 images and so doesn't use smoothing due to specific of previous implementation.
I could update tests to disable smoothing explicitly and enable smoothing for generic case if you consider it a better default option.

ofirbb:
I think this is a welcome change. It can also be documented that on certain platforms the blur takes significant percentage of the descriptor's running time and not that in some scenarios it can increase performance. 

What about the tests? Are you going to edit them?

terfendail:
@ofirbb 
It looks like existing documentation doesn't state anything about performance(or I've just unable to find it).
Since smoothing could either increase performance ( @GilLevi cases ) or decrease it ( @ofirbb cases ) I think it would be better not to mention performance at all so one have to check both options and make personal decision whether smoothing fits the given task or not.

I'm not going to modify tests in any meaningful way. The only change I could do at the moment is explicit disabling of smoothing, but such change IMHO will make the code more confusing.
Since there is no sample dedicated to LATCH descriptor the only place to get the sample code from is the code of relevant tests. So it will seem strange why the default mode is used nowhere.

GilLevi:
Hi @terfendail, I'm not sure I understand. I think Gaussian Blur should be optional , but done in default. I think this is the current implementation. Why change it?

Thanks!
Gil

terfendail:
At the moment both way of processing could be considered default depending on the use case. Current implementation use smoothing if it is called for CV_8UC1, but doesn't use smoothing if called for CV_8UC3. Actually for CV_8UC3 image smoothing operation is performed, but result of the operation is discarded immediately so in effect smoothing isn't used.
IMHO updated implementation should have uniform default for both types of images. However while I consider the way current implementation acts on CV_8UC3 rather an occasional mistake it is the only way of processing referenced by other code and it's the only way that is tested since all existing tests are implemented for `CV_8UC3`. As a consequence existing tests will fail if the smoothing is made default.
So at the moment processing without smoothing provides expected and verified result from accuracy point of view and there are no performance tests to check how the speed is affected by smoothing. That is the reason I set the smoothing disabled by default.

GilLevi:
All the test I made in our original paper are with smoothing (on CV_8UC3 and then converting to grayscale). It improved performance and didn't affect running times very drastically. Therefore I think Gaussian smoothing should be made default.

StevenPuttemans:
> All the test I made in our original paper are with smoothing (on CV_8UC3 and then converting to grayscale). It improved performance and didn't affect running times very drastically. Therefore I think Gaussian smoothing should be made default.

I agree with this for several reasons

1. The original contribution was linked to the publication, so it should maintain that functionality.
2. If we can prove that disabling Gaussian blur helps in specific situations, then it should be possible to disable it, but this should NOT be switched of by default.
3. Furthermore, to ensure compatibility we cannot all of a sudden change the default behavior of an implementation. This could have terrible consequences for companies using the actual implementation and doing a software upgrade.

terfendail:
@GilLevi @StevenPuttemans

Original implementation doesn't perform smoothing for CV_8UC3 images. That could be considered a typo from the end user point of view but the accuracy test for this descriptor process CV_8UC3 image and ensure the result. So processing without smoothing by default is assured by the accuracy test.

However I will set default to enabled smoothing if you consider that a proper way.

modules/xfeatures2d/src/latch.cpp
```diff
@@ -502,15 +503,27 @@ namespace cv
 
 
             Mat grayImage;
-            GaussianBlur(image, grayImage, cv::Size(3, 3), 2, 2);
-
-            if (image.type() != CV_8U) cvtColor(image, grayImage, COLOR_BGR2GRAY);
```
alalek: BTW, Problem is here: `GaussianBlur()` result was erased by `cvtColor()` for non-gray images.

alalek:
@GilLevi @StevenPuttemans
Any thoughts about the current proposal (with "sigma = 2.0" by default)?

vpisarev:
let me merge the patch in. The previous behaviour can easily be reproduced :+1:

PR #1421, opened 2017-10-23
Fix name extraction from path in multicalib

### LINKED ISSUE #1420: Fix name extraction from path in multicalib
At https://github.com/opencv/opencv_contrib/blob/master/modules/ccalib/src/multicalib.cpp#L144 instead of:
`std::string filename = file_list[i].substr(0, file_list[i].find('.'));`
should be:
`std::string filename = file_list[i].substr(0, file_list[i].rfind('.'));`
in order to support all types of path, including path names with points or relative paths.

cdcseacave:
resolves #1420


sovrasov:
:+1:

PR #1423, opened 2017-10-24
tracking: adding a mosse tracker

berak:
this pr adds a MOSSE tracker to the tracking module, quite similar to the python sample in opencv main.

### DIFF
--- a/modules/tracking/src/mosseTracker.cpp
+++ b/modules/tracking/src/mosseTracker.cpp
@@ -1,0 +1,249 @@
+// This file is part of the OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+//
+//[1] David S. Bolme et al. "Visual Object Tracking using Adaptive Correlation Filters"
+//    http://www.cs.colostate.edu/~draper/papers/bolme_cvpr10.pdf
+//
+
+//
+// credits:
+// Kun-Hsin Chen: for initial c++ code
+// Cracki: for the idea of only converting the used patch to gray
+//
+
+#include "opencv2/tracking.hpp"
+
+namespace cv {
+namespace tracking {
+
+struct DummyModel : TrackerModel
+{
+     virtual void modelUpdateImpl(){}
+     virtual void modelEstimationImpl( const std::vector<Mat>& ){}
+};
+
+
+const double eps=0.00001;      // for normalization
+const double rate=0.2;         // learning rate
+const double psrThreshold=5.7; // no detection, if PSR is smaller than this
+
+
+struct MosseImpl : TrackerMOSSE
+{
+protected:
+
+    Point2d center; //center of the bounding box
+    Size size;      //size of the bounding box
+    Mat hanWin;
+    Mat G;          //goal
+    Mat H, A, B;    //state
+
+    //  Element-wise division of complex numbers in src1 and src2
+    Mat divDFTs( const Mat &src1, const Mat &src2 ) const
+    {
+        Mat c1[2],c2[2],a1,a2,s1,s2,denom,re,im;
+
+        // split into re and im per src
+        cv::split(src1, c1);
+        cv::split(src2, c2);
+
+        // (Re2*Re2 + Im2*Im2) = denom
+        //   denom is same for both channels
+        cv::multiply(c2[0], c2[0], s1);
+        cv::multiply(c2[1], c2[1], s2);
+        cv::add(s1, s2, denom);
+
+        // (Re1*Re2 + Im1*Im1)/(Re2*Re2 + Im2*Im2) = Re
+        cv::multiply(c1[0], c2[0], a1);
+        cv::multiply(c1[1], c2[1], a2);
+        cv::divide(a1+a2, denom, re, 1.0 );
+
+        // (Im1*Re2 - Re1*Im2)/(Re2*Re2 + Im2*Im2) = Im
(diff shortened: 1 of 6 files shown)


berak:
```
benchmark -gt=david.gt.txt -video=david.webm -start=300 MOSSE

==========
mosse
       Overlap > 0                   100%                 469
       Overlap > 0.5             57.1429%                 268
           Precision             57.1429%
           Recall                57.1429%
           f-measure             57.1429%
                 AUC            0.509765
         Performance             2.23582 ms/frame             447.262 fps

```

(tests were added, too. since it (visually) behaves similar to the MIL one, i copied the resp. values from there, maybe someone can help finding better ones.)

sovrasov:
@berak to find the threshold values just run the tests locally and see the actual values in error messages:
```
Tracking/DistanceAndOverlap.Scaled_Data_MOSSE/0 ["david"] : 
/build/precommit-contrib_linux64/opencv_contrib/modules/tracking/test/test_trackers.cpp:312: Failure
Failed
Incorrect overlap: curr = 0.4690278172492981, min = 0.5

Google Test trace:
/build/precommit-contrib_linux64/opencv_contrib/modules/tracking/test/test_trackers.cpp:407: A
Tracking/DistanceAndOverlap.Scaled_Data_MOSSE/1 ["dudek"] : 
/build/precommit-contrib_linux64/opencv_contrib/modules/tracking/test/test_trackers.cpp:306: Failure
Failed
Incorrect distance: curr = 44.406688690185547, max = 25

Google Test trace:
/build/precommit-contrib_linux64/opencv_contrib/modules/tracking/test/test_trackers.cpp:407: A
```
For overlap I recommend to use actual - 0.05
for distance: actual + 10
Sometimes they differ on different platforms, then use the buildbot for that.

modules/tracking/src/mosseTracker.cpp
```diff
@@ -0,0 +1,291 @@
+/*///////////////////////////////////////////////////////////////////////////////////////
```
sovrasov: Please use short license header for the new files:
```
/*
```

sovrasov:
Also MOSSE should be added in [this](https://github.com/opencv/opencv_contrib/blob/6ffdd0f4ba9ab39cba200d487efac3f73ac44937/modules/tracking/samples/samples_utility.hpp) function to be able to use the new tracker from the `tracker` sample.

berak:
@sovrasov , thanks for kindly looking at it !

>  run the tests locally 
had some problems, obtaining the data, initially, but that's solved now. let's see.

> use short license header
sure.

> Also MOSSE should be added in this function
i think, i have it [covered already](https://github.com/opencv/opencv_contrib/pull/1423/files#diff-1519ded1b562802bd2ee88c326058ac7R23)

modules/tracking/test/test_trackers.cpp
```diff
@@ -523,7 +536,7 @@ TEST_P(DistanceAndOverlap, Scaled_Data_KCF)
 
 TEST_P(DistanceAndOverlap, DISABLED_Scaled_Data_TLD)
 {
-  TrackerTest test( TrackerTLD::create(), dataset, 120, .45f, Scale_1_1);
+  TrackerTest test( TrackerTLD::create(), dataset, 30, .7f, Scale_1_1);
```
sovrasov: Please revert this change. It's not related to mosse and will conflict withe the existed pull request related to TLD.

berak: oops, ofc. !

sovrasov:
I've obtained two different results on the dudek video:
```
MOSSE
       Overlap > 0                20.542%                 235
       Overlap > 0.5             17.7448%                 203
           Precision             17.7914%
           Recall                18.9366%
           f-measure             18.3461%
                 AUC            0.874668
         Performance             1.25915 ms/frame             794.188 fps

```
and
```
MOSSE
       Overlap > 0               93.7063%                1072
       Overlap > 0.5             66.6958%                 763
           Precision             66.6958%
           Recall                71.1754%
           f-measure             68.8628%
                 AUC            0.474895
         Performance             1.23888 ms/frame             807.184 fps
```
In first case the tracker just failed. Are there any ideas why results are non-deterministic?

modules/tracking/test/test_trackers.cpp
```diff
@@ -534,6 +547,13 @@ TEST_P(DistanceAndOverlap, DISABLED_GOTURN)
   test.run();
 }
 
+TEST_P(DistanceAndOverlap, Scaled_Data_MOSSE)
+{
+  TrackerTest test( TrackerMOSSE::create(), dataset, 22, -0.05f, Scale_1_1, 1);
```
sovrasov: Why do you have a negative value here? If the tracker failed, 0 is enough. I have good results on Ubuntu 16.04 in this test (min overlap is 0.69).

berak: oh, i simply must have misread your comment about the values above.

i'll try again ;)

berak:
indeed, [randomWarp, here](https://github.com/opencv/opencv_contrib/pull/1423/files#diff-9df63a86c375d7d9982b9379892e61eaR111)

should i use a fixed seed instead ?

sovrasov:
Yes, because the tests can accidentally fail on buldbot (in case of "bad" seed parameter). This is no good.

berak:
ok, got it. 

dudek is unfair, gt is off a lot ;)

sovrasov:
:+1:

PR #1429, opened 2017-10-31
add Ogre3D based visualizer module

paroj:
requires Ogre 1.10.9

### DIFF
--- a/modules/ovis/include/opencv2/ovis.hpp
+++ b/modules/ovis/include/opencv2/ovis.hpp
@@ -1,0 +1,226 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#ifndef _OPENCV_OVIS_H_
+#define _OPENCV_OVIS_H_
+
+#include <opencv2/core.hpp>
+
+/**
+@defgroup ovis OGRE 3D Visualiser
+*/
+
+namespace cv {
+namespace ovis {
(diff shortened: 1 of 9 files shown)


modules/ovis/include/opencv2/ovis.hpp
```diff
@@ -0,0 +1,262 @@
...
+// warranties of merchantability and fitness for a particular purpose are disclaimed.
+// In no event shall the Intel Corporation or contributors be liable for any direct,
+// indirect, incidental, special, exemplary, or consequential damages
+// (including, but not limited to, procurement of substitute goods or services;
+// loss of use, data, or profits; or business interruption) however caused
+// and on any theory of liability, whether in contract, strict liability,
+// or tort (including negligence or otherwise) arising in any way out of
+// the use of this software, even if advised of the possibility of such damage.
+//
+//M*/
```
terfendail: Could you please use the new license header for the new files:
```
```

modules/ovis/include/opencv2/ovis.hpp
```diff
@@ -0,0 +1,226 @@
...
+    /// draw coordinate system crosses for debugging
+    SCENE_SHOW_CS_CROSS = 4
+};
+
+enum MaterialProperty
+{
+    MAT_PROP_POINT_SIZE,
+    MAT_PROP_OPACITY,
+    MAT_PROP_EMISSIVE,
+    MAT_PROP_TEXTURE
```
alalek: MAT -> MATERIAL ? (to avoid confusing with cv::Mat)

PR #1430, opened 2017-11-03
Fix for TrackerMedianFlowModel internal image memory leak

parsifal-47:
Internal image of the tracker model need to be released on reassignment or object destruction

### This pullrequest changes

`TrackerMedianFlowModel` which is not used outside of trackerMedianFlow.cpp


terfendail:
AFAIK `copyTo` should release destination image if it's not empty. Image should be released on destruction as well. So the change doesn't seem necessary.
Even if the memory leak exists due to improper image memory deallocation on reassignment or destruction such issue should be resolved by `Mat` implementation fix rather than workaround.

parsifal-47:
oh, got it, thanks and sorry for bothering

PR #1432, opened 2017-11-05
aruco: fix dictionaries format description

okalachev:
The description of aruco dictionaries format is incorrect: in fact the rotation number of the marker is before the marker's bytes themselves.

```c++
static unsigned char DICT_4X4_1000_BYTES[][4][2] =
```

In example above array of length 4 is rotations and array of length 2 is marker's bytes.


terfendail:
👍

PR #1433, opened 2017-11-05
Convert to gray instead of throwing error

Sahloul:
Similar to other descriptors, if the input image is colored, covert it silently to gray-level rather than terminating the execution

### DIFF
--- a/modules/xfeatures2d/src/freak.cpp
+++ b/modules/xfeatures2d/src/freak.cpp
@@ -333,8 +333,13 @@ void FREAK_Impl::compute( InputArray _image, std::vector<KeyPoint>& keypoints, O
     ((FREAK_Impl*)this)->buildPattern();
 
     // Convert to gray if not already
-    Mat grayImage = image;
-    CV_Assert(grayImage.channels() == 1);
+    Mat grayImage;
+    if( image.channels() == 3 || image.channels() == 4 )
+        cvtColor(image, grayImage, COLOR_BGR2GRAY);
+    else {
+        CV_Assert(image.channels() == 1);
+        grayImage = image;
+    }
 
     // Use 32-bit integers if we won't overflow in the integral image
     if ((image.depth() == CV_8U || image.depth() == CV_8S) &&


modules/xfeatures2d/src/freak.cpp
```diff
@@ -333,8 +333,11 @@ void FREAK_Impl::compute( InputArray _image, std::vector<KeyPoint>& keypoints, O
     ((FREAK_Impl*)this)->buildPattern();
 
     // Convert to gray if not already
-    Mat grayImage = image;
-    CV_Assert(grayImage.channels() == 1);
+    Mat grayImage;
+    if( image.channels() == 3 || image.channels() == 4 )
+        cvtColor(image, grayImage, COLOR_BGR2GRAY);
```
sturkmen72: does cvtColor work if image.channels() == 4

modules/xfeatures2d/src/freak.cpp
```diff
@@ -333,8 +333,11 @@ void FREAK_Impl::compute( InputArray _image, std::vector<KeyPoint>& keypoints, O
...
+    else
+        image.copyTo(grayImage);
```
sturkmen72: is copyTo() needed? see https://github.com/opencv/opencv_contrib/pull/915

modules/xfeatures2d/src/freak.cpp
sturkmen72: yes, it works. i tested to be sure.

modules/xfeatures2d/src/freak.cpp
```diff
@@ -333,8 +333,11 @@ void FREAK_Impl::compute( InputArray _image, std::vector<KeyPoint>& keypoints, O
     ((FREAK_Impl*)this)->buildPattern();
 
     // Convert to gray if not already
-    Mat grayImage = image;
-    CV_Assert(grayImage.channels() == 1);
+    Mat grayImage;
+    if( image.channels() == 3 || image.channels() == 4 )
+        cvtColor(image, grayImage, COLOR_BGR2GRAY);
+    else
+        grayImage = image;
```
terfendail: There also could be image.channels() == 2 case that now will fall throw here.

alalek: `CV_Assert()` should be restored (on line 337) to check supported number of channels (updated variant).

modules/xfeatures2d/src/freak.cpp
```diff
@@ -333,8 +333,12 @@ void FREAK_Impl::compute( InputArray _image, std::vector<KeyPoint>& keypoints, O
...
 
     // Convert to gray if not already
-    Mat grayImage = image;
-    CV_Assert(grayImage.channels() == 1);
+    Mat grayImage;
+    if( image.channels() == 3 || image.channels() == 4 )
+        cvtColor(image, grayImage, COLOR_BGR2GRAY);
+    else
+        CV_Assert(image.channels() == 1);
+        grayImage = image;
```
terfendail: Assignment is now out of else branch scope so in case image.channels() == 3 it will overwrite produced greyscale image back to RGB

Sahloul: Excuse my Python habits! Thank you..

terfendail:
👍

PR #1438, opened 2017-11-09
Adding Bright edges contribution

fiammante:
This pull request adds a new way of detecting edges used is low resolution image object recognition in real projects (e.g. 50cm per pixel). It corrects surfaces for objects partially under a lighting shadow, and reveal low visibility edges. The result is used to feed further object detection processes particularly in context where small edges or details are important for object discrimination.
The BrightEdges.cpp provides an implementation of an equalized absolute difference of blurs, and an optional further treatment to contrast edges based on finding local minimum along at least two directions. The local minimum detection contrast is a parameter, using 10 as default (10 on 255 maximum).
The kernel size for the gaussian and the average blur can be passed as parameters too.
brightedgesdemo.cpp provides a sample of different results with or without edge contrasting. Default image used the ml.png .
testbrightedges

```
force_builders=Docs,Linux x64
```


fiammante:
Corrected hpp file

modules/ximgproc/src/brightedges.cpp
```diff
@@ -0,0 +1,233 @@
...
+	}
+	
+	int contrastEdges(Mat &minput, Mat &mouput, int contrast) {
+
+		unsigned char *input = (unsigned char*)(mouput.data);
+
+		Mat mwork = minput.clone();
+		mwork.setTo(Scalar(255));
+
+		mouput.setTo(Scalar(255));
```
terfendail: This assignment could be done later when `mwork` is already filled to allow inplace processing as well.

modules/ximgproc/src/brightedges.cpp
```diff
@@ -0,0 +1,233 @@
...
+		return input[iedge.cols * (row)+(col)];
+
+
+	}
+	
+	int contrastEdges(Mat &minput, Mat &mouput, int contrast) {
+
+		unsigned char *input = (unsigned char*)(mouput.data);
+
+		Mat mwork = minput.clone();
```
terfendail: Cloning is excessive here. It would be better to just create the matrix of given size and type.
i.e. `Mat mwork(minput.size(), minput.type(), Scalar(255));`

modules/ximgproc/src/brightedges.cpp
```diff
@@ -0,0 +1,233 @@
...
+#include <signal.h>
+#include <opencv2/ximgproc.hpp>
+namespace cv
+{
+	namespace ximgproc
+	{
+	bool isPixelMinimum(Mat &edge, int row, int col, int contrast) {
+		int count = 0;
+		unsigned char *input = (unsigned char*)(edge.data);
+		int pixel = input[edge.cols * row + col] + contrast;
```
terfendail: It would be better to use `edge.at<unsigned char>(row, col)` or `edge.ptr(row)[col]` to consider the case of non-continuous image(when `edge.step != edge.cols`)

modules/ximgproc/src/brightedges.cpp
fiammante: @terfendail Corrected the code to only explicitly work on the border instead of setting all matrix to 255. Also corrected the minimum detection. (FYI in my project we don't use contrasting as convolutions do the job for us). See the new sample image (I tested on many different images).
testbrightedges

modules/ximgproc/src/brightedges.cpp
```diff
@@ -219,7 +260,7 @@ namespace cv
 
 		absdiff(gblur, bblur, diff);
 
-		cvtColor(diff, gray, COLOR_BGR2GRAY);
+		cvtColor(diff, gray, CV_BGR2GRAY);
```
sturkmen72: please use COLOR_BGR2GRAY

fiammante:
@sturkmen72 Thank you, fixed.

modules/ximgproc/src/brightedges.cpp
fiammante: Replaced clone thank you.

fiammante:
removed all clones

modules/ximgproc/src/brightedges.cpp
```diff
@@ -0,0 +1,260 @@
...
+#include "precomp.hpp"
+#include <iostream>
+#include <signal.h>
+namespace cv
+{
+
+	bool isPixelMinimum(Mat &edge, int row, int col, int contrast) {
+		int count = 0;
+		unsigned char *input = (unsigned char*)(edge.data);
+		int pixel = input[edge.cols * row + col] + contrast - 1;
```
terfendail: It would be better to use edge.at<unsigned char>(row, col) or edge.ptr(row)[col] to consider the case of non-continuous image(when edge.step != edge.cols)

modules/ximgproc/include/opencv2/ximgproc.hpp
```diff
@@ -56,26 +56,21 @@
...
 
-
 /** @defgroup ximgproc Extended Image Processing
   @{
     @defgroup ximgproc_edge Structured forests for fast edge detection
 
 This module contains implementations of modern structured edge detection algorithms,
 i.e. algorithms which somehow takes into account pixel affinities in natural images.
 
-    @defgroup ximgproc_edgeboxes EdgeBoxes
```
terfendail: Changes in this file looks unrelated to PR purpose. Are these changes necessary?

terfendail:
@fiammante
Could you please move the change to other branch?
There is an issue with existing CI infrastructure that makes impossible to test changes proposed from `master` branch.

modules/ximgproc/include/opencv2/ximgproc.hpp
fiammante: @tenderfail I added the interface to the brightedge method if any one wants to use it just as the other  do to expose their methods from the src folder. I can leave this hpp alone if you wish.

fiammante:
@terfendail  ok closing this pull request. Which branch should I use?

fiammante:
@terfendail  I only see a master branch here https://github.com/opencv/opencv_contrib how can I use another branch?

modules/ximgproc/src/brightedges.cpp
fiammante: @terfendail Sounds strange to me. I used to develop device driver and pointer arithmetic (array indexing) was always much compact faster looking at the assembler generated by C++ (I debugged assembler from device drivers written in C++). Is there a particular code with edge.at() that makes it safer?

alalek:
It is about the source branch of PR, not the target branch which is always master for opencv_contrib.
```
	fiammante  wants to merge 7 commits into opencv:master from fiammante:master
                                                                               ^^^
                                                                             fix this
```

Unfortunately GitHub doesn't allow to do this without opening a new pull request from proper branch.

fiammante:
@alalek can you please tell me which source branch I should use?

modules/ximgproc/src/brightedges.cpp
terfendail: OpenCV allows matrix data to be non-continuous so the next matrix row isn't always located just after the previous in memory(it could be due to alignment reasons or specific preprocessing of source matrix). There is a stride field in the matrix header that specify step between rows and edge.at() consider the matrix specific when performs address arithmetic.

modules/ximgproc/include/opencv2/ximgproc.hpp
terfendail: Lines changed in the file are related to the documentation structure. Also removed documentation groups are used in descriptions of other algorithms so the change could even affect documentation build. Thus IMHO these changes should be reverted.

terfendail:
@fiammante you could use any branch name other than master as a source PR branch.
i.e you could create `bright_edges` branch in your repository that will point to the same commit as your master branch

modules/ximgproc/src/brightedges.cpp
fiammante: @tenderfail thank you understood. I shall change code accordingly, but then which branch do I put it in? In the ximgproc samples folder? Thank you in advance

PR #1439, opened 2017-11-09
Cvv parser compatibility issues

Sahloul:
### This pullrequest changes
- Replace getopt with CommandLineParser
- Workaround for VS compatibility with Unicode strings [Ü, Ö, Ä]

### DIFF
--- a/modules/cvv/samples/cvv_demo.cpp
+++ b/modules/cvv/samples/cvv_demo.cpp
@@ -1,5 +1,4 @@
 // system includes
-#include <getopt.h>
 #include <iostream>
 
 // library includes
@@ -16,6 +15,7 @@
 #include <opencv2/cvv/dmatch.hpp>
 #include <opencv2/cvv/final_show.hpp>
 
+using namespace std;
 using namespace cv;
 
 template<class T> std::string toString(const T& p_arg)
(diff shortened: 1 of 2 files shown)


Sahloul:
I have pushed two commits because each addresses different and independent issue.

modules/cvv/samples/cvv_demo.cpp
```diff
@@ -42,29 +42,21 @@ main(int argc, char** argv)
...
+      "{ help h usage ? |    | show this message }"
+      "{ resolution r   |0x0| resolution to width and height in the format WxH }";
+  CommandLineParser parser(argc, argv, keys);
+  string res(parser.get<string>("resolution"));
+  if (parser.has("help") || res == "0x0") {
+    usage();
+    return 0;
+  }
+  char dummych;
+  resolution = new cv::Size();
```
terfendail: Looks like `resolution` isn't released

Sahloul: @terfendail You might be right, but I have not introduced or modified `resolution` allocation/deallocation myself, and it is been there in the original code the exact way. Is it my responsibility to address ALL issues, or just to avoid introducing new ones?

All I care about in this PR is to fix Visual Studio issues, so that it can compile. Of course you can push another PR to do the release.

alalek: "Resolution" parameter were optional in original sample and using of pointer had some sense. Currently this parameter is required, so pointer can be dropped completely.

BTW, Calling parse function (scanf) after CommandLineParser doesn't look very good. Perhaps we could split this into two standalone parameters: "w" and "h". What do you think?

Sahloul: I see! Sorry for the confusion...

modules/cvv/samples/cvv_demo.cpp
```diff
@@ -78,6 +72,7 @@ main(int argc, char** argv)
     printf("Setting resolution to %dx%d\n", resolution->width, resolution->height);
     capture.set(CV_CAP_PROP_FRAME_WIDTH, resolution->width);
     capture.set(CV_CAP_PROP_FRAME_HEIGHT, resolution->height);
+    delete resolution;
```
terfendail: @Sahloul Could you please replace resolution definition to be `cv::Size` structure instead of pointer as @alalek suggests.

alalek: It is optional again (checked here `if (resolution)`)

terfendail: Sorry, missed that.

terfendail:
👍

PR #1449, opened 2017-11-17
Bright edges source code and demo

fiammante:
A new approach for visualizing edges. It uses the absolute difference of 2 blurs to compute gradients creating local minimas for edges. The histogram equalization makes the edges appears darker within a bright zone.
If a contrast value of zero is given no additional process is performed. 
Otherwise a contrasting correction finds the local minimas and sets the edge to black, while setting the surrounding to white. A pixel correction restores line continuity and removes dark spots.
The default contrast value is 1 over a range of 255. This default contrast ensures visualizing edges on chessboard like images or low resolution images where small objects edges may be needed.
chessprocess
Other contrast values will look for deeper local minimas, removing smaller details and only leaving higher gradient variations.
testbrightedges
The images in this entry are the result of brightedgesexample.cpp .


fiammante:
@tenderfail I applied all of your recommendations to this pull request new branch, edge.ptr() etc.  Please let me know if it is now complete.

modules/ximgproc/src/brightedges.cpp
```diff
@@ -0,0 +1,216 @@
...
+		if (line < weight) lines += 1;
+		if (line == 1) return 0;
+		int surround = iedge.ptr(row - 1)[col - 1] +
+			iedge.ptr(row - 1)[col] +
+			iedge.ptr(row - 1)[col + 1] +
+			iedge.ptr(row)[col - 1] +
+			iedge.ptr(row)[col + 1] +
+			iedge.ptr(row + 1)[col - 1] +
+			iedge.ptr(row + 1)[col] +
+			iedge.ptr(row - 1)[col + 1];
```
terfendail: Shouldn't it be `iedge.ptr(row + 1)[col + 1]`?

fiammante:
@terfendail Good catch! Just fixed it.

fiammante:
@terfendail I don't understand why sometimes the build fails on test_stereo which I don't use. Can you help me?

modules/ximgproc/include/opencv2/ximgproc/brightedges.hpp
```diff
@@ -0,0 +1,51 @@
...
+// the use of this software, even if advised of the possibility of such damage.
+//
+//M*/
+#include "opencv2/core.hpp"
+namespace cv
+{
+	CV_EXPORTS_W void  BrightEdges(Mat &_original, Mat &_edgeview, int contrast = 1, int shortrange = 3, int longrange = 9);
+	int correctPixel(Mat &iedge, int row, int col);
+	int contrastEdges(Mat &minput, Mat &mouput, int contrast);
+	bool isPixelMinimum(Mat &edge, int row, int col, int contrast);
```
terfendail: Last three functions look like internal.
Could you please remove their definition from the header?

modules/ximgproc/include/opencv2/ximgproc/brightedges.hpp
```diff
@@ -0,0 +1,51 @@
...
+// loss of use, data, or profits; or business interruption) however caused
+// and on any theory of liability, whether in contract, strict liability,
+// or tort (including negligence or otherwise) arising in any way out of
+// the use of this software, even if advised of the possibility of such damage.
+//
+//M*/
+#include "opencv2/core.hpp"
+namespace cv
+{
+	CV_EXPORTS_W void  BrightEdges(Mat &_original, Mat &_edgeview, int contrast = 1, int shortrange = 3, int longrange = 9);
```
terfendail: Could you please move definition inside nested namespace `xfeatures2d`

terfendail:
At least I could try to find out the reason of test_stereo failures but at the moment it isn't failed yet.

modules/ximgproc/include/opencv2/ximgproc/brightedges.hpp
fiammante: @terfendail doing it now. I have added them because build complained about missing prototype.

terfendail:
Could you please add inclusion of `ximgproc/brightedges.hpp` to `ximgproc.hpp`

modules/ximgproc/samples/brightedgesexample.cpp
```diff
@@ -0,0 +1,112 @@
...
+int main(int argc, const char** argv)
+{
+	CommandLineParser parser(argc, argv, keys);
+	if (parser.has("help"))
+	{
+		help();
+		return 0;
+	}
+	string filename = parser.get<string>(0);
+	image = imread(filename, 1);
```
terfendail: It would be better to use `IMREAD_COLOR` constant here instead of exact value

modules/ximgproc/samples/brightedgesexample.cpp
```diff
@@ -0,0 +1,112 @@
...
+	if (image.empty())
+	{
+		printf("Cannot read image file: %s\n", filename.c_str());
+		help();
+		return -1;
+	}
+	cedge.create(image.size(), image.type());
```
terfendail: `cedge` matrix isn't used inside sample

modules/ximgproc/samples/brightedgesexample.cpp
```diff
@@ -0,0 +1,112 @@
...
+	cvtColor(image, gray, COLOR_BGR2GRAY);
+	// Create a window
+	// //  " original ";
+	namedWindow("Original");
```
terfendail: It isn't necessary to precreate the window if there is no special requirement for window properties. `imshow` could create the window on the own.
IMHO it would be better to shorten the service part of the sample code to make cleaner the usage of described functionality

modules/ximgproc/include/opencv2/ximgproc/brightedges.hpp
fiammante: done. When I move out the prototypes of functions to the cpp I get "warning: no previous declaration for"  xxx. But I have them on top of the file. I don't understand why I get this warning.

modules/ximgproc/samples/brightedgesexample.cpp
fiammante: done

modules/ximgproc/samples/brightedgesexample.cpp
fiammante: removed

modules/ximgproc/samples/brightedgesexample.cpp
```diff
@@ -0,0 +1,112 @@
...
+	namedWindow("Contrast 5");
+	imshow("Contrast 5", edge);
+	// " Contrast 10  \n";
+	BrightEdges(image, edge, 10);
+	namedWindow("Contrast 10");
+	imshow("Contrast 10", edge);
+	//  "wait key ";
+	waitKey(0);
+	// //  "Destroy ";
+	destroyAllWindows();
```
terfendail: Windows destruction is also unnecessary since it will be performed automatically on application termination

modules/ximgproc/samples/brightedgesexample.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include "opencv2/core/utility.hpp"
+#include "opencv2/imgproc.hpp"
+#include "opencv2/imgcodecs.hpp"
+#include "opencv2/highgui.hpp"
+#include <stdio.h>
+#include <iostream>
+#include "opencv2/ximgproc/brightedges.hpp"
+using namespace cv;
+using namespace std;
+int edgeThresh = 1;
```
terfendail: edgeThresh isn't used inside sample

modules/ximgproc/samples/brightedgesexample.cpp
fiammante: removed

alalek:
> why sometimes the build fails on test_stereo

Because test_stereo is buggy and has sporadic failures (at least it uses `rand()` call instead of reproducible `cv::RNG`).

modules/ximgproc/samples/brightedgesexample.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include "opencv2/imgcodecs.hpp"
+#include "opencv2/highgui.hpp"
+#include <stdio.h>
+#include <iostream>
+#include "opencv2/ximgproc.hpp"
+using namespace cv;
+using namespace ximgproc;
+using namespace std;
+int edgeThresh = 1;
+Mat image, gray, gblur, bblur, edge, cedge,diff;
```
terfendail: IMHO it would be better to move matrix declarations closer to the place they are used.

fiammante:
@terfendail moved Mat declaration before place they are used.

modules/ximgproc/src/brightedges.cpp
```diff
@@ -0,0 +1,224 @@
...
+#include "precomp.hpp"
+#include <iostream>
+#include <signal.h>
+namespace cv
+{
+	namespace ximgproc {
+
+	int correctPixel(Mat &iedge, int row, int col);
+	int contrastEdges(Mat &minput, Mat &mouput, int contrast);
+	bool isPixelMinimum(Mat &edge, int row, int col, int contrast);
```
terfendail: IMHO it would be better to define functions as static instead of forward declarations since they aren't used outside of the module

fiammante: Removed unused Mat and moved declaration before where they are used.

fiammante:
@terfendail changed local functions to static

modules/ximgproc/samples/brightedgesexample.cpp
terfendail: Could you please remove `edgeThresh` variable since it is unused.
Also could you please replace tabs with spaces. OpenCV standard for indentation is [4 spaces](https://github.com/opencv/opencv/wiki/Coding_Style_Guide)

modules/ximgproc/samples/brightedgesexample.cpp
terfendail: I think it would be better to remove`namedWindow` calls which are unrelated to the algorithm and just clutter the code

fiammante: @terfendail without namedwindow the windows just do not stay displayed on Windows 10. Just like described here http://answers.opencv.org/question/135303/namedwindow-imshow-not-showing-on-the-screen/  am I missing something?

modules/ximgproc/samples/brightedgesexample.cpp
fiammante: @terfendail and removed unused variable

modules/ximgproc/samples/brightedgesexample.cpp
terfendail: @fiammante mentioned question is related to an issue of specific setup rather than inappropriate way to create window an moreover it doesn't work properly even with `namedWindow`.
I've tried to reproduce the issue on Win10 and at least on my machine window appears normally both with `namedWindow` and without it. However I understand that doesn't mean it works everywhere.
Anyway image creation is unrelated to this PR so I think it could be merged.
BTW `imshow` should be able to create the window on the own so if it doesn't could you please create the issue describing it with details of your setup

modules/ximgproc/samples/brightedgesexample.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "opencv2/imgproc.hpp"
+#include "opencv2/imgcodecs.hpp"
+#include "opencv2/highgui.hpp"
+#include <stdio.h>
+#include <iostream>
+#include "opencv2/ximgproc.hpp"
+using namespace cv;
+using namespace ximgproc;
+using namespace std;
+// define a trackbar callback
```
terfendail: This comment looks unrelated

modules/ximgproc/samples/brightedgesexample.cpp
```diff
@@ -0,0 +1,107 @@
...
+    //  " absdiff ";
+    Mat edge;
+    BrightEdges(image, edge, 0); //  No contrast
+    namedWindow("Absolute Difference");
+    imshow("Absolute Difference", edge);
+    // " default contrast 1 ";
+    BrightEdges(image, edge);
+    namedWindow("Default contrast");
+    imshow("Default contrast", edge); // Default contrast 1
+    // Wait for a key stroke; the same function arranges events processing
```
terfendail: This comment looks unrelated

fiammante:
@terfendail fixed 2 unrelated comments (leftover from previous code).

terfendail:
👍  I think it could be merged now
@fiammante Thanks for contribution

PR #1451, opened 2017-11-17
Fix the runtime bug again.

klchang:
Some time ago, I provided a [PR](https://github.com/opencv/opencv_contrib/pull/1370/files) for fixing the runtime bug. But I unintentionally found the bug had not been solved thoroughly, the commited codes changed maybe when merged. I don't know why it happened. Could someone explain it? Thanks in advance.


alalek:
> I don't know why it happened

It is just another file. Don't worry.

klchang:
@alalek OK. But I have a question that, for merging other branches into master branch in this project, is there any differences compared with generic `git merge`? Because in my project, I had not encountered such case before. Thanks in advance.

PR #1456, opened 2017-11-20
OCL: bioinspired: refactor OpenCL conditions

alalek:
https://github.com/opencv/opencv_contrib/pull/1318#issuecomment-342971194

### DIFF
--- a/modules/bioinspired/src/opencl/retina_kernel.cl
+++ b/modules/bioinspired/src/opencl/retina_kernel.cl
@@ -25,7 +25,7 @@
 //
 //   * Redistribution's in binary form must reproduce the above copyright notice,
 //     this list of conditions and the following disclaimer in the documentation
-//     and/or other oclMaterials provided with the distribution.
+//     and/or other materials provided with the distribution.
 //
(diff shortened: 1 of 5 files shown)


alalek:
@amroamroamro Could you check this with your case?

alalek:
:+1:

amroamroamro:
@alalek 
sorry it seems I never replied..
Yes the changes work for me (tested v3.4.0) 👍 

---

One comment, I can see you replaced `ocl::RetinaOCLImpl* _ocl_retina` with `cv::Ptr<ocl::RetinaOCLImpl> _ocl_retina`. Is it still correct to compare the smart pointer to 0? i.e shouldn't we do this:

    if (!_ocl_retina.empty()) ...

instead of this:

    if (_ocl_retina != 0) ...

or is there an implicit cast from `cv::Ptr` to the raw pointer?

alalek:
Cast operator is here:
https://github.com/opencv/opencv/blob/3.4.0/modules/core/include/opencv2/core/cvstd.hpp#L373

but `.empty()` should be more clear.

amroamroamro:
got it, thanks.

PR #1458, opened 2017-11-21
ppf_icp.py: Python 3 support

nszceta:

### DIFF
--- a/modules/surface_matching/samples/ppf_icp.py
+++ b/modules/surface_matching/samples/ppf_icp.py
@@ -20,8 +20,8 @@ noise = np.random.normal(0.0, 0.1, height * width * 3).reshape((-1, 3))
 noise2 = np.random.normal(0.0, 1.0, height * width)
 
 x, y = np.meshgrid(
-    range(-width/2, width/2),
-    range(-height/2, height/2),
+    range(-width//2, width//2),
+    range(-height//2, height//2),
     sparse=False, indexing='xy'
 )
 z = np.zeros((height, width))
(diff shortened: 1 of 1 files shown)


modules/surface_matching/samples/ppf_icp.py
```diff
@@ -1,27 +1,32 @@
...
+cloud, rotated_cloud = [None] * 3, [None] * 3
+retval, residual, pose = [None] * 3, [None] * 3, [None] * 3
 noise = np.random.normal(0.0, 0.1, height * width * 3).reshape((-1, 3))
 noise2 = np.random.normal(0.0, 1.0, height * width)
 
 x, y = np.meshgrid(
-    range(-width/2, width/2),
-    range(-height/2, height/2),
+    range(-width // 2, width // 2),
+    range(-height // 2, height // 2),
```
alalek: Please eliminate other unnecessary changes. No need to touch unrelated code.
Also proper description for this is: "ppf_icp.py: Python 3 support"

nszceta: / to // is necessary. Python 3 range requires int not float

nszceta: I am verifying why I needed to make this change to make the code work

nszceta: Although not immediately apparent at a glance, type(-width/2) is float not int

alalek: You need this change for Python3, but you don't need other 16 changes related to whitespace modifications.

nszceta:
@alalek fixed. squashed. 

https://github.com/opencv/opencv_contrib/compare/master...nszceta:master

PR #1463, opened 2017-11-24
Affine feature2D - Bug fix

Sahloul:
### This pullrequest changes
Prevent crashing due to invalid co-variant matrix


alalek:
Are there some simple reproducer for this crash?

What is about to add this on line 173:
```
if (Lxmysmooth.empty())
{
    //??? i++;
    continue;
}
```

Sahloul:
@alalek You can reproduce it using the code provided in https://github.com/opencv/opencv_contrib/pull/933#issuecomment-293780764 and the following image:
affine_detector_test

Sahloul:
@alalek A test called `bioinspired` (mostly unrelated to this PR) is keeping to fail on Mac :/

PR #1475, opened 2017-12-02
stereo: fix crash

alalek:
- uninitialized variable 'previous_size'
- incorrect Mat element access pattern via address arithmetic
- invalid 'for' range


alalek:
:+1:

PR #1480, opened 2017-12-09
ximgproc: fix paillou_demo

berak:
### This pullrequest changes
* better cmdline arg check
* remove global pointer to image


terfendail:
👍

PR #1483, opened 2017-12-11
Python: 'cv2.' -> 'cv.' via 'import cv2 as cv'

alalek:
related https://github.com/opencv/opencv/pull/10280
related https://github.com/opencv/opencv/pull/9260


vpisarev:
:+1:

PR #1484, opened 2017-12-14
Fix some wrong comments in tracker.hpp

klchang:
Fix some wrong comments in `modules/tracking/include/opencv2/tracking/tracker.hpp`.

### DIFF
--- a/modules/tracking/include/opencv2/tracking/tracker.hpp
+++ b/modules/tracking/include/opencv2/tracking/tracker.hpp
@@ -527,9 +527,9 @@ class CV_EXPORTS_W Tracker : public virtual Algorithm
 
   virtual ~Tracker();
 
-  /** @brief Initialize the tracker with a know bounding box that surrounding the target
+  /** @brief Initialize the tracker with a known bounding box that surrounded the target
     @param image The initial frame
-    @param boundingBox The initial boundig box
+    @param boundingBox The initial bounding box
 
     @return True if initialization went succesfully, false otherwise
      */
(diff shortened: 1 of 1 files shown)


modules/tracking/include/opencv2/tracking/tracker.hpp
```diff
@@ -1227,12 +1227,12 @@ class CV_EXPORTS_W TrackerKCF : public Tracker
...
 
     /**
-    * \brief Read parameters from file, currently unused
+    * \brief Read parameters from file
     */
     void read(const FileNode& /*fn*/);
 
     /**
-    * \brief Read parameters from file, currently unused
+    * \brief Write parameters in a file
```
terfendail: IMHO here should be `Write parameters to a file`

klchang: IMHO, the principles to fix these wrong comments is as few changes as possible, at the same time, consistent with existed comments. In the line 1116 of tracker.hpp, there is a comment, i.e. `* \brief Write parameters in a file`, so I copyed it. If only from the view of its meaning, 'Write parameters to a file' should be better.

terfendail:
@klchang Thank you for contribution.
The change looks fine for me now, however it couldn't be tested properly due to an issue with existing CI infrastructure that makes impossible to test changes proposed from master branch.
Could you please move the change to other branch?

klchang:
@terfendail OK. I have moved the changes to a new branch in my repository. Thank you for your attention.

PR #1487, opened 2017-12-18
Improved Noise reduction

fiammante:
This pull request improves Brightedges.cpp with linear correction and noise reduction particularly useful for videos coming from webcams.
It also adds webcam support in Brightedgeexamples with side by side comparison with Canny with adjustable threshold and contrast.
noise


fiammante:
@terfendail Good Morning Vitaly, could you help me with the pull request? I have improved my code both algorithm and demo, but my pull request seems unnoticed. What should I do?
Thank you very much in advance and happy year end celebrations. Marc F

alalek:
Some accuracy / validation tests should be added. For example, check that noise reduction really reduces number of edges.

fiammante:
@alalek if I change noise reduction value from -1 to -4 here is a visual test. If ok with you I shall update the brightedgesdemo code. See below the result
image

alalek:
No, I mean the test code of this algorithm. Code without regular tests becomes dead code very quickly.

PR #1496, opened 2017-12-26
bgsegm: add java bindings

### LINKED ISSUE #1495: Can we use the new background subtraction algorithms in Java wrapping?

- OpenCV =>3.4.0
- Operating System / Platform =>Windows 64 Bit
- Compiler =>Visual Studio 2017

##### Detailed description

I want to use **new background subtraction algorithms** in Java wrapping, but when I build OpenCV from source with extra modules for Java, but I cannot see any newly added algorithms for background subtraction in Java Jar file. But when I open `opencv_contrib\modules\bgsegm` then opened `CMakeLists.txt` and inside I did not see wrapping for Java. 

> set(the_description "Background Segmentation Algorithms")
> ocv_define_module(bgsegm opencv_core opencv_imgproc opencv_video opencv_calib3d WRAP python)

So I want to ask that, can I manually add Java or not? Something like this:

> set(the_description "Background Segmentation Algorithms")
> ocv_define_module(bgsegm opencv_core opencv_imgproc opencv_video opencv_calib3d WRAP python **java**)

berak:
resolves #1495

added a bgsegm/misc/java/gen_dict.json file with a reference to video.BackgroundSubtractor, so the classes in bgsegm find their proper base class 

also added some java testcases

### DIFF
--- a/modules/bgsegm/CMakeLists.txt
+++ b/modules/bgsegm/CMakeLists.txt
@@ -1,2 +1,2 @@
 set(the_description "Background Segmentation Algorithms")
-ocv_define_module(bgsegm opencv_core opencv_imgproc opencv_video opencv_calib3d WRAP python)
+ocv_define_module(bgsegm opencv_core opencv_imgproc opencv_video opencv_calib3d WRAP python java)
--- a/modules/bgsegm/misc/java/gen_dict.json
+++ b/modules/bgsegm/misc/java/gen_dict.json
@@ -1,0 +1,13 @@
+{
+    "type_dict": {
+        "BackgroundSubtractor": {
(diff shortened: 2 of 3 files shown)


berak:
hmm, none of the buildbots seems to run the java tests (is that correct ?)

alalek:
Yes (there are no support for Java tests for contrib)

berak:
i just discovered, that the xfeatures2d.SIFT and SURF classes do get the proper 

    import org.opencv.features2d.Feature2D;

without adding a dict file, so there must be another way, closing this for now.

alalek:
Some changes with Java bindings generator were applied here: https://github.com/opencv/opencv/pull/10495

PR #1500, opened 2017-12-30
Add attribute read/write supports to the HDF module.

csukuangfj:
#### Motivation

The current HDF module supports only writing datasets,
which are usually used to store data of large sizes.

Attributes are used to describe a dataset or a group.
It is, in general, small in size, such as a literal string
or a scalar.

---------

This pull request adds supports to read/write
attributes of the root group.

The following data types for attributes are implemented:
 - int
 - double
 - cv::String
 - cv::InputArray


csukuangfj:
@alalek 
Doesn't  the Mac platform in the buildbot support C++11?

csukuangfj:
The added tutorial can be viewed here
<http://pullrequest.opencv.org/buildbot/export/pr_contrib/1500/docs/d5/d10/tutorial_hdf_read_write_attributes.html>

The doc for the added functions can be viewed here
<http://pullrequest.opencv.org/buildbot/export/pr_contrib/1500/docs/d3/d23/classcv_1_1hdf_1_1HDF5.html#a7b1f9a5d6c7bfe011670757995c77f04>

modules/hdf/samples/read_write_attributes.cpp
```diff
@@ -0,0 +1,105 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
```
alalek: BTW, Most path of OpenCV samples doesn't have license header.

modules/hdf/samples/read_write_attributes.cpp
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * Although HDF supports associating attributes with both datasets and groups,
+ * only support for the root group is implemented by OpenCV at present.
+ */
+
+#ifdef __GNUC__
+#  pragma GCC diagnostic ignored "-Wmissing-declarations"
+#  if defined __clang__ || defined __APPLE__
+#    pragma GCC diagnostic ignored "-Wmissing-prototypes"
+#    pragma GCC diagnostic ignored "-Wextra"
```
alalek: Add "static":
```
static void read_write_attributes()
```

modules/hdf/tutorials/table_of_content_hdf.markdown
```diff
@@ -22,3 +22,11 @@ to use this module.
     *Author:* Fangjun Kuang
 
     You will learn how to create, read and write datasets.
+
+-   @subpage tutorial_hdf_read_write_attributes
+
+    *Compatibility:* \> OpenCV 3.0
```
alalek: 3.0 => 3.4 ?

modules/hdf/samples/read_write_attributes.cpp
csukuangfj: Thank you for the review. I have made changes according to you comments.

alalek: Great! Please remove these "#pragma" lines too ("static" should resolve these warnings)

csukuangfj: Indeed! Warnings disappear.

modules/hdf/include/opencv2/hdf/hdf5.hpp
```diff
@@ -103,13 +103,109 @@ class CV_EXPORTS_W HDF5
...
+     *
+     *  @snippet samples/read_write_attributes.cpp snippets_write_str
+     *
+     * @note CV_Error() is called if the given attribute already exists. Use atexists()
+     * to check whether it exists or not beforehand. And use atdelete() to delete
+     * it if it already exists.
+     *
+     * @sa atexists, atdelete, atread
+     */
+    CV_WRAP virtual void atwrite(const int value, const String& atlabel) = 0;
```
alalek: BTW, Current OpenCV interfaces usually pass required "label" before "value":
- imshow(window_name, mat)
- imwrite(file_name, img)
- [FileStorage](https://github.com/opencv/opencv/blob/3.4.0/modules/core/include/opencv2/core/persistence.hpp#L695) utility functions

csukuangfj: Yes, you are right. I implemented it originally like the one you suggested.
But I found that other member functions in the same file put the label after the value,
for example, 
 - [dswrite](https://github.com/opencv/opencv_contrib/blob/master/modules/hdf/include/opencv2/hdf/hdf5.hpp#L298)
 - [dsread](https://github.com/opencv/opencv_contrib/blob/master/modules/hdf/include/opencv2/hdf/hdf5.hpp#L432)

To conform the style used in the same file, I put the label after the value.

alalek:
@csukuangfj Could you take a look on [these failures](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_linux64_no_opt/builds/10065/steps/test_hdf/logs/stdio) of hdf module (debug build on Linux)?

csukuangfj:
@alalek 
It should be fixed now. It happens that [MatSize::opeator()()][1] checks the dimension
of the matrix in the debug version.

Please refer to the pull request https://github.com/opencv/opencv_contrib/pull/1516

[1]: https://github.com/opencv/opencv/blob/master/modules/core/include/opencv2/core/mat.inl.hpp#L1407

PR #1512, opened 2018-01-07
LUCID - Support RGBA

Sahloul:
RGBA are also colored images, but LUCID rejects them!

### This pullrequest changes
- Converts RGBA to RGB

### DIFF
--- a/modules/xfeatures2d/src/lucid.cpp
+++ b/modules/xfeatures2d/src/lucid.cpp
@@ -100,10 +100,15 @@ namespace cv {
         // gliese581h suggested filling a cv::Mat with descriptors to enable BFmatcher compatibility
         // speed-ups and enhancements by gliese581h
         void LUCIDImpl::compute(InputArray _src, std::vector<KeyPoint> &keypoints, OutputArray _desc) {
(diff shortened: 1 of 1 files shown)


modules/xfeatures2d/src/lucid.cpp
```diff
@@ -100,10 +100,15 @@ namespace cv {
         // gliese581h suggested filling a cv::Mat with descriptors to enable BFmatcher compatibility
         // speed-ups and enhancements by gliese581h
         void LUCIDImpl::compute(InputArray _src, std::vector<KeyPoint> &keypoints, OutputArray _desc) {
-            cv::Mat src_input = _src.getMat();
-            if (src_input.empty())
-                return;
-            CV_Assert(src_input.depth() == CV_8U && src_input.channels() == 3);
+            if (_src.empty()) return;
```
alalek: BTW, Any thoughts about this: https://github.com/opencv/opencv/issues/8300 ?

Sahloul: Let us assume the user knows what he is doing! At worst the user can iterate empty collections, but it would be very disappointing to see an exception.

PR #1515, opened 2018-01-09
Update track_vot.cpp

sturkmen72:
### This pullrequest changes

first call of  `dataset->getNextFrame(frame)`  opens 00000002.jpg. i think it should open 00000001.jpg according to track_vot.cpp https://github.com/sturkmen72/opencv_contrib/blob/master/modules/datasets/samples/track_vot.cpp#L85

### DIFF
--- a/modules/datasets/src/track_vot.cpp
+++ b/modules/datasets/src/track_vot.cpp
@@ -134,9 +134,9 @@ namespace cv
                         printf("Error to open groundtruth.txt!!!");
 
                     //Make a list of datasets lengths
-                    int currFrameID = 1;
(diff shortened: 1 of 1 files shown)


sturkmen72:
@alalek thank you. let me change the documentation. pointing out there are VOT 2015 VOT 2016 VOT 2017 datasets. https://docs.opencv.org/master/db/dc1/group__datasets__track.html

alalek:
@sturkmen72 Any updates on this?

sturkmen72:
@alalek  enrichment on dataset documentation is in my todo list. let me provide another PR later.

PR #1521, opened 2018-01-11
delete useless member and  function parameter

103yiran:


alalek:
- [Avoid merge commits or PRs during code update](https://www.google.com/search?q=github+update+from+upstream)
- Always build and test code locally before submitting them.

PR #1523, opened 2018-01-14
aurco: support custom dictionaries via python bindngs

adrianheron:
### This pullrequest changes

As described in the documentation https://docs.opencv.org/3.4.0/d5/dae/tutorial_aruco_detection.html the way to manually create custom dictionaries is to edit the `bytesList` property of the Dictionary object, typically using matrices generated via the `Dictionary::getByteListFromBits()` function.  In the current python bindings `bytesList` is read-only and `Dictionary::getByteListFromBits()` is not exported.

This PR makes it possible to manually create custom dictionaries by making `bytesList`, `markerSize`, and `maxCorrectionBits` writable and exporting `Dictionary::getByteListFromBits()` and `Dictionary::getBitsFromByteList()`.

In python creating a custom dictionary can now be accomplished via this pattern
```python
import numpy as np
from cv2.aruco import custom_dictionary, Dictionary_getByteListFromBits

my_dict = custom_dictionary(2, 3)

bits_H = np.array([[1, 0, 1],
                   [1, 1, 1],
                   [1, 0, 1]],
                  dtype=np.uint8)
bits_I = np.array([[0, 1, 0],
                   [0, 1, 0],
                   [0, 1, 0]],
                  dtype=np.uint8)

my_dict.bytesList = np.array([Dictionary_getByteListFromBits(bits_H)[0],
                              Dictionary_getByteListFromBits(bits_I)[0]])
```

Tested with Python 2.7, here are the marker images drawn from this custom dictionary:
hi_00 hi_01

### DIFF
--- a/modules/aruco/include/opencv2/aruco/dictionary.hpp
+++ b/modules/aruco/include/opencv2/aruco/dictionary.hpp
@@ -61,9 +61,9 @@ namespace aruco {
 class CV_EXPORTS_W Dictionary {
 
     public:
-    CV_PROP_RW Mat bytesList;      // marker code information
-    CV_PROP int markerSize;        // number of bits per dimension
-    CV_PROP int maxCorrectionBits; // maximum number of bits that can be corrected
+    CV_PROP_RW Mat bytesList;         // marker code information
+    CV_PROP_RW int markerSize;        // number of bits per dimension
+    CV_PROP_RW int maxCorrectionBits; // maximum number of bits that can be corrected
 
 
     /**
--- a/modules/aruco/include/opencv2/aruco/dictionary.hpp
+++ b/modules/aruco/include/opencv2/aruco/dictionary.hpp
@@ -61,7 +61,7 @@ namespace aruco {
 class CV_EXPORTS_W Dictionary {
 
     public:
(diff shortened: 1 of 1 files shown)


kamrausch:
I think this will do what I want. My goal is to take a single snapshot with a stereo pair of cameras using several unique boards scattered across the FOV. With this method, I can loop through stationary boards rather than images while moving the cameras on a gimble mount or something similar. 

Using this pull request, I can create N custom dictionaries pulling from one common dictionary like 4x4_250 where the beginning of each new custom dictionary starts with the ending of the previous custom dictionary. 

It would be convienent to be able to specify the starting marker ID in the board creation but I see that isn’t written that way in the c++ code either.

kamrausch:
Thanks for the pointer to the PR btw.

terfendail:
👍

devernay:
It would also be great if we could either have ad "id_offset" parameter to `CharucoBoard_create` to start a given charuco board at some id, or if CharucoBoard.ids could be writable (as with the C++ API). Currently, the only way to may a charuco board starting at - say - id 24 is to create a custom dict containing markers 24 and following from the original dict.

PR #1535, opened 2018-01-26
Fixed several warnings produced by clang-6.0.0

mshabunin:
### This pullrequest changes

- removed `register` specifier (deprecated in C++17)
- disabled _sfm_ module when configuring with cmake 2.x
- added `static` to several functions (_sfm_ module)
- infinite recursion in `Project` function (_sfm_ module)

### DIFF
--- a/modules/bioinspired/src/parvoretinafilter.hpp
+++ b/modules/bioinspired/src/parvoretinafilter.hpp
@@ -244,7 +244,7 @@ private:
         float *parvocellularOutputON_PTR= parvocellularOutputON+r.start;
         float *parvocellularOutputOFF_PTR= parvocellularOutputOFF+r.start;
 
-            for (register int IDpixel=r.start ; IDpixel!=r.end ; ++IDpixel)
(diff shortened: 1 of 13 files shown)


modules/sfm/CMakeLists.txt
```diff
@@ -1,6 +1,11 @@
 set(OPENCV_MODULE_IS_PART_OF_WORLD FALSE)
 set(the_description "SFM algorithms")
 
+if(${CMAKE_VERSION} VERSION_LESS 3.0)
+  message(STATUS "Module opencv_sfm has been disabled: does not support cmake version less than 3.0")
+  ocv_module_disable(sfm)
```
alalek: Actually it depends on external packages, so it can't be (and should not be) handled in OpenCV.

Ubuntu 14.04 with CMake 2.8.12.2 [works](http://pullrequest.opencv.org/buildbot/builders/master-contrib-lin64/builds/10350/steps/cmake/logs/stdio) well.

mshabunin: I've reverted this change.

PR #1536, opened 2018-01-26
Allow setting the features on the RandomPatternCornerFinder

caseymcc:
### This pullrequest changes

This allows the RandomPatterCornerFinder's features to be set rather than computed. This is useful when the pattern has already been located in some previous process and just re-used later.
Also includes minor change to the getObjectPoints and getImagePoints, changes return type to const &, since a vector is in use,  this keeps it from copying if not needed (cv::Mat likely would shallow copy).


modules/ccalib/src/randpattern.cpp
```diff
@@ -249,6 +249,16 @@ void RandomPatternCornerFinder::loadPattern(cv::Mat patternImage)
...
 }
 
+void RandomPatternCornerFinder::loadPattern(cv::Mat patternImage, std::vector<cv::KeyPoint> patternKeyPoints, Mat patternDescriptors)
+{
+	_patternImage=patternImage.clone();
+	if(_patternImage.type()!=CV_8U)
+		_patternImage.convertTo(_patternImage, CV_8U);
+	_patternImageSize=patternImage.size();
+    _keypointsPattern=patternKeyPoints;
+    _descriptorPattern=patternDescriptors.clone();
```
terfendail: I think it make sense to check provided descriptors matrix size to ensure it is compatible with internal descriptor extractor.

caseymcc: along the lines of something like this?

    CV_Assert(patternKeyPoints.size()==patternDescriptors.rows);
    CV_Assert(patternDescriptors.cols==_descriptor->descriptorSize());
    CV_Assert(patternDescriptors.type()==_descriptor->descriptorType());

or actual return code?

terfendail: I think it is enough to add assertions. But type of descriptor matrix should either be checked against CV_32F or converted.

caseymcc: Yeah, I noticed the CV_32F conversion when I went back to fix the assertions. I am curious why the descriptors are converted to CV_32F. need to check the matchers and see if it is actually a requirement on their part. But this should work for this item.

terfendail:
👍

PR #1545, opened 2018-02-06
ORB support for SIFT keypoints

Open3D-admin:
related to opencv/opencv#10556
### This pullrequest changes
- Add unit tests for ORB descriptor with SIFT detector

### DIFF
--- a/modules/xfeatures2d/test/test_rotation_and_scale_invariance.cpp
+++ b/modules/xfeatures2d/test/test_rotation_and_scale_invariance.cpp
@@ -801,6 +801,16 @@ TEST(Features2d_RotationInvariance_Descriptor_BoostDesc_BINBOOST_256, regression
     test.safe_run();
 }
 
+TEST(Features2d_RotationInvariance_Descriptor_10555, regression)
+{
+    //opencv/opencv#10555: Enforce ORB support for SIFT keypoints
+    DescriptorRotationInvarianceTest test(SIFT::create(),
+                                          ORB::create(),
+                                          NORM_HAMMING,
+                                          0.9999f);
+    test.safe_run();
+}
+
 /*
  * Detector's scale invariance check
  */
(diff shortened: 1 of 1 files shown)


modules/xfeatures2d/test/test_rotation_and_scale_invariance.cpp
```diff
@@ -978,4 +988,14 @@ TEST(Features2d_ScaleInvariance_Descriptor_BoostDesc_BINBOOST_256, regression)
     test.safe_run();
 }
 
+TEST(Features2d_ScaleInvariance_Descriptor_10555, regression)
+{
+    //opencv/opencv#10555: Enforce ORB support for SIFT keypoints
+    DescriptorScaleInvarianceTest test(SIFT::create(),
+                                       ORB::create(),
+                                       NORM_HAMMING,
+                                       0.0022f);
```
Sahloul: @terfendail Any idea why the score is that low? I wonder if this has any relation to opencv/opencv#10690

alalek: Why do you select `NORM_HAMMING` norm here?

Sahloul: In [the documentation](https://docs.opencv.org/3.0-beta/doc/py_tutorials/py_feature2d/py_orb/py_orb.html#orb-in-opencv), it is stated that:
> Another parameter, WTA_K decides number of points that produce each element of the oriented BRIEF descriptor. By default it is two, ie selects two points at a time. **In that case, for matching, NORM_HAMMING distance is used**

Am I missing something?

terfendail: PR#10690 make possible to change `firstLevel` option but doesn't change the default. Also it doesn't change behavior for `firstLevel` equal to 0 (at least it shouldn't). So I think this change is unrelated

Sahloul: @terfendail  Although `ORB::create()` uses `firstLevel=0`, passing it SIFT keypoints change `firstLevel` value (check [here](https://github.com/opencv/opencv/pull/10556/files#diff-cda2e4de89b5dc7ca181848aa0e5a461R1020) for details). Therefore, I wonder if this has any relation to the change in concern..

alalek:
Source branch was removed, closing this.

PR #1563, opened 2018-02-15
fix missing include that causes compilation fail in windows

jasjuang:
`std::array` is used in `dual_filter_view.hpp` but the header `array` wasn't included. This still compiles on linux because it somehow includes it implicitly, however this will cause compilation fail on windows. This PR fixes it.

### DIFF
--- a/modules/cvv/src/view/dual_filter_view.hpp
+++ b/modules/cvv/src/view/dual_filter_view.hpp
@@ -2,6 +2,7 @@
 #define CVVISUAL_DUAL_FILTER_VIEW_HPP
 
 // STD
+#include <array>
 #include <string>
 #include <unordered_map>
 // OpenCV


alalek:
> however this will cause compilation fail on windows

There is an issue with "std::array" detection in MSVC. See this PR: https://github.com/opencv/opencv/pull/10589 (closed due stalled non-green builds - build warnings)

jasjuang:
@alalek Is `modules/core/include/opencv2/core/cvdef.h` going to be included by everyone? If yes, are you suggesting me to update https://github.com/opencv/opencv/pull/10589 with `_MSVC_LANG`? If not, are you suggesting me to add the guard for including `array` for `dual_filter_view.hpp`?

alalek:
https://github.com/opencv/opencv/pull/10589 is integrated.
@jasjuang Please try again with the latest master.

jasjuang:
Thanks it works for me now

PR #1570, opened 2018-02-26
Apriltag

szk1509:
### Improves ArUco marker detection

Improves the accuracy of the ArUco marker detection by using the advantages of the AprilTag2 project
"AprilTag 2: Efficient and robust fiducial detection"
https://april.eecs.umich.edu/software/apriltag.html


modules/aruco/src/apriltag_quad_thresh.cpp
```diff
@@ -0,0 +1,1600 @@
+/* Copyright (C) 2013-2016, The Regents of The University of Michigan.
+All rights reserved.
+
+This software was developed in the APRIL Robotics Lab under the
+direction of Edwin Olson, ebolson@umich.edu. This software may be
+available under alternative licensing terms; contact the address above.
```
alalek: @szk1509 Could you ask original authors about using of OpenCV license for this code?

We planning to move aruco module into the main repository in OpenCV 4.0, that requires clean code as much as possible from legal point of view.

modules/aruco/src/apriltag_quad_thresh.cpp
```diff
@@ -0,0 +1,1600 @@
...
+#undef MAYBE_SWAP
+
+    // a merge sort with temp storage.
+
+    // Use stack storage if it's not too big.
+    int stacksz = sz;
+    if (stacksz > 1024)
+        stacksz = 0;
+
+    struct pt *_tmp_stack = new struct pt[stacksz];
```
alalek: Dynamic array size is C99 extension (not supported in C++).
Perhaps you need to use alloca() instead.
Or `cv::AutoBuffer<struct pt, 1024> _tmp_stack(sz);` (but it doesn't not completely equal to variable size arrays - be careful with recursive functions).

**Update**: I see "new" here. Probably you should use `cv::AutoBuffer<struct pt, 128> _tmp_stack(sz)` here without any "stacksz" trick.

modules/aruco/src/apriltag_quad_thresh.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+// limitation: image size must be <32768 in width and height. This is
+// because we use a fixed-point 16 bit integer representation with one
+// fractional bit.
+
+#ifndef _OPENCV_APRIL_QUAD_THRESH_HPP_
+#define _OPENCV_APRIL_QUAD_THRESH_HPP_
+#ifdef __cplusplus
+
+#include "precomp.hpp"
```
alalek: `#include "precomp.hpp"` should be used in cpp file only as the first preprocessor directive (without any #if). Remove it from .hpp files and add to all module `.cpp` files.

modules/aruco/src/zarray.hpp
```diff
@@ -0,0 +1,172 @@
...
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+The views and conclusions contained in the software and documentation are those
+of the authors and should not be interpreted as representing official policies,
+either expressed or implied, of the Regents of The University of Michigan.
+*/
+
+#ifndef _OPENCV_ZARRAY_HPP_
+#define _OPENCV_ZARRAY_HPP_
+#ifdef __cplusplus
```
alalek: OpenCV is C++ library, so `#ifdef __cplusplus` is not required.

modules/aruco/src/zarray.hpp
```diff
@@ -0,0 +1,172 @@
...
+
+#include <assert.h>
```
alalek: `assert()` calls should be replaced to `CV_Assert()`. Remove this include to catch `assert()` statements.

modules/aruco/src/apriltag_quad_thresh.cpp
paroj: we got the permission, here is the reply by Edwin Olson:
> Hi Pavel,
>
> The 3 clause BSD license is also acceptable. 
>
> Looking forward to seeing this in OpenCV!
>
> Best,
>
> -Ed

alalek: @paroj @edwinolson Thanks!

modules/aruco/src/apriltag_quad_thresh.cpp
```diff
@@ -518,6 +515,8 @@ int quad_segment_agg(int sz, struct line_fit_pt *lfps, int indices[4]){
 
         fit_line(lfps, sz, rv->left, rv->right, NULL, NULL, &rv->err);
 
+        //TODO is finite CV_Assert():
+        CV_DbgAssert (isfinite(-rv->err) && "zmaxheap_add: Trying to add non-finite number to heap.  NaN's prohibited, could allow INF with testing");
```
alalek: BTW, There is `cvIsNaN(double)` function (if needed to allow "INF").

modules/aruco/CMakeLists.txt
```diff
@@ -1,2 +1,3 @@
 set(the_description "ArUco Marker Detection")
 ocv_define_module(aruco opencv_core opencv_imgproc opencv_calib3d WRAP python java)
+ocv_warnings_disable(CMAKE_CXX_FLAGS -Wunused-parameter)
```
alalek: Disabling warnings globally should be avoided.
- use `CV_UNUSED(unused_parameter)` macro
- or avoid parameter name in declaration "..., int /*unused_paramer*/, ... " (doesn't work in case of conditional compilation with "#if").

modules/aruco/src/apriltag_quad_thresh.cpp
```diff
@@ -0,0 +1,1603 @@
...
+    }
+
+#undef MAYBE_SWAP
+
+    // a merge sort with temp storage.
+
+    // Use stack storage if it's not too big.
+    int stacksz = sz;
+    if (stacksz > 1024)
+        stacksz = 0;
```
alalek: `stacksz` is not needed anymore (after `cv::AutoBuffer` replacement).

modules/aruco/CMakeLists.txt
szk1509: oh, sorry... my bad, i was trying to get rid of 4244. It is ok if i disable 4244 globally?

szk1509: btw, thank you for your help :)

alalek: 4244 is "conversion from 'type1' to 'type2', possible loss of data"

OpenCV has static code scans and disabling of these warnings doesn't help really (static code analyzers has own settings).
So it is better to extend type of target variable (if it make sense and doesn't affect performance), or make an explicit conversion (like double -> float).

modules/aruco/src/apriltag_quad_thresh.hpp
```diff
@@ -0,0 +1,143 @@
...
+#include "opencv2/aruco.hpp"
+#include "unionfind.hpp"
+#include "zmaxheap.hpp"
+#include "zarray.hpp"
+
+namespace cv {
+namespace aruco {
+
+static inline int64_t u64hash_2(uint64_t x) {
+    return (2654435761 * x) >> 32;
```
paroj: this must be `uint32_t u64hash_2(uint64_t x)`. see: https://stackoverflow.com/questions/11871245/knuth-multiplicative-hash

modules/aruco/src/apriltag_quad_thresh.cpp
```diff
@@ -0,0 +1,1606 @@
...
+            }
+
+            // solve
+            double L0 = W00*B0 + W01*B1;
+
+            // compute intersection
+            quad->p[i][0] = lines[i][0] + L0*A00;
+            quad->p[i][1] = lines[i][1] + L0*A10;
+
+            if (0) {
```
alalek: Enabling of this validation code breaks tests (assertion raised). Could you take a look on this?

szk1509: :thinking: you are right, i've compare it with the original code and it also breaks at that point...

alalek: Looks like there is a typo below (A10 -> A01):
```.diff
-double x = lines[(i+1)&3][0] - L1*A10;
+double x = lines[(i+1)&3][0] - L1*A01;
```

alalek:
I pushed several commits with changes. Please take a look.

szk1509:
ups and thank you so much! :)

PR #1579, opened 2018-03-15
update structured_edge_detection

yarglawaldeg:
update structured_edge_detection to read models from updated p. dollar toolbox

### This pullrequest changes

current version returns blurry edge map with "new" models from updated p. dollar toolbox
the solution can be to set sharpen parameter from the models to 0 and update the structured_edge_detection.cpp file as suggested here : http://answers.opencv.org/question/72962/trained-model-for-opencv-structured-edge-detection/?answer=95989#post-id-95989

### DIFF
--- a/modules/ximgproc/src/structured_edge_detection.cpp
+++ b/modules/ximgproc/src/structured_edge_detection.cpp
@@ -801,9 +801,11 @@ protected:
                 {// for j,k in [0;width)x[0;nTreesEval)
 
                     int currentNode = pIndex[j*nTreesEval + k];
+                    
+                    const int nBnds = (__rf.edgeBoundaries.size() - 1) / (nTreesNodes * nTrees);
 
-                    int start  = __rf.edgeBoundaries[currentNode];
-                    int finish = __rf.edgeBoundaries[currentNode + 1];
+                    int start = __rf.edgeBoundaries[currentNode * nBnds];
+                    int finish = __rf.edgeBoundaries[currentNode * nBnds + 1];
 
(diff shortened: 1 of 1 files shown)


bommo1:
related https://github.com/opencv/opencv_contrib/issues/1282  https://github.com/opencv/opencv_contrib/issues/1559

modules/ximgproc/src/structured_edge_detection.cpp
```diff
@@ -801,9 +801,11 @@ class StructuredEdgeDetectionImpl : public StructuredEdgeDetection
                 {// for j,k in [0;width)x[0;nTreesEval)
 
                     int currentNode = pIndex[j*nTreesEval + k];
-
-                    int start  = __rf.edgeBoundaries[currentNode];
-                    int finish = __rf.edgeBoundaries[currentNode + 1];
+                    size_t sizeBoundaries = __rf.edgeBoundaries.size();
+                    int convertedBoundaries = static_cast<int>(sizeBoundaries);
+                    int nBnds = (convertedBoundaries - 1) / (nTreesNodes * nTrees);
```
alalek: Looks like `nBnds` is a constant for these loops (so it should be moved out).

Does "((convertedBoundaries - 1) % (nTreesNodes * nTrees)) == 0" always (remainder == 0)?
Or should we calculate original expression exactly (multiplication at first): `currentNode * (convertedBoundaries - 1) / (nTreesNodes * nTrees)`?

vpisarev:
:+1:

PR #1587, opened 2018-03-21
aruco: add a user supplied random seed for creating custom dictionaries

### LINKED ISSUE #1586: aruco: Inconsistent custom dictionary generation
When generating a custom dictionary, aruco uses the platform-specific rand() function to start generating a custom dictionary. Because of this, it does not consistently generate the same set of markers.

There is also no way to readily export or import markers (documentation on this point is old and outdated and doesn't appear to actually map to current practice).  This means that when using a custom dictionary, you must rely on the dictionary being generated the same way with the same parameters each time.

And the result is that markers generated on a Linux build will be different than those on a Windows build, etc. A solution would be to (a) use a known algorithm which performs consistently across platforms  like Mersenne Twister and (b) allow the user to pass in a seed.

Alternatively, adding convenience functions for exporting and importing custom generated dictionaries would also be a solution.

As it stands, the custom dictionary generation is a fundamentally broken workflow which will break in unpredictable ways. The tokens you physically construct today might start inexplicably failing tomorrow because the PRNG changed its output.

berak:
resolves #1586

use `cv::theRNG()` and `cv::setRNGSeed()` to make results reproducable.

### DIFF
--- a/modules/aruco/include/opencv2/aruco/dictionary.hpp
+++ b/modules/aruco/include/opencv2/aruco/dictionary.hpp
@@ -84,14 +84,14 @@ class CV_EXPORTS_W Dictionary {
     /**
      * @see generateCustomDictionary
      */
-    CV_WRAP_AS(create) static Ptr<Dictionary> create(int nMarkers, int markerSize);
+    CV_WRAP_AS(create) static Ptr<Dictionary> create(int nMarkers, int markerSize, int randomSeed=0);
 
 
     /**
      * @see generateCustomDictionary
      */
     CV_WRAP_AS(create_from) static Ptr<Dictionary> create(int nMarkers, int markerSize,
-            const Ptr<Dictionary> &baseDictionary);
+            const Ptr<Dictionary> &baseDictionary, int randomSeed=0);
 
     /**
      * @see getPredefinedDictionary
@@ -176,7 +176,8 @@ CV_EXPORTS_W Ptr<Dictionary> getPredefinedDictionary(int dict);
   */
 CV_EXPORTS_AS(custom_dictionary) Ptr<Dictionary> generateCustomDictionary(
         int nMarkers,
-        int markerSize);
+        int markerSize,
+        int randomSeed=0);
 
 
 /**
@@ -185,6 +186,7 @@ CV_EXPORTS_AS(custom_dictionary) Ptr<Dictionary> generateCustomDictionary(
   * @param nMarkers number of markers in the dictionary
   * @param markerSize number of bits per dimension of each markers
   * @param baseDictionary Include the markers in this dictionary at the beginning (optional)
+  * @param randomSeed a user supplied seed for theRNG()
   *
   * This function creates a new dictionary composed by nMarkers markers and each markers composed
   * by markerSize x markerSize bits. If baseDictionary is provided, its markers are directly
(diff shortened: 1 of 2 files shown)


modules/aruco/include/opencv2/aruco/dictionary.hpp
```diff
@@ -84,14 +84,14 @@ class CV_EXPORTS_W Dictionary {
     /**
      * @see generateCustomDictionary
      */
-    CV_WRAP_AS(create) static Ptr<Dictionary> create(int nMarkers, int markerSize);
+    CV_WRAP_AS(create) static Ptr<Dictionary> create(int nMarkers, int markerSize, int randomSeed=~0);
```
alalek: Please use "cv::uint64" type for seed. "int" is not a good case for cross-platform support.

Why is `~0` seed value used here?
Simple "0" should work too:
```
inline RNG::RNG(uint64 _state) { state = _state ? _state : 0xffffffff; }
```

modules/aruco/src/dictionary.cpp
```diff
@@ -377,8 +378,8 @@ static int _getSelfDistance(const Mat &marker) {
 /**
  */
 Ptr<Dictionary> generateCustomDictionary(int nMarkers, int markerSize,
-                                         const Ptr<Dictionary> &baseDictionary) {
-
+                                         const Ptr<Dictionary> &baseDictionary, int randomSeed) {
+    setRNGSeed(randomSeed); // #1586
```
alalek: In internal implementation we can pass `cv::RNG& rng` instead of initial seed value.

Perhaps we should not touch OpenCV seed in this case (this code is data generator, not a stochastic algorithm, like RANSAC).

berak: good idea, that would also solve the setRNGSeed() problem.
plan now: use size_t and cv::RNG, and pass on the latter internally.

modules/aruco/include/opencv2/aruco/dictionary.hpp
berak: > Why is ~0 seed value used here?

i was too lazy to type `0xffffffff;` , but sure, 0 is the better choice.

> why is it int, not uint64 ?
(setRNGSeed() wants it so.)
https://github.com/opencv/opencv/blob/0366c1b0932141d6f1a3d49d996d8aca22330a7e/modules/core/include/opencv2/core.hpp#L2271

hmm, there is support for `int64` in python/java, but not for `uint64`. would a `size_t` do ?

alalek: `size_t` is not good solution for cross-platform too (it can have 32/64-bits).
setRNGSeed() uses `int` but RNG() uses `uint64`. This is not consistent behavior in OpenCV for now.
Perhaps you are right and the "int" type is required for bindings.

berak: so, int64 (truncates sign) or int ? (truncates 32bits)

alalek: Lets use "int" to avoid issues with bindings. 32-bit is enough for RNG generator for now (and it can be easily extended to 64-bit without huge breaking of existed code)

berak: ok. will do that.

PR #1606, opened 2018-04-11
linemod: export Template::features for python

paroj:


alalek:
@paroj Is this patch is for master (4.x) branch only or should be included into 3.4+ too?

paroj:
it can actually be used for both. I should target the 3.4 branch then?

alalek:
Changed base branch to '3.4' (it will be merged into 'master' on next week after PR merge)

PR #1613, opened 2018-04-16
ximgproc: fix weightedMedianFilter() hang

### LINKED ISSUE #1610: weightedMedianFilter hangs with some input images
- OpenCV => 3.4.1-312-g7e5581c
- opencv_contrib => 3.4.1-70-g2aaa67a
- Operating System / Platform => GNU/Linux x86_64 (Linux Mint 18.3 Sylvia)
- Compiler => gcc  5.4.0 20160609

##### Detailed description
The weightedMedianFilter () sometimes hangs with some particular input images.
The hang seems not depend on particular parameters (sigma, radius, weight type or mask image) as well as on joint image, but just on input image content.

It works fine with total majority of input images but fails with some particulars.
   
##### Steps to reproduce
Attached is an example of 16-bit single-channel grayscale png image (uint16_t), which causes weightedMedianFilter() hangup . The image was converted to CV_32FC1 before passing to  weightedMedianFilter().

If I change pixel values in this image in a some way  (for example convert the image to 8 bit, make some brighness/contrast stretch, etc) then no hangups happened.

ss-2 003

The outline of code example is below:

    ```.cpp

     // C++ code example
     Mat input_image = cv::imread("ss-2.003.png", IMREAD_UNCHANGED);
     
     input_image.convertTo(input_image, CV_32F);
   
     weightedMedianFilter(Mat::ones(input_image.size(), CV_8U),
        input_image, output_image, raduis, sigma,  WMF_OFF);
    
     // No return: hang up inside of weightedMedianFilter()
 
    ```
Attached is complete code exampe to reproduce.
[weightedMedianFilter-hangup.tar.gz](https://github.com/opencv/opencv_contrib/files/1911165/weightedMedianFilter-hangup.tar.gz)

Just unpack, make and run:
$ ./weightedMedianFilter-hangup ss-2.003.png 

Will see 100% cpu load with no return.

alalek:
resolves #1610

### DIFF
--- a/modules/ximgproc/src/weighted_median_filter.cpp
+++ b/modules/ximgproc/src/weighted_median_filter.cpp
@@ -55,51 +55,64 @@ using namespace cv::ximgproc;
 void from32FTo32S(Mat &img, Mat &outImg, int nI, float *mapping)
 {
     int rows = img.rows, cols = img.cols;
-    int alls = rows * cols;
+    size_t alls = (size_t)rows * cols;
 
+    CV_Assert(img.isContinuous());
     float *imgPtr = img.ptr<float>();
     typedef pair<float,int> pairFI;
-    pairFI *data = (pairFI *)malloc(alls*sizeof(pairFI));
+    std::vector<pairFI> data(alls);
 
     // Sort all pixels of the image by ascending order of pixel value
-    for(int i=0;i<alls;i++){
-        data[i].second = i;
-        data[i].first = imgPtr[i];
+    for (size_t i = 0; i < alls; i++)
+    {
+        pairFI& d = data[i];
+        d.second = i;
+        d.first = imgPtr[i];
     }
-    sort(data,data+alls);
+
+    struct PixelValueOrder {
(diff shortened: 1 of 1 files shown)


modules/ximgproc/src/weighted_median_filter.cpp
```diff
@@ -55,51 +55,64 @@ using namespace cv::ximgproc;
...
+    const float maxRange = (float)(maxVal - minVal);
 
     float l = 0, r = maxRange*2.0f/nI;
     // Perform binary search on error bound
-    while(r-l > th)
+    while (r > l)
     {
-        float m = (r+l)*0.5f;
+        float m = (r + l) * 0.5f;
+        if (m == r || m == r)
```
berak: is that a typo here ?

also, `m`  is set to either `l` or `r` below, based on the result in `suc`, so it it meant:

    if (m==r || m==l)

it will bail out in the 2nd iteration already.

alalek:
:+1:

PR #1617, opened 2018-04-20
cleanup: drop dnn_modern

alalek:
- almost all functionality of this module is covered by OpenCV dnn module
- module build has been broken for the long time due strong conflict with OpenCV dnn module
- requires external tinydnn library
- there are no tests


vpisarev:
:+1:

isgursoy:
Why dont we replace it with the dlib instead?

PR #1619, opened 2018-04-22
changed error messages from cout to cerr

### LINKED ISSUE #1618: line_descriptor module error messages print to stdout
- OpenCV => 3.4.1

##### Detailed description
All error messages in the line_descriptor module prints to stdout.  They should instead print to stderr.

kevinabrandon:
resolves #1618

### This pullrequest changes
This replaces all instances of cout with cerr in the line_descriptor module.

### DIFF
--- a/modules/line_descriptor/src/binary_descriptor.cpp
+++ b/modules/line_descriptor/src/binary_descriptor.cpp
@@ -416,7 +416,7 @@ void BinaryDescriptor::detect( const Mat& image, CV_OUT std::vector<KeyLine>& ke
 {
   if( image.data == NULL )
   {
-    std::cout << "Error: input image for detection is empty" << std::endl;
+    std::cerr << "Error: input image for detection is empty" << std::endl;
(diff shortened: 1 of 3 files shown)


vpisarev:
@kevinabrandon, thank you! Since you are working on this PR, may I ask you to replace those printfs with the call to `CV_Error()`?

kevinabrandon:
I’m not sure simply replacing them with cv_error is what you want, that would end up changing the programs output.  It seems that a lot of these were originally intended as debug messages, so there might be a better argument to remove them altogether.

kevinabrandon:
For example, if you try to detect lines on an image with no lines in it, you get [here](https://github.com/opencv/opencv_contrib/blob/master/modules/line_descriptor/src/binary_descriptor.cpp#L2204).  Right now it returns correctly without error that there are no lines in the image, but if the cout was replaced with cv_error then you'd have an exception thrown.

PR #1623, opened 2018-04-25
android: fix contrib build

alalek:
Main PR: https://github.com/opencv/opencv/pull/11391

- simplified big/little-endian detection


alalek:
:+1:

PR #1634, opened 2018-05-04
remove bug when src is vector Point2i in FourierDescriptors

LaurentBerger:
Resolves http://answers.opencv.org/question/190847/how-to-use-opencv-contrib-python-fourierdescriptor/?comment=190856#post-id-190856

Add python sample

Add opencv test

### DIFF
--- a/modules/ximgproc/src/fourier_descriptors.cpp
+++ b/modules/ximgproc/src/fourier_descriptors.cpp
@@ -5,8 +5,7 @@
 #include "precomp.hpp"
 #include <math.h>
 #include <vector>
-#include <iostream>
-#include <fstream>
+
 /*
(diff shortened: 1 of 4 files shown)


LaurentBerger:
@berak Can you check my first python example?

modules/ximgproc/src/fourier_descriptors.cpp
```diff
@@ -295,6 +298,7 @@ void transformFD(InputArray _src, InputArray _t,OutputArray _dst,  bool fdContou
     if (!fdContour)
     {
         Mat ctr1 = _src.getMat();
+        std::ofstream fs;
```
alalek: Do we need this here? (and fstream include)

PR #1642, opened 2018-05-15
Rename `near` to `zNear`

fegorsch:
### This pullrequest changes

`near` is an illegal variable name on Windows (if `windef.h` is included),
because a macro with the same name is defined.

Someone else already put your rage into words, see
http://lolengine.net/blog/2011/3/4/fuck-you-microsoft-near-far-macros.


alalek:
/cc @paroj 

BTW, Does `#undef near` works?

PR #1645, opened 2018-05-21
Check if faces detected or not.

krshrimali:
Added if condition, whether faces are detected or not, will help to keep program running for some images.
If imwrite used along with imshow, it may produce error for :  No faces found.Aborting.. in function 'fit' (Bad Argument).

### This pullrequest changes
Adds an if condition, for the condition when no face detected. Thus, proper flow of the code maintained.


alalek:
I removed the trailing whitespace and rebased to 3.4 branch.

alalek:
Error is not related to this patch. Fix is in progress (to the main opencv repository): https://github.com/opencv/opencv/pull/11555

krshrimali:
Okay, great!

PR #1649, opened 2018-05-22
Adding valuable comments for params

krshrimali:
Related to comment : https://github.com/opencv/opencv_contrib/issues/1614#issuecomment-390930873 on issue 1614, resolved by setting parameters. Valuable comments can help users identify proper usage of the code, also not mentioned in the source code in detail. 

### This pullrequest changes
Added important comments for users to work around with parameters to configure FacemarkLBF.

### DIFF
--- a/modules/face/samples/facemark_demo_lbf.cpp
+++ b/modules/face/samples/facemark_demo_lbf.cpp
@@ -43,6 +43,9 @@ Mentor: Delia Passalacqua
 #include "opencv2/imgproc.hpp"
 #include "opencv2/face.hpp"
 
+// include this header if drawLandmarks function is to be used
+// #include "drawLandmarks.hpp" 
+
 using namespace std;
 using namespace cv;
 using namespace cv::face;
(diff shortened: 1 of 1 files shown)


modules/face/samples/facemark_demo_lbf.cpp
```diff
@@ -43,6 +43,9 @@ Mentor: Delia Passalacqua
 #include "opencv2/imgproc.hpp"
 #include "opencv2/face.hpp"
 
+// include this header if drawLandmarks function is to be used
+// #include "drawLandmarks.hpp"
```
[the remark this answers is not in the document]
krshrimali: So sorry. This function was referred from : https://github.com/spmallick/learnopencv/blob/master/FacialLandmarkDetection/drawLandmarks.hpp and https://www.learnopencv.com/facemark-facial-landmark-detection-using-opencv/

Kindly ignore this functions' pull request. Consider the params - one only.

modules/face/samples/facemark_demo_lbf.cpp
```diff
@@ -63,6 +66,14 @@ int main(int argc, char** argv)
     FacemarkLBF::Params params;
     params.model_filename = model_path;
     params.cascade_face = cascade_path;
+    /* Change given parameters according to need. (if not given by user, set by default - default arguments given below).
```
berak: params are only needed for training a new model, they have no effect at all here

modules/face/samples/facemark_demo_lbf.cpp
```diff
@@ -63,6 +66,14 @@ int main(int argc, char** argv)
...
+    // FacemarkLBF supports 29 and 68 point landmarks currently.
```
berak: unfortunately, as it is now, the lbf model does not support any other than 68 points

please [see the answer to your question here](http://answers.opencv.org/question/192167/facemarklbf-training-fewer-landmark-points-than-68/)

PR #1669, opened 2018-06-21
kinfu: allow basic wrapping for bindings

paroj:
for this
- move Params struct out of class
- use static create instead of pimpl
- allow demo to be compiled without VIZ


savuor:
Thank you for the PR
Can you resolve the conflicts? I rewrote the demo a bit recently.

vpisarev:
thank you! :+1:

PR #1671, opened 2018-06-22
KinectFusion demo: live input support added

savuor:
### This pullrequest changes
* Added support for live input from Kinect2 device
* fixed several minor bugs


vpisarev:
looks great! thank you! :+1:

PR #1679, opened 2018-07-02
python3 compatible

SongChiYoung:
### This pullrequest changes

Python3 compatible. 

Python3 does not support accessing "map" like the "list".

Instead, it propose set to "the list".

I was test it work python3 and python2 also and it work smooth.


alalek:
Thank you!

I rebased this patch to 3.4 branch. We will merge changes from 3.4 into master regularly (weekly/bi-weekly).

PR #1682, opened 2018-07-03
xfeatures2d: throw CV_Error if not built with OPENCV_ENABLE_NONFREE

### LINKED ISSUE #1668: xfeatures2d module built with SURF and SIFT even when OPENCV_ENABLE_NONFREE is disabled
When setting `OPENCV_ENABLE_NONFREE` to false, the `SURF` and `SIFT` bits of the `xfeatures2d` module are still built. It is my understanding that these bits are patented and require licensing if they are to be used, so they should only be built if `OPENCV_ENABLE_NONFREE` is set to true.

berak:
resolves #1668

unless opencv was built with OPENCV_ENABLE_NONFREE, 
the following algorithms are disabled (throw a runtime error):

* SIFT
* SURF
* SURF_OCL
* SURF / CUDA

* xphoto/bm3d_image_denoising (this already existed)

### DIFF
--- a/modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
+++ b/modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
@@ -51,7 +51,7 @@ This section describes experimental algorithms for 2d feature detection.
     @defgroup xfeatures2d_nonfree Non-free 2D Features Algorithms
 
 This section describes two popular algorithms for 2d feature detection, SIFT and SURF, that are
-known to be patented. Use them at your own risk.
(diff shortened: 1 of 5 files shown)


alalek:
The main idea of this macro is to exclude non-free code **completely** from the binaries.
Just disabling it is not enough.

It should be something like this (in `.cpp` files):
```
#include public files only

#ifdef OPENCV_ENABLE_NONFREE
#include of implementation files should be guarded
... current SIFT/SURF code including "Impl" classes is here ...
#else
... add wrappers of public functions with exception throwing ...
#endif
```

PR #1709, opened 2018-08-02
Process InputOutputArray when BM3D_STEPALL is set

LaurentBerger:
solves http://answers.opencv.org/question/196559/after-applying-cvxphotobm3ddenoising-imshow-is-giving-insertion-error/

### DIFF
--- a/modules/xphoto/src/bm3d_image_denoising.cpp
+++ b/modules/xphoto/src/bm3d_image_denoising.cpp
@@ -148,6 +148,7 @@ void bm3dDenoising(
         _dst.create(srcSize, type);
         break;
     case BM3D_STEPALL:
+        _basic.create(srcSize, type);
         _dst.create(srcSize, type);
         break;
     default:
--- a/modules/xphoto/src/bm3d_image_denoising.cpp
+++ b/modules/xphoto/src/bm3d_image_denoising.cpp
@@ -148,7 +148,8 @@ void bm3dDenoising(
         _dst.create(srcSize, type);
         break;
     case BM3D_STEPALL:
-        _basic.create(srcSize, type);
+        if (_basic.needed())
+            _basic.create(srcSize, type);
         _dst.create(srcSize, type);
         break;
     default:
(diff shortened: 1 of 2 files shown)


indr0:
certainly with lack of experience & understanding, now  should I rebuild opencv with opencv_contrib module ??

modules/xphoto/src/bm3d_image_denoising.cpp
```diff
@@ -148,6 +148,7 @@ void bm3dDenoising(
         _dst.create(srcSize, type);
         break;
     case BM3D_STEPALL:
+        _basic.create(srcSize, type);
         _dst.create(srcSize, type);
```
alalek: Please take a look on line 159.

Perhaps this should be guarded: `if (_basic.needed()) basic.create(...);`

modules/xphoto/test/test_denoise_bm3d.cpp
```diff
@@ -76,7 +76,7 @@ namespace opencv_test { namespace {
 
         // BM3D: two different calls doing exactly the same thing
         cv::Mat result, resultSec;
-        cv::xphoto::bm3dDenoising(original, cv::Mat(), resultSec, 10, 4, 16, 2500, 400, 8, 1, 0.0f, cv::NORM_L2, cv::xphoto::BM3D_STEPALL);
+        cv::xphoto::bm3dDenoising(original, result, resultSec, 10, 4, 16, 2500, 400, 8, 1, 0.0f, cv::NORM_L2, cv::xphoto::BM3D_STEPALL);
         cv::xphoto::bm3dDenoising(original, result, 10, 4, 16, 2500, 400, 8, 1, 0.0f, cv::NORM_L2, cv::xphoto::BM3D_STEPALL);
```
alalek: You should not use "result" twice. It is not guaranteed that it is updated by second call.
Perhaps we should leave tests "as is" or replace `Mat()` to `noArray()`.

PR #1722, opened 2018-08-16
opencv-core: avoid using of multi-argument CV_Assert()

alalek:
**Main PR**: https://github.com/opencv/opencv/pull/12219

replace to CV_Assert_N()


alalek:
:+1:

PR #1724, opened 2018-08-18
replace internal usage of deprecated estimateRigidTransform in face

hrnr:
* face module now links to calib3d instead of video module

merge with https://github.com/opencv/opencv/pull/12248


hrnr:
Amended to restart the build.

PR #1730, opened 2018-08-23
int to proper enum type

cv3d:
Merge with opencv/opencv#

### This pullrequest changes
There are hundreds of unnamed enums, and they are being passed as int in the C++ interface.
In order to avoid passing a wrong value, and to maintain good practices, the hope is to name most of these enums, and change the int arguments into the proper enum types wherever possible.


modules/optflow/src/sparse_matching_gpc.cpp
```diff
@@ -278,7 +278,7 @@ bool ocl_getAllDCTDescriptorsForImage( const Mat *imgCh, std::vector< GPCPatchDe
            (int)globSize[0], (int)globSize[1], (int)patchRadius )
     .run( 2, globSize, 0, true ) == false )
     return false;
-  Mat cpuOut = out.getMat( 0 );
```
alalek: `ACCESS_READ` should be here.

alalek:
@cv3d Please rebase this patch onto 3.4 branch.

cv3d:
@alalek Done!

cv3d:
@alalek I opened a new PR #1733 instead.
I will keep this PR for future commits related to the int to enum changes

PR #1736, opened 2018-08-25
Correction for computeBinaryMap crash:

### LINKED ISSUE #1735: cv::saliency::StaticSaliencyFineGrained::computeBinaryMap crashes everytime

- OpenCV => branch 3.4
- Operating System / Platform => Archlinux 64bits
- Compiler => GCC

##### Detailed description
When we detect a saliency map with StaticSaliencyFineGrained and want to compute the binary map with the computeBinaryMap() function, it crashes everytime because:
1) StaticSaliencyFineGrained::computeSaliency() outputs an 8 bits image (CV_8UC1)
2) computeBinaryMap() needs an floating point image (CV_32F)

gdemarcq:
resolves #1735

### This pullrequest changes

This PR corrects the function cv::saliency::staticSaliencyFineGrained::computeSaliencyImpl() by changing the output image depth (CV_8UC1 to CV_32F) and normalizing it by the max value (as in staticSaliencySpectralResidual). It also adds an assert in cv::saliency::staticSaliency::computeBinaryMap() in order to check the input image depth which must be CV_32F.

### DIFF
--- a/modules/saliency/src/staticSaliency.cpp
+++ b/modules/saliency/src/staticSaliency.cpp
@@ -53,7 +53,7 @@ namespace saliency
 bool StaticSaliency::computeBinaryMap( InputArray _saliencyMap, OutputArray _binaryMap )
 {
   Mat saliencyMap = _saliencyMap.getMat();
-  CV_Assert( saliencyMap.depth() == CV_32F );
+  CV_CheckTypeEQ(saliencyMap.type(), CV_32FC1, "");
   Mat labels = Mat::zeros( saliencyMap.rows * saliencyMap.cols, 1, 1 );
   Mat samples = Mat_<float>( saliencyMap.rows * saliencyMap.cols, 1 );
   Mat centers;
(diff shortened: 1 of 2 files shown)


modules/saliency/src/staticSaliency.cpp
```diff
@@ -53,6 +53,7 @@ namespace saliency
 bool StaticSaliency::computeBinaryMap( InputArray _saliencyMap, OutputArray _binaryMap )
 {
   Mat saliencyMap = _saliencyMap.getMat();
+  CV_Assert( saliencyMap.depth() == CV_32F );
```
alalek:
```
CV_CheckTypeEQ(saliencyMap.type(), CV_32FC1, "");
```

modules/saliency/src/staticSaliencyFineGrained.cpp
```diff
@@ -66,7 +66,9 @@ bool StaticSaliencyFineGrained::computeSaliencyImpl(InputArray image, OutputArra
 {
     Mat dst(Size(image.getMat().cols, image.getMat().rows), CV_8UC1);
     calcIntensityChannel(image.getMat(), dst);
-    dst.copyTo(saliencyMap);
+    double minVal, maxVal;
+    minMaxLoc( dst, &minVal, &maxVal );
+    dst.convertTo(saliencyMap, CV_32F, 1.0f/maxVal);
```
alalek: `minMaxLoc` approach doesn't look reliable here.

dst type is `CV_8UC1`, so just apply 1/255.0f scale:
```
dst.convertTo(saliencyMap, CV_32F, 1.0f/255.0f);  // values are in range [0; 1]
```

PR #1739, opened 2018-08-27
cleanup stl string replacement (drop implementation of cv::String)

alalek:
**Main PR**: https://github.com/opencv/opencv/pull/12293

```
allow_multiple_commits=1
```


alalek:
:+1:

PR #1745, opened 2018-09-02
Fix initial guide image of rolling guidance filter

### LINKED ISSUE #1738: Initial guide image in cv::ximgproc::rollingGuidanceFilter() is wrong

- OpenCV => 3.4.2
- Operating System / Platform => Windows 10 64 bit

##### Detailed Description

A guide image of rollingGuidanceFilter() is initialized as a copy of a src image, as shown in the following code.
https://github.com/opencv/opencv_contrib/blob/5d3928e4f501aca9a9e6bf7791ecdf271f29c20f/modules/ximgproc/src/rolling_guidance_filter.cpp#L49

But the guide image must be initialized as a constant image (please see Algorithm 1 of [the original paper](http://www.cse.cuhk.edu.hk/leojia/projects/rollguidance/#downloads)).

This weakens the scale-aware property of the filter.
So expected results cannot be obtained.

**Example**

Below is an input image used in Fig. 10 of the original paper.
input

Below is the expected result of the filter, generated with the [executable in the author's page](http://www.cse.cuhk.edu.hk/leojia/projects/rollguidance/#downloads) with command `RollingGuidanceFilter.exe -i input.png -sr 10 -ss 40 -iter 5`.
input png out

Below is the result of `cv2.ximgproc.rollingGuidanceFilter(img, d=-1, sigmaColor=10, sigmaSpace=40, numOfIter=5)`.
output

As shown in the above images, small squares in the input image should be removed, but actually none of the squares are removed.

dai1741:
resolves #1738

### This pullrequest changes

This fixes a bug that a src image instead of a constant image is used as a guide image in the first iteration of rolling guidance filter.
This uses gaussian filter in the first iteration because it is theoretically equivalent to joint bilateral filter with a constant guide image and is more efficient (thanks to @LaurentBerger).
This also fixes a corresponding testcase.
In addition, this also includes a demo of the filter created by @LaurentBerger.

### DIFF
--- a/modules/ximgproc/samples/rollingGuidanceFilter_demo.cpp
+++ b/modules/ximgproc/samples/rollingGuidanceFilter_demo.cpp
@@ -1,0 +1,95 @@
+/*
+ *  By downloading, copying, installing or using the software you agree to this license.
+ *  If you do not agree to this license, do not download, install,
+ *  copy or use the software.
+ *
+ *
+ *  License Agreement
+ *  For Open Source Computer Vision Library
+ *  (3 - clause BSD License)
+ *
+ *  Redistribution and use in source and binary forms, with or without modification,
+ *  are permitted provided that the following conditions are met :
+ *
+ *  * Redistributions of source code must retain the above copyright notice,
+ *  this list of conditions and the following disclaimer.
+ *
+ *  * Redistributions in binary form must reproduce the above copyright notice,
+ *  this list of conditions and the following disclaimer in the documentation
+ *  and / or other materials provided with the distribution.
+ *
+ *  * Neither the names of the copyright holders nor the names of the contributors
+ *  may be used to endorse or promote products derived from this software
+ *  without specific prior written permission.
+ *
+ *  This software is provided by the copyright holders and contributors "as is" and
+ *  any express or implied warranties, including, but not limited to, the implied
+ *  warranties of merchantability and fitness for a particular purpose are disclaimed.
+ *  In no event shall copyright holders or contributors be liable for any direct,
+ *  indirect, incidental, special, exemplary, or consequential damages
+ *  (including, but not limited to, procurement of substitute goods or services;
+ *  loss of use, data, or profits; or business interruption) however caused
+ *  and on any theory of liability, whether in contract, strict liability,
+ *  or tort(including negligence or otherwise) arising in any way out of
+ *  the use of this software, even if advised of the possibility of such damage.
+ */
+
+#include <opencv2/core.hpp>
+#include <opencv2/core/utility.hpp>
+#include <opencv2/highgui.hpp>
+#include <opencv2/ximgproc.hpp>
+
+using namespace cv;
+using namespace cv::ximgproc;
+
+#include <iostream>
+using namespace std;
+
+int sColor = 10, sSpace = 3,nbIter=4;
+
+const char* window_name = "Rolling guidance filter";
+
+
+
+/**
+ * @function paillouFilter
+ * @brief Trackbar callback
+ */
+static void rollingFilter(int, void *pm)
+{
+    Mat img = *((Mat*)pm);
+    double sigmaColor(sColor/10.0), sigmaSpace(sSpace);
+    Mat dst;
+    rollingGuidanceFilter(img, dst, -1, sigmaColor, sigmaSpace, nbIter);
+    imshow("rollingGuidanceFilter",dst );
+}
+
+
+int main(int argc, char* argv[])
+{
+    if (argc < 2)
+    {
+        cout << "usage: rollingGuidanceFilter_demo [image]" << endl;
+        return 1;
+    }
+    Mat img = imread(argv[1]);
+    if (img.empty())
+    {
+        cout << "File not found or empty image\n";
+        return 1;
+    }
+
+    imshow("Original",img);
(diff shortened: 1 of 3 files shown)


modules/ximgproc/src/rolling_guidance_filter.cpp
```diff
@@ -59,18 +57,24 @@ namespace ximgproc
...
         if (dst.data == src.data)
             src = src.clone();
 
         int srcCnNum = src.channels();
 
         if (srcCnNum == 1 || srcCnNum == 3)
         {
-            while(numOfIter--){
-                jointBilateralFilter(guidance, src, guidance, d, sigmaColor, sigmaSpace, borderType);
+            if (numOfIter > 0) {
```
LaurentBerger: gaussianBlur must be used only to initialize image

dai1741: Yes, gaussianBlur is used only at iteration 0.
This is intended according to Algorithm 1 in p. 10 of [the paper](http://www.cse.cuhk.edu.hk/leojia/projects/rollguidance/paper/[ECCV2014]RollingGuidanceFilter_5M.pdf).
The filter returns a constant image when `numOfIter==0`, and returns a gaussianBlur result when `numOfIter==1`.

modules/ximgproc/test/test_rolling_guidance_filter.cpp
```diff
@@ -137,9 +137,9 @@ INSTANTIATE_TEST_CASE_P(TypicalSet1, RollingGuidanceFilterTest,
 //////////////////////////////////////////////////////////////////////////
 
 typedef tuple<double, string, int> RGFBFParam;
-typedef TestWithParam<RGFBFParam> RollingGuidanceFilterTest_BilateralRef;
+typedef TestWithParam<RGFBFParam> RollingGuidanceFilterTest_GaussianRef;
 
-TEST_P(RollingGuidanceFilterTest_BilateralRef, Accuracy)
+TEST_P(RollingGuidanceFilterTest_GaussianRef, Accuracy)
```
LaurentBerger: I think there is a problem in previous test and problem is still here : test uses gaussainblur to test gaussianblur. May be add result data in opencv_extra is necessary (see https://github.com/opencv/opencv_contrib/blob/master/modules/ximgproc/test/test_joint_bilateral_filter.cpp#L170-L199)

dai1741: Thank you for pointing it out! I modified the test to use naive gaussian.
Please note, however, that I had to use [a large tolerance value of 0.02](https://github.com/opencv/opencv_contrib/pull/1745/commits/84265bd9d667885b24d718c653fecd6e6f2c01d4#diff-54b7f4ae0ded9b81e2867156cd335a39R16) because of low accuracy of cv::GaussianBlur() for uchar type. And I removed tests for `sigmaS==8.0` because it takes too much time (about 10 sec for CV_32FC3).

LaurentBerger: Just and idea : You can run original code using this link with lena (img) as data  and save result (imgref). Don't you think that it will a better test compare opencv result with imgref?
@alalek

modules/ximgproc/samples/rollingGuidanceFilter_demo.cpp
```diff
@@ -0,0 +1,95 @@
...
+using namespace cv::ximgproc;
+
+#include <iostream>
+using namespace std;
+
+int sColor = 10, sSpace = 3,nbIter=4;
+
+const char* window_name = "Rolling guidance filter";
+
+
```
LaurentBerger: As rolling guidance filter can be slow, you can add this code too : 
void ProgressWindow(String windowName, Mat x)
{
    Mat y = Mat::zeros(x.size(), CV_8UC1);
    putText(y, "rollingGuidanceFilter in progress!", Point(10, x.rows / 2), FONT_HERSHEY_SIMPLEX, 1, Scalar(255));
    imshow(windowName, y);
    waitKey(1);
}

modules/ximgproc/samples/rollingGuidanceFilter_demo.cpp
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * @function paillouFilter
+ * @brief Trackbar callback
+ */
+static void rollingFilter(int, void *pm)
+{
+    Mat img = *((Mat*)pm);
+    double sigmaColor(sColor/10.0), sigmaSpace(sSpace);
+    Mat dst;
```
LaurentBerger: progressWindow("rollingGuidanceFilter", img);

modules/ximgproc/samples/rollingGuidanceFilter_demo.cpp
```diff
@@ -0,0 +1,95 @@
+/*
+ *  By downloading, copying, installing or using the software you agree to this license.
```
alalek: Please avoid license headers in samples.

modules/ximgproc/test/test_rolling_guidance_filter.cpp
alalek: IMHO, comparing with OpenCV gaussian blur is valid too (we checking that it is called on the first iteration)

PR #1765, opened 2018-09-10
core: replace cv::Ptr implementation => std::shared_ptr

alalek:
**Merge with**: https://github.com/opencv/opencv/pull/12467


alalek:
:+1:

PR #1772, opened 2018-09-13
Fixed cnn_3dobj module build

mshabunin:
### This pullrequest changes

This module can be built in Ubuntu 18 with installed `libcaffe-cpu-dev` package.


alalek:
BTW, Module's readme requires special caffe build from the separate branch. Not sure that they are compatible. Or required changes were merged into mainline caffe repository. Need to check that.

mshabunin:
No, it seems that PR has not been merged: BVLC/caffe#3123. Howether, it looks like that branch adds only training support, while inference should work with regular Caffe. 

I don't see any exotic layers in the test network: https://github.com/opencv/opencv_contrib/blob/master/modules/cnn_3dobj/testdata/cv/3d_triplet_testIMG.prototxt

PR #1804, opened 2018-10-01
ovis: add getCompositorTexture for reading custom render targets 

paroj:
not sure how API breaks are supposed to be handled in contrib, therefore SCENE_RENDER_FLOAT is a separate commit.

To review this change, take a look at:
https://ogrecave.github.io/ogre/api/latest/_compositor-_scripts.html#Format-1

with this change we get access to "rt0" of user defined content and format. Previously we hard-coded all of this via SCENE_RENDER_FLOAT.

```
allow_multiple_commits=1
```


alalek:
There are no ABI requirements for modules in opencv_contrib.

PR #1815, opened 2018-10-07
tabulation -> 4 spaces

cv3d:
Merge with opencv/opencv#TODO

### This pullrequest changes
Tabulation -> 4 spaces! Although huge, review should be easy via `git diff -w` & `grep`


cv3d:
@alalek It seems tabulation check is not enforced in the docs checks, at least in contrib repo!

PR #1816, opened 2018-10-07
Cleanup to maintain consistency between master and 3.4 repos

cv3d:


LaurentBerger:
if (tracker == NULL) becomes if (!tracker) ??
and p = arcLength(_src, true); becomes p = arcLength(ctr, true);

https://github.com/opencv/opencv/wiki/How_to_contribute
Scope your PR to one issue. Before submitting, make sure the diff contains no unrelated changes. If you want to cover more than one issue, submit your changes for each as separate pull requests.

cv3d:
It is not a significant contribution, but rather a merge that got missing somewhere...

/cc @alalek If you can point out the original commits, I would gladly cherry-pick, as I find it a lot of pain to do so, for three lines of code!

PR #1317, opened 2017-08-10
Gsoc opencv project : The Fast Bilateral Filter 

THUKey:
Add the source file for FastBilateralSolverFilter(located in modules/ximgproc/src/fbs_filter.cpp), and add the class declaration to the file modules/ximgproc/include/opencv2/ximgproc/edge_filter.hpp.

the test file and some sample(depth and colorization) for FastBilteralSolverFilter will be add next week.

TODO list:
- [ ] performance test

PR #1819, opened 2018-10-08
The Fast Bilateral Filter

berak:
an attempt of a "friendly takeover" of the gsoc 2017 project, #1317, in the hope to finally get it merged.


ytyytyyt:
There are some data type related issues with the implementation of the Fast Bilateral Solver and the disparity filtering demo code. The main issue is that the fbs code cannot handle disparity images but converts them into 8 bit images. I have a fix for this.

Another "issue" is that the fbs code uses fast global smoothing filter for initializing the solver. This may improve the filtering outcome but this is not part of the original fbs,

kuan-wang:
Thank you all!!  I'm sorry I didn't notice that this project is not merged after GSOC2017.

PR #1839, opened 2018-10-14
rgbd: run non-free tests if OPENCV_ENABLE_NONFREE is defined

alalek:


alalek:
:+1:
[Validated](http://pullrequest.opencv.org/buildbot/builders/master-contrib-lin64/builds/10600).

PR #1840, opened 2018-10-14
change area() emptiness checks to empty()

tompollok:
relates to opencv https://github.com/opencv/opencv/pull/12823
###This pullrequest changes

changes area() emptiness checks to use empty()

cases:
.area() == 0 to empty()
.area() != 0 to !empty()
.area() > 0 to !empty()

_src.size().area() > 0 to !_src.empty() //no need to create size here

if( r.area() ) to if( !r.empty() )


tompollok:
@alalek here the same PR for contrib

PR #1841, opened 2018-10-14
The kinfu_demo should open the default camera if I don't specify the camera in the cli

soyersoyer:


soyersoyer:
Thanks for the advices. I modified the code to create the DepthSource only once.

PR #1847, opened 2018-10-15
Checker for empty region

### LINKED ISSUE #1846: Segmentation fault at trackerKCF.cpp

- OpenCV => 4.0.0-alpha

##### Detailed description

`TrackerKCFImpl::getSubWindow()` hits a null pointer when the object moves out of the image borders and there is only a fraction of a pixel overlap. The issue is very easy to follow in the code:

In my example the the `roi` of this tracker is:

x: -122.5,
y:  297.5, 
width: 123.0, 
height: 112.0

So, the checker in [line 667](https://github.com/opencv/opencv_contrib/blob/9735ec666c92f7853fd7f20468e7ac701d3a2df0/modules/tracking/src/trackerKCF.cpp#L667) does not return false because the result of the `(roi & Rect2d(0,0, img.cols, img.rows))` is **[0.0, 297.5, 0.5, 112]**. As you can see, the width is 0.5 (less than a pixel). Please note that the [`&`](https://github.com/opencv/opencv/blob/68942affdbc4677aa845bc4307d4752182324a0e/modules/core/include/opencv2/core/types.hpp#L1958) operator does not return an empty `Rect()` object to pass this checker. It only returns an empty `Rect()` when the width or height are <= 0 ([See](https://github.com/opencv/opencv/blob/68942affdbc4677aa845bc4307d4752182324a0e/modules/core/include/opencv2/core/types.hpp#L1958)). 

So, in this example, although the `region.width` is **0**, there will be no returns and the execution will move on to create the `patch` :

`patch=img(region).clone();` [here](https://github.com/opencv/opencv_contrib/blob/9735ec666c92f7853fd7f20468e7ac701d3a2df0/modules/tracking/src/trackerKCF.cpp#L678).

Since `region` does not have any width, the `patch` is an empty image and `patch.data` points to a **nullptr**. This will cause a segmentation error in [Line 688](https://github.com/opencv/opencv_contrib/blob/9735ec666c92f7853fd7f20468e7ac701d3a2df0/modules/tracking/src/trackerKCF.cpp#L688) when it wants to get the `rows` parameters of this object:

`if(patch.rows==0 || patch.cols==0)return false;`

A checker like this at [line 677](https://github.com/opencv/opencv_contrib/blob/9735ec666c92f7853fd7f20468e7ac701d3a2df0/modules/tracking/src/trackerKCF.cpp#L677), before creating the patch, prevents this issue:

```
if (region.width < 1 || region.height < 1) return false;
```

##### Steps to reproduce

This exception is very specific and it rarely happens. But it is easy to see the problem from the code.

mhaghighat:
### This pullrequest changes
This pullrequest adds a checker to the trackerKCF.cpp to avoid creating patches from regions with 0 width/height. Please refer to issue #1846 for more details.


alalek:
Thank you for contribution!

As a bugfix this patch should [go](https://github.com/opencv/opencv/wiki/How_to_contribute#faq) into 3.4 branch. We will merge changes from 3.4 into master regularly (weekly/bi-weekly).

So, please:

tompollok:
You can use
If(region.empty())

PR #1860, opened 2018-10-30
scan vector from end to beginning when removing elements

### LINKED ISSUE #1859: Bug in file line_descriptor/src/binary_descriptor.cpp
In function BinaryDescriptor::computeImpl
Line 628:

    for ( size_t j = 0; j < sl[i].size(); j++ )
    {
      //if( (int) ( sl[i][j] ).octaveCount > params.numOfOctave_ )
      if( (int) ( sl[i][j] ).octaveCount > octaveIndex )
        ( sl[i] ).erase( ( sl[i] ).begin() + j );
    }

After erasing, the size of the vector sl[i] will change and the last element will not be processed.

bbeco:
resolves #1859

### This pullrequest changes
this PR changes the inner loop for removing useless `OctaveSingleLines` so that the index goes from the vector's end to its beginning thus preventing element skipping when an `erase` operation occurs.

### DIFF
--- a/modules/line_descriptor/src/binary_descriptor.cpp
+++ b/modules/line_descriptor/src/binary_descriptor.cpp
@@ -623,9 +623,9 @@ void BinaryDescriptor::computeImpl( const Mat& imageSrc, std::vector<KeyLine>& k
   }
 
   /* delete useless OctaveSingleLines */
-  for ( size_t i = 0; i < sl.size(); i++ )
+  for ( int i = 0; i < sl.size(); i++ )
   {
-    for ( size_t j = sl[i].size() - 1; j >= 0; --j )
+    for ( int j = sl[i].size() - 1; j >= 0; --j )
     {
       //if( (int) ( sl[i][j] ).octaveCount > params.numOfOctave_ )
       if( (int) ( sl[i][j] ).octaveCount > octaveIndex )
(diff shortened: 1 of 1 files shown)


Varvrar:
size_t variable cannot be < 0

Solution to this issue can be like this:
    for ( size_t j = 0; j < sl[i].size(); )
    {
      //if( (int) ( sl[i][j] ).octaveCount > params.numOfOctave_ )
		if ((int)(sl[i][j]).octaveCount > octaveIndex)
			(sl[i]).erase((sl[i]).begin() + j);
		else j++;
    }

Now I am trying to do PR with this.

bbeco:
> size_t variable cannot be < 0

That's true!

I fixed this by replacing `j > -1` with `j >= 0`.

> Solution to this issue can be like this:
> for ( size_t j = 0; j < sl[i].size(); )
> {
> //if( (int) ( sl[i][j] ).octaveCount > params.numOfOctave_ )
> if ((int)(sl[i][j]).octaveCount > octaveIndex)
> (sl[i]).erase((sl[i]).begin() + j);
> else j++;
> }

This is also correct but it introduces an `if` statement inside the loop (which may reduce performances)

Varvrar:
Ok, but I still cant find elegant solution...
Your fix will not work. j >= 0 will never be false:

size_t j = 0;
--j;
// Now j = 18446744073709551615

bbeco:
that's also true. I changed the type to `int`

alalek:
Thank you for contribution!
Fix is merged from here: #1861

PR #1863, opened 2018-10-31
minor changes

sturkmen72:
### This pullrequest changes

minor changes

### DIFF
--- a/modules/tracking/src/trackerBoosting.cpp
+++ b/modules/tracking/src/trackerBoosting.cpp
@@ -126,7 +126,7 @@ bool TrackerBoostingImpl::initImpl( const Mat& image, const Rect2d& boundingBox 
   Mat_<int> intImage;
   Mat_<double> intSqImage;
   Mat image_;
-  cvtColor( image, image_, CV_RGB2GRAY );
+  cvtColor( image, image_, COLOR_BGR2GRAY );
   integral( image_, intImage, intSqImage, CV_32S );
   TrackerSamplerCS::Params CSparameters;
   CSparameters.overlap = params.samplerOverlap;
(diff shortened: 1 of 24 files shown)


modules/tracking/src/trackerBoosting.cpp
```diff
@@ -126,7 +126,7 @@ bool TrackerBoostingImpl::initImpl( const Mat& image, const Rect2d& boundingBox
   Mat_<int> intImage;
   Mat_<double> intSqImage;
   Mat image_;
-  cvtColor( image, image_, CV_RGB2GRAY );
+  cvtColor( image, image_, COLOR_BGR2GRAY );
```
sturkmen72: (RFC) should i keep it `COLOR_RGB2GRAY`

alalek: Yes, lets avoid using of legacy "_c.h" headers and their symbols.

sturkmen72: R-G-B & B-G-R

alalek: Oops...
Looks good to me.
OpenCV default format for I/O is `BGR`
(except Android - RGBA - they should explicitly pass gray images for such algorithms).

modules/xphoto/samples/inpainting.cpp
```diff
@@ -41,22 +38,22 @@ int main( int argc, const char** argv )
...
-    cv::Mat src = cv::imread(inFilename, -1);
+    cv::Mat src = cv::imread(inFilename, cv::IMREAD_UNCHANGED);
     if ( src.empty() )
     {
         printf( "Cannot read image file: %s\n", inFilename.c_str() );
         return -1;
     }
 
-    cv::cvtColor(src, src, CV_RGB2Lab);
+    cv::cvtColor(src, src, cv::COLOR_RGB2Lab);
```
alalek: BGR too here and in `COLOR_Lab2RGB`

PR #1877, opened 2018-11-05
fbs disparity filtering fix

ytyytyyt:
### This pullrequest changes

- Fixes Fast Bilateral Solver (FBS) disparity filtering demo, i.e. FBS works now with `int16` disparity images
- Revises data type handling for both input and confidence map images
- Adds option to give command line arguments to FBS in addition to WLS
- Moves Fast Global Smoothing (FGS) pre-filtering from `fbs_filter.cpp` to `disparity_filtering.cpp` demo code

```
force_builders=linux,docs,windows,macosx
```

### DIFF
--- a/modules/ximgproc/src/fbs_filter.cpp
+++ b/modules/ximgproc/src/fbs_filter.cpp
@@ -66,6 +66,8 @@
     typedef std::unordered_map<long long /* hash */, int /* vert id */>  mapId;
 #endif
 
+#define EPS 1e-43f
+
 namespace cv
 {
 namespace ximgproc
@@ -87,7 +89,7 @@ namespace ximgproc
         void filter(InputArray src, InputArray confidence, OutputArray dst) CV_OVERRIDE
         {
 
-            CV_Assert(!src.empty() && (src.depth() == CV_8U || src.depth() == CV_16S || src.depth() == CV_32F) && src.channels()<=4);
+            CV_Assert(!src.empty() && (src.depth() == CV_8U || src.depth() == CV_16U || src.depth() == CV_32F) && src.channels()<=4);
             CV_Assert(!confidence.empty() && (confidence.depth() == CV_8U || confidence.depth() == CV_16S || confidence.depth() == CV_32F) && confidence.channels()==1);
             if (src.rows() != rows || src.cols() != cols)
             {
@@ -114,8 +116,6 @@ namespace ximgproc
             for(int i=0;i<src.channels();i++)
             {
                 Mat cur_res = src_channels[i].clone();
-                if(src.depth() != CV_8UC1)
-                    cur_res.convertTo(cur_res, CV_8UC1);
 
                 solve(cur_res,conf,cur_res);
                 cur_res.convertTo(cur_res, src.type());
@@ -206,6 +206,7 @@ namespace ximgproc
 
     void FastBilateralSolverFilterImpl::init(cv::Mat& reference, double sigma_spatial, double sigma_luma, double sigma_chroma, int num_iter, double max_tol)
     {
+
         guide = reference.clone();
 
         bs_param.cg_maxiter = num_iter;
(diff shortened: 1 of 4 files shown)


modules/ximgproc/src/fbs_filter.cpp
```diff
@@ -487,12 +488,18 @@ namespace ximgproc
...
+        else if(target.depth() == CV_8U)
+            target.convertTo(x, CV_32FC1, 1.0f/255.0f);
+
+        confidence.convertTo(w, CV_32FC1);
+
         xw = x.mul(w);
         cv::ximgproc::fastGlobalSmootherFilter(guide, xw, filtered_xw, fgs_spatialSigma, fgs_colorSigma);
         cv::ximgproc::fastGlobalSmootherFilter(guide, w, filtered_w, fgs_spatialSigma, fgs_colorSigma);
-        cv::divide(filtered_xw, filtered_w, filtered_disp, 1.0f, CV_32FC1);
+        cv::divide(filtered_xw, filtered_w+EPS, filtered_disp, 1.0f, CV_32FC1);
```
alalek: Please add some simple regression test to for this issue into test_fbs.cpp file.
New test images can be added into opencv_extra/testdata (but prefer to reuse existed).

Is it related for master branch only? Or 3.4 too? OpenCV handles division by zero differently for floating-point numbers (3.4 has zero pre-check, master uses IEEE 754 rules).

ytyytyyt: This was related to master branch only as I was originally working with 3.4 and noticed the issue while switching to master. Like you pointed out, I had already a suspicion that the 3.4 handles division by zero differently but didn't check this.

Would it be better to remove the Fast Global Filtering because it is not used in the original paper for initializing the Fast Bilateral Solver? In fbs_filter.cpp, the parameters for FGS are also fixed. I would prefer to remove the FGS part and maybe add it to disparity filtering demo for comparison.

In the case, you think its better to keep the FGS in fbs_filter.cpp, I can add the requested test.

alalek: > it is not used in the original paper

It make sense to remove that.
From user/applications side it is more easy to add pre-filter than to remove extra stage.

@berak Do you have any objections?

alalek: BTW, This code is not a part of any OpenCV release yet - you can work on 3.4 branch (3.4.4 expected on this week).

berak: no objections, wise deccision ;)

ytyytyyt: Alright, thanks! I will try to revise the code as soon as possible.

ytyytyyt: In its present state, the fbs filter expects float confidence map values to be in [0,255] range like in disparity_filters.cpp. Would it be better to change the float confidence values to [0,1] range instead?

Besides this issue my revision should be complete.

alalek: > float confidence values to [0,1] range

Sounds good to me.

ytyytyyt: Now the FGS (i.e. WLS) filtering is moved from fbs_filter.cpp to disparity_filtering.cpp as a preprocessing option. Consequently, there is no division by zero issue even between master and 3.4 branches.

ytyytyyt: @alalek I added the fixed FBS filter to 3.4 branch (hopefully wasn't too late for 3.4.4).

ytyytyyt: > Please add some simple regression test to for this issue into test_fbs.cpp file.
> New test images can be added into opencv_extra/testdata (but prefer to reuse existed).

I checked the `test_fbs.cpp` as well. It is comparing the output to a reference image produced with a different filter (FGS) with a very loose error threshold. I updated the test with a new reference image produced with the same filter and will commit it soon.

PR #1887, opened 2018-11-09
xfeatures2d: validate keypoints mask for SURF

alalek:
Avoid keypoints from masked region.

relates https://github.com/opencv/opencv/pull/12827


alalek:
:+1:

PR #1891, opened 2018-11-09
fix fbs test with new reference image

ytyytyyt:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/547

Fix test for Fast Bilateral Solver with a new reference image to be included in `opencv_extra`.

```
opencv_extra=3.4
```


alalek:
Please reopen PR with meaningful source branch name instead of 3.4. For example, "fbs_test".
Branches with names "3.4" / "master" have restricted functionality (they can't be tested with opencv_extra).

It is enough to open PR into 3.4 branch only. We will merge changes from 3.4 into master later.

alalek:
#1892

PR #1892, opened 2018-11-09
fix fbs test with new reference image

ytyytyyt:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/547

Fix test for Fast Bilateral Solver with a new reference image to be included in opencv_extra.

```
opencv_extra=3.4
```


ytyytyyt:
Thanks! 👍 If there are any issues with the Fast Bilateral Solver in future, just let me know.

PR #1898, opened 2018-11-11
fix include and use #if CV_SIMD128 instead of #ifdef CV_SIMD128

### LINKED ISSUE #1894: A way to use SIMD in dis optical flow

- OpenCV => :grey_question:3.4
- Operating System / Platform => :grey_question:ubuntu16.04
- Compiler => :grey_question: gcc5.4

##### Detailed description

Dis optical flow cannot use simd in devices that have simd capacity, even when sse or neon is enabled when compilation. I find the solution after some attempts. It cannot use simd because it does not include the head file <opencv2/core/hal/intrin.hpp>, after I include this head file, dis optical flow can use simd if sse or neon is enabled. I think some other contribe functions may have the same problem.

##### Steps to reproduce
nothing

tompollok:
resolves #1894 in 3.4 opencv_contrib branch
relates https://github.com/opencv/opencv/pull/13115

### DIFF
--- a/modules/optflow/src/dis_flow.cpp
+++ b/modules/optflow/src/dis_flow.cpp
@@ -41,7 +41,9 @@
 //M*/
 
 #include "precomp.hpp"
+#include "opencv2/core/hal/intrin.hpp"
 #include "opencl_kernels_optflow.hpp"
+
 
 using namespace std;
 #define EPS 0.001F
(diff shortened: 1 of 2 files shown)


alalek:
@tompollok BTW, Please use descriptive source branch name (like "optflow_fix_simd"):
- it allows to submit multiple PRs (with different names)
- no need to "hold" your generic branches "3.4" / "master" (until PR merge into upstream)

tompollok:
@alalek just for clarification: what do you mean with "no need to "hold" your generic branches "3.4" / "master" (until PR merge into upstream)"

alalek:
Updating commits of your branch from which PR was created, automatically updates PR.
To avoid unintended PR updates you need to "hold" changes on your branch until PR has been reviewed and merged.

PR #1901, opened 2018-11-13
ovis: allow re-initialization by destroying all windows

paroj:


modules/ovis/src/ovis.cpp
```diff
@@ -236,6 +236,8 @@ struct Application : public OgreBites::ApplicationContext, public OgreBites::Inp
         return ret;
     }
 
+    size_t numWindows() { return mWindows.size(); }
```
alalek: "const" method?

PR #1936, opened 2018-12-07
linemod: add drawFeatures method

paroj:
<cut/>

draws features like:
image 0ocstz

(Model Based Training, Detection and Pose
Estimation of Texture-Less 3D Objects in
Heavily Cluttered Scenes, Figure 2)

### DIFF
--- a/modules/rgbd/src/linemod.cpp
+++ b/modules/rgbd/src/linemod.cpp
@@ -207,6 +207,29 @@ void colormap(const Mat& quantized, Mat& dst)
   }
 }
 
+void drawFeatures(InputOutputArray img, const std::vector<Template>& templates, const Point2i& tl, int size)
+{
+#ifdef HAVE_OPENCV_IMGPROC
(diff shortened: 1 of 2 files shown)


modules/rgbd/src/linemod.cpp
```diff
@@ -207,6 +207,29 @@ void colormap(const Mat& quantized, Mat& dst)
...
+    static int markers[] = {MARKER_SQUARE, MARKER_DIAMOND};
+
+    int modality = 0;
+    for(const Template& t : templates)
+    {
+        if(t.pyramid_level != 0) continue;
+
+        for(const Feature& f : t.features)
+        {
+            drawMarker(img, tl + Point(f.x, f.y), colors[int(modality != 0)], markers[int(modality != 0)], size);
```
alalek: `drawMarker()` is imgproc feature.
It can be useful to save "result" image instead of showing it.

paroj: changed the check to imgproc. However this does not show anything, but rather draws on whatever image you give it.

PR #1959, opened 2019-01-06
Implementation of an Image Quality Analysis (IQA) API 

clunietp:
### This pullrequest implements an image quality analysis (IQA) API for OpenCV

See readme.md for details on the interface.  Includes tests and code documentation, feedback welcomed.  Thanks!


modules/quality/include/opencv2/quality/QualityGMSD.hpp
```diff
@@ -0,0 +1,89 @@
...
+            CV_WRAP cv::Scalar compute(InputArrayOfArrays cmpImgs) CV_OVERRIDE;
+
+            /** @brief Implements Algorithm::empty()  */
+            CV_WRAP bool empty() const CV_OVERRIDE { return _qualityMaps.empty() && _refImgData.empty(); }
+
+            /** @brief Implements Algorithm::clear()  */
+            CV_WRAP void clear() CV_OVERRIDE { _qualityMaps.clear(); _refImgData.clear(); Algorithm::clear(); }
+
+            /** @brief Returns pointer to output quality maps images that were generated during computation, if supported by the algorithm.  */
+            CV_WRAP const std::vector<quality_map_type>& getQualityMaps() const CV_OVERRIDE { return _qualityMaps; }
```
alalek: OpenCV approach is:
```
CV_WRAP void getQualityMaps(OutputArrayOfArrays dst);
```

In the current form it is not handled properly by Python/Java bindings generator.

modules/quality/include/opencv2/quality/QualityMSE.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include "QualityBase.hpp"
+
+namespace cv {
+	namespace quality {
+
+		/**
+        @brief Full reference mean square error algorithm  https://en.wikipedia.org/wiki/Mean_squared_error
+        */
+		class CV_EXPORTS_W QualityMSE : public QualityBase {
+		public:
```
alalek: Do we really need class to compute this?
Why function doesn't work?

modules/quality/src/QualityGMSD.cpp
```diff
@@ -0,0 +1,215 @@
...
+    {
+        if ( !cv::ocl::useOpenCL() || src.depth() == CV_64F)   // nothing more to do
+            return filter_2D<UMat, DstMat>(src, dst, kernel, anchor, delta, border_type);
+
+        auto dst_type = dst.type() == 0 ? src.type() : dst.type();
+
+        // UMat conversion to 64F
+        UMat src_converted = {};
+        src.copyTo(src_converted);
+        src_converted.convertTo(src_converted, CV_64F);
```
alalek: > copyTo()
> convertTo()

`.convertTo()` performs copy too. `.copyTo()` is redundant.

modules/quality/src/QualitySSIM.cpp
```diff
@@ -0,0 +1,152 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#include "opencv2/quality/QualitySSIM.hpp"
```
alalek: "precomp.hpp" first in `.cpp` files.

See contribution/coding style guides on Wiki: https://github.com/opencv/opencv/wiki

modules/quality/test/test_precomp.hpp
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        // execute quality test for a pair of images
+        template <typename TMat>
+        inline void quality_test( cv::Ptr<quality::QualityBase> ptr, const TMat& cmp, const Scalar& expected, const std::size_t quality_maps_expected = 1, const bool disable_ocl = false )
+        {
+#ifdef HAVE_OPENCL
+            auto prev = cv::ocl::useOpenCL();
+            if ( disable_ocl )
+                cv::ocl::setUseOpenCL(false);
```
alalek: Take a look on these macros:
- OCL_ON(stmt)/OCL_OFF(stmt) 

Usage example: modules/core/test/opencl

clunietp:
@alalek I have made some updates, could you please re-review when you get a chance?  Thanks

modules/quality/include/opencv2/quality/QualityBase.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+	/************************************ Quality Base Class ************************************/
+
+	class CV_EXPORTS_W QualityBase
+        : public virtual Algorithm
+	{
+	public:
+
+        /**
+		@brief Destructor
```
alalek: Please fix indentation:
- no additional spaces in **namespaces**
- **no tabs** (anywhere)

modules/quality/include/opencv2/quality/QualityBase.hpp
```diff
@@ -0,0 +1,68 @@
...
+			*/
+		virtual ~QualityBase() = default;
+
+        /**
+        @brief Compute quality score per channel, see specific algorithm for interpreting result scores
+        @param cmpImgs comparison image(s), can be empty for no-reference quality algorithms
+        */
+        virtual CV_WRAP cv::Scalar compute( InputArrayOfArrays cmpImgs ) = 0;
```
alalek: What is the meaning of `cv::Scalar` results here?

modules/quality/include/opencv2/quality/QualityGMSD.hpp
```diff
@@ -0,0 +1,82 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#ifndef OPENCV_QUALITY_QUALITYGMSD_HPP
+#define OPENCV_QUALITY_QUALITYGMSD_HPP
+
+#include "QualityBase.hpp"
```
alalek: Use lower case for file names: https://github.com/opencv/opencv/wiki/Coding_Style_Guide

modules/quality/include/opencv2/quality/QualityGMSD.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+namespace cv {
+namespace quality {
+namespace detail {
+namespace gmsd {
+
+    using mat_type = UMat;
+
+    // holds computed values for an input mat
+    struct mat_data
```
alalek: 1. Exposing internal implementation details in public API headers.

2.
> mat_data

Refer to OpenCV naming policy: https://github.com/opencv/opencv/wiki/Coding_Style_Guide

modules/quality/include/opencv2/quality/QualityPSNR.hpp
```diff
@@ -0,0 +1,117 @@
...
+#include <limits>   // numeric_limits
+#include "QualityBase.hpp"
+#include "QualityMSE.hpp"
+
+namespace cv {
+namespace quality {
+namespace detail {
+
+    // convert mse to psnr
+    inline CV_CONSTEXPR double mse_to_psnr( double mse, double max_pixel_value )
```
alalek: > CV_CONSTEXPR

Why?

.gitignore
```diff
@@ -1,12 +0,0 @@
-*.autosave
```
alalek: Restore "original" file.

modules/quality/include/opencv2/quality/QualityBase.hpp
clunietp: The `cv::Scalar` returns the quality score for each channel in the input image.  For example, if the user provides a BGR image, the scalar returns the quality score for Blue --> scalar[0], Green --> scalar[1], Red --> scalar[2].  Alternatively, if the user only wants to provide a grayscale image, the result will be scalar[0].

modules/quality/include/opencv2/quality/QualityPSNR.hpp
clunietp: for c++ `constexpr` support, in case the compiler sees an opportunity to take advantage of it.  I typically add this for small utility functions like this one.  Should it be removed?

modules/quality/include/opencv2/quality/QualityGMSD.hpp
clunietp: @alalek  OK I will move it to a private class instead

modules/quality/include/opencv2/quality/QualityPSNR.hpp
alalek: Does it work here? With "partial" specialization of const arguments (`max_pixel_value` is compile-time constant, but `mse` - doesn't)?

clunietp: @alalek you're right in that this particular library won't use it as a `constexpr` since `mse` won't be known at compile-time.  However, `mse` is known at compile time in the test app but I'm not concerned about test app performance.  Typically I'll mark something as `constexpr`  if it could potentially be used in such a context (like in this case), and leave it up to the caller/compiler to identify those situations as appropriate.  However, since you feel it is unnecessary in this context, I will remove it and add it again later if there is a legitimate use case in the module.

clunietp:
Needed to rebranch.  Continued in https://github.com/opencv/opencv_contrib/pull/1990

PR #1964, opened 2019-01-09
sfm: avoid variable-names starting with underscore

### LINKED ISSUE #1963: sfm: conflicting identifiers

- OpenCV => 3.4
- Operating System / Platform => OpenBSD 6.4
- Compiler => clang 6.0.0

##### Detailed description

The following error occurs when compiling sfm on OpenBSD:

```
/usr/ports/pobj/opencv-3.4.5/opencv-3.4.5/opencv_contrib/modules/sfm/src/projection.cpp:58:36: error: expected ')'
homogeneousToEuclidean(const Mat & _X, Mat & _x)
                                   ^
/usr/include/ctype.h:51:12: note: expanded from macro '_X'
#define _X      0x40
```

rrrapha:
resolves #1963

### DIFF
--- a/modules/sfm/src/projection.cpp
+++ b/modules/sfm/src/projection.cpp
@@ -55,34 +55,34 @@ namespace sfm
 
 template<typename T>
 void
-homogeneousToEuclidean(const Mat & _X, Mat & _x)
+homogeneousToEuclidean(const Mat & X_, Mat & x_)
 {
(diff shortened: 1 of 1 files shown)


rrrapha:
See also:
https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier/228797#228797

PR #1972, opened 2019-01-23
Add link directories for freetype

mleotta:
### This pullrequest changes

This change allows the freetype module to be built against external
libraries that are not found on the standard system path.  I ran into
this issue building OpenCV on MacOS against freetype installed by
macports.  CMake found the libraries at configure time but linking
failed at build time.


alalek:
> pkg_search_module(FREETYPE freetype2)

relates https://github.com/opencv/opencv/pull/13534

rrrapha:
I can confirm this fixes the linking issue on OpenBSD too. 
(Tested with OpenCV 3.4.5)

PR #1981, opened 2019-01-30
linemod: allow creating modalities from bindings

paroj:


alalek:
Could you please rebase this patch onto 3.4 branch?

PR #2010, opened 2019-02-20
New Sample Text Detection

MaicolGomez:
This pull request adds a new sample to obtain text detection and recognition.

This sample is based on the code of Lluis Gomez I. Bigorda.
In this new sample, It is changed the use of the channels in the image and also is created 
a new function to sort the regions detects in the first two filters to make 
a faster algorithm to group into a bounding box.
Also is add an option to use pramga directives to parallelize the compute of the regions 
for each channel.

### DIFF
--- a/modules/text/samples/textdetection_2.cpp
+++ b/modules/text/samples/textdetection_2.cpp
@@ -1,0 +1,183 @@
+/*
+ * textdetection_2.cpp
+ *
+ * A demo program based on the code of Lluis Gomez i Bigorda provide by opencv_contrib
+ *
+ * Created on: August, 2018
+ *     Author: UNICAMP-Samsung group - Contact: mpereira at ic.unicamp.br
+ */
+
+#include "opencv2/text.hpp"
+#include "opencv2/highgui.hpp"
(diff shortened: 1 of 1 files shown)


modules/text/samples/textdetection_2.cpp
```diff
@@ -0,0 +1,183 @@
...
+		else{
+			int j;
+			for(j = 0; j < strlen(text); j++)
+				if( text[j] != '\n' )
+					printf("%c",text[j]);
+				else
+					printf(" ");
+			printf("\n");
+		}
+    }
```
berak: `delete[] text;` is missing here

modules/text/samples/textdetection_2.cpp
```diff
@@ -0,0 +1,183 @@
...
+			printf("\n");
+		else{
+			int j;
+			for(j = 0; j < strlen(text); j++)
+				if( text[j] != '\n' )
+					printf("%c",text[j]);
+				else
+					printf(" ");
+			printf("\n");
+		}
```
berak: missing `delete[] text` here..

PR #2036, opened 2019-03-12
Invert marker

szk1509:
### This pullrequest introduce the detection of inverted markers.
i.e. it allows detecting white markers on a black background


alalek:
Add tests for new feature:
- to ensure that it works really
- to ensure that it will not be broken by mistake in the future
- simple test is one of the best documentation for the code

P.S. Please "cleanup" your commits - squash them into one. Also check "Author"'s email of the first commit at least (GitHub doesn't recognize it).

alalek:
BTW, No need to re-open PRs. You can push changes on existed branches (with "--force" option).

PR #2041, opened 2019-03-14
Fix CUDA filter2d multithreaded bug (from opencv3.4)

flyingfalling:
This fixes the CUDA filter2d issue from opencv (moved here after opencv 4.0).

This has already been accepted/merged into opencv, but the source was moved to opencv_contrib after 4.0. Apologies for multiple pull requests if it was unnecessary.


alalek:
Ported here: https://github.com/opencv/opencv_contrib/pull/2040

PR #2046, opened 2019-03-17
Minor changes to the quality module to support Visual Studio 2013

JoeHowse:
This pull request makes minor changes to the quality module to support Visual Studio 2013.


modules/quality/include/opencv2/quality/quality_utils.hpp
```diff
@@ -16,7 +16,7 @@ namespace quality_utils
 {
 
 // default type of matrix to expand to
-static CV_CONSTEXPR const int EXPANDED_MAT_DEFAULT_TYPE = CV_32F;
+static CV_CONSTEXPR int EXPANDED_MAT_DEFAULT_TYPE = CV_32F;
```
alalek: Please split this by `#ifdef` too:
```
#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900/*MSVS 2015*/)
static constexpr const int EXPANDED_MAT_DEFAULT_TYPE = CV_32F;
#else
// support MSVS 2013
static const int EXPANDED_MAT_DEFAULT_TYPE = CV_32F;
#endif
```

or leave this line unchanged with this patch from main repository: https://github.com/opencv/opencv/pull/14101

PR #2050, opened 2019-03-20
Addded an option to visualize features used for image stabilization

SBCV:
While using videostab.cpp to perform video stabilization, I experienced several times sub-optimal results. Using the visualization of this PR showed that in my sequences too many feature points are being detected on independently moving objects (in my case vehicles).  

I tried to stay as close as possible to the original implementation and determined the output (hopefully) with a minimal amount of changes. 

### This pull request changes

This PR adds an option to videostab/samples/videostab.cpp, which allows to define an additional video output paths for an overlay of the original video and feature locations used for image stabilization.


SBCV:
Just discovered that no output is generated, when transformations are written to disc ...
I'll check what is going wrong and update the PR

SBCV:
Fixed bug, improved performance & reduced number of new lines using a callback

asmorkalov:
Looks good for me :+1: . @alalek What do you say?

alalek:
1. Lifetime of callback-related objects. `std::function` with lambdas must be used very accurate.
2. OpenCV bindings (Python, Java). We need robust solution to support them too.

SBCV:
Hi, thank you for having a look. 
Unfortunately, your last message was a bit unclear to me. Are there some callback-related lines I that must be changed?
I'm not familiar with Python/Java bindings. What kind of lines could cause troubles? 
Side note: At the time when I wrote this PR, the video stabilization tool of OpenCV had no python bindings.

PR #2052, opened 2019-03-21
Added a parameter to mask features used for video stabilization

SBCV:
This PR adds an additional video path parameter to modules/videostab/samples/videostab.cpp, which allows to mask features during video stabilization. 

The mask video must contain black (0 values) for parts, which are ignored during stabilization.

### DIFF
--- a/modules/videostab/include/opencv2/videostab/frame_source.hpp
+++ b/modules/videostab/include/opencv2/videostab/frame_source.hpp
@@ -86,6 +86,28 @@ private:
     Ptr<IFrameSource> impl;
 };
 
+class MaskFrameSource : public IFrameSource
+{
+public:
+    MaskFrameSource(const Ptr<IFrameSource>& source): impl(source) {};
+
+    virtual void reset() CV_OVERRIDE { impl->reset(); }
+    virtual Mat nextFrame() CV_OVERRIDE {
+        Mat nextFrame = impl->nextFrame();
+        maskCallback_(nextFrame);
+        return nextFrame;
+    }
+
+    void setMaskCallback(std::function<void(Mat&)> MaskCallback)
+    {
+        maskCallback_ = std::bind(MaskCallback, std::placeholders::_1);
+    };
+
+private:
+    Ptr<IFrameSource> impl;
+    std::function<void(Mat&)> maskCallback_;
+};
+
 //! @}
 
 } // namespace videostab
--- a/modules/videostab/include/opencv2/videostab/global_motion.hpp
+++ b/modules/videostab/include/opencv2/videostab/global_motion.hpp
@@ -180,6 +180,12 @@ public:
     virtual void setMotionModel(MotionModel val) { motionModel_ = val; }
     virtual MotionModel motionModel() const { return motionModel_; }
 
+    virtual void setFrameMask(InputArray mask)
+    {
+        if (!mask.empty())
+            CV_Error(Error::StsNotImplemented, "Mask support is not implemented.");
+    }
+
     virtual Mat estimate(const Mat &frame0, const Mat &frame1, bool *ok = 0) = 0;
 
 protected:
@@ -208,6 +214,8 @@ public:
     virtual void setMotionModel(MotionModel val) CV_OVERRIDE { motionEstimator_->setMotionModel(val); }
     virtual MotionModel motionModel() const CV_OVERRIDE { return motionEstimator_->motionModel(); }
 
+    virtual void setFrameMask(InputArray mask) CV_OVERRIDE { motionEstimator_->setFrameMask(mask); }
+
     virtual Mat estimate(const Mat &frame0, const Mat &frame1, bool *ok = 0) CV_OVERRIDE;
 
 private:
@@ -235,6 +243,8 @@ public:
     void setOutlierRejector(Ptr<IOutlierRejector> val) { outlierRejector_ = val; }
     Ptr<IOutlierRejector> outlierRejector() const { return outlierRejector_; }
 
+    virtual void setFrameMask(InputArray mask) CV_OVERRIDE { mask_ = mask.getMat(); }
+
     virtual Mat estimate(const Mat &frame0, const Mat &frame1, bool *ok = 0) CV_OVERRIDE;
     Mat estimate(InputArray frame0, InputArray frame1, bool *ok = 0);
 
@@ -243,6 +253,7 @@ private:
     Ptr<FeatureDetector> detector_;
     Ptr<ISparseOptFlowEstimator> optFlowEstimator_;
     Ptr<IOutlierRejector> outlierRejector_;
+    Mat mask_;
 
     std::vector<uchar> status_;
     std::vector<KeyPoint> keypointsPrev_;
@@ -263,6 +274,8 @@ public:
     void setOutlierRejector(Ptr<IOutlierRejector> val) { outlierRejector_ = val; }
     Ptr<IOutlierRejector> outlierRejector() const { return outlierRejector_; }
 
+    virtual void setFrameMask(InputArray mask) CV_OVERRIDE { mask_ = mask.getMat(); }
+
     virtual Mat estimate(const Mat &frame0, const Mat &frame1, bool *ok = 0) CV_OVERRIDE;
     Mat estimate(const cuda::GpuMat &frame0, const cuda::GpuMat &frame1, bool *ok = 0);
 
@@ -271,6 +284,7 @@ private:
     Ptr<cuda::CornersDetector> detector_;
     SparsePyrLkOptFlowEstimatorGpu optFlowEstimator_;
     Ptr<IOutlierRejector> outlierRejector_;
+    GpuMat mask_;
 
     cuda::GpuMat frame0_, grayFrame0_, frame1_;
     cuda::GpuMat pointsPrev_, points_;
--- a/modules/videostab/include/opencv2/videostab/stabilizer.hpp
+++ b/modules/videostab/include/opencv2/videostab/stabilizer.hpp
@@ -77,6 +77,9 @@ public:
     void setFrameSource(Ptr<IFrameSource> val) { frameSource_ = val; }
     Ptr<IFrameSource> frameSource() const { return frameSource_; }
 
+    void setMaskSource(const Ptr<IFrameSource>& val) { maskSource_ = val; }
+    Ptr<IFrameSource> maskSource() const { return maskSource_; }
+
     void setMotionEstimator(Ptr<ImageMotionEstimatorBase> val) { motionEstimator_ = val; }
     Ptr<ImageMotionEstimatorBase> motionEstimator() const { return motionEstimator_; }
 
@@ -110,6 +113,7 @@ protected:
 
     Ptr<ILog> log_;
     Ptr<IFrameSource> frameSource_;
+    Ptr<IFrameSource> maskSource_;
     Ptr<ImageMotionEstimatorBase> motionEstimator_;
     Ptr<DeblurerBase> deblurer_;
     Ptr<InpainterBase> inpainter_;
--- a/modules/videostab/src/global_motion.cpp
+++ b/modules/videostab/src/global_motion.cpp
@@ -725,7 +725,7 @@ Mat KeypointBasedMotionEstimator::estimate(const Mat &frame0, const Mat &frame1,
 Mat KeypointBasedMotionEstimator::estimate(InputArray frame0, InputArray frame1, bool *ok)
 {
     // find keypoints
-    detector_->detect(frame0, keypointsPrev_);
+    detector_->detect(frame0, keypointsPrev_, mask_);
     if (keypointsPrev_.empty())
         return Mat::eye(3, 3, CV_32F);
 
@@ -815,7 +815,7 @@ Mat KeypointBasedMotionEstimatorGpu::estimate(const cuda::GpuMat &frame0, const 
     }
 
     // find keypoints
-    detector_->detect(grayFrame0, pointsPrev_);
+    detector_->detect(grayFrame0, pointsPrev_, mask_);
 
     // find correspondences
     optFlowEstimator_.run(frame0, frame1, pointsPrev_, points_, status_);
--- a/modules/videostab/src/stabilizer.cpp
+++ b/modules/videostab/src/stabilizer.cpp
@@ -391,6 +391,9 @@ void TwoPassStabilizer::runPrePassIfNecessary()
         {
             if (frameCount_ > 0)
             {
+                if (maskSource_)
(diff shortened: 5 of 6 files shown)


alalek:
Design-related questions:
1. Has been `mask` applied for `frame0` or `frame1`?
2. Did you consider alternative variant with `setFrameMask(InputArray mask)`?
3. Do we need support for per-frame mask? What is about "global" mask?

SBCV:
Thank you for having a look at this. 

Concerning 1.
The mask parameter is applied to frame0.
See `detector_->detect(frame0, keypointsPrev_, mask)` in global_motion.cpp. 
This is consistent with the indexing of the computed transformations.

Concerning 2.
What do you exactly mean? 
The PR adds a `InputArray mask` parameter to `ImageMotionEstimatorBase` and the corresponding inheriting classes, e.g.
`KeypointBasedMotionEstimator::estimate(InputArray frame0, InputArray frame1, bool *ok, InputArray mask)`
...
`Mat KeypointBasedMotionEstimatorGpu::estimate(const cuda::GpuMat &frame0, const cuda::GpuMat &frame1, bool *ok, InputArray mask)`
This allows a user to provide for each call a mask parameter.
If you mean with `setFrameMask(InputArray mask)` a global mask, then this is probably similar to `setMaskSource(maskSource)` in the `StabilizerBase` class. Which makes sense (see below)

Note the difference, where these functions are defined, i.e. `ImageMotionEstimatorBase` vs. `StabilizerBase` 

Concerning 3.
In my case, I use a vehicle detector to remove vehicles in a video sequence (since I want to use the background for stabilization and not the vehicles). This results in different masks for different frames. But I agree, for a static mask, one should not be forced to apply a video sequence with masks. Probably it is reasonable to allow the user to provide a single mask image (or a single `Mat()`) in addition to a video sequence. 
The current approach was the most elegant way I found to integrate this functionality in `videostab.cpp`. Since there is no convenient way to call `estimate(InputArray frame0, InputArray frame1, bool *ok, InputArray mask)` from  `videostab.cpp`

alalek:
> Concerning 2.
What do you exactly mean?

I mean what is about this alternative:
- preserve current interface of "estimate()" virtual method
- add separate virtual method `setFrameMask()`. Which should be called before `estimate()` (once for global masks, on each frame for different masks). Default implementation throws `StsNotImplemented` error if mask is not "empty".

Main goal of this - minimize/avoid breaking of existed code.

SBCV:
Ah I see your point, I'll update the PR accordingly.

SBCV:
Adjusted the PR according to your comments w.r.t. a separate `setFrameMask()` call. (I hope this is what you have been talking about)
Any other thoughts?

modules/videostab/include/opencv2/videostab/global_motion.hpp
```diff
@@ -180,6 +180,12 @@ class CV_EXPORTS ImageMotionEstimatorBase
     virtual void setMotionModel(MotionModel val) { motionModel_ = val; }
     virtual MotionModel motionModel() const { return motionModel_; }
 
+    virtual void setFrameMask(Mat /*val*/)
```
alalek:
```
virtual void setFrameMask(InputArray mask)
{
    if (!mask.empty())
        CV_Error(...);
}
```

modules/videostab/include/opencv2/videostab/global_motion.hpp
```diff
@@ -263,6 +274,8 @@ class CV_EXPORTS KeypointBasedMotionEstimatorGpu : public ImageMotionEstimatorBa
     void setOutlierRejector(Ptr<IOutlierRejector> val) { outlierRejector_ = val; }
     Ptr<IOutlierRejector> outlierRejector() const { return outlierRejector_; }
 
+     virtual void setFrameMask(Mat val) CV_OVERRIDE { mask_ = val; }
```
alalek:
```
mask.copyTo(mask_);
```
or
```
mask_ = mask.getMat();
```

The second requires responsibility of caller to avoid "used" masks modification.
The first way is slower but more reliable.

modules/videostab/include/opencv2/videostab/global_motion.hpp
```diff
@@ -271,6 +284,7 @@ class CV_EXPORTS KeypointBasedMotionEstimatorGpu : public ImageMotionEstimatorBa
     Ptr<cuda::CornersDetector> detector_;
     SparsePyrLkOptFlowEstimatorGpu optFlowEstimator_;
     Ptr<IOutlierRejector> outlierRejector_;
+    Mat mask_;
```
alalek: `GpuMat`?

modules/videostab/src/global_motion.cpp
```diff
@@ -711,6 +711,7 @@ KeypointBasedMotionEstimator::KeypointBasedMotionEstimator(Ptr<MotionEstimatorBa
     setDetector(GFTTDetector::create());
     setOpticalFlowEstimator(makePtr<SparsePyrLkOptFlowEstimator>());
     setOutlierRejector(makePtr<NullOutlierRejector>());
+    setFrameMask(Mat());
```
alalek: Probably we don't need this call.

modules/videostab/src/stabilizer.cpp
```diff
@@ -391,6 +391,16 @@ void TwoPassStabilizer::runPrePassIfNecessary()
         {
             if (frameCount_ > 0)
             {
+                Mat feature_mask = Mat();
```
alalek: ` = Mat()` is not needed.

modules/videostab/src/stabilizer.cpp
```diff
@@ -391,6 +391,16 @@ void TwoPassStabilizer::runPrePassIfNecessary()
...
+                if (maskSource_) {
+                    Mat mask_frame = maskSource_->nextFrame();
+                    if (!mask_frame.empty())
+                    {
+                        feature_mask = Mat(mask_frame.size(), CV_8UC1);
```
alalek: Initializer here is not needed too. `dst` can be allocated by processing functions.

modules/videostab/src/stabilizer.cpp
```diff
@@ -391,6 +391,16 @@ void TwoPassStabilizer::runPrePassIfNecessary()
...
+                        cv::cvtColor(mask_frame, feature_mask, cv::COLOR_BGR2GRAY);
```
alalek: It is better to move such logic into "maskSource_" implementation.

which should produce binary images (0, 255) instead of grayscale.

modules/videostab/src/stabilizer.cpp
```diff
@@ -391,6 +391,16 @@ void TwoPassStabilizer::runPrePassIfNecessary()
...
+                    }
+                }
+                motionEstimator_->setFrameMask(feature_mask);
```
alalek: This line should be under `if (maskSource_)` only.

modules/videostab/src/stabilizer.cpp
SBCV: Any suggestions? `maskSource_` is of type `IFrameSource`, which is defined in opencv (not opencv_contrib). Do you want to introduce a new class inheriting from `IFrameSource`, which overwrites `nextFrame()`?

alalek: Users should define their own class for mask generator, derived from `IFrameSource` interface (with inner conversions).
This class is not needed in OpenCV library itself (except probably tests).

SBCV:
Adjusted the code according to your comments, I hope I understood you correctly.
If not, let me know

modules/videostab/include/opencv2/videostab/global_motion.hpp
```diff
@@ -180,6 +180,12 @@ class CV_EXPORTS ImageMotionEstimatorBase
     virtual void setMotionModel(MotionModel val) { motionModel_ = val; }
     virtual MotionModel motionModel() const { return motionModel_; }
 
+    virtual void setFrameMask(InputArray mask)
+    {
+        if (!mask.empty())
+            CV_Error(Error::StsNotImplemented, "Mask must not be empty.");
```
alalek: > "Mask must not be empty."

"Mask support is not implemented".


This implementation is a stub for other (outside of OpenCV) implemented algorithms (they were implemented without mask).

modules/videostab/include/opencv2/videostab/frame_source.hpp
```diff
@@ -86,6 +86,23 @@ class CV_EXPORTS VideoFileSource : public IFrameSource
     Ptr<IFrameSource> impl;
 };
 
+class CV_EXPORTS MaskVideoFileSource : public IFrameSource
```
alalek: No need to copy implementation of VideoFileSource:
- try to derive this class from `VideoFileSource`
- or better try to make proxy over any other IFrameSource:
  * ctor: `MaskFrameSource(const Ptr<IFrameSource>& source);`
  * usage: `MaskFrameSource(makePtr<VideoFileSource>(...));`

Also threshold value should vary in case of conditions (nightly frames are darker). Or user want to avoid lighter areas (sky region). There are too many cases really.
There is no "rocket science" to implement it, but too many application specific details which OpenCV is not able to cover (and should not - lets keep OpenCV as simple as possible).
So it is better to let user to implement masking by yourself and remove it from OpenCV modules code (you can still put this implementation in samples, see notes above).

modules/videostab/include/opencv2/videostab/global_motion.hpp
```diff
@@ -263,6 +274,8 @@ class CV_EXPORTS KeypointBasedMotionEstimatorGpu : public ImageMotionEstimatorBa
     void setOutlierRejector(Ptr<IOutlierRejector> val) { outlierRejector_ = val; }
     Ptr<IOutlierRejector> outlierRejector() const { return outlierRejector_; }
 
+     virtual void setFrameMask(InputArray mask) CV_OVERRIDE { mask_ = mask.getMat(); }
```
alalek: indentation -> 4 spaces

modules/videostab/include/opencv2/videostab/frame_source.hpp
SBCV: Updated the corresponding part. Used a proxy to VideoFileSource (i.e. the second recommended option). Used a callback too generalize the functionality of maskframesource

modules/videostab/src/frame_source.cpp
```diff
@@ -116,5 +116,15 @@ int VideoFileSource::height() { return ((VideoFileSourceImpl*)impl.get())->heigh
 int VideoFileSource::count() { return ((VideoFileSourceImpl*)impl.get())->count(); }
 double VideoFileSource::fps() { return ((VideoFileSourceImpl*)impl.get())->fps(); }
 
+MaskFrameSource::MaskFrameSource(Ptr<IFrameSource> source): impl(source) {}
+
+void MaskFrameSource::reset() { impl->reset(); }
+Mat MaskFrameSource::nextFrame() {
+    Mat nextFrame = impl->nextFrame();
+    if (!nextFrame.empty())
+        maskCallback_(nextFrame);
```
alalek: > if (!nextFrame.empty())

This condition is implementation-specific too.
I believe callback should handle this instead:
- callback can generate some pre-defined mask
- or keep passing of the last observed mask

modules/videostab/include/opencv2/videostab/frame_source.hpp
```diff
@@ -86,6 +86,24 @@ class CV_EXPORTS VideoFileSource : public IFrameSource
     Ptr<IFrameSource> impl;
 };
 
+class CV_EXPORTS MaskFrameSource : public IFrameSource
```
alalek: I would prefer to remove "CV_EXPORTS" from here and define all methods inline in the header file.

modules/videostab/include/opencv2/videostab/stabilizer.hpp
```diff
@@ -77,6 +77,9 @@ class CV_EXPORTS StabilizerBase
     void setFrameSource(Ptr<IFrameSource> val) { frameSource_ = val; }
     Ptr<IFrameSource> frameSource() const { return frameSource_; }
 
+    void setMaskSource(Ptr<IFrameSource> val) { maskSource_ = val; }
```
alalek: Please use `const Ptr<IFrameSource>&` here.

modules/videostab/include/opencv2/videostab/frame_source.hpp
```diff
@@ -86,6 +86,24 @@ class CV_EXPORTS VideoFileSource : public IFrameSource
...
+{
+public:
+    MaskFrameSource(Ptr<IFrameSource> source);
```
alalek: `const Ptr<IFrameSource>&`

modules/videostab/include/opencv2/videostab/frame_source.hpp
SBCV: One quick question, your recommendation breaks the style in frame_source.hpp. 
Currently it is consistent, i.e.
`class CV_EXPORTS NullFrameSource : public IFrameSource`
`class CV_EXPORTS VideoFileSource : public  IFrameSource`
`class CV_EXPORTS MaskFrameSource : public IFrameSource`

modules/videostab/include/opencv2/videostab/stabilizer.hpp
SBCV: This also breaks the style in stabilizer.hpp, e.g.
`void setFrameSource(Ptr<IFrameSource> val) { frameSource_ = val; }`
`void setMaskSource(Ptr<IFrameSource>& val) { maskSource_ = val; }`

SBCV: I can change it no problem, just wanted to point that out

modules/videostab/include/opencv2/videostab/frame_source.hpp
alalek: These is no reason to make exportable API (like `NullFrameSource`) with very simple implementation.
`MaskFrameSource` looks like a proxy with "callback".

modules/videostab/include/opencv2/videostab/stabilizer.hpp
alalek: Right way is using of "const reference" (in all these cases).

> to point that out

:+1:

modules/videostab/include/opencv2/videostab/frame_source.hpp
SBCV: Just wanted to point that out, I'll change this as recommended

SBCV:
Changed the code according to your comments  and squashed all previous commits

SBCV:
Thank you for revising! Sry about the GPU errors, I thought the automatic builds would cover that :(

PR #2053, opened 2019-03-21
Update t_hash_int.cpp

LanPaHenDa:
### This pullrequest changes

Bug in the hashtableWrite.
fwrite the &node->data need to change to node->data


modules/surface_matching/src/t_hash_int.cpp
```diff
@@ -292,7 +292,7 @@ int hashtableWrite(const hashtable_int * hashtbl, const size_t dataSize, FILE* f
         while (node)
         {
             fwrite(&node->key, sizeof(KeyType), 1, f);
-            fwrite(&node->data, dataSize, 1, f);
```
alalek: Fix is almost correct, but it is not complete.

Take a look on this [code](https://github.com/opencv/opencv_contrib/blob/3.4.5/modules/surface_matching/src/t_hash_int.cpp#L346-L357) for reading.

asmorkalov:
@LanPaHenDa any news on the Alexander's proposal?

asmorkalov:
@LanPaHenDa Friendly reminder.

PR #2055, opened 2019-03-22
added detection of inverted Markers, tests included

szk1509:
### This pullrequest introduces the detection of inverted markers
i.e. it allows to detect white markers on a black background

<cut/>

Inverted Marker
inverted

Enclosed-Inverted Marker
img

### DIFF
--- a/modules/aruco/src/aruco.cpp
+++ b/modules/aruco/src/aruco.cpp
@@ -516,12 +516,13 @@ static bool _identifyOneCandidate(const Ptr<Dictionary>& dictionary, InputArray 
 
     // check if it is a white marker
     if(params->detectInvertedMarker){
-	// to get from 255 to 1
-        int invBError = _getBorderErrors(~candidateBits-254, dictionary->markerSize, params->markerBorderBits);
-	// white marker
+        // to get from 255 to 1
+        Mat invertedImg = ~candidateBits-254;
+        int invBError = _getBorderErrors(invertedImg, dictionary->markerSize, params->markerBorderBits);
+        // white marker
         if(invBError<borderErrors){
             borderErrors = invBError;
-            candidateBits=~candidateBits-254;
+            invertedImg.copyTo(candidateBits);
         }
     }
     if(borderErrors > maximumErrorsInBorder) return false;
--- a/modules/aruco/src/aruco.cpp
+++ b/modules/aruco/src/aruco.cpp
@@ -86,7 +86,8 @@ DetectorParameters::DetectorParameters()
       aprilTagCriticalRad( (float)(10* CV_PI /180) ),
       aprilTagMaxLineFitMse(10.0),
       aprilTagMinWhiteBlackDiff(5),
-      aprilTagDeglitch(0){}
+      aprilTagDeglitch(0),
+      detectInvertedMarker(false){}
 
 
 /**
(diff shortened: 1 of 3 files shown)


szk1509:
hi @alalek , 
Thank you for your advice :)
I've already corrected the issues (made the changes from #2036 on this new branch), so I will delete the other branch later.
Sorry about the delay and the inconveniences :apple:

modules/aruco/test/test_arucodetection.cpp
```diff
@@ -257,32 +280,45 @@ void CV_ArucoDetectionPerspective::run(int aprilDecimate) {
...
 
-                if(aprilDecimate == 1){
-                        params->cornerRefinementMethod = cv::aruco::CORNER_REFINE_APRILTAG;
+                /// create synthetic image
+                Mat img;
+                // marker :: Inverted
+                if(CV_ArucoDetectionPerspective::DETECT_INVERTED_MARKER == tryWith){
+                    img =
+                    projectMarker(dictionary, currentId, cameraMatrix, deg2rad(yaw), deg2rad(pitch),
+                                      distance, imgSize, markerBorder, groundTruthCorners, szEnclosed);
```
alalek: `img = projectMarker()` can be moved out of "if" block.
(Left "as is" original code of `img` initialization)

```
if(CV_ArucoDetectionPerspective::DETECT_INVERTED_MARKER == tryWith)
{
    img = ~img;
    params->detectInvertedMarker = true;
}
```

modules/aruco/src/aruco.cpp
```diff
@@ -512,7 +513,18 @@ static bool _identifyOneCandidate(const Ptr<Dictionary>& dictionary, InputArray
...
     int borderErrors =
         _getBorderErrors(candidateBits, dictionary->markerSize, params->markerBorderBits);
-    if(borderErrors > maximumErrorsInBorder) return false; // border is wrong
+
+    // check if it is a white marker
+    if(params->detectInvertedMarker){
+	// to get from 255 to 1
+        int invBError = _getBorderErrors(~candidateBits-254, dictionary->markerSize, params->markerBorderBits);
+	// white marker
+        if(invBError<borderErrors){
```
alalek: Please fix indentation (it is "jumping").

modules/aruco/src/aruco.cpp
```diff
@@ -512,7 +513,18 @@ static bool _identifyOneCandidate(const Ptr<Dictionary>& dictionary, InputArray
...
+            borderErrors = invBError;
+            candidateBits=~candidateBits-254;
```
alalek: > `~candidateBits-254`

This expression has been used multiple times.
It is better to "cache" its result.

modules/aruco/test/test_arucodetection.cpp
```diff
@@ -257,32 +280,45 @@ void CV_ArucoDetectionPerspective::run(int aprilDecimate) {
     Ptr<aruco::Dictionary> dictionary = aruco::getPredefinedDictionary(aruco::DICT_6X6_250);
 
     // detect from different positions
-    for(double distance = 0.1; distance <= 0.5; distance += 0.2) {
-        for(int pitch = 0; pitch < 360; pitch += 60) {
-            for(int yaw = 30; yaw <= 90; yaw += 50) {
+    for(double distance = 0.1; distance < 0.7; distance += 0.2) {
+        for(int pitch = 0; pitch < 360; pitch += (distance == 0.1? 60:180)) {
+            for(int yaw = 70; yaw <= 120; yaw += 40){
```
alalek: > int yaw = 70; yaw <= 120; yaw += 40

yaw: 70, 110 ?

Old yaw values are: 30, 80

modules/aruco/test/test_arucodetection.cpp
```diff
@@ -257,32 +280,45 @@ void CV_ArucoDetectionPerspective::run(int aprilDecimate) {
...
+                    params->cornerRefinementMethod = cv::aruco::CORNER_REFINE_APRILTAG;
                 }
 
+                /// detect markers
+                vector< vector< Point2f > > corners;
+                vector< int > ids;
                 aruco::detectMarkers(img, dictionary, corners, ids, params);
 
-                // check results
+                /// check results
```
alalek: `///` is usually used for Doxygen comments (in header files).

modules/aruco/test/test_arucodetection.cpp
szk1509: even though the detection of a default marker happens using that yaw value, it was not the case for the inverted marker, however, it also occurs in the opposite way, that is, where the inverted marker is detected more easily than the default marker.

hamddan4:
The dictionary has to be inverted too?

PR #2072, opened 2019-04-02
Added sample python code for facemark landmark detection

shardulparab97:
Have added sample python code for the fit function for facemark landmark detection.

There will be a case of std::bad_alloc() error which will be solved via - (https://github.com/opencv/opencv_contrib/issues/1661#issuecomment-397802106 ) or the PR which is underway for the same.

Also please let me know if this is the right location to add this sample file.

### DIFF
--- a/modules/face/samples/python/facemark_fit_demo.py
+++ b/modules/face/samples/python/facemark_fit_demo.py
@@ -1,0 +1,66 @@
+'''
+This example shows the using the Facemark API(fitting) for python.
+
+USAGE: facemark_fit_demo.py <method - kazemi / aam / lbf > <path_to_input_image> <path to cascade classifier> <path to the pre-trained model>
+'''
+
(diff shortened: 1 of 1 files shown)


asmorkalov:
:+1:

asmorkalov:
@alalek May I merge the patch?

asmorkalov:
@shardulparab97 Please squash commits before merge.

alalek:
@asmorkalov This should not be merged without proper fix in bindings of face module:

> the above code will give a std::bad_alloc() error (due to the fit function)

Perhaps something has been improved in GSoC project, but this sample should be rechecked again.

asmorkalov:
@shardulparab97 Could you check the same with current master branch. The sample works for me:

asmorkalov:
@shardulparab97 Do you have any news on sample code. Does it work with upstream branch correctly?

PR #2079, opened 2019-04-09
ovis: addResourceLocation - only retain unique locations

### LINKED ISSUE #2078: ovis: addResourceLocation

- OpenCV => 4.1
- Operating System / Platform => Windows
- Compiler => vc14 (Visual Studio 2015)

##### Detailed description

All `ovis`-resources are managed globally by a `ResourceGroupManager`-singleton. When `cv2::ovis::addResourceLocation(const String& path)` is called, `path` is added to the list of resource locations to load resources from. If the same path is added twice, `ovis` will try to load all contained resources twice, which results in an `ItemIdentityException`.

It would be nice, if adding resources location twice would not lead to a crash. Perhaps by making `_extraResourceLocations` a `std::set` of normalized paths.

##### Steps to reproduce

```.python
cv2.ovis.addResourceLocation("./resources")
cv2.ovis.addResourceLocation("./resources")
win = cv2.ovis.createWindow("name", (1024,768), cv2.ovis.SCENE_SEPERATE)
```

paroj:
fixes #2078

### DIFF
--- a/modules/ovis/src/ovis.cpp
+++ b/modules/ovis/src/ovis.cpp
@@ -22,7 +22,7 @@ const char* RESOURCEGROUP_NAME = "OVIS";
 Ptr<Application> _app;
 
 static const char* RENDERSYSTEM_NAME = "OpenGL 3+ Rendering Subsystem";
-static std::vector<String> _extraResourceLocations;
+static std::set<String> _extraResourceLocations;
 
 // convert from OpenCV to Ogre coordinates:
 static Quaternion toOGRE(Degree(180), Vector3::UNIT_X);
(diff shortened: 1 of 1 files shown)


alalek:
@paroj 3.4 branch?

paroj:
started using c++11. If its fine in 3.4, then I can rebase

alalek:
In contrib we can use something like [this](https://github.com/opencv/opencv_contrib/blob/3.4.6/modules/cvv/CMakeLists.txt#L1-L2) (disables module completely if there is no C++11)

BTW, C++14 is enabled by default in modern GCC.

paroj:
OGRE >= 1.11 also requires C++11. However there seem to be some users e068b62a1432d4d5688693a9e20bf175dfaa9a3e that are still using older OGRE versions. Therefore I would leave 3.4 as is, s.t. there is an usable C++03 OGRE/ OpenCV combination.

PR #2080, opened 2019-04-09
aruco: forward drawAxis to drawFrameAxes

paroj:


alalek:
@paroj Thank you!

I believe we can put this patch into 3.4 branch first. Could you please rebase it?

PR #2083, opened 2019-04-11
fix Text module build with Tesseract

Bleach665:
resolves https://github.com/opencv/opencv/issues/14302

### DIFF
--- a/modules/text/src/erfilter.cpp
+++ b/modules/text/src/erfilter.cpp
@@ -47,6 +47,9 @@
 #include <fstream>
 #include <queue>
 
+#undef min
+#undef max
+
 #if defined _MSC_VER && _MSC_VER == 1500
     typedef int int_fast32_t;
 #else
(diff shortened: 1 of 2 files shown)


Bleach665:
Resolve this https://github.com/opencv/opencv/issues/14302

modules/text/src/erfilter.cpp
```diff
@@ -47,6 +47,9 @@
 #include <fstream>
 #include <queue>
 
+#undef min
+#undef max
```
alalek: It doesn't look reliable, especially there are no tesseract headers on the lines above.
Other files could be broken in the similar way easily by using normal `std::min()` / `std::max()`.

Please try this patch instead: https://github.com/alalek/opencv_contrib/commit/issue_2084

Bleach665: Your patch resolves this bug. Build was successful.

PR #2089, opened 2019-04-12
Tesseract and Leptonica configuration using cmake find_package added.

Bleach665:
### This pullrequest changes

Added Tesseract and Leptonica configuration using find_package. Is also possible directly specifying libs and include dir.
Configuring Leptonica using find_package makes it possible to link image processing libraries  which was used to build Leptonica to Text module. It resolve VS linker errors like this:
````
leptonica-1.79.0d.lib(tiffio.obj) : error LNK2019: unresolved external symbol TIFFCleanup referenced in function pixReadStreamTiff [C:\000ttt\prebuild_x64\modules\text\opencv_text.vcxproj]
leptonica-1.79.0d.lib(tiffio.obj) : error LNK2019: unresolved external symbol TIFFClose referenced in function
pixWriteTiffCustom [C:\000ttt\prebuild_x64\modules\text\opencv_text.vcxproj]
leptonica-1.79.0d.lib(tiffio.obj) : error LNK2019: unresolved external symbol TIFFGetField referenced in function extractG4DataFromFile [C:\000ttt\prebuild_x64\modules\text\opencv_text.vcxproj]
````

### CODE REVIEW

**Comment by Bleach665** [comment]
@alalek, please take a look at this. I need to create issue or this pull request is enough?
Buildbot report that build was unsuccessful in Docs and it's strange. Changes are only in CMakeLists.txt file.

**Comment by alalek** [comment]
1. Don't re-open PRs. All stuff can be changed without reopening PR (except source branch name).
2. Add more details how did you installed tesseract/leptonica on your machine. Did you use static builds? Did your follow [this guide](https://docs.opencv.org/3.4/db/d4c/tutorial_install_tesseract.html)? What are the differences?
3. OpenCV doesn't use Leptonica directly. It is used by tesseract only, so such dependency should be provided by tesseract build scripts.

I have Fedora 29:
- installed packages: `tesseract-devel` with **automatic dependency** on:
  * `tesseract` (3.05.02)
  * `leptonica` (1.77)
  * `leptonica-devel`
- CMake's `find_package(Tesseract)` (without OpenCV code) doesn't work (`Tesseract_FOUND=0`)
- only pkgconfig information is available, so we must to call `pkg_check_modules()`.
- `pkg_check_modules(Tesseract tesseract)` is enough. Without "lept" requirement due no direct dependency OpenCV on leptonica:
  * `Tesseract_INCLUDE_DIRS=/usr/include/tesseract;/usr/include/leptonica`
  * `Tesseract_LIBRARIES=tesseract`
  * `Tesseract_LINK_LIBRARIES=/usr/lib64/libtesseract.so`
  * `Tesseract_STATIC_LIBRARIES=tesseract;pthread;lept;z;png16;z;jpeg;gif;tiff;webp` (**lept** is propagated here automatically) - it is not used in my case, just for info.
- `opencv_test_text` has been built succesfuly.
- `example_text_end_to_end_recognition` works well.

This patch **brokes** this config.

Ubuntu 16.04:
- package [libtesseract-dev](https://packages.ubuntu.com/xenial/amd64/libtesseract-dev/filelist)
  * no CMake's "find_package()" scripts
  * provides `/usr/lib/pkgconfig/tesseract.pc` only

Ubuntu 14.04 doesn't have even [tesseract.pc](https://packages.ubuntu.com/trusty/amd64/libtesseract-dev/filelist) file.

Ubuntu 18.04 still uses [tesseract.pc](https://packages.ubuntu.com/bionic/amd64/libtesseract-dev/filelist) file only (without CMake's scripts).

This patch can't detect Tesseract on Ubuntu properly (looks like Ubuntu 14.04 has been broken before that).

File: modules/text/CMakeLists.txt [hunk 1]
```diff
@@ -4,11 +4,42 @@ ocv_define_module(text opencv_ml opencv_imgproc opencv_core opencv_features2d op
 if(NOT CMAKE_CROSSCOMPILING OR OPENCV_FIND_TESSERACT)
   set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
   find_package(Tesseract QUIET)
+  find_package(Leptonica QUIET)
```
alalek: Try this code:
```
if(NOT CMAKE_CROSSCOMPILING OR OPENCV_FIND_TESSERACT)
  find_package(Tesseract QUIET)  # Prefer CMake's standard locations (including Tesseract_DIR)
  if(NOT Tesseract_FOUND)
    include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/FindTesseract.cmake")  # OpenCV's fallback
  endif()
  if(Tesseract_FOUND)
```
(other changes should be reverted)

You can use `ocv_cmake_dump_vars(tesseract)` for debugging of search results.

Bleach665: About my build:
OS Windows 10, VS 2017, CMake 3.12.
To properly work Tesseract require that Leptonica must be built with image proc. libs. At least with tiff and libpng. In turn, libpng require zlib, therefore, except Tesseract and Leptonica we need to build it too.
I didnt use guide, just configured each lib through CMake GUI and build via msbuild.
Shared build opencv and libs was successful and work fine, but in statical build, we receive unresolved external errors to libs with which Leptonica was built.
In case that Leptonica was configured using find_package ${Leptonica_LIBRARIES} consist image proc. dependencies and ocv_target_link_libraries successfully added it to opencv_text module VS project.
Goal this patch first was resolve Windows static build via VS.
I have now took into account of your comments and make changes according to them.
This 
> Try this code:
> 
> if(NOT CMAKE_CROSSCOMPILING OR OPENCV_FIND_TESSERACT)
>   find_package(Tesseract QUIET)  # Prefer CMake's standard locations (including Tesseract_DIR)
>   if(NOT Tesseract_FOUND)
>     include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/FindTesseract.cmake")  # OpenCV's fallback
>   endif()
>   if(Tesseract_FOUND)
code I not tested because it not fix Win static build.

After update I test this patch in Windows 10 and Xubuntu 18.04.1 LTS.
In Windows I try 2 options: 
- specify Tesseract_DIR and Leptonica_DIR (static build)
- specify Tesseract_INCLUDE_DIR, Tesseract_LIBRARY and Lept_LIBRARY (shared build).
Both of them works fine. Build was successful and example_text_end_to_end_recognition work correct.
In Xubuntu I install libtesseract-dev and it dependencies. CMake properly detect Tesseract_INCLUDE_DIR, Tesseract_LIBRARY and Lept_LIBRARY. Build also was successful and example_text_end_to_end_recognition work correct.

modules/text/CMakeLists.txt
```diff
@@ -3,12 +3,46 @@ ocv_define_module(text opencv_ml opencv_imgproc opencv_core opencv_features2d op
...
+  endif()
+ 
+  if(Tesseract_INCLUDE_DIR AND Tesseract_LIBRARY AND Lept_LIBRARY)
+    message(STATUS "Tesseract:   YES ${Tesseract_VERSION}")
     set(HAVE_TESSERACT 1)
-    ocv_include_directories(${Tesseract_INCLUDE_DIRS})
-    ocv_target_link_libraries(${the_module} ${Tesseract_LIBRARIES})
+ 
+    ocv_include_directories(${Tesseract_INCLUDE_DIR})
+    ocv_target_link_libraries(${the_module} ${Tesseract_LIBRARY})
```
alalek: > Tesseract_LIBRARIES

Replacing to `Tesseract_LIBRARY` is wrong.
Proper `find_package(Tesseract)` script brings into `LIBRARIES` variable all dependencies required for **static** build.
"Proper" means created correctly.

OpenCV doesn't use leptonica directly, so it should not care about it. The same note is about mentioned libpng, zlib, tiff (no conflicts?) libraries too.


Example (Fedora 29):
```
$ pkgconf --libs tesseract
-ltesseract
$ pkgconf --static --libs tesseract
-ltesseract -lpthread -llept -lz -lpng16 -lz -ljpeg -lgif -ltiff -lwebp
```

All transitive dependencies are handled via `pkgconfig`.

alalek:
> in statical build

In static builds all dependencies including "internal" should be provided by library itself (Tesseract) and should be applied automatically (by CMake "find"/"config" scripts or similar).
This can't / should not be resolved by OpenCV (we should not know anything about leptonica dependency because we don't use any API from there).

Workaround is: specify required variable via CMake command-line.

Issue #2094, opened 2019-04-16
SparseRLOFOpticalFlow unintended assertions   C++ and Python-Binding


tsenst:
##### System information (version)

- OpenCV => 4.1.0

##### Detailed description

There are two problems and an issue with the SparseRLOF implementation. I am working on this issue but wanted to open this one to report the known problems.

- The first refers to the closed issue #2091. Referring  https://github.com/opencv/opencv_contrib/issues/2087#issuecomment-483261683 the code unintended crash if the input images have very flat textures. Problem is at this position where the Roi gets to large. https://github.com/opencv/opencv_contrib/blob/eb7c61fa18040e0aaebd990c9cea58da83c928c1/modules/optflow/src/rlof/rlof_invokerbase.hpp#L111
- Second is an unintended assertion which is thrown when using the python-binding. The problem lies here: https://github.com/opencv/opencv_contrib/blob/eb7c61fa18040e0aaebd990c9cea58da83c928c1/modules/optflow/src/rlofflow.cpp#L313
-Third as discussed in issue #2091 at https://github.com/opencv/opencv_contrib/issues/2087#issuecomment-482949038 the blur image of the image buffers are not connected to the cross-based segmentation. It should be checked if this results into a loss of accuracy. If not, the computation of blur images can be removed

##### Steps to reproduce

- for the first issue refer to  https://github.com/opencv/opencv_contrib/issues/2087#issuecomment-483261683
- for the second the following python example : 
```
x = np.linspace(0, img1.shape[1], 10)
y = np.linspace(0, img1.shape[0], 10)
posx, posy = np.meshgrid(x,y)
posx = np.reshape(posx, (posx.shape[0] * posx.shape[1]))
posy = np.reshape(posy, (posy.shape[0] * posy.shape[1]))
prevPts = np.squeeze(np.dstack((posx, posy)))
prevPts = prevPts.astype(np.float32)
currPts = prevPts.copy()
rlof = cv2.optflow.SparseRLOFOpticalFlow_create()
ret = rlof.calc(img1, img2, prevPts, currPts)
```
will result in:
> OpenCV(4.1.0-dev) Error: Assertion failed (channels() == CV_MAT_CN(dtype)) in copyTo, file /home/senst/workspace/RLOFatOpencv/opencv/modules/core/src/copy.cpp, line 254
> Traceback (most recent call last):
>   File "/home/senst/workspace/RLOFatOpencv/pyexample/example-py", line 19, in <module>
[...]

PR #2091, opened 2019-04-14
optflow: add a check for 3 channel images for sparse RLOF

berak:
(hopefully) resolves #2087

3 channel images are required for the (default) SR_CROSS RLOF optflow implementation, add a check for the sparse version similar [to the dense one here](https://github.com/opencv/opencv_contrib/blob/master/modules/optflow/src/rlofflow.cpp#L117)

PR #2095, opened 2019-04-16
optflow: fix RLOF SR_CROSS check

alalek:
relates #2091
relates #2094 (1)


tsenst:
The patch looks good. Now an proper error message should appear in only gray-scaled image is used with SR_CROSS.

alalek:
:+1:

PR #2103, opened 2019-04-19
Added a fix for issue 2102.

### LINKED ISSUE #2102: Selective Search crashes on images with a width:height or height:width ratio greater than about 2.43
- OpenCV => 4.1
- Operating System / Platform => Linux Mint 18.1 Serena
- Installed opencv-contrib-python 4.1.0 module from pip

##### Detailed description

As described in the title, when using opencv in python 3, the Selective Search Segmentation module raises an exception on an image with a width-to-height or height-to-width ratio greater than about 2.43. The exact ratio where the problem starts differs depending on the size of the input image, but is close to 2.4 or 2.5 in every case I've tried.

##### Steps to reproduce

```.py
import cv2
import selectivesearch as ss
import skimage.data
import numpy as np

# image is 300 x 451 x 3
img = skimage.data.chelsea()
cvss = cv2.ximgproc.segmentation.createSelectiveSearchSegmentation()

# this works
cvss.setBaseImage(img)
cvss.switchToSingleStrategy()
single_cvss_results = cvss.process()

# this works too
img = cv2.resize(img, (300, 731))
cvss.setBaseImage(img)
cvss.switchToSingleStrategy()
single_cvss_results = cvss.process()

# this raises an exception when cvss.process() is called!
img = cv2.resize(img, (300, 732))
cvss.setBaseImage(img)
cvss.switchToSingleStrategy()
single_cvss_results = cvss.process()

# any greater h:w ratio will cause the same exception... 
img = cv2.resize(img, (300, 1000))
cvss.setBaseImage(img)
cvss.switchToSingleStrategy()
single_cvss_results = cvss.process()

# this works...
img = cv2.resize(img, (30, 79))
cvss.setBaseImage(img)
cvss.switchToSingleStrategy()
single_cvss_results = cvss.process()

# this raises the same exception when cvss.process() is called!
img = cv2.resize(img, (30, 80))
cvss.setBaseImage(img)
cvss.switchToSingleStrategy()
single_cvss_results = cvss.process()
```

The exception raised is:
```
cv2.error: OpenCV(4.1.0) /io/opencv/modules/core/src/matrix.cpp:466: error: (-215:Assertion failed) 0 <= roi.x && 0 <= roi.width && roi.x + roi.width <= m.cols && 0 <= roi.y && 0 <= roi.height && roi.y + roi.height <= m.rows in function 'Mat'
```

An identical exception is raised if the height and width values in the above code are swapped.  The same exception is also raised regardless of whether the single, fast, or quality SS strategy is selected.

I've also been able to replicate this on another system running Ubuntu 16.04, using opencv 3.4.2, and with different input images.

I've cross-posted this on the issues page for the opencv_contrib repo, since I wasn't sure where it belonged.

AHolliday:
resolves #2102

### This pullrequest changes

During one of the stages of selective search, the image is rotated 45 degrees and gradients are computed along the X and Y directions in the rotated image.  The gradient image is then rotated back to the original orientation and cropped to remove surrounding empty space, leaving it with the same shape as the original image.  However, the implementation of the crop will specify an ROI with negative x or y values if the image's native aspect ratio is too high or too low (basically, if the image is far from square).

This is because the bounding box of the rotated image is strictly bigger than the original image if the original is sufficiently close to square; but if the original is too "skinny", the bounding box of its rotation will be less tall or wide than the original.

The fix is to threshold the x and y computed for the ROI at 0.

Note: while fixing this bug, I also discovered that parts of "narrow" gradient images were being clipped at their edges due to improper cropping, again due to the apparent assumption that the bounding box of the rotated image was always strictly bigger in both axes than the original image.  This PR includes a fix for that as well.


alalek:
@AHolliday Thank you for investigating of this problem!

@the-glu Could you take a look on this patch?

AHolliday:
Anyone had a chance to take a look at this?

alalek:
@AHolliday Looks like there is no objections about the fix (also there are no automatic tests for this functionality too), so lets apply proposed fix :+1:

As a "bugfix" this patch should [go](https://github.com/opencv/opencv/wiki/How_to_contribute#faq) into [3.4 branch](https://github.com/opencv/opencv/wiki/Branches) first. We will merge changes from 3.4 into master regularly (weekly/bi-weekly).

So, please:

alalek:
Something goes wrong here with git, so GitHub closes PR (and it can't be reopened for now at least from our/maintainers side).

Please revert changes back, something like:
```
git checkout -B fix_2102 54bac53
git push origin fix_2102 -f
```
(some extra commands to recover source tree: `git stash`, `git rebase --abort`)

Then try re-open this PR.

---

If anything fails, then open new PR (from 3.4 branch) with cross-link on this PR.
Backup of your patch is [here](https://github.com/alalek/opencv_contrib/commits/pr2103).

AHolliday:
Alright, I made a new PR: https://github.com/opencv/opencv_contrib/pull/2132

AHolliday:
Although it seems now that the patch is failing to pass the cmake tests.

PR #2113, opened 2019-05-06
Simplified Quality API

### LINKED ISSUE #2111: cv.quality.QualitySSIM_compute(mat2,mat1) return only one chanel for rgb image in python language

- OpenCV => 4.1
- Operating System / Platform => macos 10.12.6
- Compiler => pip install opencv-contrib-python

##### Detailed description

my code is like this 
```
import cv2 as cv

im1 = "/Users/admin/09.temp_src/476d375d4bd3980.jpg";
im2 = "/Users/admin/work/l75d4bda68432523980.jpg";

mat1 = cv.imread(im1)
mat2 = cv.imread(im2)

print("shape of image1: (%s,%s,%s)" % mat1.shape)
print("shape of image2: (%s,%s,%s)" % mat2.shape)

ssim_scalar ,map= cv.quality.QualitySSIM_compute(mat2,mat1)
print("ssim of per channel:")
print(ssim_scalar)

```
the images are rgb image with 3 channels
but after run the code , it show only one chanel like this 
```
shape of image1: (1554,2560,3)
shape of image2: (1554,2560,3)
ssim of per channel:
(0.9802379911743933, 0.0, 0.0, 0.0)

Process finished with exit code 0
```

i am sorry for boring you !

clunietp:
### This pullrequest changes
Simplifies the quality API and brings it in line with the majority of OpenCV.  This change allows more intuitive usage, especially for Python users

resolves #2111


modules/quality/test/test_brisque.cpp
```diff
@@ -21,13 +21,33 @@ const cv::Scalar
...
+    }
+    catch (const SkipTestException&)
+    {
+        throw;//rethrow
+    }
+    catch (...)
+    {
+        if (required)
+            throw;
+        throw SkipTestException("File not found: " + relative_path);
```
alalek: > cvtest::findDataFile sometimes throws even if setting required=false

It throws on CI (via `OPENCV_TEST_REQUIRE_DATA=1`).
To ensure that all test data files are property committed / deployed (CI should be properly - there were some incidents of using of missing files in tests).

clunietp: That makes sense.  I'll need to add the BRISQUE model data to `opencv_extra` to get the tests to pass, is that OK with you?  Otherwise I'm good skipping the BRISQUE tests if the files aren't there, like we were doing before, and this is my hack to make it happen.  Which do you prefer?

clunietp: Alternatively, if there's a way to reference the OpenCV build directory/samples when the tests are run, we can keep additional data out of `opencv_extra`.  I'll poke around a bit to see if this can happen, but if you have any pointers for me that would be great.  Thanks

alalek: What is size of this model/models?
We prefer don't commit large blobs into opencv_extra (>10Mb). If these models may be useful for samples, then `opencv_extra` is not a right place too (we prefer to store test data only in opencv_extra).

Additional test data directory can be referenced separately like [DNN tests does](https://github.com/opencv/opencv/blob/3.4.6/modules/dnn/test/test_main.cpp#L11) (files are deployed on CI by maintainers, but we need a link).

For large files (<100Mb) or which should be versioned/updated, we have [opencv_3rdparty](https://github.com/opencv/opencv_3rdparty) repository (we using separate branch for "case" branched from "root" branch). I will provide extra instruction if this is that case.

clunietp: @alalek I have two files, total of 566KB.  

The models are useful not only for the samples, but also for anyone that would want to use the algorithm in production.  This is the more likely scenario vs a user training their own model.  So it seems that `opencv_extra` is probably not the correct place.

`opencv_3rdparty` seems to be the right place if the models must be deployed outside of the `opencv_contrib` build infrastructure

Previously, I had added:

    ocv_add_testdata(samples/ contrib/quality FILES_MATCHING PATTERN "*.yml" )

to my CMakeLists.txt.  This allowed me to search for the model in `{$OPENCV_DIR}/testdata/contrib/quality/` but that only seemed to work on Linux and not Windows, IIRC.  (ie, the `/testdata/` directory isn't being created on Windows during a build/install, unless I'm doing it wrong)

Can you please provide instructions for contributing to the `opencv_3rdparty` repo, if this is still the best path forward?  Thank you

alalek: Currently check is forced on Linux builders only.
Windows builders just print "skip" message:
```
[ RUN      ] CV_Quality_BRISQUE.static_
[     SKIP ] OpenCV tests: Can't find data file: brisque_range_live.yml
[       OK ] CV_Quality_BRISQUE.static_ (4 ms)
```

OK, files are already committed (opencv_contrib is fine too).

> `ocv_add_testdata(samples/ contrib/quality FILES_MATCHING PATTERN "*.yml" )`

Good!

> brisque_range_live.yml

Should be used as: `contrib/quality/brisque_range_live.yml`

alalek: or add [searching subdirectory](https://github.com/opencv/opencv_contrib/blob/3.4.6/modules/text/test/test_main.cpp#L8).

clunietp: Thanks @alalek, this seems to be working well for both Win and Linux, at least on my machine.  Hopefully the buildbot agrees when it comes back online

modules/quality/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 set(the_description "Image Quality Analysis API")
 ocv_define_module(quality opencv_core opencv_imgproc opencv_ml WRAP python)
-ocv_add_testdata(samples/ contrib/quality
-    FILES_MATCHING PATTERN "*.yml"
-)
\ No newline at end of file
+
+# add brisque model, range files to installation.  the tests and quality module use these files unless otherwise directed
+file(GLOB QUALITY_MODEL_DATA samples/*.yml)
+install(FILES ${QUALITY_MODEL_DATA} DESTINATION ${OPENCV_OTHER_INSTALL_PATH}/quality COMPONENT libs)
```
alalek: Please keep using of `ocv_add_testdata()` macro.
Tests should run from the build directory too (without any installation path).

modules/quality/src/qualitybrisque.cpp
```diff
@@ -53,6 +54,16 @@ namespace
     // brisque intermediate matrix element type.  float if BRISQUE_CALC_MAT_TYPE == CV_32F, double if BRISQUE_CALC_MAT_TYPE == CV_64F
     using brisque_calc_element_type = float;
 
+    // returns path to a file located in the opencv install/data path
+    cv::String get_install_data_path(const cv::String& file_name)
+    {
+#if !defined( OPENCV_INSTALL_PREFIX ) || !defined( OPENCV_DATA_INSTALL_PATH )
+        return cv::String();
+#else
+        return cv::String(OPENCV_INSTALL_PREFIX) + "/" + OPENCV_DATA_INSTALL_PATH + "/quality/" + file_name;
```
alalek: It is broken due missing files in "guessed" location.

Installation prefix makes no sense on Windows (final location on users system is not fixed).

Lets avoid handling of "empty" model case (tests / samples should pass valid path).

clunietp:
Thank you @alalek !  👍

PR #2131, opened 2019-05-31
docs: remove redundant `return`

asashour:
This change is to remove redundant `return` doc, for a method with `void` type.

It prevents the success of https://github.com/opencv/opencv/pull/14667, and the corresponding error is [here](https://pullrequest.opencv.org/buildbot/builders/precommit_docs/builds/20631/steps/make%20doxygen/logs/warnings%20%282%29)

Please suggest if other text should be written instead.

```
opencv=javadoc
**WIP**
force_builders=Docs
```

### DIFF
--- a/modules/bioinspired/src/transientareassegmentationmodule.cpp
+++ b/modules/bioinspired/src/transientareassegmentationmodule.cpp
@@ -166,7 +166,7 @@ public:
 
     /**
      * access function
-     * @return the last segmentation result: a boolean picture which is resampled between 0 and 255 for a display purpose
+     * return the last segmentation result: a boolean picture which is resampled between 0 and 255 for a display purpose
      */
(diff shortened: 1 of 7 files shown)


modules/bioinspired/src/transientareassegmentationmodule.cpp
```diff
@@ -166,7 +166,6 @@ class TransientAreasSegmentationModuleImpl : protected BasicRetinaFilter
 
     /**
      * access function
-     * @return the last segmentation result: a boolean picture which is resampled between 0 and 255 for a display purpose
```
alalek: Just remove `@` specifier.

Internal code comments must not be a part of public API documentation.

alalek:
> Internal code comments must not be a part of public API documentation.

In other words - fixing `.cpp` can't help with documentation.
You should fix public `.hpp` file instead.

PR #2143, opened 2019-06-07
Add AprilTag 3 module

catree:
I would like to add AprilTag 3 for fast, robust and accurate fiducial tags detection, [code](https://github.com/AprilRobotics/apriltag) by @mkrogius

The idea is:

- no change in the AprilTag 3 source code. Modifications should be done [upstream](https://github.com/AprilRobotics/apriltag). This way, updating the AprilTag code should be straightforward.
- it needs `pthreads`. On Windows with MSVC, [pthreads4w](https://sourceforge.net/projects/pthreads4w/) should be built automatically to support at least Linux and Windows

I don't really know CMake. The first thing should be to check CMake code.
The interface should be quite simple, example code:

        apriltag::AprilTagDetector detector(apriltag::TAG_36h11);
        vector<vector<Point2d> > corners;
        vector<int> ids;
        detector.detectTags(img, corners, ids);

        vector<pair<Matx31d, Matx31d> > rvecs, tvecs;
        detector.estimateTagsPosePnP(corners, tagSize, cameraMatrix, distCoeffs, rvecs, tvecs);

To compute the pose, two options:
- `estimateTagsPoseAprilTag()` that returns at most two poses, using the AprilTag library
- `estimateTagsPosePnP()` that uses `solvePnPGeneric()` and uses `SOLVEPNP_IPPE_SQUARE` by default

---

Main "issue" is that with some compiler, the build takes lots of time due to some big source file and the compiler trying to optimize the code ([related](https://github.com/AprilRobotics/apriltag/issues/16)).

One way to reproduce the issue, for instance on Ubuntu 16.04:

```
gcc --version
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.11) 5.4.0 20160609
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

Then run:

```
git clone https://github.com/AprilRobotics/apriltag.git
cd apriltag/
mkdir build
cd build/
cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo
make -j4
```

Build should takes lots of time (> 10 min).

To overcome this:
- add a CMake option to enable explicitly the build of `CIRCLE49h12`, `CUSTOM48h12`, `STANDARD52h13`. These should be tag families that take lots of time to build with some compiler / configuration.
- introduce a new mechanism to load these families at runtime?

Probably the first solution should be fine until a new mechanism is introduced upstream.

---

TODO:

- fix potential warnings, memory leaks, ...
- validate build on buildbot
- validate API
- documentation
- check that Java and Python interfaces work

Suggestions, modifications, feedbacks are welcome.


mkrogius:
I'm very glad to see this happening! Also thanks for the method for reproducing the build issue, I hope to have enough free time to look into that this week

paroj:
I dont really like to have this as a separate module. We already have an unified marker detection API through the aurco module, which already includes apriltag2.

This approach allows seamless integration with language bindings, ChAruco patterns and various marker dictionaries.
I dont think it makes much sense to introduce and document a new API here.

The apriltag2 code in aruco intentionally diverges from upstream to use opencv primitives instead of the ad-hoc apriltag solutions.
The opencv implementations are both more tested and better optimized (threading, SIMD, OpenCL). The replaced primitives include
- gaussian blur
- PnP
- SVD
- pthreads (by parallell_for_)

The modified code was verified to reproduce upstream results. Therefore I would prefer to upgrade the apriltag2 code in aruco to apriltag3. The modification should be straightforward, given both codebases share the same ancestor.

catree:
Please refer to the original [AprilTag 3 repository](https://github.com/AprilRobotics/apriltag) and [apriltag_ros](https://github.com/AprilRobotics/apriltag_ros).
Robotics community are already using it, e.g. [here](https://discuss.ardupilot.org/t/indoor-non-gps-flight-using-apriltags-ros-based/42878), [here](https://github.com/IntelRealSense/librealsense/pull/4340) or [here](https://packages.debian.org/source/sid/apriltag).

Personally, I see the possible following improvements, features:
  - specific code to improve performance on ARM architecture
  - a mode to allow occluded tags detection
  - something similar to ArUco 3 fractal markers?

---

OpenCV ArUco with `CORNER_REFINE_APRILTAG` gives the same corner accuracy than AprilTag 2 but the whole pipeline still give terrible results compared to AprilTag 2. Click on the image for the full video:

[output](https://youtu.be/8G1d65kWVWY)

Code is [here](https://gist.github.com/catree/dd901325a6bb8cc5d8bce12eba70a794), one image [here](https://user-images.githubusercontent.com/8229425/60726568-2c347400-9f3c-11e9-9644-6ac6f096931a.png) and AprilTag 2016-12-01 can be downloaded [here](https://april.eecs.umich.edu/media/apriltag/apriltag-2016-12-01.tgz).

paroj:
@szk1509 take a look

PR #2152, opened 2019-06-16
issue-2060: implement ability to return scores for edgeboxes implemen…

saskatchewancatch:
…tation

### DIFF
--- a/modules/ximgproc/src/edgeboxes.cpp
+++ b/modules/ximgproc/src/edgeboxes.cpp
@@ -48,7 +48,6 @@ OpenCV port by: Leonardo Lontra <lhe dot lontra at gmail dot com>
 */
 
 #include "precomp.hpp"
-
 using namespace cv;
 using namespace std;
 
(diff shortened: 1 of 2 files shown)


modules/ximgproc/src/edgeboxes.cpp
```diff
@@ -910,13 +910,14 @@ void EdgeBoxesImpl::boxesNms(Boxes &boxes, float thr, float eta, int maxBoxes)
...
 
-void EdgeBoxesImpl::getBoundingBoxes(InputArray edge_map, InputArray orientation_map, std::vector<Rect> &boxes)
+void EdgeBoxesImpl::getBoundingBoxes(InputArray edge_map, InputArray orientation_map, std::vector<Rect> &boxes, OutputArray scores)
 {
     CV_Assert(edge_map.depth() == CV_32F);
     CV_Assert(orientation_map.depth() == CV_32F);
 
     Mat E = edge_map.getMat().t();
     Mat O = orientation_map.getMat().t();
+    std::vector<float> * _scores = (std::vector<float> *) scores.getObj();;
```
alalek: Don't use `.getObj()`:
- without OutputArray type checks. It is just unsafe.
- with external interfaces. This doesn't work with Python/java bindings (`CV_WRAP` forces bindings generation).

Use `.getMat()` and check that it `CV_32FC1` or use `Mat(your_vector).copyTo(scores)` instead.

saskatchewancatch: @alalek I took your second suggestion to use `Mat(_scores).copyTo(scores)` instead. Onus is on caller to ensure that vector of floats is passed in. Autogenerated doc also makes it clear that vector of floats is expected to be passed in if scores are desired. Tested locally on my machine.

PR #2160, opened 2019-06-20
ovis: add methods (+ sample) for setting entity animations

dnandha:
### This pullrequest changes
__ovis__ currently lacks the ability to animate entities.
This PR adds two new methods to __ovis__: `setEntityAnimation` to set animation states for a loaded mesh/entity and `updateEntityAnimation` to update the time delta for the animation state in order to progress the animation.

Basic guides on animation setup in Ogre can be found both found [here](https://wiki.ogre3d.org/Sinbad+Model#Usage) and [here](https://wiki.ogre3d.org/Intermediate+Tutorial+1#Animation). The implemented methods are based on the information presented in those guides.

### Tested on
OS: `5.1.7-arch1-1-ARCH`
Platform: `x86_64 GNU/Linux`
Compiler: `g++ (GCC) 8.3.0`
OpenCV: `4.1.0`


modules/ovis/include/opencv2/ovis.hpp
```diff
@@ -164,6 +164,25 @@ class CV_EXPORTS_W WindowScene {
...
+
+    /**
+     * set entity animation state
+     * @param name enitity name
+     * @param animname animation state name
+     * @param enable activate or deactivate animation
+     * @param loop activate or deactivate animation loop
+     */
+    CV_WRAP virtual void setEntityAnimation(const String& name, const String& animname,
+                                            bool enable = true, bool loop = true) = 0;
```
paroj: as a rule of thumb, if your function takes multiple boolean arguments, your API is flawed.

e.g. guess what:
```cpp
Ogre::SceneNode::_update(true, false)
```
will do?

modules/ovis/samples/ovis_anim_demo.cpp
```diff
@@ -0,0 +1,46 @@
...
+  ovis::addResourceLocation("packs/Sinbad.zip"); // shipped with Ogre
+
+  Ptr<ovis::WindowScene> win = ovis::createWindow(String("ovis"), imsize, ovis::SCENE_INTERACTIVE | ovis::SCENE_AA);
+  win->createEntity("sinbad", "Sinbad.mesh", Vec3i(0, 0, 0), Vec3f(0, 0, 0));
+  win->setCameraPose(Vec3f(0, 0, 20));
+  win->setCameraLookAt("sinbad");
+  win->createLightEntity("sun", Vec3i(0, 0, 100));
+
+  // play idle animation
+  win->setEntityAnimation("sinbad", "IdleBase");
```
paroj: instead of creating a whole new sample, copying the existing one, just add the animation to the ovis_demo.{py, cpp}

modules/ovis/include/opencv2/ovis.hpp
```diff
@@ -164,6 +164,25 @@ class CV_EXPORTS_W WindowScene {
     CV_WRAP virtual void setEntityPose(const String& name, InputArray tvec = noArray(),
                                        InputArray rot = noArray(), bool invert = false) = 0;
 
+    /**
+     * update entity animation state
+     * @param name enitity name
+     * @param animname animation state name
+     * @param offset amount of time, in seconds, to extend the animation
+     */
+    CV_WRAP virtual void updateEntityAnimation(const String& name, const String& animname,
```
paroj: this API as too low-level as it forces the user to use a timer to actually animate something. Instead automatically connect the animatonstate to the default ogre timer when the animation is enabled.
c.f.
https://github.com/OGRECave/ogre-meshviewer/blob/19.01/ogre_mesh_viewer.py#L229-L231

modules/ovis/include/opencv2/ovis.hpp
```diff
@@ -164,6 +164,25 @@ class CV_EXPORTS_W WindowScene {
...
+                                               double offset) = 0;
+
+    /**
+     * set entity animation state
+     * @param name enitity name
+     * @param animname animation state name
+     * @param enable activate or deactivate animation
+     * @param loop activate or deactivate animation loop
+     */
+    CV_WRAP virtual void setEntityAnimation(const String& name, const String& animname,
```
paroj: change this to `void enableEnitityAnimation(str name, str animname, bool)` implementing the behavior outlined above.
Move the loop flag to the `cv::ovis::EntityProperty` enum as `ENTITY_ANIMATION_LOOP`

dnandha: Thanks, but this one I couldn't agree on: An entity can carry multiple animations and you want to be able to tell which one of them should be played in a loop and which one not. Having a property on the entity would restrict us to a global behavior.

Instead to changing this to `enableEntityAnimation` I changed it to `playEntityAnimation`, which by default plays / enables the animation state. The only flag left to set is whether to loop the animation or not.

Another method `stopEntityAnimation` has been added to stop the animation playback and clean up, similar to the reset function in the _ogre mesh viewer_.

modules/ovis/include/opencv2/ovis.hpp
dnandha: Fully agreed and changed. Animation updates are now controlled by the frame controller.
To prevent multiple frame controllers for the same animation state to be created, and to allow us to clean up frame controllers, a new variable has been created (an unordered map bookkeeping over all created frame controllers).

modules/ovis/samples/ovis_anim_demo.cpp
dnandha: I'd also like to have just one example. However, I did want to showcase the animation playback and the updated demo allows to list and play all available animations on the _Sinbad_ mesh.

Please tell me if it still makes sense to merge both demos.

modules/ovis/include/opencv2/ovis.hpp
dnandha: Thanks! I've noticed this flaw while programming and could resolve it now thanks to your suggestions! (See comments below)

dnandha:
@paroj Thanks for your suggestions. Each one of them was very helpful!

To sum up the changes:
- The method `updateEntityAnimation` has been removed
- The method `setEntityAnimation` has been renamed to `playEntityAnimation` (with only one flag left to set) to start/enable the animation state and create a frame controller
- The method `stopEntityAnimation` has been added to stop/disable the animation state and clean up the corresponding frame controller
- The variable `frameCtrlrs` has been added to keep track of all created frame controllers
- The method `getEntityAnimations` has been added which returns a list (_String_) of all available animations on the mesh

modules/ovis/src/ovis.cpp
```diff
@@ -576,6 +578,57 @@ class WindowSceneImpl : public WindowScene
...
+    }
+
+    void playEntityAnimation(const String& name, const String& animname, bool loop = true) CV_OVERRIDE
+    {
+        SceneNode& node = _getSceneNode(sceneMgr, name);
+        Entity* ent = dynamic_cast<Entity*>(node.getAttachedObject(name));
+        CV_Assert(ent && "invalid entity");
+        AnimationState* animstate = ent->getAnimationState(animname);
+
+        if (ent->getSkeleton()->getBlendMode() != ANIMBLEND_CUMULATIVE)
```
paroj: I think we should respect what is specified in the .skelton file instead of randomly overwriting it.

dnandha: I see that. It's not completely random though, see [here](https://github.com/OGRECave/ogre/blob/6089c69af600a92288b077e738968a9850fe8279/Samples/Character/include/SinbadCharacterController.h#L192).
The blend mode defaults to 'AVERAGE' if not serialized in the skeleton file, as it's the case for the __Ogre__ supplied meshes (the animations look terrible if the mode isn't properly set).

I'm adding the new _EntityProperty_ __ENTITY_ANIMBLEND_MODE__ and a corresponding `setEntityProperty` function override so that the user is able to override/set the blend mode accordingly (successfully tested).

modules/ovis/include/opencv2/ovis.hpp
```diff
@@ -105,7 +106,10 @@ class CV_EXPORTS_W WindowScene {
     CV_WRAP virtual void setEntityProperty(const String& name, int prop, const Scalar& value) = 0;
 
     /// @overload
-    CV_WRAP virtual void setEntityProperty(const String& name, int prop, const String& value) = 0;
+    CV_WRAP_AS(setEntityPropertyScale) virtual void setEntityProperty(const String& name, int prop, const String& value) = 0;
```
dnandha: needs to be done for python, so that it knows which function to call

dnandha: needs to be done for python, so that it knows which function to call

modules/ovis/include/opencv2/ovis.hpp
```diff
@@ -105,7 +106,10 @@ class CV_EXPORTS_W WindowScene {
...
+
+    /// @overload
+    CV_WRAP_AS(setEntityPropertyAnim) virtual void setEntityProperty(const String& name, int prop, const int& value) = 0;
```
dnandha: needs to be done for python, so that it knows which function to call

dnandha: CV_WRAP_AS is needed for python, so that it knows which function to call

dnandha:
@paroj Latest push removes the additional sample and adds a simple idle animation to the existing __ovis_demo.cpp__ and __ovis_demo.py__.
Futher, the API now allows to set the animation blend mode via `setEntityProperty` (the hard-coded blend mode override has been removed).

modules/ovis/include/opencv2/ovis.hpp
paroj: this should be avoided as it will break existing python code.
actually the python bindings know how to differentiate `String` and `Scalar`.
Where things fail is `int` vs. `Scalar` as the latter can be implicitly constructed from `int`.

Just remove the `int` overload and use value[0] in the Scalar variant.

dnandha: done

dnandha:
@paroj `int` overload for `setEntityProperty` removed and consolidated into existing function.

paroj:
:+1:

PR #2165, opened 2019-06-25
Nvidia optical flow

vchiluka5:
**Main PR**: https://github.com/opencv/opencv/pull/14888

Adds source code for OpenCV-OF interface implementation, unit test and performance test.
This change implements a OpenCV interface for the NVIDIA Optical Flow SDK 1.0. The SDK exposes the latest hardware
capability of Turing GPUs dedicated to computing the relative motion of pixels between images.

The hardware uses sophisticated algorithms to yield highly accurate flow vectors, with robust frame-to-frame intensity variations and
tracks the true object motion faster and more accurately. The optical flow hardware accelerator generates block-based optical flow vectors.
The size of the block depends on hardware in use and can be queried using the function getGridSize(). The block-based flow vectors generated
by the hardware can be converted to dense representation (i.e. per-pixel flow vectors) using upSampler() helper function, if needed.
The flow vectors are stored in CV_16SC2 format with x and y components of each flow vector in 16-bit signed fixed point representation S10.5. 
This change also adds a unit test and a performance test for the new interface.

```
force_builders_only=Custom,docs
buildworker:Custom=linux-1,linux-2,linux-4
docker_image:Custom=ubuntu-cuda:16.04
```


.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "modules/cudaoptflow/include/opencv2/NVIDIAOpticalFlowSDK"]
+	path = modules/cudaoptflow/include/opencv2/NVIDIAOpticalFlowSDK
+	url = https://github.com/NVIDIA/NVIDIAOpticalFlowSDK.git
```
alalek: Git submodules are not allowed.

This significantly increases contribution barrier.

vchiluka5: Submodule is only for pulling NVOF SDK headers. Headers are tied to specific version of SDK and hence we thought getting them via submodule was the best option.
Can you please suggest any other alternatives? Headers don't follow opencv coding convention and SDK is already available publicly at NVIDIA devzone.

alalek: if these headers are under appropriate license (BSD-like), then you can copy them into this repository.

> Headers are tied to specific version of SDK

This may be a problem (even with submodule).
What is the process to deliver compatible headers to users?


---

One more alternative is download and unpack sources by CMake. [Example](https://github.com/opencv/opencv_contrib/blob/3.4.0/modules/dnn_modern/CMakeLists.txt#L16-L31)
Several sources can be applied depending on users configuration (CUDA SDK version).

---

Also we need CMake checks for supported CUDA versions only (and disable functionality for unsupported versions - throw exception).

vchiluka5: > if these headers are under appropriate license (BSD-like), then you can copy them into this repository.
> Headers are under  BSD-3 license but they don't follow opencv coding styles and we can't change that since SDK is already public
> > Headers are tied to specific version of SDK
> So right now SDK 1.0 is public and these headers are from SDK 1.0. Submodule is linked at particular branch(nvof_1_0). So in future for SDK 2.0, we will have new branch(nvof_2_0) with SDK 2.0 with new headers and we will update the submodule dependency here.
> This may be a problem (even with submodule).
> What is the process to deliver compatible headers to users?
> Headers will be as per SDK version. Now its SDK 1.0. Later with SDK 2.0 we will update new headers.
> One more alternative is download and unpack sources by CMake. [Example](https://github.com/opencv/opencv_contrib/blob/3.4.0/modules/dnn_modern/CMakeLists.txt#L16-L31)
> Several sources can be applied depending on users configuration (CUDA SDK version).
> 
> Also we need CMake checks for supported CUDA versions only (and disable functionality for unsupported versions - throw exception).
Mostly all CUDA versions are supported. Backward compatibility is maintained. Only requirement is Turing based GPU for HW acceleration and it its not present then appropriate error message is returned.

vchiluka5: > >  if these headers are under appropriate license (BSD-like), then you can copy them into this repository.
> Headers are under  BSD-3 license but they don't follow opencv coding styles and we can't change that since SDK is already public

vchiluka5: >>Headers are tied to specific version of SDK
> So right now SDK 1.0 is public and these headers are from SDK 1.0. Submodule is linked at particular branch(nvof_1_0). So in future for SDK 2.0, we will have new branch(nvof_2_0) with SDK 2.0 with new headers and we will update the submodule dependency here.
>>This may be a problem (even with submodule).
>>What is the process to deliver compatible headers to users?
> Headers will be as per SDK version. Now its SDK 1.0. Later with SDK 2.0 we will update new headers.
>>One more alternative is download and unpack sources by CMake. [Example](https://github.com/opencv/opencv_contrib/blob/3.4.0/modules/dnn_modern/CMakeLists.txt#L16-L31)
>> Several sources can be applied depending on users configuration (CUDA SDK version).
> We felt managing headers via submodule is the best way.
>>Also we need CMake checks for supported CUDA versions only (and disable functionality for unsupported versions - throw exception).
> Mostly all CUDA versions are supported. Backward compatibility is maintained. Only requirement is Turing based GPU for HW acceleration and it its not present then appropriate error message is returned.

alalek: You can put (copy) these headers "as is" (without modifications) - like other 3rdparty components sources.

> they don't follow opencv coding styles

This is not a problem, because these headers are internal (not exposed as OpenCV public API).

vchiluka5:
Is it OK if we discuss all necessary changes and then il make all changes at once? 
Il move sample app and merge in commit 1 of this PR once we finalize about how headers can be used

.gitmodules
vchiluka5: Sure. Thanks .Let me discuss this within my team. And il update it here as soon as possible.

vchiluka5: Hey @alalek 
I had a discussion with my team and we decided to go with downloading via cmake. Thanks for giving a sample. I updated CMakeLists.txt in same way to fetch our headers. I force pushed so as to avoid un-necessary CL's. I hope this is fine? Can we proceed with further review?

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
```diff
@@ -102,6 +105,47 @@ class CV_EXPORTS_W SparseOpticalFlow : public Algorithm
...
+    @param stream Stream for the asynchronous version.
+    @param hint Hint buffer if client provides external hints. Must have same size as flow buffer.
+                Caller can provide flow vectors as hints for optical flow calculation.
+    @param cost Cost buffer contains numbers indicating the confidence associated with each of the generated flow vectors.
+                Higher the cost, lower the confidence. Cost buffer is of type CV_32SC1.
+
+    @note
+    - Client must use critical sections around each calc() function if calling it from multiple threads.
+    */
+    CV_WRAP virtual void calc(
```
alalek: Please remove CV_WRAP here, or replace `CV_EXPORTS` => `CV_EXPORTS_W` on the line with class declaration.
This breaks the build (see [here](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_custom_linux64/builds/131)).

modules/cudaoptflow/CMakeLists.txt
```diff
@@ -6,4 +6,23 @@ set(the_description "CUDA-accelerated Optical Flow")
...
+# ----------------------------------------------------------------------------
+# MODULE REQUIREMENTS
+# ----------------------------------------------------------------------------
+
+set(NVIDIA_OPTICAL_FLOW_1_0_HEADERS_PATH "${CMAKE_CURRENT_SOURCE_DIR}/include/opencv2/NVIDIAOpticalFlowSDK_1_0_Headers")
+set(NVIDIA_OPTICAL_FLOW_1_0_HEADERS_ROOT "${NVIDIA_OPTICAL_FLOW_1_0_HEADERS_PATH}/NVIDIAOpticalFlowSDK_1_0_Headers")
+ocv_download(FILENAME "NVIDIAOpticalFlowSDK_1_0_Headers.tar.gz"
+               HASH "c9e06b032b9839e7106baf42cb1b429c"
+               URL
+                 "https://raw.githubusercontent.com/NVIDIA/NVIDIAOpticalFlowSDK/nvof_1_0_bsd/"
```
alalek: BTW, It is possible to download using specific commit hash (no need for archive in git repo):
https://github.com/NVIDIA/NVIDIAOpticalFlowSDK/archive/79c6cee80a2df9a196f20afd6b598a9810964c32.zip

(these files are usually stable, but may change if GitHub changes packing options - I remember only one case when archive's contents are changed (due compression parameters)).

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
```diff
@@ -43,6 +43,9 @@
 #ifndef OPENCV_CUDAOPTFLOW_HPP
 #define OPENCV_CUDAOPTFLOW_HPP
 
+#include "NVIDIAOpticalFlowSDK_1_0_Headers/nvOpticalFlowCommon.h"
+#include "NVIDIAOpticalFlowSDK_1_0_Headers/nvOpticalFlowCuda.h"
```
alalek: Do we really need that headers as a part of public OpenCV API?

Usually we hide all implementation details. In general, inclusion of OpenCV header should not affect names outside of "cv::" namespace or `CV_` macros/defines
For enums we "duplicate" in corresponding classes (at least such duplication is needed for bindings generator).

modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
```diff
@@ -0,0 +1,548 @@
...
+        }                                                                           \
+    } while (0)
+
+using namespace std;
+using namespace cv;
+using namespace cv::cuda;
+
+namespace
+{
+    class NvidiaOpticalFlowImpl : public cv::cuda::NvidiaOpticalFlow_1_0
```
alalek: Please remove indentation in namespaces.

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
```diff
@@ -342,6 +386,56 @@ class CV_EXPORTS_W OpticalFlowDual_TVL1 : public DenseOpticalFlow
...
+    @param perfPreset Optional parameter. Refer [NV OF SDK documentation](https://developer.nvidia.com/opticalflow-sdk) for details about presets.
+                      Defaults to NV_OF_PERF_LEVEL_SLOW.
+    @param enableTemporalHints Optional parameter. Flag to enable temporal hints. When set to true, the hardware uses the flow vectors
+                               generated in previous call to calc() as internal hints for the current call to calc().
+                               Useful when computing flow vectors between successive video frames. Defaults to false.
+    @param enableExternalHints Optional Parameter. Flag to enable passing external hints buffer to calc(). Defaults to false.
+    @param enableCostBuffer Optional Parameter. Flag to enable cost buffer output from calc(). Defaults to false.
+    @param gpuId Optional parameter to select the GPU ID on which the optical flow should be computed. Useful in multi-GPU systems. Defaults to 0.
+    */
+    CV_WRAP static Ptr<NvidiaOpticalFlow_1_0> create(
```
alalek: will this call raise exception or return empty pointer if specified `gpuId` H/W doesn't support this feature? (please clarify this case in documentation)

(also we probably need some check for that in tests)

modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
```diff
@@ -0,0 +1,548 @@
...
+//  By downloading, copying, installing or using the software you agree to this license.
+//  If you do not agree to this license, do not download, install,
+//  copy or use the software.
+//
+//
+//                           License Agreement
+//                For Open Source Computer Vision Library
+//
+// Copyright (C) 2000-2008, Intel Corporation, all rights reserved.
+// Copyright (C) 2009, Willow Garage Inc., all rights reserved.
```
alalek: Please use [short license header](https://github.com/opencv/opencv/wiki/Coding_Style_Guide#file-structure) in new files:
```
//
// ... optional file's author/copyright lines ...
```

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
vchiluka5: Done.

modules/cudaoptflow/CMakeLists.txt
vchiluka5: Oh Thanks. I was not aware about it.
Il make it download via hash.

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
vchiluka5: We will need those headers since all the enums and function pointers are declared there and every release of new SDK will have some parameters/enums/functions added. So yes, headers are must.

modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
vchiluka5: Sure. Noted

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
vchiluka5: So if specific GPU is not compatible then HMODULE hModule = LoadLibrary(MODULENAME); wont be able to get the library file or dll and it will return appropriate error..

modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
vchiluka5: Noted

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
alalek: You need headers, but exposing them as public OpenCV API is not necessary.

If you add "duplicate" of enum `NV_OF_PERF_LEVEL` here (renamed, but keep the same values, so just cast is enough), then you can use headers from .cpp files only (move headers into `src/` subtree) and don't expose them for users.

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
alalek: There is `gpuId`.
2 gpus in system. They are different.

Is it clear now why approach with `LoadLibrary` will not work?

vchiluka5: I got your point. In that case nvCreateOpticalFlowCuda() will fail and it will return exception. It wont create empty pointer. Il try out this situation and add appropriate checks in tests.

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
vchiluka5: True. Probably we can do it but again once i may need to discuss it within my team. Il move the headers and update the CL  once confirmed.

modules/cudaoptflow/CMakeLists.txt
vchiluka5: Any suggestions on this @alalek

alalek: > d41d8cd98f00b204e9800998ecf8427e

Why did you using MD5 of empty file?
Take a look on CMake logs, error message is there.

It should be:
```
set(NVIDIA_OPTICAL_FLOW_1_0_HEADERS_MD5 "ca5acedee6cb45d0ec610a6732de5c15")
```

BTW, files are unpacked with extra sub-directory:
```
include/opencv2/NVIDIAOpticalFlowSDK_1_0_Headers/
└── NVIDIAOpticalFlowSDK-79c6cee80a2df9a196f20afd6b598a9810964c32
    ├── nvOpticalFlowCommon.h
    ├── nvOpticalFlowCuda.h
    └── README.md
```

So, we need:
```
ocv_include_directories(SYSTEM "${NVIDIA_OPTICAL_FLOW_1_0_HEADERS_PATH}/NVIDIAOpticalFlowSDK-${NVIDIA_OPTICAL_FLOW_1_0_HEADERS_MD5}")
```

and remove these headers from public API location:

```
set(NVIDIA_OPTICAL_FLOW_1_0_HEADERS_PATH "${CMAKE_CURRENT_BINARY_DIR}/include/NVIDIAOpticalFlowSDK_1_0_Headers")
```

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
vchiluka5: I added try catch block so that test suite wont crash and stop. So if specified HW is not supported test will fail and return silently. Will this be Okay?

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
vchiluka5: I moved headers to src and duplicated NV_OF_PERF_LEVEL as NVIDIA_OF_PERF_LEVEL

modules/cudaoptflow/CMakeLists.txt
vchiluka5: Done

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
alalek: Silent errors is a bad practice.
Exceptions is more useful for users (at least they have error message, and sometimes stack trace).

Looks like, there is no way to detect support during `create()` function.
Checking API support is still actual for tests (to skip test on unsupported H/W instead of failures).

modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
```diff
@@ -0,0 +1,581 @@
+//
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+//M*/
+#include "precomp.hpp"
+#include "NVIDIAOpticalFlowSDK_1_0_Headers/NVIDIAOpticalFlowSDK-79c6cee80a2df9a196f20afd6b598a9810964c32/nvOpticalFlowCommon.h"
+#include "NVIDIAOpticalFlowSDK_1_0_Headers/NVIDIAOpticalFlowSDK-79c6cee80a2df9a196f20afd6b598a9810964c32/nvOpticalFlowCuda.h"
```
alalek: With `ocv_include_directories()` it is enough to include:
```
#include "nvOpticalFlowCommon.h"
#include "nvOpticalFlowCuda.h"
```

vchiluka5: I thought the same and tried including like this
#include "nvOpticalFlowCommon.h"
#include "nvOpticalFlowCuda.h"
But compilation failed and hence i added them this way

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
vchiluka5: I meant it wont crash abruptly now rather it will give the exception and exit.
OpenCV uses gtest framework right? So cant we put restriction that particular tests should run only on turing and above gpu archs?

alalek: OpenCV uses patched Google Tests framework.

You can use this code to "skip" test.
```
if (<not_supported_configuration_condition>)
    throw SkipTestException("Current configuration is not supported");
```
Main question here is "condition" implementation.

---

Alternative (exception variant):
```
cv::Ptr<cv::cuda::NvidiaOpticalFlow_1_0> d_nvof;
try
{
    d_nvof =cv::cuda::NvidiaOpticalFlow_1_0::create(...);
}
catch (... not supported exception ...)
{
    throw SkipTestException("Current configuration is not supported");
}
```

modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
alalek: Include directory is not added properly.

Use this:
```
ocv_define_module(cudaoptflow opencv_video opencv_optflow opencv_cudaarithm opencv_cudawarping opencv_cudaimgproc OPTIONAL opencv_cudalegacy WRAP python)

set(NVIDIA_OPTICAL_FLOW_1_0_HEADERS_COMMIT "79c6cee80a2df9a196f20afd6b598a9810964c32")
set(NVIDIA_OPTICAL_FLOW_1_0_HEADERS_MD5 "ca5acedee6cb45d0ec610a6732de5c15")
set(NVIDIA_OPTICAL_FLOW_1_0_HEADERS_PATH "${OpenCV_BINARY_DIR}/3rdparty/NVIDIAOpticalFlowSDK_1_0_Headers")
ocv_download(FILENAME "${NVIDIA_OPTICAL_FLOW_1_0_HEADERS_COMMIT}.zip"
               HASH ${NVIDIA_OPTICAL_FLOW_1_0_HEADERS_MD5}
               URL
                 "https://github.com/NVIDIA/NVIDIAOpticalFlowSDK/archive/"
               DESTINATION_DIR "${NVIDIA_OPTICAL_FLOW_1_0_HEADERS_PATH}"
               STATUS NVIDIA_OPTICAL_FLOW_1_0_HEADERS_DOWNLOAD_SUCCESS
               ID "NVIDIA_OPTICAL_FLOW"
               UNPACK RELATIVE_URL)

if(NOT NVIDIA_OPTICAL_FLOW_1_0_HEADERS_DOWNLOAD_SUCCESS)
  message(STATUS "Failed to download NVIDIA_Optical_Flow_1_0 Headers")
else()
  add_definitions(-DHAVE_NVIDIA_OPTFLOW=1)
  ocv_include_directories(SYSTEM "${NVIDIA_OPTICAL_FLOW_1_0_HEADERS_PATH}/NVIDIAOpticalFlowSDK-${NVIDIA_OPTICAL_FLOW_1_0_HEADERS_COMMIT}")
endif()
```

alalek: Also take a look on `add_definitions(-DHAVE_NVIDIA_OPTFLOW=1)` line.
You should properly guard implementation code (if downloading is failed/disabled/not available).

vchiluka5: Thanks alot :-)
I bounded implementation code as if headers are not downloaded then return HeaderIs Null error

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
vchiluka5: I seperated error code for each specific error to define that condition. We mapped opencv error to as much similar error can be to nvof error.

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
```diff
@@ -342,6 +383,69 @@ class CV_EXPORTS_W OpticalFlowDual_TVL1 : public DenseOpticalFlow
...
+- A sample application demonstrating the use of NVIDIA Optical Flow can be found at
+opencv_source_code/samples/gpu/nvidia_optical_flow.cpp
+- An example application comparing accuracy and performance of NVIDIA Optical Flow with other optical flow algorithms in OpenCV can be found at
+opencv_source_code/samples/gpu/optical_flow.cpp
+*/
+
+/**
+* Supported optical flow/stereo disparity performance levels.
+*/
+typedef enum _NVIDIA_OF_PERF_LEVEL
```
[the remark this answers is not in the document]
vchiluka5: NVIDIA_OF_PERF_LEVEL enum is not needed in NvidiaOpticalFlow_1_0 right? since it is casted to NV_OF_PERF_LEVEL

modules/cudaoptflow/perf/perf_optflow.cpp
```diff
@@ -326,4 +326,56 @@ PERF_TEST_P(ImagePair, OpticalFlowDual_TVL1,
...
+        cv::cuda::GpuMat d_flow;
+        cv::Ptr<cv::cuda::NvidiaOpticalFlow_1_0> d_nvof;
+        try
+        {
+            d_nvof = cv::cuda::NvidiaOpticalFlow_1_0::create(width, height, preset,
+                enableTemporalHints, enableExternalHints, enableCostBuffer, gpuid);
+        }
+        catch (Error::Code &e)
+        {
+            if(e == Error::StsBadFunc || e == Error::StsBadArg || e == Error::StsNullPtr)
```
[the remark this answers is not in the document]
vchiluka5: Noted

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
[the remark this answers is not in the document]
vchiluka5: Done.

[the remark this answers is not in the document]
vchiluka5: If i move it inside class then everywhere i need to refer it like this cuda::NvidiaOpticalFlow_1_0::NVIDIA_OF_PERF_LEVEL otherwise it gives build errors.

vchiluka5:
Hi @alalek 
I see this error in some python bindings. What is this about? Can you please check and also suggest possible solutions to this?

  File "/build/precommit-contrib_custom_linux64/opencv/modules/python/bindings/..//src2/gen2.py", line 1102, in <module>
    generator.gen(srcfiles, dstdir)
  File "/build/precommit-contrib_custom_linux64/opencv/modules/python/bindings/..//src2/gen2.py", line 991, in gen
    self.add_func(decl)
  File "/build/precommit-contrib_custom_linux64/opencv/modules/python/bindings/..//src2/gen2.py", line 909, in add_func
    func_map = self.classes[classname].methods
KeyError: 'cuda_NvidiaHWOpticalFlow'

alalek:
All base classes must be wrapped into bindings too:
```.diff
-class CV_EXPORTS NvidiaHWOpticalFlow 
+class CV_EXPORTS_W NvidiaHWOpticalFlow 
```

vchiluka5:
@alalek  Thanks for reviewing till now,
Updated with all above discussed changes. Please Let me know if anymore changes are needed. Il try to complete all changes asap.

vchiluka5:
Hi @alalek 

This error here is due to CUDA_CUDA_LIBRARY not linked to opencv_world module for which i submitted a pull request in opencv main branch
https://pullrequest.opencv.org/buildbot/builders/precommit-contrib_custom_linux64/builds/138/steps/compile%20release/logs/stdio

alalek:
This build tests both patches together (both patches are applied).
So something in these patches is wrong and should be get fixed/removed.

vchiluka5:
> This build tests both patches together (both patches are applied).
> So something in these patches is wrong and should be get fixed/removed.

Looks like its not picking properly from opencv main branch. NVIDIA_Optical_Flow branch is not used here:
https://pullrequest.opencv.org/buildbot/builders/precommit-contrib_custom_linux64/builds/140

alalek:
[Here](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_custom_linux64/builds/141/steps/Merge%20opencv%20with%20test%20branch/logs/stdio):
```
From https://github.com/vchiluka5/opencv
 * branch            NVIDIA_Optical_Flow -> FETCH_HEAD
Removing samples/gpu/optical_flow.cpp
Merge made by the 'recursive' strategy.
 CMakeLists.txt               |   2 +-
 samples/gpu/optical_flow.cpp | 240 -------------------------------------------
 2 files changed, 1 insertion(+), 241 deletions(-)
 delete mode 100644 samples/gpu/optical_flow.cpp
```

vchiluka5:
> [Here](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_custom_linux64/builds/141/steps/Merge%20opencv%20with%20test%20branch/logs/stdio):
> 
> ```
> From https://github.com/vchiluka5/opencv
>  * branch            NVIDIA_Optical_Flow -> FETCH_HEAD
> Removing samples/gpu/optical_flow.cpp
> Merge made by the 'recursive' strategy.
>  CMakeLists.txt               |   2 +-
>  samples/gpu/optical_flow.cpp | 240 -------------------------------------------
>  2 files changed, 1 insertion(+), 241 deletions(-)
>  delete mode 100644 samples/gpu/optical_flow.cpp
> ```

Hi @alalek 
I re-built everything. Steps i did are 
git clone -b NVIDIA_Optical_Flow  https://github.com/vchiluka5/opencv.git
git clone -b NVIDIA_Optical_Flow  https://github.com/vchiluka5/opencv_contrib.git
mkdir build && mkdir install
cd build
cmake -D WITH_CUDA=ON -D BUILD_EXAMPLES=ON -D BUILD_TESTS=ON -D BUILD_PERF_TESTS=ON -D CMAKE_BUILD_TYPE=RELEASE -D CMAKE_INSTALL_PREFIX=/home/kp/Desktop/opencv_master/install -D OPENCV_EXTRA_MODULES_PATH=/home/kp/Desktop/opencv_master/opencv_contrib/modules  /home/kp/Desktop/opencv_master/opencv/
make all -j 40 && make install

and i was able to build opencv without any errors. 
Error says warning: libcuda.so.1, needed by ../../lib/libopencv_cudaoptflow.so.4.1.1, not found
and hence unresolved references to driver APIs

Is CUDA installed on that machine where the compilation is going on?

alalek:
Build is going in Linux docker container (without any CUDA device, just build, no tests).

Dockerfile with CUDA related installations is [here](https://github.com/opencv-infrastructure/opencv-worker-config/blob/linux/images/ubuntu-cuda--16.04/docker/Dockerfile).

alalek:
These files are available from container:
```
# find / -name libcuda*
/usr/local/cuda-8.0/targets/x86_64-linux/lib/libcudart.so.8.0
/usr/local/cuda-8.0/targets/x86_64-linux/lib/libcudart.so.8.0.61
/usr/local/cuda-8.0/targets/x86_64-linux/lib/libcudart.so
/usr/local/cuda-8.0/targets/x86_64-linux/lib/libcudadevrt.a
/usr/local/cuda-8.0/targets/x86_64-linux/lib/stubs/libcuda.so
/usr/local/cuda-8.0/targets/x86_64-linux/lib/libcudart_static.a
/usr/local/cuda-8.0/extras/Debugger/include/libcudacore.h
/usr/local/cuda-8.0/extras/Debugger/lib64/libcudacore.a
```

So only "stubs" are available in build container.

[Here](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_custom_linux64/builds/141/steps/cmake/logs/vars):
```
CUDA_CUDA_LIBRARY=/usr/local/cuda/lib64/stubs/libcuda.so
```

It seems correct. Need to check what linker does wrong.

alalek:
Related issue: https://github.com/NVIDIA/nvidia-docker/issues/775

vchiluka5:
> **Update**: Looks like stub file is malformed and can't be used with shared libraries.
> I re-links on non-existed `libcuda.so.1` (from `libcuda.so`)

Does re-linking solved the build issue? I still see failure with same error.

alalek:
There is typo in my previous comment (`I` -> `it`).

I tried building with extra symbolic link (from [here](https://github.com/NVIDIA/nvidia-docker/issues/775#issuecomment-400178344)), it seems able to build.

However it is not robust solution in general (due patching content in CUDA SDK directories).

---

Changes are not added into CI infrastructure yet.

vchiluka5:
> There is typo in my previous comment (`I` -> `it`).
> 
> I tried building with extra symbolic link (from [here](https://github.com/NVIDIA/nvidia-docker/issues/775#issuecomment-400178344)), it seems able to build.
> 
> However it is not robust solution in general (due patching content in CUDA SDK directories).
> 
> Changes are not added into CI infrastructure yet.

What should be our next step then? Since you were able to build that indicates everything is fine in this patch right?

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
```diff
@@ -342,6 +383,69 @@ class CV_EXPORTS_W OpticalFlowDual_TVL1 : public DenseOpticalFlow
...
+    @param perfPreset Optional parameter. Refer [NV OF SDK documentation](https://developer.nvidia.com/opticalflow-sdk) for details about presets.
+                      Defaults to NV_OF_PERF_LEVEL_SLOW.
+    @param enableTemporalHints Optional parameter. Flag to enable temporal hints. When set to true, the hardware uses the flow vectors
+                               generated in previous call to calc() as internal hints for the current call to calc().
+                               Useful when computing flow vectors between successive video frames. Defaults to false.
+    @param enableExternalHints Optional Parameter. Flag to enable passing external hints buffer to calc(). Defaults to false.
+    @param enableCostBuffer Optional Parameter. Flag to enable cost buffer output from calc(). Defaults to false.
+    @param gpuId Optional parameter to select the GPU ID on which the optical flow should be computed. Useful in multi-GPU systems. Defaults to 0.
+    */
+    CV_WRAP static Ptr<NvidiaOpticalFlow_1_0> create(
```
alalek: Lets remove `CV_WRAP` here to workaround bindings issue (look like there is bug in bindings generator).
This would disable Python bindings for this functionality (not critical for now - can be enabled later with corresponding Python tests)

vchiluka5: Sure. I got python bindings error repro locally and i was able to fix it. Hopefully it compiles fine now.
Thanks @alalek  for all suggestions.. Appreciated :-)
And your test commit works so will you merge it here in opencv and then il update my PR or how should we integrate?

vchiluka5:
> Looks good to me! Thank you 👍

Thanks alot @alalek

PR #2171, opened 2019-06-28
Added LookAt-methods and a GetPose-method for entities 

AlexJ95:
**This pullrequest changes**
This PR adds 2 overloaded methods setEntityLookAt, which add the option to make an entity look at another one or at a given point in world coordinates.

Also it adds a GetEntityPose method, which just was missing. It gives you the position and rotation of an entity.

Tested on:

OpenCV => 4.1.0
Operating System / Platform => Windows 10 64 Bit
Compiler => Visual Studio 2017

### DIFF
--- a/modules/ovis/include/opencv2/ovis.hpp
+++ b/modules/ovis/include/opencv2/ovis.hpp
@@ -165,6 +165,16 @@ public:
     CV_WRAP virtual void setEntityPose(const String& name, InputArray tvec = noArray(),
                                        InputArray rot = noArray(), bool invert = false) = 0;
 
+	/**
+     * Retrieves the current pose of an entity
+	 * @param name entity name
+     * @param R 3x3 rotation matrix
+     * @param tvec translation vector
+     * @param invert return the inverted pose
+     */
+	CV_WRAP virtual void getEntityPose(const String& name, OutputArray R = noArray(), OutputArray tvec = noArray(),
+                                       bool invert = false) = 0;
+
     /**
      * get a list of available entity animations
      * @param name entity name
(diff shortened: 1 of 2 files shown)


modules/ovis/include/opencv2/ovis.hpp
```diff
@@ -165,6 +165,15 @@ class CV_EXPORTS_W WindowScene {
...
                                        InputArray rot = noArray(), bool invert = false) = 0;
 
+	/**
+     * Retrieves the current pose of an entity
+	 * @param name entity name
+     * @param R 3x3 rotation matrix
+     * @param tvec translation vector
+     * @param invert return the inverted pose
+     */
+	CV_WRAP virtual void getEntityPose(const String& name, OutputArray R, OutputArray tvec, bool invert) = 0;
```
paroj: add default arguments as in `getCameraPose`. This is useful with python.

modules/ovis/include/opencv2/ovis.hpp
```diff
@@ -236,6 +245,20 @@ class CV_EXPORTS_W WindowScene {
      */
     CV_WRAP virtual void setCameraLookAt(const String& target, InputArray offset = noArray()) = 0;
 
+	/**
+     * convenience method to orient an entity to a point in world coordinates
+	 * @param origin entity name
+     * @param target targetpoint vector
+     */
+	CV_WRAP virtual void setEntityLookAt(const String& origin, InputArray target) = 0;
```
paroj: combine both functions into one as in `setCameraLookAt` to allow an lookAt offset. Conversely you can implement lookat for an arbitrary destination if `target == ""`.

modules/ovis/src/ovis.cpp
```diff
@@ -847,6 +877,26 @@ class WindowSceneImpl : public WindowScene
 
         camNode->lookAt(tgt->_getDerivedPosition() + _offset, Ogre::Node::TS_WORLD);
     }
+
+	void setEntityLookAt(const String& origin, InputArray target) CV_OVERRIDE
+    {
+
+		SceneNode* or = sceneMgr->getEntity(origin)->getParentSceneNode();
```
paroj: as the diff suggests or is a bad variable name

paroj:
@alalek :+1:

PR #2175, opened 2019-06-30
ovis, fixes: windows build, play multiple animations

### LINKED ISSUE #2177: Casting Error in ovis
Bug

Casting-Error 

Ocurred on:
- OpenCV => 4.1.0
- Operating System / Platform => Windows 10 64 Bit

dnandha:
Fixes #2177 

##### This pullrequest fixes two issues
1. Animations only played (/stopped) for one entity instance
- Fix: Save each animation **state** instead of the animation name
2. Build failure on the Windows platform because MSVC can't handle a static type conversion (#2177)
- Fix: Add explicit cast to *int* before the static type cast

@paroj


paroj:
:+1:

PR #2176, opened 2019-06-30
ovis: add functionality to attach entities to bones

dnandha:
### This pullrequest changes
Adds functionality to the *ovis* module which allows creating and removing entities attached to other entities, for example items held by a character mesh.

The PR adds three new methods:
1. `createEntityAttached` create entity attached to the bone of another entity
2. `removeEntityAttached` remove entity previously attached to another entity
3. `getBoneList` get list of all bones available on entity mesh

The existing *ovis_demo* samples have been updated to demonstrate the added functionality.

@paroj


paroj:
There is no reason why attaching should be limited to bones. I would rather introduce a general concept of attaching/ detaching entities with optional bone support.
* an detached entity would have its own transform (aka SceneNode)
* an attached entity would share the transform with its parent (or inherit the bone transform)

all entites would then be created detached and there would be
1. `attachEntity(str entity, str parent, str bone = "")`
2. `detachEntity(str entity)` (an entity can only ever have one parent)

the `removeEntity`, `setEntityPose` etc. methods will have to be updated to handle the introduced complexity.

asmorkalov:
@dnandha Please rebase the topic branch and resolve conflict.

PR #2179, opened 2019-07-01
de-hardcode some important parameters of TrackerMOSSE

amorvincitomnia53:
### This pullrequest changes

* Added cv::TrackerMOSSE::Params struct containing some important parameters for MOSSE tracker that was previously hard-coded (such as learning rate, detection threshold, etc.)

### DIFF
--- a/modules/tracking/include/opencv2/tracking/tracker.hpp
+++ b/modules/tracking/include/opencv2/tracking/tracker.hpp
@@ -42,11 +42,11 @@
 #ifndef __OPENCV_TRACKER_HPP__
 #define __OPENCV_TRACKER_HPP__
 
+#include "feature.hpp"
+#include "onlineBoosting.hpp"
+#include "onlineMIL.hpp"
 #include "opencv2/core.hpp"
 #include "opencv2/imgproc/types_c.h"
-#include "feature.hpp"
-#include "onlineMIL.hpp"
-#include "onlineBoosting.hpp"
 
 /*
  * Partially based on:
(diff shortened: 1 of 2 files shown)


modules/tracking/include/opencv2/tracking/tracker.hpp
```diff
@@ -1312,12 +1267,35 @@ The implementation is based on @cite MOSSE Visual Object Tracking using Adaptive
...
+        double eps;            // for normalization
+        double rate;           // learning rate
+        double psr_threshold;  // no detection, if PSR is smaller than this
+    };
+    /** @brief Constructor
   */
-  CV_WRAP static Ptr<TrackerMOSSE> create();
+    CV_WRAP static Ptr<TrackerMOSSE> create();
+
+    static Ptr<TrackerMOSSE> create(const TrackerMOSSE::Params& params);
```
alalek: `Params` are not accessible from Python, so probably we need one more `CV_WRAP` overload with "unpacked" parameters (which creates `Params` internally).

modules/tracking/src/mosseTracker.cpp
```diff
@@ -15,35 +15,30 @@
...
-     virtual void modelUpdateImpl() CV_OVERRIDE {}
-     virtual void modelEstimationImpl( const std::vector<Mat>& ) CV_OVERRIDE {}
-};
-
 
-const double eps=0.00001;      // for normalization
-const double rate=0.2;         // learning rate
-const double psrThreshold=5.7; // no detection, if PSR is smaller than this
+namespace tracking
+{
```
alalek: Revert back unnecessary changes.

PR #2180, opened 2019-07-02
OpenCV fix 14774 breaks cudacodec

cudawarped:
resolves https://github.com/opencv/opencv/issues/14850

In https://github.com/opencv/opencv/pull/14774 `cap_ffmpeg_api.hpp` was replaced with `cap_ffmpeg_legacy_api.hpp` and all of the cuda encoder decoder parts were removed from  `cap_ffmpeg_legacy_api.hpp`, e.g.

<cut/>

```
/*
 * For CUDA decoder
 */

OPENCV_FFMPEG_API struct InputMediaStream_FFMPEG* create_InputMediaStream_FFMPEG(const char* fileName, int* codec, int* chroma_format, int* width, int* height);
OPENCV_FFMPEG_API void release_InputMediaStream_FFMPEG(struct InputMediaStream_FFMPEG* stream);
OPENCV_FFMPEG_API int read_InputMediaStream_FFMPEG(struct InputMediaStream_FFMPEG* stream, unsigned char** data, int* size, int* endOfFile);

typedef struct InputMediaStream_FFMPEG* (*Create_InputMediaStream_FFMPEG_Plugin)(const char* fileName, int* codec, int* chroma_format, int* width, int* height);
typedef void (*Release_InputMediaStream_FFMPEG_Plugin)(struct InputMediaStream_FFMPEG* stream);
typedef int (*Read_InputMediaStream_FFMPEG_Plugin)(struct InputMediaStream_FFMPEG* stream, unsigned char** data, int* size, int* endOfFile);
```
additionaly the CUDA encoder/decoder implementation was removed from `cap_ffmpeg_impl.hpp`
meaning that simply updating the include in [precomp.hpp](https://github.com/opencv/opencv_contrib/blob/297e8be9fc747ee5a8516ad237f3eb5b46373a05/modules/cudacodec/src/precomp.hpp#L84) fails when compiling cudacodec.

I don't have a deep understanding of the intention of the changes in #14774 or the recommended way that all the pieces of OpenCV fit together so I have submitted this hack which I used to get things working on my system as a PR to further explain the problem, and get some guidance as to the correct solution.

```
opencv=ffmpeg_raw_retrieve
force_builders=Custom
buildworker:Custom=linux-4
docker_image:Custom=ubuntu-cuda:16.04
```


modules/cudacodec/src/cuvid_ffmpeg_api.hpp
```diff
@@ -0,0 +1,49 @@
+#ifndef __OPENCV_CUVID_FFMPEG_API_H__
```
alalek: Please add short license header (here and for other files): https://github.com/opencv/opencv/wiki/Coding_Style_Guide#file-structure
```
```

modules/cudacodec/src/cuvid_ffmpeg_impl.hpp
```diff
@@ -0,0 +1,575 @@
...
+using namespace cv;
+
+/*
+ * For CUDA encoder
+ */
+
+struct OutputMediaStream_FFMPEG
+{
+	bool open(const char* fileName, int width, int height, double fps);
+	void close();
```
alalek: Please replace tabs => 4 spaces

modules/cudacodec/src/cuvid_ffmpeg_api.hpp
```diff
@@ -0,0 +1,49 @@
...
+//#   define OPENCV_FFMPEG_API
+//#endif
+//#endif
+	/*
+	 * For CUDA encoder
+	 */
+
+	OPENCV_FFMPEG_API struct OutputMediaStream_FFMPEG* create_OutputMediaStream_FFMPEG(const char* fileName, int width, int height, double fps);
+	OPENCV_FFMPEG_API void release_OutputMediaStream_FFMPEG(struct OutputMediaStream_FFMPEG* stream);
+	OPENCV_FFMPEG_API void write_OutputMediaStream_FFMPEG(struct OutputMediaStream_FFMPEG* stream, unsigned char* data, int size, int keyFrame);
```
alalek: Please avoid indentation in namespace/extern "C" blocks.

modules/cudacodec/src/cuvid_ffmpeg_api.hpp
```diff
@@ -0,0 +1,49 @@
...
+//
+//#ifndef OPENCV_FFMPEG_API
+//#if defined(__OPENCV_BUILD) || defined(BUILD_PLUGIN)
+//#   define OPENCV_FFMPEG_API
+//#elif defined _WIN32
+//#   define OPENCV_FFMPEG_API __declspec(dllexport)
+//#elif defined __GNUC__ && __GNUC__ >= 4
+//#   define OPENCV_FFMPEG_API __attribute__ ((visibility ("default")))
+//#else
+//#   define OPENCV_FFMPEG_API
```
alalek: I believe `OPENCV_FFMPEG_API` modifier can be removed.

modules/cudacodec/CMakeLists.txt
```diff
@@ -13,6 +13,10 @@ ocv_glob_module_sources()
 
 set(extra_libs "")
 
+if(HAVE_FFMPEG)
```
alalek: Perhaps this should not work with `HAVE_FFMPEG_WRAPPER` (Windows)

modules/cudacodec/src/precomp.hpp
```diff
@@ -81,7 +81,7 @@
     #include "video_decoder.hpp"
     #include "video_parser.hpp"
 
-    #include "../src/cap_ffmpeg_api.hpp"
+    #include "cuvid_ffmpeg_api.hpp"
```
alalek: We need guard: `#if defined(HAVE_FFMPEG)` here and in places where this API is used.

cudawarped: I am not that familiar with CMake so I don't know how easy it would be to achieve, but wouldn't it be an idea to forced Build_opencv_cudacodec=FALSE if -DWITH_FFMPEG=OFF in addition to or instead of  placing guards around all the encode decode source files?

modules/cudacodec/CMakeLists.txt
cudawarped: Would it be possible to quickly explain the difference between HAVE_FFMPEG and HAVE_FFMPEG_WRAPPER

cudawarped:
I have added the changes but I am still unsure if I have used the correct preprocessor.  What is the difference in purpose between WITH_FFMPEG and WITH_FFMPEG_WRAPPER?

I have also added a guard on the tests as the smaller video requires ffmpeg to demux the file.   The guards are not properly implemented yet because I can't work out how to include HAVE_FFMPEG_WRAPPER, should it be as a  preprocessor or cvconfig.h, or should I remove the guard and let the tests fail?

alalek:
> WITH_FFMPEG_WRAPPER

There is no such option/variable.

HAVE_FFMPEG - OpenCV can use FFmpeg SDK (headers/libs)
HAVE_FFMPEG_WRAPPER - OpenCV can't use FFmpeg, but it can load wrapper dynamically. Implies `HAVE_FFMPEG` option.

So your condition should be `HAVE_FFMPEG && !HAVE_FFMPEG_WRAPPER` (similar to that).
But, `HAVE_FFMPEG_WRAPPER` is not exposed via CMake to modules outside of `videoio`.

Also to use FFmpeg header/libs your should use proper dependency target (`ocv.3rdparty.ffmpeg`).
Are you sure that your `cuvid_ffmpeg_impl.hpp` code has been compiled on your side? I don't see any FFmpeg `#include`.
You mention file `cap_ffmpeg_legacy_api.hpp`, but it is missing in the patch.

cudawarped:
Your correct, I wasn't correctly including cuvid_ffmpeg_impl.hpp and now I have, I realize everything depends on FFmpeg headers/libs.  

I still am unsure as to why the cudacodec routines were removed in #14774 so I'm not entirely sure if I should be directly using FFmpeg headers/libs in cudacodec or somehow using the dynamic wrapper?  If I can use the FFmpeg headers/libs what is the recommended way to include them in cudacodec?  If not I will have to leave this and it may be an idea to just disable the FFmpeg fallback in cudacodec.

alalek:
Dynamic wrapper doesn't contain CUDA-related code anymore.
Used classes should be bundled into cudacodec sources itself (and they depend/use FFmpeg SDK).

cudawarped:
By the used classes do you mean everything which was contained in cap_ffmpeg_impl.hpp and cap_ffmpeg_api.hpp which I am trying to move to cuvid_ffmpeg_imp.hpp and cuvid_ffmpeg_api.hpp?

So this is my first time working with CMake, how do I directly link to the FFmpeg SDK in cudacodec?

cudawarped:
So on reading the FFmpeg [readme ](https://github.com/opencv/opencv/blob/master/3rdparty/ffmpeg/readme.txt) and [this](https://github.com/opencv/opencv/blob/master/cmake/OpenCVModule.cmake)  CMake file I think I can see what is going on.

Now the only issue is that the shared library on windows has changed from `opencv_ffmpeg411_64.dll` to `opencv_videoio_ffmpeg411_64.dll` and the symbols which allow direct access to the bit stream have been removed

- create_InputMediaStream_FFMPEG

- release_InputMediaStream_FFMPEG

- read_InputMediaStream_FFMPEG

so I cannot see a way to read the video stream from a file with FFmpeg api any more.  This is currently required as a fall back option when cuvidCreateVideoSource() fails.  Is there another way to achieve this?

I had a quick look at the MFX implementation but it appears this only accepts raw streams as input.

mshabunin:
@cudawarped , current CUDA codec design is flawed, because it uses private methods and definitions of _videoio_ module. It is better to add raw stream output support to VideoCapture to make it work instead of duplicating the code. Or, even better, new class: `VideoContainer`.

cudawarped:
@mshabunin that sounds reasonable, I may not have time to implement this for a while, it was not my intention to implement a fix just to highlight the changes which broke cudacodec after 4.1.0, if someone else wants to take over so that cudacodec is not broken in 4.1.1 then feel free.

cudawarped:
@mshabunin everything should be working as well as it was before the break.  

Ideally cudacodec should be updated to mirror the [AppDec](https://github.com/NVIDIA/video-sdk-samples/tree/master/Samples/AppDecode/AppDec) sample from Nvidia, adding support for a wider variety of color space formats etc.  Unfortunately because cudacodec needs to work with all CUDA API's from the last 5 years mirroring the sample code and maintaining compatibility would be very difficult.

modules/cudacodec/src/ffmpeg_video_source.cpp
```diff
@@ -44,97 +44,116 @@
...
+void FourccToChromaFormat(const int pixelFormat, ChromaFormat &chromaFormat, int & nBitDepthMinus8)
 {
-    CV_Assert( init_MediaStream_FFMPEG() );
+    switch (pixelFormat)
+    {
+    case CV_FOURCC_MACRO('I', '4', '2', '0'):
+        chromaFormat = YUV420;
+        nBitDepthMinus8 = 0;
+        break;
+    default:
```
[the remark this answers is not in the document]
cudawarped: I think your change makes everything clearer because I am pretty sure the historic version of [AppDec](https://github.com/NVIDIA/video-sdk-samples/tree/master/Samples/AppDecode/AppDec) which cudacodec is based on cannot deal with any formats other than YUV420. Therefore any other cases would be redundant.

modules/cudacodec/include/opencv2/cudacodec.hpp
```diff
@@ -320,7 +323,7 @@ class CV_EXPORTS_W RawVideoSource
     @param size Size in bytes of current frame.
     @param endOfFile Indicates that it is end of stream.
      */
-    virtual bool getNextPacket(unsigned char** data, int* size, bool* endOfFile) = 0;
+    virtual bool getNextPacket(unsigned char** data, size_t* size, bool* endOfFile) = 0;
```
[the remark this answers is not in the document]
cudawarped: I agree should I change this or are you happy to?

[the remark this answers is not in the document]
alalek: It would be nice if you can add commit for that.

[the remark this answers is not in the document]
cudawarped: commit added

PR #2187, opened 2019-07-14
Data augmentation module (augment)

FadyEssam:
# Data augmentation module

* ## Overview
This projects aims mainly to implement module augment which provides simple api (as intuitive and easy as keras) to be able to create Augmenter instance and add to it transformations with random ranges and application probability and apply these transformations randomly on both images and their corresponding ground truth data

* ## Goals
1. Augmenter class to add transformations and their application probability
2. Allow each transformation to have range of random values 
3. Apply the augmenter to batches of images
4. Apply the same transformations to ground truth data
5. Support mask, points and bounding boxes (rectangles) as ground truth data

* ## Supported ground truth data data structures
1. Key points
2. Bounding boxes
3. Polygons

* ## Supported transformations
**note : b means it is implemented in the base class with general functionality, o means overridden in child**
**note: base rectangle transformation utilizes the point transformation of the transformation**
**note: base mask transformation utilizes the image transformation implemented per transformation**
**note: base point transformation doesn't change its coordinates**

| Transformation | image | Points | Rectangles | Masks |
|---|---|---|---|---|
| rotate | o | o | b | b |
| resize | o | o | b | o |
| crop to fixed size | o | o | b | b |
| vertical flipping | o | b | b | b|
| horizontal flipping | o | b | b | b | 
|  Gaussian blur |  o | b | b | b |
| average blur | o | b | b | b |


dkurt:
Please adjust the table with origin names of transformations (i.e. AverageBlur) and add information about supported data types:

transform | image | rect | mask | point
---|---|---|---|---|---
AverageBlur | + | = | = | =
FlipHorizontal | + |  | | +

So we can see what else should be implemented. In example, there is no flip for rectangles and masks for now (due in case of flipping `mask` is the same as image, you can apply `image` transform for masks by default).

modules/augment/misc/python/test/test_augmenter_images.py
```diff
@@ -0,0 +1,111 @@
...
+class augmenter_test(NewOpenCVTests):
+    def test_augmenter_images(self):
+        numberOfImages = np.random.randint(MIN_NUMBER_OF_TESTS, MAX_NUMBER_OF_TESTS)
+
+        aug = cv.augment_Augmenter()
+        aug.add(t=cv.augment_FlipHorizontal(), prob=0.7)
+        aug.add(t=cv.augment_FlipVertical(), prob=0.5)
+        aug.add(t=cv.augment_GaussianBlur(kernelSize=5, sigma=12), prob=0.7)
+        aug.add(t=cv.augment_Rotate(minAngle=0, maxAngle=180), prob=0.3)
+        aug.add(t=cv.augment_Resize(size=(1200, 900)), prob=0.4)
```
dkurt: Please remove redundant `t=` aliases.

modules/augment/include/opencv2/augment.hpp
```diff
@@ -0,0 +1,18 @@
...
+#ifndef OPENCV_AUGMENT_AUGMENT_HPP
+#define OPENCV_AUGMENT_AUGMENT_HPP
+#include <opencv2/augment/transform.hpp>
+#include <opencv2/augment/flip.hpp>
+#include <opencv2/augment/rotate.hpp>
+#include <opencv2/augment/gaussianblur.hpp>
+#include <opencv2/augment/averageblur.hpp>
+#include <opencv2/augment/resize.hpp>
+#include <opencv2/augment/augmenter.hpp>
+#include <opencv2/augment/croptofixedsize.hpp>
```
dkurt: I think that we need to move all the transformations into a single file and name it somethind like `all_transformations.hpp`.

modules/augment/include/opencv2/augment/croptofixedsize.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+
+#ifndef OPENCV_AUGMENT_CROPTOFIXEDSIZE_HPP
+#define OPENCV_AUGMENT_CROPTOFIXEDSIZE_HPP
+#include <opencv2/augment/transform.hpp>
+
+namespace cv {
+    namespace augment {
+
+        class CV_EXPORTS_W CropToFixedSize : public Transform
```
dkurt: Please fix indentation.

modules/augment/include/opencv2/augment/flip.hpp
```diff
@@ -0,0 +1,55 @@
...
+public:
+    /* @brief Constructor
+    */
+    CV_WRAP FlipVertical();
+
+    /* @brief Apply the vertical flipping to a single image
+        @param _src Input image to be flipped
+        @param _dst Output (flipped) image
+    */
+    CV_WRAP virtual void image(InputArray _src, OutputArray _dst);
```
dkurt: Please remove extra `_` because it's not needed.

modules/augment/src/rotate.cpp
```diff
@@ -0,0 +1,40 @@
...
+// of this distribution and at http://opencv.org/license.html.
+
+#include "precomp.hpp"
+namespace cv { namespace augment {
+
+Rotate::Rotate(float minAngle, float maxAngle)
+{
+    this->minAngle = minAngle;
+    this->maxAngle = maxAngle;
+}
```
dkurt:
```cpp
Rotate::Rotate(float _minAngle, float _maxAngle) : minAngle(_minAngle), maxAngle(_maxAngle) {}
```

modules/augment/src/transform.cpp
```diff
@@ -0,0 +1,132 @@
...
+    Point2f br_transformed = this->point(br);
+
+    float x1_transformed = std::min({ tl_transformed.x, bl_transformed.x, tr_transformed.x, br_transformed.x });
+    float y1_transformed = std::min({ tl_transformed.y, bl_transformed.y, tr_transformed.y, br_transformed.y });
+    float x2_transformed = std::max({ tl_transformed.x, bl_transformed.x, tr_transformed.x, br_transformed.x });
+    float y2_transformed = std::max({ tl_transformed.y, bl_transformed.y, tr_transformed.y, br_transformed.y });
+
+    Vec4f output_box({ x1_transformed ,y1_transformed ,x2_transformed ,y2_transformed });
+    return output_box;
+}
```
dkurt: Can you briefly explain what this method does?

modules/augment/src/rotate.cpp
FadyEssam: I will add _ to the members of the classes instead of the functions arguments to remove the warnings and because the functions parameters are the ones visible to the user

dkurt: This is not a header file. You can use different names between function prototype and body.

modules/augment/src/transform.cpp
FadyEssam: We need the output bounding box to include the same object included by the original box and still represent it by 2 points, so sometimes when the box is rotated and shifted, We can get the minimum x, y and maximum x,y of the transformed points to ensure including the whole object.

This is the same approach using in data augmentation library imaug used in pytorch

modules/augment/include/opencv2/augment.hpp
FadyEssam: why? I think it is clearer that each transformation has its own file

modules/augment/include/opencv2/augment/gaussianblur.hpp
```diff
@@ -0,0 +1,73 @@
...
+    Size _minKernelSize;
+    Size _maxKernelSize;
+    float _minSigmaX;
+    float _minSigmaY;
+    float _maxSigmaX;
+    float _maxSigmaY;
+    Size _kernelSize;
+    float _sigmaX;
+    float _sigmaY;
+    bool _sameXY;
```
dkurt: Why you did these changes? All the members should be named without extra characters.

modules/augment/include/opencv2/augment/flip.hpp
```diff
@@ -0,0 +1,55 @@
...
+        @param _src Input image to be flipped
+        @param _dst Output (flipped) image
+    */
+    CV_WRAP void image(InputArray _src, OutputArray _dst);
+
+    /* @brief Apply the flipping for a single point
+        @param src Input point to be flipped
+    */
+    Point2f point(const Point2f& src);
+
```
dkurt: Please remove extra empty line here and in other places

dkurt:
@FadyEssam, please consider https://github.com/opencv/opencv_contrib/pull/2187#issuecomment-511276705. This contribution requires such kind of table to see what's done and what should be implemented next. Thanks!

FadyEssam:
> Please adjust the table with origin names of transformations (i.e. AverageBlur) and add information about supported data types:
> 
> transform	image	rect	mask	point
> AverageBlur	+	=	=	=
> FlipHorizontal	+			+
> So we can see what else should be implemented. In example, there is no flip for rectangles and masks for now (due in case of flipping `mask` is the same as image, you can apply `image` transform for masks by default).
> 
> `+` means supported
> `=` means identity

I am working on removing code warnings first, then doc errors, then I will work on the description

modules/augment/include/opencv2/augment/transform.hpp
```diff
@@ -0,0 +1,78 @@
...
+    CV_WRAP virtual void image(InputArray _src, OutputArray _dst);
+
+    /* @brief Apply the transformation for a single point (this is overridden by transformations implementation)
+        @param src Input point to be tranformed
+    */
+    virtual Point2f point(const Point2f& src);
+
+
+    /* @brief Apply the transformation for a rectangle
+        @param box Vec4f consisting of (x1, y1, x2, y2) corresponding to (top left, bottom right)
```
dkurt: Fix a documentation line please as well

modules/augment/include/opencv2/augment/averageblur.hpp
```diff
@@ -50,6 +38,8 @@ private :
     Size maxKernelSize;
     Size kernelSize;
     bool sameXY;
+    std::vector<int> borderTypes;
+    int borderType;
```
dkurt: Please avoid duplication

modules/augment/misc/python/test/test_average_blur.py
```diff
@@ -7,10 +7,8 @@
 np.random.seed(seed=1)
 cv.setRNGSeed(seed=1)
 
-transformations = [cv.augment_AverageBlur(minKernelSize=3, maxKernelSize=15),
-                   cv.augment_AverageBlur(minKernelSize=(5,7), maxKernelSize=(13,15)),
-                   cv.augment_AverageBlur(kernelSize=7),
-                   cv.augment_AverageBlur(kernelSize=(5,7))]
+transformations = [cv.augment_AverageBlur(minKernelSize=(5,7), maxKernelSize=(13,15)),
+                   cv.augment_AverageBlur(kernelSize=(5,7), borderType=cv.BORDER_REPLICATE)]
```
dkurt: Please keep tests for all the cases.

modules/augment/src/averageblur.cpp
```diff
@@ -5,41 +5,41 @@
 #include "precomp.hpp"
 namespace cv { namespace augment {
 
-AverageBlur::AverageBlur(Size _minKernelSize, Size _maxKernelSize)
+AverageBlur::AverageBlur(Size _minKernelSize, Size _maxKernelSize, const std::vector<int>& _borderTypes)
 {
+    if (_borderTypes.size() > 0) borderTypes = _borderTypes;
+
+    else borderTypes = { BORDER_CONSTANT , BORDER_REPLICATE , BORDER_REFLECT};
```
dkurt: Please fix coding style:
```cpp
    if (_borderTypes.size() > 0)
        borderTypes = _borderTypes;
    else
        borderTypes = { BORDER_CONSTANT, BORDER_REPLICATE, BORDER_REFLECT };
```

dkurt: What about the rest of types? 
https://docs.opencv.org/master/d2/de8/group__core__array.html#ga209f2f4869e304c82d07739337eae7c5

modules/augment/src/averageblur.cpp
```diff
@@ -5,41 +5,41 @@
...
+    borderTypes = { _borderType };
     minKernelSize = maxKernelSize = _kernelSize;
     sameXY = false;
 }
 
 void AverageBlur::init(const Mat&)
 {
-    kernelSize.height = Transform::rng.uniform(minKernelSize.height / 2, maxKernelSize.height / 2 + maxKernelSize.height % 2) * 2 + 1; //generate only random odd numbers
-    kernelSize.width = sameXY? kernelSize.height : Transform::rng.uniform(minKernelSize.width / 2, maxKernelSize.width / 2 + maxKernelSize.width % 2) * 2 + 1; //generate only random odd numbers
+    if (minKernelSize == maxKernelSize) kernelSize = minKernelSize;
```
dkurt: If `minKernelSize == maxKernelSize`, `kernelSize` equals `minKernelSize` (see constructor)

modules/augment/include/opencv2/augment/averageblur.hpp
FadyEssam: This is not a duplicatio.. borderType is the current random instance of borderTypes

modules/augment/misc/python/test/test_average_blur.py
FadyEssam: The constructor(int kernelSize, int borderType = default) was removed because it conflicts with (int minKernelSize, int maxKernelSize , int borderType = default)

the constructor(Size kernelSize, borderType) is sufficient for the same purpose

modules/augment/src/averageblur.cpp
FadyEssam: They are not supported in blur and give exception of non supported borderType

modules/augment/src/averageblur.cpp
FadyEssam: what do you mean?

modules/augment/src/averageblur.cpp
dkurt:
```
cv.blur(img, (3, 3), borderType=cv.BORDER_REFLECT_101)
```
works

The rest of them throw exception, I agree.

modules/augment/src/averageblur.cpp
dkurt: I mean that every iteration this code assign `minKernelSize` to `kernelSize` however we already did it once in the constructor:
```cpp
    minKernelSize = _minKernelSize;
    maxKernelSize = _maxKernelSize;
    kernelSize = _minKernelSize;
    sameXY = false;
}
```

BTW, please check that it's initialized in all the constructors.

modules/augment/include/opencv2/augment/averageblur.hpp
dkurt: Oh, sorry. That's OK.

modules/augment/src/averageblur.cpp
FadyEssam: Yes I see

modules/augment/src/averageblur.cpp
FadyEssam: Ok I will add it

modules/augment/misc/python/test/test_average_blur.py
dkurt: This way we need rather resolve a conflict:

```python
cv.augment_AverageBlur(kernelSize=3)
cv.augment_AverageBlur(minKernelSize=3, minKernelSize=5)
```
are the most important methods. Let's keep the following constructors:

```cpp
AverageBlur(int kernelSize)
AverageBlur(int minKernelSize, int maxKernelSize, int borderType = BORDER_DEFAULT)
```

modules/augment/src/averageblur.cpp
FadyEssam: done

modules/augment/src/averageblur.cpp
FadyEssam: done

modules/augment/misc/python/test/test_average_blur.py
FadyEssam: done

modules/augment/src/averageblur.cpp
```diff
@@ -0,0 +1,61 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#include "precomp.hpp"
+namespace cv { namespace augment {
+
+AverageBlur::AverageBlur(Size _minKernelSize, Size _maxKernelSize, const std::vector<int>& _borderTypes)
+{
+    if (_borderTypes.size() > 0) borderTypes = _borderTypes;
```
dkurt: Please fix code style:
```cpp
if (_borderTypes.size() > 0)
    borderTypes = _borderTypes;
else
    borderTypes = { BORDER_CONSTANT , BORDER_REPLICATE , BORDER_REFLECT, BORDER_REFLECT_101};
```

modules/augment/src/rotate.cpp
```diff
@@ -0,0 +1,68 @@
...
+Rotate::Rotate(float _minAngle, float _maxAngle, std::vector<int> _interpolations, std::vector<int> _borderTypes, const Scalar&  _borderValue)
+{
+    minAngle = _minAngle;
+    maxAngle = _maxAngle;
+    angle = minAngle;
+
+    if (_borderTypes.size() > 0)
+        borderTypes = _borderTypes;
+    else
+        borderTypes = { BORDER_CONSTANT , BORDER_REPLICATE , BORDER_WRAP , BORDER_REFLECT_101 };
```
dkurt: What about the rest of border types? https://docs.opencv.org/master/d2/de8/group__core__array.html#ga209f2f4869e304c82d07739337eae7c5

FadyEssam: The rest of border types when combined with the rest of interpolations they throw exception at warpAffine()

I had to try all the combinations to detect them

modules/augment/src/transform.cpp
```diff
@@ -0,0 +1,149 @@
...
+    float x1_transformed = std::min({ tl_transformed.x, bl_transformed.x, tr_transformed.x, br_transformed.x });
+    float y1_transformed = std::min({ tl_transformed.y, bl_transformed.y, tr_transformed.y, br_transformed.y });
+    float x2_transformed = std::max({ tl_transformed.x, bl_transformed.x, tr_transformed.x, br_transformed.x });
+    float y2_transformed = std::max({ tl_transformed.y, bl_transformed.y, tr_transformed.y, br_transformed.y });
+
+    Rect2f output_box(Point2f(x1_transformed ,y1_transformed),  Point2f(x2_transformed ,y2_transformed));
+    return output_box;
+}
+
+Rect2f Transform::basicRectangle(const Rect2f& src)
```
dkurt: How this method differs from `Transform::rectangle`? When we need to use `basicRectangle` and when `rectangle`?

modules/augment/src/transform.cpp
```diff
@@ -0,0 +1,149 @@
...
+    dst.create(int(src.size()), 1, 0, -1, true);
+
+    for (unsigned int i = 0; i < src.size(); i++)
+    {
+        Mat src_row = src[i];
+        dst.create(src_row.size(), CV_32F, i, true);
+        Mat dst_row = dst.getMat(i);
+        this->points(src_row, dst_row);
+    }
+}
```
dkurt: What's the difference between polygons and vector of points (Transform::points)?

modules/augment/src/resize.cpp
```diff
@@ -0,0 +1,53 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#include "precomp.hpp"
+namespace cv { namespace augment {
+
+Resize::Resize(Size _size,const std::vector<int>& _interpolations)
```
dkurt: Please add a whitespace between arguments

modules/augment/src/augmenter.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+            }
+
+            dstImgs.create(img.size(), img.type(), i, true);
+            Mat dstImg = dstImgs.getMat(i);
+            img.copyTo(dstImg);
+
+    }
+
+}
```
dkurt: Please fix indentation

modules/augment/src/augmenter.cpp
```diff
@@ -0,0 +1,169 @@
...
+        dstImgs.create(augmentedImg.size(), augmentedImg.type(), i, true);
+        Mat dstImg = dstImgs.getMat(i);
+        augmentedImg.copyTo(dstImg);
+
+        dstRects.create(augmentedRects.size(), augmentedRects.type(), i, true);
+        Mat dstImgRects = dstRects.getMat(i);
+        augmentedRects.copyTo(dstImgRects);
+
+    }
+}
```
dkurt: Please reduce code duplication: every method looks very similar to each other.

modules/augment/src/flip.cpp
```diff
@@ -0,0 +1,43 @@
...
+}
+
+Point2f FlipHorizontal::point(const Point2f& src)
+{
+    return Point2f(srcImageCols - 1 - src.x, src.y);
+}
+
+Rect2f FlipHorizontal::rectangle(const Rect2f& src)
+{
+    return Rect2f(srcImageCols - 1 - (src.x + src.width), src.y, src.width, src.height);
```
dkurt: This is not correct implementation. In example, if `x == 0` and `width == 2`, it'll be transformed to `srcImageCols - 3` however it should be `srcImageCols - 2`.

modules/augment/src/transform.cpp
FadyEssam: rectangle() transforms the 4 corners and then calculates the new top left and bottom right even if the rectangle was deformed, that's why it's the general case, but basicRectangle is much faster because it only transforms the top left and bottom right and it depends on the fact that the rectangle shape will not change

FadyEssam: I think that we implemented specific rectangle() for each transformation so can I delete the basicRectangle() ?

modules/augment/src/transform.cpp
FadyEssam: polygons is designed to take vector<vector<point>> because the image can have more than one polygon

modules/augment/src/flip.cpp
FadyEssam: yes it shoulde be src.width-1 instead of src.width .. will fix it

FadyEssam: fixed

modules/augment/src/augmenter.cpp
FadyEssam: that;s because the only difference is the type of ground truth data

modules/augment/src/augmenter.cpp
FadyEssam: fixed

PR #2193, opened 2019-07-20
Moved aruco dictionary globals into getPredefinedDictionary to allow lazy-loading of the opencv DLL

SoylentGraham:
### This pullrequest changes

I want to lazy-load my opencv dll, to which I've added the aruco module.

But as these objects are constructed globally the system crashes as I've not loaded the DLL yet.

Moving them into the only place they're used, constructed statically, they construct on first use, and only after I've loaded the DLL.

Having statics hidden in a function isn't the greatest of solutions, but then I wouldn't have made them globals anyway :)

### DIFF
--- a/modules/aruco/src/dictionary.cpp
+++ b/modules/aruco/src/dictionary.cpp
@@ -268,37 +268,37 @@ Mat Dictionary::getBitsFromByteList(const Mat &byteList, int markerSize) {
 
 
 
+Ptr<Dictionary> getPredefinedDictionary(PREDEFINED_DICTIONARY_NAME name)
+{
+    // DictionaryData constructors calls
+    //    moved out of globals so construted on first use, which allows lazy-loading of opencv dll
+    static const Dictionary DICT_ARUCO_DATA = Dictionary(Mat(1024, (5 * 5 + 7) / 8, CV_8UC4, (uchar*)DICT_ARUCO_BYTES), 5, 0);
 
-// DictionaryData constructors calls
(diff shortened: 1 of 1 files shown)


modules/aruco/src/dictionary.cpp
```diff
@@ -269,36 +269,38 @@ Mat Dictionary::getBitsFromByteList(const Mat &byteList, int markerSize) {
...
-const Dictionary DICT_ARUCO_DATA = Dictionary(Mat(1024, (5*5 + 7)/8, CV_8UC4, (uchar*)DICT_ARUCO_BYTES), 5, 0);
 
-const Dictionary DICT_4X4_50_DATA = Dictionary(Mat(50, (4*4 + 7)/8, CV_8UC4, (uchar*)DICT_4X4_1000_BYTES), 4, 1);
-const Dictionary DICT_4X4_100_DATA = Dictionary(Mat(100, (4*4 + 7)/8, CV_8UC4, (uchar*)DICT_4X4_1000_BYTES), 4, 1);
-const Dictionary DICT_4X4_250_DATA = Dictionary(Mat(250, (4*4 + 7)/8, CV_8UC4, (uchar*)DICT_4X4_1000_BYTES), 4, 1);
-const Dictionary DICT_4X4_1000_DATA = Dictionary(Mat(1000, (4*4 + 7)/8, CV_8UC4, (uchar*)DICT_4X4_1000_BYTES), 4, 0);
+Ptr<Dictionary> getPredefinedDictionary(PREDEFINED_DICTIONARY_NAME name)
+{
+	// DictionaryData constructors calls
+	//	moved out of globals so construted on first use, which allows lazy-loading of opencv dll
```
alalek: tabs => 4 spaces

PR #2215, opened 2019-08-05
Lazy loading nvcuda.dll

### LINKED ISSUE #2210: opencv_cudaoptflow 4.1.1 has a strong dependency to nvcuda.dll
Starting from OpenCV 4.1.1, the NVidia Optflow sdk is taken into account in cudaoptflow.
However, because of the use of
cuCtxGetCurrent(), cuCtxPopCurrent(), cuCtxPushCurrent(), cuCtxSynchronize(),
then opencv_cudaoptflow411.dll fails to load, and prevents a program from running if nvcuda.dll is not available on the host machine.

A solution under windows+Visual Studio is to "Delay load" opencv_cudaoptflow411.dll, but it is less convenient than unders OpenCV <= 4.1.0, where checking at runtime for "cv::cuda::getCudaEnabledDeviceCount()" was enough to decide if cuda was available.

vchiluka5:
Added lazy loading of nvcuda.dll. Removed unwanted driver APIs as the final output remains same and replaced driver API with runtime API wherever possible

resolves #2210 

```
force_builders=Custom
buildworker:Custom=linux-4
docker_image:Custom=ubuntu-cuda:18.04
```

### DIFF
--- a/modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
+++ b/modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
@@ -29,11 +29,14 @@ CV_Error(cv::Error::HeaderIsNull, "Nvidia Optical Flow headers not found. Make s
 
 //macro for dll loading
 #if defined(_WIN64)
-#define MODULENAME TEXT("nvofapi64.dll")
+#define OF_MODULENAME TEXT("nvofapi64.dll")
+#define CUDA_MODULENAME TEXT("nvcuda.dll")
 #elif defined(_WIN32)
-#define MODULENAME TEXT("nvofapi.dll")
+#define OF_MODULENAME TEXT("nvofapi.dll")
+#define CUDA_MODULENAME TEXT("nvcuda.dll")
 #else
-#define MODULENAME "libnvidia-opticalflow.so.1"
+#define OF_MODULENAME "libnvidia-opticalflow.so.1"
+#define CUDA_MODULENAME "libcuda.so"
 #endif
 
 #define NVOF_API_CALL(nvOFAPI)                                                                      \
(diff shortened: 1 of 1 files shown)


modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
```diff
@@ -169,7 +201,8 @@ class NvidiaOpticalFlowImpl : public cv::cuda::NvidiaOpticalFlow_1_0
     NvOFHandle GetHandle() { return m_hOF; }
 
 protected:
-    HMODULE m_hModule; //module handle to load nvof dll
+    HMODULE m_hOFModule; //module handle to load nvof dll
+    HMODULE m_hCudaModule; //module handle to load cuda dll
```
alalek: It should loaded once per process. Don't load .dll per class instance.

vchiluka5: Noted

alalek: Please use singleton pattern for loading of required functions.

vchiluka5: Does my current changes address it?
I am loading it only when necessary

chacha21:
In your latest patch, GetProcAddress/dlsym are used to identify "cuCtxGetCurrent" and it is fine.

However, you can notice that "cuCtxPushCurrent" and "cuCtxPopCurrent" are defined as macros in cuda.h and are aliasing cuCtxPushCurrent_v2 and cuCtxPopCurrent_v2.
Thus, we can't be sure that in the future, "cuCtxGetCurrent" won't evolve the same.

That's why I suggest using a stringify macro to build the "cuCtxGetCurrent" symbol name, because it will protect from such changes an compatibility breaking.
`#define FUNCTION_NAME(name) MACRO_STRINGIFY(name)
#define MACRO_STRINGIFY(name) #name
...
const char* cuCtxGetCurrent_name = FUNCTION_NAME(cuCtxGetCurrent);
GetProcAddress(hNVCuda, cuCtxGetCurrent_name);
`

vchiluka5:
> In your latest patch, GetProcAddress/dlsym are used to identify "cuCtxGetCurrent" and it is fine.
> 
> However, you can notice that "cuCtxPushCurrent" and "cuCtxPopCurrent" are defined as macros in cuda.h and are aliasing cuCtxPushCurrent_v2 and cuCtxPopCurrent_v2.
> Thus, we can't be sure that in the future, "cuCtxGetCurrent" won't evolve the same.
> 
> That's why I suggest using a stringify macro to build the "cuCtxGetCurrent" symbol name, because it will protect from such changes and compatibility breaking.
> `#define FUNCTION_NAME(name) MACRO_STRINGIFY(name)
> 
> #define MACRO_STRINGIFY(name) #name
> 
> ...
> 
> const char* cuCtxGetCurrent_name = FUNCTION_NAME(cuCtxGetCurrent);
> 
> GetProcAddress(hNVCuda, cuCtxGetCurrent_name);
> `

cuCtxGetCurrent() is used to only fetch the runtime context and there is only 1 runtime context. This function has no further evolution of v2 and above and hence we can keep it simple for this function. 
Remaining functions like cuCtxPopCurrent and cuCtxPushCurrent were not needed in this scenario and hence removed.
This implementation is for SDK 1.0 and for next release we are trying to completely get rid of driver APIs so that integration becomes more easy.

chacha21:
:+1:

modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
alalek: > I am loading it only when necessary

Yes, you loading this on-demand.

> once

No, you loading libraries every time when class object is created.

vchiluka5: Okay Got it
So few doubts here @alalek 
1. We want a scenario where user can create multiple instances of same class. May be with bigger buffers and other stuff. In this case, we need to module handles global and load them only once. These global handles are within same namespace so is this approach OK?
There will be 2 global HMODULES 1 for cuda and 2 for opticalflow and one boolean to check if modules are loaded.

2. If i make it singleton then user will not be able to create mutiple instamces which blocks above usecase. Is ther any other way you suggest here?

alalek: 3rd variant should be straightforward.

You may skip library unloading process (a lot of code don't care about that and it may lead to unpredictable behavior during process termination).

[Example](https://github.com/opencv/opencv/blob/4.1.1/modules/videoio/src/cap_gstreamer.cpp#L184). Singleton class can provide pointers onto required functions (fields initialized in constructor).

vchiluka5: Is there any specific reason to skip library unloading part?. I am asking this since we faced some issues in other libraries due to not unloading.
I was thinking about having a refcount of class instances and when its 0, we can unload the libraries which are loaded and make them Null

alalek: Process termination is tricky part on Win32: https://github.com/opencv/opencv/issues/12750
Especially working with asynchronous multi-threaded libraries like OpenCL or CUDA.
If unloading works, then - OK.

> having a refcount

Not needed. Also this would break "once" behavior again.
Unload once in singleton destructor (see example, add `printf` to debug that).

vchiluka5: Sure. Thanks :-)
Will add in destructor

vchiluka5:
Hi @alalek 
Can you please elaborate build failure? I dont see error regarding nvidia optical flow and its on IOS.

modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
```diff
@@ -112,6 +115,109 @@ using namespace cv::cuda;
...
+        m_hOFModule = NULL;
+        m_cudaDriverAPIGetCurrentCtx = NULL;
+        m_NvOFAPICreateInstanceCuda = NULL;
+    }
+
+    static LoadNvidiaModules* Instance()
+    {
+        static LoadNvidiaModules* m_pInstance;
+        if (!m_pInstance)
+            m_pInstance = new LoadNvidiaModules;
```
alalek: > pointer

which code should call object destruction?

vchiluka5: Object destruction will happen by default right?

vchiluka5: if i declare this static LoadNvidiaModules* m_pInstance; outside this function then i am getting linker errors

vchiluka5: Added object destruction code.

alalek: Destructor code has never been called. You can ensure by adding `printf()` there.

Please follow the **example** (link is above):
- no instance pointers
- just static (global) object in function

vchiluka5: Okay. Done.
Please take a look

PR #2228, opened 2019-08-15
JavaScript face support

kavikode:

### DIFF
--- a/modules/aruco/CMakeLists.txt
+++ b/modules/aruco/CMakeLists.txt
@@ -1,2 +1,2 @@
 set(the_description "ArUco Marker Detection")
-ocv_define_module(aruco opencv_core opencv_imgproc opencv_calib3d WRAP python java)
+ocv_define_module(aruco opencv_core opencv_imgproc opencv_calib3d WRAP python java js)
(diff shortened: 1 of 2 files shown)


asmorkalov:
@kavikode The patch makes JS module mandatory otherwise face module is turned off as JS dependency is not satisfied.

kavikode:
> @kavikode The patch makes JS module mandatory otherwise face module is turned off as JS dependency is not satisfied.

Will you please help me understand which version you are using because it is not working for me @asmorkalov ? 

I already added the bindings for facemark. I even asked the same question on the OpenCV forum, link: https://answers.opencv.org/question/217298/opencvjs-building-wasm-cvfacecreatefacemarklbf/. 

Can you please let me know what am I doing wrong?

Please help!

asmorkalov:
Hello @kavicode. OpenCV builds tree of modules with dependencies. If some dependencies are not satisfied then the module is disabled. In your case Face module depends on JS. It means that if I build OpenCV for regular platforms without JS, Face module is always disabled. Please take a look on other modules wrapped to JS, Java, Python.

alalek:
@asmorkalov Please take a look on https://github.com/opencv/opencv/issues/15299.
This PR is just reproducer for JS bindings generator problem (PR can be merged if generator got fixed).

asmorkalov:
@alalek thanks for clarification.

asenyaev:

asenyaev:

PR #2276, opened 2019-09-27
Ensures that l0smooth input arguments are valid

developedby:
### This pullrequest changes

Adds two asserts to ensure that the input arguments 'lambda' and 'kappa' are valid.
If 'lambda <= 0' or 'kappa <= 1.0', the stop condition for the solver is never met because beta would not be increasing.


saskatchewancatch:
Nice nice. This kind of check should be more prevalent in these algos.

@developedby I think maintainers prefer that for enhancements like this, the change be based off of 3.4 branch. The problem would seem to exist in 3.4 as well.

After that whitespace fix, probably best to squash changes and rebase change onto 3.4. Maintainers merge regularly from 3.4 => master and the change will end up in master.

developedby:
@saskatchewancatch I think it's fixed now

PR #2278, opened 2019-09-30
Implementation of Global sampling based matting 

Nerdyvedi:


modules/ximgproc/CMakeLists.txt
```diff
@@ -1,2 +1,2 @@
 set(the_description "Extended image processing module. It includes edge-aware filters and etc.")
-ocv_define_module(ximgproc opencv_core opencv_imgproc opencv_calib3d opencv_imgcodecs WRAP python java)
+ocv_define_module(ximgproc opencv_core opencv_highgui opencv_imgproc opencv_calib3d opencv_imgcodecs WRAP python java)
```
asmorkalov: `highgui` dependency is redundant for the module. It's already there for testing.

modules/ximgproc/include/opencv2/ximgproc/globalmatting.hpp
```diff
@@ -0,0 +1,86 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+#ifndef GLOBAL_MATTING_H
+#define GLOBAL_MATTING_H
+
+#include <opencv2/highgui.hpp>
```
asmorkalov: Please remove it. Highgui module is not used in the header file.

asmorkalov:
@Nerdyvedi Do you have any progress on the patch?

Nerdyvedi:
@asmorkalov Could you please take a look at what's wrong, I don't think the issue is with my code

modules/ximgproc/include/opencv2/ximgproc/globalmatting.hpp
```diff
@@ -0,0 +1,85 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+#ifndef GLOBAL_MATTING_H
+#define GLOBAL_MATTING_H
```
alalek: Add `OPENCV_XIMGPROC_` prefix and "_HPP" suffix

modules/ximgproc/include/opencv2/ximgproc/globalmatting.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+#include <opencv2/imgproc.hpp>
+#include <opencv2/ximgproc/edge_filter.hpp>
+
+#include <string>
+#include <cstdlib>
```
alalek: Unnecessary in header file too.

modules/ximgproc/src/globalmatting.cpp
```diff
@@ -0,0 +1,568 @@
...
+	result += sqr(i - (alpha * f + (1 - alpha) * b));
+    }
+
+    return sqrt(result);
+}
+
+// Eq. 4
+float GlobalMatting::distCost(const cv::Point &p0, const cv::Point &p1, float minDist)
+{
+    int dist = sqr(p0.x - p1.x) + sqr(p0.y - p1.y);
```
alalek: `normL2Sqr<int>(p0 - p1);`

modules/ximgproc/src/globalmatting.cpp
```diff
@@ -0,0 +1,568 @@
...
+
+void GlobalMatting::expansionOfKnownRegions(const cv::Mat_<cv::Vec3b> &image,
+	                            cv::Mat_<uchar> &trimap,
+	                            int r, float c)
+{
+    int w = image.cols;
+    int h = image.rows;
+
+    for (int x = 0; x < w; ++x)
+	for (int y = 0; y < h; ++y)
```
alalek: "(x,y)" for loops have slowest memory access pattern. Any reason for that?

modules/ximgproc/src/globalmatting.cpp
```diff
@@ -0,0 +1,568 @@
...
+	        }
+	}
+
+    for (int x = 0; x < trimap.cols; ++x)
+	for (int y = 0; y < trimap.rows; ++y)
+	{
+	    if (trimap(y, x) == 1)
+	        trimap(y, x) = 0;
+	    else if (trimap(y, x) == 254)
+	        trimap(y, x) = 255;
```
alalek: Consider using `cv::threshold()` here instead of manual loop.

modules/ximgproc/src/globalmatting.cpp
```diff
@@ -0,0 +1,568 @@
...
+}
+
+// erode foreground and background regions to increase the size of unknown region
+void GlobalMatting::erodeFB(cv::Mat_<uchar> &trimap, int r)
+{
+    int w = trimap.cols;
+    int h = trimap.rows;
+
+    cv::Mat_<uchar> foreground(trimap.size(), (uchar)0);
+    cv::Mat_<uchar> background(trimap.size(), (uchar)0);
```
alalek: > (uchar)0

`Scalar::all(0)`

modules/ximgproc/src/globalmatting.cpp
```diff
@@ -0,0 +1,568 @@
...
+
+    std::vector<cv::Point> coords(w * h);
+    for (int y = 0; y < h; ++y)
+	for (int x = 0; x < w; ++x)
+	    coords[x + y * w] = cv::Point(x, y);
+
+    for (int iter = 0; iter < 10; ++iter)
+    {
+	// propagation
+	std::random_shuffle(coords.begin(), coords.end());
```
alalek: Avoid using of std::random functions.
- their implementations may be not thread safe
- so their results are not reproducible (hard to reproduce test failures)

Use OpenCV's functions instead (they are thread-safe with thread-local RNG state, which is propagated through `cv::parallel_for()`).

modules/ximgproc/src/globalmatting.cpp
```diff
@@ -0,0 +1,568 @@
...
+
+void GlobalMatting::expansionOfKnownRegions(cv::InputArray _img, cv::InputOutputArray _trimap, int niter)
+{
+    cv::Mat img = _img.getMat();
+    cv::Mat &trimap = _trimap.getMatRef();
+
+    if (img.empty())
+	CV_Error(CV_StsBadArg, "image is empty");
+    if (img.type() != CV_8UC3)
+	CV_Error(CV_StsBadArg, "image mush have CV_8UC3 type");
```
alalek: `CV_CheckTypeEQ(img.type(), CV_8UC3, "...")`

modules/ximgproc/test/test_globalmatting.cpp
```diff
@@ -0,0 +1,48 @@
...
+namespace opencv_test
+{
+  namespace
+  {
+    const std::string INPUT_DIR       = "cv/ximgproc";
+    const std::string IMAGE_FILENAME  = "input/doll.png";
+    const std::string TRIMAP_FILENAME = "trimap/doll.png";
+
+
+    void runModel()
```
alalek: `static` to make function local.

modules/ximgproc/tutorials/globalmatting.md
```diff
@@ -0,0 +1,31 @@
...
+| Mean square error           |       28.3                      | 29.5                 |
+| Gradient error              |       25                        | 26.3                 |
+| Connectivity error          |       28                        | 36.3                 |
+
+
+Some of the outputs with of this implementation are as follows :
+
+|    Image                    | Trimap                          | Alpha matte(this implementation)  |
+|  --------------             | --------------                  | ------------------------          |
+|![alt text](https://raw.githubusercontent.com/Nerdyvedi/GSOC-Opencv-matting/master/Global-Sampling-Method-for-Alpha-Matting/Input/doll.png "img1" ) |![alt text](https://raw.githubusercontent.com/Nerdyvedi/GSOC-Opencv-matting/master/Global-Sampling-Method-for-Alpha-Matting/Trimap/doll.png "trimap1" ) |![alt text](https://raw.githubusercontent.com/Nerdyvedi/GSOC-Opencv-matting/master/Global-Sampling-Method-for-Alpha-Matting/Results/doll.png "results1" ) |
```
alalek: External links on embedded images are not allowed.

modules/ximgproc/src/globalmatting.cpp
Nerdyvedi: Scalar::all(0) works on my local machine but is producing errors here. @alalek

alalek: Right, it is template `Mat_` - my mistake. I though it is workaround for 0 implicit cast to `void*`.

Nerdyvedi:
@alalek , Buildbot is giving a patch size error, Even though I have squashed my commits. Is it because I have commited 9 images for the tutorial

alalek:
> modules/ximgproc/tutorials/images/doll-input.png

PNG format is not very good for real-life content.
Consider converting these "input" tutorials images into JPEG format.
Using PNG makes sense in tests only (lossless input).

modules/ximgproc/src/globalmatting.cpp
Nerdyvedi: @alalek , I couldn't find documentation of normL2Sqr anywhere

modules/ximgproc/src/globalmatting.cpp
Nerdyvedi: Could you suggest an alternative ?

modules/ximgproc/src/globalmatting.cpp
Nerdyvedi: I don't think cv::threshold function can be used here

modules/ximgproc/src/globalmatting.cpp
alalek:
```
for (int y = 0; y < h; ++y)
    for (int x = 0; x < w; ++x)
```

---

Don't use "tabs" in code

modules/ximgproc/src/globalmatting.cpp
alalek: [Code is here](https://github.com/opencv/opencv/blob/4.1.1/modules/core/include/opencv2/core/types.hpp#L1414).

Nerdyvedi:
@alalek I have made the changes you asked for, There is a patch size error, Could you let me know how to solve it?

alalek:
I rebased patch and cleanup code according OpenCV coding style guide.

Added reference test data check into test shows that result is unstable. Please take a look.

P.S. Fetch code from your fork directly, don't try to "rebase" or "merge".

modules/ximgproc/src/globalmatting.cpp
```diff
@@ -0,0 +1,616 @@
...
+    for (int y = 0; y < h; ++y)
+    {
+        for (int x = 0; x < w; ++x)
+        {
+            if (trimap(y, x) == 128)
+            {
+                Point p(x, y);
+
+                samples[y][x].fi = rand() % foregroundBoundary.size();
+                samples[y][x].bj = rand() % backgroundBoundary.size();
```
alalek: Avoid using `rand()` anywhere.
It is not stable and sometimes it is not thread safe. We need reproducible results for automated tests.

Consider using of OpenCV's thread local `theRNG()` implementation.

PR #2285, opened 2019-10-07
Update pd_inria.cpp

sturkmen72:
### This pullrequest changes
adds a new option to save bounding boxes as resized positive images.

### DIFF
--- a/modules/datasets/samples/pd_inria.cpp
+++ b/modules/datasets/samples/pd_inria.cpp
@@ -10,11 +10,20 @@ using namespace cv::datasets;
 int main(int argc, char *argv[])
 {
     const char *keys =
-            "{ help h usage ? |    | show this message }"
-            "{ path p         |true| path to dataset }";
+            "{ help h usage ? |     | show this message }"
+            "{ path p         |true | path to dataset }"
+            "{ save s         |false| save resized positive images }"
+            "{ rwidth rw      |64   | width of resized positive images }"
+            "{ rheight rh     |128  | height of resized positive images }"
+            "{ padding        |8    | vertical padding of resized positive images }";
 
     CommandLineParser parser(argc, argv, keys);
+    bool savebbox = parser.get<bool>("save");
+    int rwidth = parser.get<int>("rwidth");
+    int rheight = parser.get<int>("rheight");
+    int padding = parser.get<int>("padding");
     string path(parser.get<string>("path"));
+
     if (parser.has("help") || path=="true")
     {
         parser.printMessage();
@@ -28,6 +37,8 @@ int main(int argc, char *argv[])
     size_t test_size = dataset->getTest().size();
     cout << "train size: " << train_size << endl;
     cout << "test size: " << test_size << endl;
+
+    int bbox_count = 0;
 
     for( size_t i = 0; i < train_size; i++ )
     {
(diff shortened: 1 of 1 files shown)


sturkmen72:
a screenshot of actual sample


proposed changes adds ability of saving bounding boxes like images below





the saved images almost same as the images in /INRIAPerson/96X160H96/Train/pos when default values used. also the user can create his custom sized images.

modules/datasets/samples/pd_inria.cpp
```diff
@@ -53,14 +64,41 @@ int main(int argc, char *argv[])
...
+            int bottom = top;
+            int left = r0.x - r.x;
+            int right = left;
+
+            copyMakeBorder(img(r0), dst, top, bottom, left, right, BORDER_REFLECT);
+            resize(dst, dst, Size(rwidth, rheight), 0, 0, INTER_AREA);
+            imwrite(path + format("person_%04d.png", bboxcount++), dst);
+            }
+            else
             rectangle( img, example->bndboxes[j], Scalar( 0, 0, 255 ), 2 );
```
alalek: Please fix indentation

modules/datasets/samples/pd_inria.cpp
```diff
@@ -53,14 +64,41 @@ int main(int argc, char *argv[])
...
         }
 
+        if (!savebbox)
+        {
```
alalek:
```
if (!savebbox)
    contrinue;  // skip UI updates
...
```

sturkmen72:
@alalek thank you for your review and feedback.

 indentation is fixed.

"skip UI updates" no, on the contrary here we show the image with bounding boxes.

By the way there was a bug that i fixed it. ( when width of the bounding box wider )

the code may be a bit sloppy but i did not care cos it is just a sample

modules/datasets/samples/pd_inria.cpp
alalek: indentation of `rectangle` statement after `else` should be fixed too

modules/datasets/samples/pd_inria.cpp
alalek: > "skip UI updates" no, on the contrary here we show the image with bounding boxes.

Sure, need to invert condition:

```
if (savebbox)
    continue;  // skip UI updates
```

Or please update indentation of "if body".

modules/datasets/samples/pd_inria.cpp
```diff
@@ -53,14 +64,47 @@ int main(int argc, char *argv[])
...
+
+            if(savebbox)
+            {
+                Mat dst;
+                int top = r0.y - r.y;
+                int bottom = top;
+                int left = r0.x - r.x;
+                int right = left;
+
+                copyMakeBorder(img(r0), dst, top, bottom, left, right, BORDER_REFLECT);
```
alalek: Why do we not just using `vpadding, vpadding, padding, padding` (without calculations with rectangles above)?
Or with scaled padding: `padding *= rwidth / r.width; vpadding *= rheight / r.height` (if applicable, it is not clear from command line description and why it is not zero by default)

sturkmen72:
sorry again to my sloppiness :)

my main motivation was to recreate the images in `/INRIAPerson/96X160H96/Train/pos` which could not be opened with OpenCV (after updating libPNG). I was intending to use them with train_HOG.cpp. The bounding boxes in the dataset has different w/h ratios so i wanted to calculate padding.

let me try to add some explanation on the code later.

PR #2292, opened 2019-10-11
require OGRE 1.11+ for ovis

huangqinjin:
https://github.com/opencv/opencv_contrib/blob/83e98d2424bbe3854d4686dc6c9cf9a15812e8d7/modules/ovis/src/ovis.cpp#L638

The `create` API is introduced in commit [63906c8](https://github.com/OGRECave/ogre/commit/63906c8cc1b3d40d829296f1636a8e33bfe59b0d), so requires v1.11.5+.


paroj:
:+1:

PR #2293, opened 2019-10-13
Update .travis.yml

sturkmen72:
### This pullrequest changes
sets ` -DBUILD_PERF_TESTS=OFF -DBUILD_TESTS=OFF`
to avoid time out of Travis CI

actually it gives
> The job exceeded the maximum time limit for jobs, and has been terminated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,15 @@ compiler:
   - clang
 before_script:
   - cd ../
-  - git clone https://github.com/opencv/opencv.git
+  - git clone --branch 3.4 --depth=1 https://github.com/opencv/opencv.git
   - mkdir build-opencv
   - cd build-opencv
-  - cmake -DOPENCV_EXTRA_MODULES_PATH=../opencv_contrib/modules ../opencv
+  - cmake 
+    -DBUILD_EXAMPLES=ON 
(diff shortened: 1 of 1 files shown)


asmorkalov:
@alalek Is Travis-CI used somehow for validation?

alalek:
We don't use Travis for OpenCV repositories

alalek:
BTW,

Probably it would be nice to add `BUILD_EXAMPLES=ON`

> make -j5

Not sure that Travis allows multiple cores per Job on free accounts. Oversubscribing will cause terrible results.
Does `make -j$(nproc)` work?
Or we can switch to ninja build tool (`cmake -GNinja ...` + `ninja`)

asmorkalov:
The default host configuration for travis is 2-cores WM. So `-j5` is redundant.

sturkmen72:
> Probably it would be nice to add BUILD_EXAMPLES=ON

i think this will cause timeout again

another issue is indeed OpenCV 3.4 branch should be cloned in  Contrib 3.4 branch  and OpenCV master branch should be cloned in  Contrib master branch

alalek:
`git clone --branch 3.4 --depth=1 https://github.com/opencv/opencv.git`

PR #2301, opened 2019-10-18
Build DoG Pyramid if useProvideKeypoints is false

### LINKED ISSUE #2238: xfeatures2d SIFT needlessly computes DoG pyramid during compute() call
https://github.com/opencv/opencv_contrib/blob/fd004893cbc8c5c3f7dd45a17f0e34673027320c/modules/xfeatures2d/src/sift.cpp#L1154-L1162

In `SIFT` extractor when `detectAndCompute()` is called with `useProvidedKeypoints=true` (e.g. by calling `compute()`) the difference of gaussians pyramid (`dogpyr`) is computed, but it is not used at all - `dogpyr` is used only for `detect()` calls.

Fixing it - e.g. by moving line `buildDoGPyramid(gpyr, dogpyr);` into the if just below - would increase performance of `compute()` calls by 10%-20% (rough estimate on 2 tested images).

ab-dragon:
The buildDoGPyramid operation need not be performed unconditionally. In cases where it is not needed, both memory and speed performance can be improved

resolves #2238 

### This pullrequest changes

Build DoG Pyramid conditionally during detectAndCompute in SIFT

### DIFF
--- a/modules/xfeatures2d/src/sift.cpp
+++ b/modules/xfeatures2d/src/sift.cpp
@@ -1153,13 +1153,13 @@ void SIFT_Impl::detectAndCompute(InputArray _image, InputArray _mask,
     //double t, tf = getTickFrequency();
     //t = (double)getTickCount();
     buildGaussianPyramid(base, gpyr, nOctaves);
-    buildDoGPyramid(gpyr, dogpyr);
 
     //t = (double)getTickCount() - t;
     //printf("pyramid construction time: %g\n", t*1000./tf);
 
     if( !useProvidedKeypoints )
     {
+        buildDoGPyramid(gpyr, dogpyr);
         //t = (double)getTickCount();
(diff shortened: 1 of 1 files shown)


modules/xfeatures2d/src/sift.cpp
```diff
@@ -1151,13 +1151,13 @@ void SIFT_Impl::detectAndCompute(InputArray _image, InputArray _mask,
...
     //t = (double)getTickCount();
     buildGaussianPyramid(base, gpyr, nOctaves);
-    buildDoGPyramid(gpyr, dogpyr);
 
     //t = (double)getTickCount() - t;
     //printf("pyramid construction time: %g\n", t*1000./tf);
 
     if( !useProvidedKeypoints )
     {
+        buildDoGPyramid(gpyr, dogpyr);
```
alalek: Please reduce scope of `dogpyr` variable (move it here).

PR #2310, opened 2019-10-25
ovis: updates

paroj:
see individual commits

```
allow_multiple_commits=1
```


modules/ovis/samples/ovis_demo.py
```diff
@@ -31,3 +31,5 @@
 while cv.ovis.waitKey(1) != 27:
     R, t = iwin.getCameraPose()
     owin.setEntityPose("cam", t, R)
+
+del iwin # must be destroyed in reverse creation order
```
fegorsch: Note that `del` is guaranteed to remove the binding, but it does not necessarily free the underlying memory. This should work in CPython, however, as garbage collection is reference-count-based. Perhaps add a comment.

[Thread with more info.](https://stackoverflow.com/a/1316788)

PR #2320, opened 2019-10-30
s/seperate/separate/

xerus:
```
allow_multiple_commits=1
```

### DIFF
--- a/modules/ovis/include/opencv2/ovis.hpp
+++ b/modules/ovis/include/opencv2/ovis.hpp
@@ -19,7 +19,7 @@ namespace ovis {
 enum SceneSettings
 {
     /// the window will use a separate scene. The scene will be shared otherwise.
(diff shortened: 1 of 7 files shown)


modules/ovis/include/opencv2/ovis.hpp
```diff
@@ -18,8 +18,8 @@ namespace ovis {
 
 enum SceneSettings
 {
-    /// the window will use a seperate scene. The scene will be shared otherwise.
-    SCENE_SEPERATE = 1,
```
alalek: /cc @paroj 

Do we need some backward compatibility code?
```.diff
     SCENE_SEPARATE = 1,
+#ifndef CV_DOXYGEN
+    SCENE_SEPERATE = SCENE_SEPARATE,  // 20191101: backward compatibility
+#endif
```

PR #2330, opened 2019-11-04
PyrLKOpticalFlowBase::buildImagePyramid() is public

### LINKED ISSUE #2324: Make SparsePyrLKOpticalFlow::buildImagePyramid() public or free

- OpenCV => 4.1.2
- Operating System / Platform => Ubuntu 64 Bit
- Compiler => GCC

##### Detailed description

The overridden virtual function `SparsePyrLKOpticalFlow::calc()` accepts image pyramids as a possible input. However, there is currently no way to create and subsequently pass image pyramids as arguments because the member function `SparsePyrLKOpticalFlow::buildImagePyramid()` is protected. `SparsePyrLKOpticalFlow::buildImagePyramid()` should either be made public or made a free function such that the user can better manage system memory and compute resources and actually pass pyramids to `SparsePyrLKOpticalFlow::calc()`.

##### Steps to reproduce
N/A

czgdp1807:
resolves #2324 

### This pullrequest changes

I have changed the scope of PyrLKOpticalFlowBase::buildImagePyramid() from protected to public. The issue suggested to do SparsePyrLKOpticalFlow::buildImagePyramid() to public or free. AFAICT, SparsePyrLKOpticalFlowImpl is a friend of PyrLKOpticalFlowBase, and hence it is able to access buildImagePyramid(). So if that's the case then making it free is also a good choice. 
Please let me know if I have done or said anything wrong. Thanks.

### DIFF
--- a/modules/cudaoptflow/src/pyrlk.cpp
+++ b/modules/cudaoptflow/src/pyrlk.cpp
@@ -89,13 +89,14 @@ namespace
 
         void dense(const GpuMat& prevImg, const GpuMat& nextImg, GpuMat& u, GpuMat& v, Stream& stream);
 
+        void buildImagePyramid(const GpuMat& prevImg, std::vector<GpuMat>& prevPyr, const GpuMat& nextImg, std::vector<GpuMat>& nextPyr, Stream stream);
+
     protected:
         int winSize_[2];
         int halfWinSize_[2];
         int maxLevel_;
         int iters_;
         bool useInitialFlow_;
-        void buildImagePyramid(const GpuMat& prevImg, std::vector<GpuMat>& prevPyr, const GpuMat& nextImg, std::vector<GpuMat>& nextPyr, Stream stream);
     private:
         friend class SparsePyrLKOpticalFlowImpl;
         std::vector<GpuMat> prevPyr_;


czgdp1807:
Ah! I think, I should have checked out from 3.4 branch on my system will correct it here.

czgdp1807:
I observe that, `cudaoptflow` are not there in `3.4` so changed the base again to `master`.

czgdp1807:
ping @llschloesser

czgdp1807:
I will look into the failures. I have to figure out how my change is affecting the tests.

czgdp1807:
I just tested `opencv_contrib:master` locally on my system and observed that the same testing i.e., `opencv_test_tracking` is failing on it too. So, the failure is not due to my changes. Please let me know how to proceed in these cases. 
PS - Rebasing will not help due to the reasons suggested [here](https://github.com/opencv/opencv_contrib/pull/2330#issuecomment-549405060).

alalek:
Files from "src" is not a part of OpenCV API.

czgdp1807:
@alalek Can you please elaborate? I cannot figure out how your comment is associated with the issue or the PR.

czgdp1807:
ping @llschloesser

alalek:
Request from issue is about extending of "public OpenCV API" available for user application.

There is high level list of steps:
- design API. Building pyramids for **both** prev/next frames is useless for pipelining. At least check existed API for "CPU" from the main OpenCV repository (see `buildPyramid()` at docs.opencv.org ).
- add declaration into header file from "include" module's directory
- add documentation
- add implementation
- add corresponding tests (and fix implementation...)
- add sample (optional)
- add/update tutorial (optional)

So, just changing "protected:" => "public:" in internal private files does **NOT** work.

czgdp1807:
I looked at [buildPyramid](https://docs.opencv.org/2.4/modules/imgproc/doc/filtering.html). I think that may be we don't need to make `buildImagePyramid`, public or independent because, there is already a function fulfilling that purpose. So probably, we can close this PR as well as the associated issue. However, if the issue is still needed to be resolved, please let me know, I will make updates to this PR.

I have the following queries regarding the Lucas-Kanade Optical Flow Algorithm using Pyramids,

1. In the following code snippet, may be, we can add `prevPyr.pop_back()` and `nextPyr.pop_back()` as we never access those layers of the pyramid once we are done with them, provided we start from the top most level(that is from the back of the vector, apparently which is being done in the code).  
https://github.com/opencv/opencv_contrib/blob/abfcde006f3ec7fe5748deec35905fa89802596b/modules/cudaoptflow/src/pyrlk.cpp#L209-L216

2.  > Building pyramids for both prev/next frames is useless for pipelining

AFAICT, you meant to say that, building pyramids is not needed to estimate the optical flow. IMO, it should be possible to do that and it will save the memory space, however, in that case, we will be first required to go to the topmost level, using `pyrDown` and then moving backwards, from topmost level, computing each lower layer. Overall, we will be having at most two layers in the memory. The speed will be affected by a constant factor because we will be required to compute each layer twice. Please let me know if I have interpreted your comment correctly. Correct me, if possible. **I suspect the possibility of this**, I haven't looked into the details yet. Your comments on this logic will be helpful. Thanks.

3. Can I add a function to build Laplacian Pyramid of an image. Though the documentation states that it is possible build the same, using `pyrUp` and `pyrDown` but I observed that people face problems like [here](https://www.edureka.co/community/643/how-to-make-a-laplacian-pyramid-using-opencv-python). It would be quite good to provide such a functionality by `OpenCV` itself. I will raise an issue and create a PR if you support this. 

Thanks.

alalek:
> existed `buildPyramid()`

Main problem: it doesn't utilize CUDA. Or its results can't be used in with other CUDA calls.

> 2

Idea is to avoid pyramid computations twice per incoming frame (or for both prev/next images) during video processing.

Pipeline example (CUDA-like as original issue is about CUDA):
```
GpuMat prevFrame = videosource.readFrame();
std::vector<GpuMat> prevFramePyramid = buildPyramid(prevFrame);  // API/parameters can be adjusted
while (true)
{
    GpuMat nextFrame = videosource.readFrame();

    std::vector<GpuMat> nextFramePyramid = buildPyramid(nextFrame);
    //                                     ^^^^^^^^^^^^
    // called once per frame, another pyramid is got from previous iteration

    calcOpticalFlow(prevFramePyramid, nextFramePyramid, ...);
    ...

    // save frame & pyramid for the next iteration
    prevFrame = nextFrame;  // if needed
    prevFramePyramid = nextFramePyramid;  // or "swap()" if faster
}
```

czgdp1807:
Thanks for the clarification, I will look into the problem.

czgdp1807:
Closing in favour of https://github.com/opencv/opencv/pull/15889

PR #2296, opened 2019-10-15
new algorithm Rapid Frequency Selective Reconstruction (FSR) added

si40wiga:
### Submission
This pullrequest adds the implementation of the high-quality extrapolation algorithm Rapid Frequency Selective Reconstruction (FSR) with two quality profiles. As one of its fields of applications is inpainting, it was added to the existing inpainting interface in xphoto.
Please refer to
https://github.com/opencv/opencv_contrib/files/3730212/inpainting_comparison.pdf
for a detailed description of the algorithm and a comparison to existing inpainting methods in OpenCV.

```
#disable_ipp=ON
```

PR #2332, opened 2019-11-05
Add short tutorial for FSR-based image inpainting

si40wiga:
### This pullrequest changes
adds tutorial for #2296


alalek:
Thank you for contribution!

Usually we using JPEG images in documentation for real-life images.
Lossless PNG format is used for tests (or for diagrams / masks).

Also original image is stored twice (approach with 3 images would be better).

Could you please convert used images into JPEG format?
In case of non-acceptable artifacts, please use some image patch (ROI) in PNG format to show algorithm results (up to 64x64, you may upscale it by x4/x8).

asmorkalov:
@si40wiga Friendly reminder.

alalek:
You need to squash commits into one (we don't need full history of changes here - git still stores .png files).

si40wiga:
Thanks! I hope it's fine now.

> Also original image is stored twice (approach with 3 images would be better).
I'm a bit confused because I didn't store the original image twice. The reconstruction image shows the results with the two different quality profiles. The differences between the profiles are still visible in jpg format (e.g. at the front edge of the roof).

PR #2356, opened 2019-11-20
add rapid module for silhouette based 3D object tracking

paroj:
the core algorithm is in `rapid::rapid()`. However most of the internal functions were also exposed to allow different variations of this algorithm.
E.g. this can suit as base for more recent variants like:
https://www.youtube.com/watch?v=trX_AEn2lp0

### DIFF
--- a/modules/rapid/include/opencv2/rapid.hpp
+++ b/modules/rapid/include/opencv2/rapid.hpp
@@ -1,0 +1,126 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#ifndef OPENCV_RAPID_HPP_
+#define OPENCV_RAPID_HPP_
+
+#include <opencv2/core.hpp>
+#include <opencv2/imgproc.hpp>
+
+/**
+@defgroup rapid silhouette based 3D object tracking
+
+implements "RAPID-a video rate object tracker" @cite harris1990rapid with the dynamic control point extraction of @cite drummond2002real
+*/
+
+namespace cv
+{
+namespace rapid
+{
+//! @addtogroup rapid
+//! @{
+
+/**
+ * Debug draw markers of matched correspondences onto a lineBundle
+ * @param bundle the lineBundle
+ * @param srcLocations the according source locations
+ * @param newLocations matched source locations
+ * @param colors colors for the markers. Defaults to white.
+ */
+CV_EXPORTS_W void drawCorrespondencies(InputOutputArray bundle, InputArray srcLocations,
+                                       InputArray newLocations, InputArray colors = noArray());
+/**
+ * Debug draw search lines onto an image
+ * @param img the output image
+ * @param locations the source locations of a line bundle
+ * @param color the line color
+ */
+CV_EXPORTS_W void drawSearchLines(InputOutputArray img, InputArray locations, const Scalar& color);
+
+/**
+ * Draw a wireframe of a triangle mesh
+ * @param img the output image
+ * @param pts2d the 2d points obtained by @ref projectPoints
+ * @param tris triangle face connectivity
+ * @param color line color
+ * @param type line type. See @ref LineTypes.
+ * @param cullBackface enable back-face culling based on CCW order
+ */
+CV_EXPORTS_W void drawWireframe(InputOutputArray img, InputArray pts2d, InputArray tris,
+                                const Scalar& color, int type = LINE_8, bool cullBackface = false);
+/**
+ * Extract control points from the projected silhouette of a mesh
+ *
+ * see @cite drummond2002real Sec 2.1, Step b
+ * @param num number of control points
+ * @param len search radius (used to restrict the ROI)
+ * @param pts3d the 3D points of the mesh
+ * @param rvec rotation between mesh and camera
+ * @param tvec translation between mesh and camera
+ * @param K camera intrinsic
+ * @param imsize size of the video frame
+ * @param tris triangle face connectivity
+ * @param ctl2d the 2D locations of the control points
+ * @param ctl3d matching 3D points of the mesh
+ */
+CV_EXPORTS_W void extractControlPoints(int num, int len, InputArray pts3d, InputArray rvec, InputArray tvec,
+                                       InputArray K, const Size& imsize, InputArray tris, OutputArray ctl2d,
+                                       OutputArray ctl3d);
+/**
+ * Extract the line bundle from an image
+ * @param len the search radius. The bundle will have `2*len + 1` columns.
+ * @param ctl2d the search lines will be centered at this points and orthogonal to the contour defined by
+ * them. The bundle will have as many rows.
+ * @param img the image to read the pixel intensities values from
+ * @param bundle line bundle image with size `ctl2d.rows() x (2 * len + 1)` and the same type as @p img
+ * @param srcLocations the source pixel locations of @p bundle in @p img as CV_16SC2
+ */
+CV_EXPORTS_W void extractLineBundle(int len, InputArray ctl2d, InputArray img, OutputArray bundle,
+                                    OutputArray srcLocations);
+
+/**
+ * Find corresponding image locations by searching for a maximal sobel edge along the search line (a single
+ * row in the bundle)
+ * @param bundle the line bundle
+ * @param srcLocations the according source image location
+ * @param newLocations image locations with maximal edge along the search line
+ * @param response the sobel response for the selected point
+ */
+CV_EXPORTS_W void findCorrespondencies(InputArray bundle, InputArray srcLocations, OutputArray newLocations,
+                                       OutputArray response = noArray());
+
+/**
+ * Filter corresponding 2d and 3d points based on mask
+ * @param pts2d 2d points
+ * @param pts3d 3d points
+ * @param mask mask containing non-zero values for the elements to be retained
+ */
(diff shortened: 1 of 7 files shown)


modules/rapid/src/rapid.cpp
```diff
@@ -0,0 +1,359 @@
...
+    for (int i = 0; i < pts2d.rows; i++) {
+        if (img_rect.contains(pts2d(i))) {
+            img1(pts2d(i)) = i + 1;
+        }
+    }
+
+    // look up indices on contour
+    int offset = -1;
+    for (size_t i = 0; i < contour.size(); i++) {
+        if (int idx = img1(contour[i])) {
```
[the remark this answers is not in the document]
paroj: this is the C++ flavour of the python `with foo:` statement

modules/rapid/src/rapid.cpp
```diff
@@ -0,0 +1,359 @@
...
+namespace cv
+{
+namespace rapid
+{
+
+static int getSilhoutteVertices(const Size& imsize, const std::vector<Point>& contour,
+                                const Mat_<Point2f>& pts2d, std::vector<int>& v_idx)
+{
+    // store indices
+    Mat_<int> img1(imsize, 0);
```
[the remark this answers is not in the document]
paroj: 0, is faster then -1 as it uses memset

modules/rapid/src/rapid.cpp
```diff
@@ -0,0 +1,359 @@
...
+}
+
+void drawSearchLines(InputOutputArray img, InputArray _locations, const Scalar& color)
+{
+    Mat locations = _locations.getMat();
+    CV_Assert(locations.checkVector(2, CV_16S));
+
+    for (int i = 0; i < locations.rows; i++) {
+        Point pt1(locations.at<Vec2s>(i, 0));
+        Point pt2(locations.at<Vec2s>(i, locations.cols - 1));
```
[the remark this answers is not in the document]
paroj: there is other data..

modules/rapid/src/rapid.cpp
```diff
@@ -0,0 +1,347 @@
...
+
+    // sobel.cols = 2*LL + 1
+    const int LL = sobel.cols / 2;
+    const int ct = LL + 1;
+
+    // find closest maximum to center
+    for (int i = 0; i < sobel.rows; i++) {
+        int pos = ct;
+        uchar mx = sobel.at<uchar>(i, ct);
+        for (int j = 0; j < LL; j++) {
```
fegorsch: What about (untested pseudo-code):
```c++
uchar mx;
int pos;
minMaxIdx(sobel.rows(i), NULL, mx, NULL, pos);
```
?

modules/rapid/src/rapid.cpp
```diff
@@ -0,0 +1,347 @@
...
+    }
+}
+
+void findCorrespondencies(InputArray bundle, InputArray _srcLocations, OutputArray _newLocations,
+                          OutputArray _response)
+{
+    Mat_<uchar> sobel;
+    compute1DSobel(bundle.getMat(), sobel);
+
+    _newLocations.create(sobel.rows, 1, CV_32SC2);
```
fegorsch: Why does this have a different type than `srcLocations`?

modules/rapid/include/opencv2/rapid.hpp
```diff
@@ -0,0 +1,124 @@
...
+ */
+CV_EXPORTS_W void extractLineBundle(int len, InputArray ctl2d, InputArray frame, OutputArray bundle,
+                                    OutputArray srcLocations);
+
+/**
+ * Find corresponding image locations by searching for a maximal sobel edge along the search line (a single
+ * row in the bundle)
+ * @param bundle the line bundle
+ * @param srcLocations the according source image location
+ * @param newLocations image locations with maximal edge along the search line
```
fegorsch:
```suggestion
 * @param newLocations image locations with maximal edge along the search line as N x 1 Mat of type CV_32SC2
```

modules/rapid/src/rapid.cpp
```diff
@@ -0,0 +1,347 @@
...
+        }
+
+        if (!response.empty())
+            response(i) = mx;
+
+        newLocations.at<Point>(i, 0) = Point(srcLocations.at<Vec2s>(i, pos));
+    }
+}
+
+void drawCorrespondencies(InputOutputArray _img, InputArray _srcLocations, InputArray _newLocations,
```
fegorsch: The line bundle image is called `bundle` everywhere else. `img` is confusing here, because this is usually the name for the source image.

modules/rapid/src/rapid.cpp
```diff
@@ -0,0 +1,347 @@
...
+    Mat img = _img.getMat();
+    Mat_<Vec2s> srcLocations = _srcLocations.getMat();
+    Mat_<Point> newLocations = _newLocations.getMat();
+    Mat_<Vec4d> colors = _colors.getMat();
+
+    for (int i = 0; i < img.rows; i++) {
+        Vec2s ref(Vec2i(newLocations(i)));
+        for (int j = 1; j < img.cols - 1; j++) {
+            if (ref == srcLocations(i, j)) {
+                img(Rect(Point(j, i), Size(1, 1))) = colors.empty() ? Scalar::all(255) : colors(i);
```
fegorsch: Why can't you just do `img(Point(j, i)) = ...`?

modules/rapid/src/rapid.cpp
```diff
@@ -0,0 +1,347 @@
...
+
+    Mat img = _img.getMat();
+    Mat_<Vec2s> srcLocations = _srcLocations.getMat();
+    Mat_<Point> newLocations = _newLocations.getMat();
+    Mat_<Vec4d> colors = _colors.getMat();
+
+    for (int i = 0; i < img.rows; i++) {
+        Vec2s ref(Vec2i(newLocations(i)));
+        for (int j = 1; j < img.cols - 1; j++) {
+            if (ref == srcLocations(i, j)) {
```
fegorsch: I found this very confusing. I got what you're doing now, but would it perhaps be simpler to give `findCorrespondencies` another optional output containing the correspondency-coordinates in the bundle image?

modules/rapid/include/opencv2/rapid.hpp
```diff
@@ -0,0 +1,124 @@
...
+ * them. The bundle will have as many rows.
+ * @param frame the image to read the pixel intensities values from
+ * @param bundle the resulting bundle
+ * @param srcLocations the source locations in the image for each bundle location
+ */
+CV_EXPORTS_W void extractLineBundle(int len, InputArray ctl2d, InputArray frame, OutputArray bundle,
+                                    OutputArray srcLocations);
+
+/**
+ * Find corresponding image locations by searching for a maximal sobel edge along the search line (a single
```
fegorsch: If a sentence is the first sentence in the line, it starts with a lower-case letter.

(My preferred solution would of course be to stick to English and start all sentences with an upper-case letter and end all sentences with a period. That would also look much nicer in the generated documentation :) )

modules/rapid/src/rapid.cpp
```diff
@@ -0,0 +1,348 @@
...
+        const Point2f diff = contour((i + 1) % N) - contour((i - 1 + N) % N);
+        Point2f n(normalize(Vec2f(-diff.y, diff.x))); // perpendicular to diff
+        // make it cover L pixels
+        n *= len / std::max(std::abs(n.x), std::abs(n.y));
+
+        LineIterator li(_frame, contour(i) - n, contour(i) + n);
+        CV_DbgAssert(li.count == W);
+
+        for (int j = 0; j < li.count; j++, ++li) {
+            _srcLocations(i, j) = Vec2i(li.pos());
```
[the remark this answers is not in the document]
paroj: except that Point does not convert to Vec2s

modules/rapid/src/rapid.cpp
paroj: because that must assume the type of img

modules/rapid/src/rapid.cpp
paroj: which are only needed for the debug drawing?

modules/rapid/src/rapid.cpp
paroj: we must search from the center outwards here

fegorsch: why?

modules/rapid/src/rapid.cpp
fegorsch: why?

modules/rapid/src/rapid.cpp
paroj: in case there are multiple equal valued maxima

modules/rapid/src/rapid.cpp
```diff
@@ -0,0 +1,348 @@
...
+    std::vector<std::vector<Point>> contours;
+    findContours(img, contours, RETR_EXTERNAL, CHAIN_APPROX_NONE);
+    CV_Assert(!contours.empty());
+
+    Contour3DSampler sampler(pts2d, pts3d.getMat(), contours[0], imsize);
+    Rect valid_roi(Point(len, len), imsize - Size(2 * len, 2 * len));
+    sampleControlPoints(num, sampler, valid_roi, ctl2d, ctl3d);
+}
+
+void extractLineBundle(int len, InputArray ctl2d, InputArray frame, OutputArray bundle,
```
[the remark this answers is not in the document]
fegorsch: You only renamed the argument in the header.

paroj:
@alalek this one is ready from my side

paroj:
replaced the edge sampling reference by an older one

PR #2362, opened 2019-11-26
Fix cudacodec::VideoReader::nextFrame python bindings

cudawarped:
### This pullrequest fixes nextFrame python bindings

Makes the frame argument to cudacodec::VideoReader::nextFrame() optional when called from python.

```
force_builders=Custom
buildworker:Custom=linux-4
```


mshabunin:
Just note: this PR depends on opencv/opencv#15957

PR #2365, opened 2019-11-27
Do not use VLA in dynafu.cpp

malfet:
VLA is part of C99, but still not part of C++ standard
Replace float variable-length-arrays with unique_ptr arrays (which are
part of C++14)

This makes code compilable by MSVC


paroj:
I think the proper fix would be to move e.g.
`Mat depthData(params.frameSize.height, params.frameSize.width, CV_32F);` on top and then use
`depthData.ptr()` in `glReadPixels`

malfet:
Thank you for the suggestion. Updated commit accordingly.

PR #2383, opened 2019-12-11
Add LOGOS features matching strategy

catree:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/690

### This pullrequest changes
Original code: [LOGOS](https://github.com/short-circuitt/LOGOS)

Regression tests consist in comparing the output of the current implementation w.r.t. the original Matlab code, see opencv/opencv_extra#690

Note:
- pointers are used as in the original code. I did not manage to use `cv::Ptr` successfully (got some memory leak issues)
- returned distance for the matches will be zero
- LOGOS parameters cannot be set
- parallelization could be used
- main changes should be code convention and float type instead of double

---

This method requires to build a BoW for a considered image dataset. Something like:

```
        std::string pattern = "Oxford City Centre/Images/*.jpg";
        vector<String> results;
        glob(pattern, results);

        Ptr<Feature2D> surf = SURF::create();
        vector<KeyPoint> bowKeypoints;
        Mat bowDescriptors;
        for (size_t i = 0; i < results.size(); i++)
        {
            Mat img = imread(results[i]);

            std::vector<KeyPoint> keypoints;
            Mat descriptors;
            surf->detectAndCompute(img, noArray(), keypoints, descriptors);

            bowKeypoints.insert(bowKeypoints.end(), keypoints.begin(), keypoints.end());
            bowDescriptors.push_back(descriptors);
        }

        int clusterCount = 512;
        BOWKMeansTrainer bowTrainer(clusterCount);
        Mat bowCenters = bowTrainer.cluster(bowDescriptors);

        FileStorage file("Oxford_BOW_512.yml", FileStorage::WRITE);
        file << "matName" << bowCenters;
    }
```

Cosine distance can be used to get the nearest BoW cluster for each descriptor.

### DIFF
--- a/modules/xfeatures2d/doc/xfeatures2d.bib
+++ b/modules/xfeatures2d/doc/xfeatures2d.bib
@@ -71,6 +71,15 @@
   publisher = {Kluwer Academic Publishers}
 }
 
+@article{Lowry2018LOGOSLG,
+  title = {LOGOS: Local Geometric Support for High-Outlier Spatial Verification},
+  author = {Stephanie Lowry and Henrik Andreasson},
+  journal = {2018 IEEE International Conference on Robotics and Automation (ICRA)},
+  year = {2018},
+  pages = {7262-7269},
+  doi = {10.1109/ICRA.2018.8460988},
+}
+
 @article{Mikolajczyk2004,
   title = {Scale \& affine invariant interest point detectors},
   author = {Mikolajczyk, Krystian and Schmid, Cordelia},
--- a/modules/xfeatures2d/src/logos.cpp
+++ b/modules/xfeatures2d/src/logos.cpp
@@ -1,0 +1,69 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html
+#include "precomp.hpp"
+#include "logos/Logos.hpp"
+
+namespace cv
+{
+namespace xfeatures2d
+{
+void matchLOGOS(const std::vector<KeyPoint>& keypoints1, const std::vector<KeyPoint>& keypoints2,
+                const std::vector<int>& nn1, const std::vector<int>& nn2, std::vector<DMatch>& matches1to2)
+{
+    CV_CheckEQ(keypoints1.size(), nn1.size(), "Number of keypoints1 must be equal to the number of nn1.");
+    CV_CheckEQ(keypoints2.size(), nn2.size(), "Number of keypoints2 must be equal to the number of nn2.");
+    if (keypoints1.empty() || keypoints2.empty())
+    {
+        return;
+    }
+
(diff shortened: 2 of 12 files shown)


modules/xfeatures2d/test/test_logos_matcher.cpp
```diff
@@ -0,0 +1,167 @@
...
+    virtual void run(int);
+};
+
+CV_LogosMatcherTest::CV_LogosMatcherTest() {}
+
+CV_LogosMatcherTest::~CV_LogosMatcherTest() {}
+
+void CV_LogosMatcherTest::run( int )
+{
+    const std::string vP1_path = ts->get_data_path() + "detectors_descriptors_evaluation/matching/LOGOS/vP1.txt";
```
alalek: Please avoid using of legacy "cvtest::BaseTest". It is not needed after migration on GoogleTest. Just write:

```
TEST(XFeatures2d_LogosMatcher, logos_matcher_regression)
{
    const std::string vP1_path ...
    ...
}
```

> ts->get_data_path() + "..."

Please use `cvtest::findDataFile(...)` instead.

modules/xfeatures2d/test/test_logos_matcher.cpp
```diff
@@ -0,0 +1,167 @@
...
+            {
+                correctMatches++;
+                break;
+            }
+        }
+    }
+
+    std::cout << "groundTruth: " << groundTruth.size() << " ; matchesLogos: " << matchesLogos.size()
+              << " ; correctMatches: " << correctMatches << std::endl;
+    ASSERT_TRUE(correctMatches == static_cast<int>(groundTruth.size()));
```
alalek:
```
ASSERT_EQ(static_cast<int>(groundTruth.size()), correctMatches)
    << "groundTruth: " << groundTruth.size()
    << " ; matchesLogos: " << matchesLogos.size()
    << " ; correctMatches: " << correctMatches
;
```

modules/xfeatures2d/doc/xfeatures2d.bib
```diff
@@ -1,89 +1,119 @@
+@incollection{Agrawal08,
```
alalek: Could you please backport sorting onto 3.4 branch too?
With some note which order should be preserved (like "cite id" string)

modules/xfeatures2d/src/logos.cpp
```diff
@@ -0,0 +1,80 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html
+/*
+ * MIT License
+ *
+ * Copyright (c) 2018 Stephanie Lowry
```
alalek: Is this file code covered by 3rdparty license too? Is where some adopted code or it has been written from scratch ?

catree: I have removed the license.

PR #2398, opened 2019-12-29
Fix order of createTonemapDurand arguments

shimat:
### This pullrequest changes
The order of the arguments of `createTonemapDurand` is different between hpp and cpp. 
The test added by this pull request fails on the master branch.

I fixed it according to the order of `cv::bilateralFilter` arguments. (sigma_color first) 

## hpp
https://github.com/opencv/opencv_contrib/blob/master/modules/xphoto/include/opencv2/xphoto/tonemap.hpp#L53
```cpp
CV_EXPORTS_W Ptr<TonemapDurand>
createTonemapDurand(float gamma = 1.0f, float contrast = 4.0f, float saturation = 1.0f, float sigma_space = 2.0f, float sigma_color = 2.0f);
```

## cpp
https://github.com/opencv/opencv_contrib/blob/master/modules/xphoto/src/tonemap.cpp#L116
```cpp
Ptr<TonemapDurand> createTonemapDurand(float gamma, float contrast, float saturation, float sigma_color, float sigma_space)
```


alalek:
Patch is rebased onto [3.4 branch](https://github.com/opencv/opencv/wiki/Branches).

PR #2399, opened 2019-12-31
new module: intensity_transform

elim-dk:
The new module, intensity_transform, implements the following intensity transformation algorithms: gamma correction, log transformation, autoscaling, and contrast stretching.

### This pullrequest changes

Adds a new module called intensity_transform.


modules/intensity_transform/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+set(the_description "Intensity transformations")
+ocv_define_module(intensity_transform opencv_core opencv_imgcodecs WRAP python)
```
alalek: > imgcodecs

Looks like this dependency is not used in module itself ("samples" has own extra dependencies list).
Please remove this from here and `precomp.hpp`

modules/intensity_transform/include/opencv2/intensity_transform.hpp
```diff
@@ -0,0 +1,78 @@
...
+//! @{
+
+/**
+ * @brief Given an input bgr or grayscale image and constant c, apply log transformation to the image
+ * and return the resulting image.
+ *
+ * @param input input bgr or grayscale image.
+ * @param output resulting image of log transformations.
+*/
+CV_EXPORTS_W void logTransform(const cv::Mat input, cv::Mat& output);
```
alalek: [code style] Please avoid using of "cv::" in OpenCV files for `Mat`.

modules/intensity_transform/src/intensity_transform.cpp
```diff
@@ -0,0 +1,76 @@
...
+void autoscaling(const Mat input, Mat& output)
+{
+    double minVal, maxVal;
+    minMaxLoc(input, &minVal, &maxVal, NULL, NULL);
+    output = 255 * (input - minVal) / (maxVal - minVal);
+}
+
+void contrastStretching(const Mat input, Mat& output, const int r1, const int s1, const int r2, const int s2)
+{
+    vector<int> table(256);
```
alalek: `std::array` ?

modules/intensity_transform/test/test_intensity_transform.cpp
```diff
@@ -0,0 +1,199 @@
...
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#include "test_precomp.hpp"
+
+namespace opencv_test { namespace {
+
+TEST(intensity_transform_logTransform, accuracy)
+{
+    cv::Mat const image = (cv::Mat_<uchar>(10, 10) <<
```
alalek: Could you please avoid using of this initializer in OpenCV? Consider using RAW array instead.
see https://github.com/opencv/opencv/pull/15959#discussion_r349847572

modules/intensity_transform/test/test_intensity_transform.cpp
```diff
@@ -0,0 +1,199 @@
...
+                                 221, 250, 220, 217, 229, 236, 236, 241, 146, 216,
+                                   0, 240, 146, 174, 216, 244, 121, 237, 254, 250,
+                                 215, 188, 198, 255, 219, 233, 246, 232, 231, 210,
+                                 148, 253, 226, 215, 242, 218, 196, 202, 244, 101);
+
+    for(int i = 0; i < res.cols; ++i)
+    {
+        for(int j = 0; j < res.rows; ++j)
+        {
+            EXPECT_EQ((int)res.at<uint8_t>(i,j), (int)expectedRes.at<uint8_t>(i,j));
```
alalek: loops =>
```
EXPECT_EQ(0, cvtest::norm(res, expectedRes, NORM_INF));
```

Please note that this check is very strong (and should be avoided due non bit-exact floating-point implementation).
It is better to relax conditions:
```
EXPECT_LE(cvtest::norm(res, expectedRes, NORM_INF), 1);
EXPECT_LE(cvtest::norm(res, expectedRes, NORM_L1), 10);
```

modules/intensity_transform/src/intensity_transform.cpp
elim-dk: Thank you for your comment. Since cv::LUT method takes in cv::_InputArray as the argument, std::array didn't work here.

modules/intensity_transform/src/intensity_transform.cpp
```diff
@@ -0,0 +1,76 @@
...
+    }
+    else
+    {
+        output = input + Scalar(1, 1, 1);
+    }
+
+    output.convertTo(output, CV_64F);
+    cv::log(output, output);
+    output *= c;
+    output.convertTo(output, CV_8UC3);
```
alalek: `Mat::convertTo()` with `alpha/beta` parameters can reduce number of operations:

```
Mat in_plus_1_64f;
input.convertTo(in_plus_1_64f, CV_64F, 1, 1.0f);  // input + 1
Mat log_64f;
log(in_plus_1_64f, log_64f);
log_64f.convertTo(output, CV_8UC3, c, 0.0f);  // log_64f * c
```

modules/intensity_transform/src/intensity_transform.cpp
alalek: It should work, related code is [here](https://github.com/opencv/opencv/blame/4.0.0/modules/core/include/opencv2/core/mat.hpp#L207).

BTW, Use `std::array<uchar, 256> table;` and remove final `.convertTo()`.

modules/intensity_transform/src/intensity_transform.cpp
elim-dk: That's useful! Thank you for the feedback!

modules/intensity_transform/src/intensity_transform.cpp
elim-dk: Thanks for the document! But, when I use `std::array<uchar, 256> table;`, I get the following error:
`candidate function not viable: no known conversion from
      'std::array<uchar, 256>' (aka 'array<unsigned char, 256>') to
      'const cv::_InputArray' for 2nd argument
CV_EXPORTS_W void LUT(InputArray src, InputArray lut, OutputArray dst);`
Any idea why?

alalek: Which compiler is used?
Which version of code from main "opencv" repository is used?

Could you try `LUT(..., InputArray(table), ...)`?

alalek: This commit works well (pass CI checks, see #2366): https://github.com/opencv/opencv_contrib/commit/b05b32850c924e3465acfcb2d637dd2765928720

elim-dk: Great, thank you! Made the change.

elim-dk:
> Well done! Thank you 👍

Thank you for all your constructive comments and help!

PR #2410, opened 2020-01-16
Fixed KinFu getCloud() at empty volume

### LINKED ISSUE #2404: Kinect Fusion crash when cover sensor
- OpenCV => 4.1.2
- Operating System / Platform => Windows 10
- Compiler => Visual Studio 2019
- Depth sensor => Azure Kinect

##### Detailed description
If you cover the sensor (to trigger a reset of the tracking) the rgbd/kinfu sample crashes.

This is unexpected behavior. Covering the sensor to reset the tracking is an established action, as seen in demos of KinectFusion elsewhere.

##### Steps to reproduce
- start the kinfu sample
- cover the sensor

The sample stops and prints the error:
```
[ERROR:0] global H:\OpenCV-4.1.2\sources\modules\core\src\ocl.cpp (3039)
cv::ocl::Kernel::set OpenCL: Kernel(fillPtsNrm)::set(arg_index=14, flags=260):
can't create cl_mem handle for passed UMat buffer (addr=0000004D17DAE480)
```

savuor:
resolves #2404 

### This pullrequest changes

OCL version of TSDF's `fetchPointsNormals()` fixed for the case when there's no data in the volume

### DIFF
--- a/modules/rgbd/src/tsdf.cpp
+++ b/modules/rgbd/src/tsdf.cpp
@@ -1422,12 +1422,6 @@ void TSDFVolumeGPU::fetchPointsNormals(OutputArray points, OutputArray normals) 
 
         // 2. fill output arrays according to per-group points count
 
-        ocl::Kernel kfill;
-        kfill.create("fillPtsNrm", source, options, &errorStr);
-
-        if(kfill.empty())
-            throw std::runtime_error("Failed to create kernel: " + errorStr);
-
         points.create(gpuSum, 1, POINT_TYPE);
         UMat pts = points.getUMat();
         UMat nrm;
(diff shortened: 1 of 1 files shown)


savuor:
👍

modules/rgbd/src/tsdf.cpp
```diff
@@ -1438,31 +1432,41 @@ void TSDFVolumeGPU::fetchPointsNormals(OutputArray points, OutputArray normals)
...
-        if(!kfill.run(3, globalSize, localSize, true))
-            throw std::runtime_error("Failed to run kernel");
+        if (gpuSum)
+        {
+            ocl::Kernel kfill;
+            kfill.create("fillPtsNrm", source, options, &errorStr);
+
+            UMat atomicCtr(1, 1, CV_32S, Scalar(0));
+
+            if(kfill.empty())
```
alalek: > kfill.empty()

Put check right after `.create()`call.

UMat creation between them looks strange.

modules/rgbd/src/tsdf.cpp
```diff
@@ -1438,31 +1432,41 @@ void TSDFVolumeGPU::fetchPointsNormals(OutputArray points, OutputArray normals)
...
+                throw std::runtime_error("Failed to create kernel: " + errorStr);
+
+            // mem size to keep pts (and normals optionally) for all work-items in a group
+            lsz = localSize[0]*localSize[1]*localSize[2]*elemSize;
```
alalek: It is better to create new one "local" variable instead of reusing that.

savuor: both fixed

PR #2417, opened 2020-02-03
Add information about BUILD_LIST variable

jgbradley1:
BUILD_LIST can be used as an alternative to BUILD_opencv_* options to customize opencv build

### This pullrequest changes

* updates the README doc only
* adds information about the `BUILD_LIST` cmake variable for people who are interested in inclusive builds (specifying only the modules they want to build) as opposed to exclusive builds (disabling modules individually).

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,12 +23,16 @@ $ cmake -DOPENCV_EXTRA_MODULES_PATH=<opencv_contrib>/modules <opencv_source_dire
 $ make -j5
 ```
 
-As the result, OpenCV will be built in the `<opencv_build_directory>` with all
-modules from `opencv_contrib` repository. If you don't want all of the modules,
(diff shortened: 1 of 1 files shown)


asmorkalov:
The option is OpenCV internal option and is not designed to used by users. BUILD_LIST breaks bindings generation and introduced other issues. The feature is discussed in https://github.com/opencv/opencv/pull/9893.

mshabunin:
I would not agree, `BUILD_LIST` is not an internal option but in the opposite it was designed to be user-friendly. To enable bindings generation corresponding module should be explicitly listed, e.g. `python3`. Some configurations may not work due to hidden inter-module dependencies or other caveats, but in majority of cases it works just fine.

README file in _opencv_contrib_ is definitely not the place for this option documentation. It should be described in cmake options reference, but we don't have such document yet.

PR #2421, opened 2020-02-06
NVOF-Cuda Bug Fix. Should not reset runtime context

vchiluka5:
m_cuContext is the CUDA runtime context which we fetch and use and we should not be destroying it by calling cudaDeviceReset(). There are many applications which pipeline NVIDIA-Optical-Flow as an intermediate block and say if any block after it uses m_cuContext then that blocks will fail because the context they were using was destroyed without their knowledge.

<cut/>

```
force_builders=Custom
buildworker:Custom=linux-4
```

### DIFF
--- a/modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
+++ b/modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
@@ -578,11 +578,6 @@ void NvidiaOpticalFlowImpl::collectGarbage()
     {
         NVOF_API_CALL(GetAPI()->nvOFDestroy(m_hOF));
     }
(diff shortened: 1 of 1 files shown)


vchiluka5:
@alalek . 
If review and all checks are passing then can we please merge this PR?
Also can we also patch this on opencv 4.1.2/4.2.0 release?

alalek:
> this on opencv 4.1.2/4.2.0 release

We don't amend already released code.
Patch will be available with next release (probably 4.3.0).

You can distribute hotfix using link on patch of this PR: https://github.com/opencv/opencv_contrib/pull/2421.diff

PR #2443, opened 2020-03-05
optflow(ocl): avoid rebuilding kernels

jansol:
Rebuilding the same static kernel every time a function is called is not
exactly idiomatic in OpenCL, and unless the CL implementation has a fast
cache setup for kernels it introduces a lot of overhead. This moves
building of the TVL1 optical flow kernels into the tvl1flow object's
constructor, allowing them to be reused within that object.

NVIDIA drivers appear to have such a cache (possibly shared with CUDA), not sure about Intel and AMD. Huge slowdowns appear with our WIP [pocl](http://www.portablecl.org/) backend though. Reusing the programs yields a significant speedup to TVL1Flow computation even though this patch is not exhaustive (there are still kernels being recompiled in core opencv).

### DIFF
--- a/modules/optflow/src/tvl1flow.cpp
+++ b/modules/optflow/src/tvl1flow.cpp
@@ -96,6 +96,9 @@ public:
         outerIterations(outerIterations_), useInitialFlow(useInitialFlow_),
         scaleStep(scaleStep_), medianFiltering(medianFiltering_)
     {
+#ifdef HAVE_OPENCL
+        kernelsOk = createOpenCLKernels();
+#endif
     }
     OpticalFlowDual_TVL1();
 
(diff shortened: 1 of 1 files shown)


alalek:
OpenCV itself has:
- OpenCL programs in-memory cache
- OpenCL programs storage-based cache

Are you sure that they don't work for your case?

Try to investigate with enabled `CV_OPENCL_TRACE_CHECK` flag:

https://github.com/opencv/opencv/blob/90a4d67e8def9250a2c5250d3f439d25b5e164e2/modules/core/src/ocl.cpp#L76

jansol:
Hmm... According to the trace logs programs are indeed only compiled once, and I can't seem to reproduce the slowdowns any more either. Will reopen this if I find something relevant.

PR #2452, opened 2020-03-08
Allows structured_light pipeline to be run from Python

r2d3:
SinusoidalPattern::unwrapPhaseMap now takes an InputArray instead of InputArrayOfArrays to correct a Python binding problem

present a scriptable HistogramPhaseUnwrapping::create

This patch solves issue #1704 and allow to write a whole pipeline including reliability computation.

### DIFF
--- a/modules/phase_unwrapping/include/opencv2/phase_unwrapping/histogramphaseunwrapping.hpp
+++ b/modules/phase_unwrapping/include/opencv2/phase_unwrapping/histogramphaseunwrapping.hpp
@@ -75,20 +75,21 @@ public:
      * @param nbrOfSmallBins Number of bins between 0 and "histThresh". Default value is 10.
      * @param nbrOfLargeBins Number of bins between "histThresh" and 32*pi*pi (highest edge reliability value). Default value is 5.
      */
-    struct CV_EXPORTS Params
+    struct CV_EXPORTS_W_SIMPLE Params
     {
-        Params();
-        int width;
-        int height;
-        float histThresh;
-        int nbrOfSmallBins;
-        int nbrOfLargeBins;
+        CV_WRAP Params();
+        CV_PROP_RW int width;
+        CV_PROP_RW int height;
+        CV_PROP_RW float histThresh;
+        CV_PROP_RW int nbrOfSmallBins;
+        CV_PROP_RW int nbrOfLargeBins;
     };
     /**
      * @brief Constructor
 
      * @param parameters HistogramPhaseUnwrapping parameters HistogramPhaseUnwrapping::Params: width,height of the phase map and histogram characteristics.
      */
+    CV_WRAP
     static Ptr<HistogramPhaseUnwrapping> create( const HistogramPhaseUnwrapping::Params &parameters =
                                                  HistogramPhaseUnwrapping::Params() );
 
--- a/modules/phase_unwrapping/misc/python/pyopencv_phase_unwrapping.hpp
+++ b/modules/phase_unwrapping/misc/python/pyopencv_phase_unwrapping.hpp
@@ -1,0 +1,3 @@
+#ifdef HAVE_OPENCV_PHASE_UNWRAPPING
+typedef cv::phase_unwrapping::HistogramPhaseUnwrapping::Params HistogramPhaseUnwrapping_Params;
+#endif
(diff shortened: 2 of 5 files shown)


modules/phase_unwrapping/include/opencv2/phase_unwrapping/histogramphaseunwrapping.hpp
```diff
@@ -92,6 +92,11 @@ class CV_EXPORTS_W HistogramPhaseUnwrapping : public PhaseUnwrapping
     static Ptr<HistogramPhaseUnwrapping> create( const HistogramPhaseUnwrapping::Params &parameters =
                                                  HistogramPhaseUnwrapping::Params() );
 
+    // alias for scripting
+    CV_WRAP
+    static Ptr<HistogramPhaseUnwrapping> create( int width, int height, float histThresh = static_cast<float>(3 * CV_PI * CV_PI),
+                                                 int nbrOfSmallBins = 10, int nbrOfLargeBins = 5);
```
alalek: Please follow the existed practice and avoid unpacking of "Params" structure.
You can check this code as an example:

https://github.com/opencv/opencv/blob/850414a501d5e6dba111c92d73fdd05794c7061d/modules/features2d/include/opencv2/features2d.hpp#L653-L654

r2d3: The existing practice is not clear as the structured_light itself is using 2 practices:
- Params unpacking in `GrayCodePattern::create`
- smart pointer for `SinusoidalPattern::create`

Some codes pass a smart pointer like this is done for `SinusoidalPattern::create`

https://github.com/opencv/opencv_contrib/blob/master/modules/structured_light/include/opencv2/structured_light/sinusoidalpattern.hpp#L100

 but I followed the solution used by `GrayCodePattern::create`

https://github.com/opencv/opencv_contrib/blob/master/modules/structured_light/include/opencv2/structured_light/graycodepattern.hpp#L89

Because changing the create API (instead of overloading it) will break current codes using structured_light.

The code you point out is not using Params struct to pass parameters.

If you just pass a `const Params&`, the Python binding generator will not generate the need code. Am I right ?

alalek: "Params" class should be marked for bindings wrapping too (use `CV_EXPORTS_W_SIMPLE`)

r2d3: OK. I managed to bind the `HistogramPhaseUnwrapping::Params`; What I was missing is the `typedef` in `pyopencv_phase_unwrapping.hpp` as it is done in `pyopencv_features2d.hpp`

I added the Python test for structured_light replicating what is done in C++. However, this shown a "bug" in `HistogramPhaseUnwrapping::unwrapPhaseMap` where the `unwrappedPhaseMap` were uninitialized in places outside of the mask.
It is not clear if this were intentional or not.

hitmanman:
Hi the following codings make my python kernel die. Can you tell me how to correct it?
thank you
---------------------------------------------
import cv2
import numpy as np
import math
import sys

cam_size = [-1,-1]

params = cv2.structured_light_SinusoidalPattern_Params()
params.width = 1024 #pixel width
params.height = 1024 #pixel height
params.nbrOfPeriods = 20 # amount of waves
params.setMarkers = False # patterns /w or /wo markers
params.horizontal = False
params.methodId = cv2.structured_light.FTP
#params.methodId = cv2.structured_light.PSP
#params.methodId = cv2.structured_light.FAPS
params.shiftValue = 2.0 * math.pi / 3.0
params.nbrOfPixelsBetweenMarkers = 70

sinus = cv2.structured_light_SinusoidalPattern.create(params)
patterns = sinus.generate()
count = 0

img = []
while count < len(patterns[1]) :
    img.append(patterns[1][count])
    count += 1

captures = []
captures.append(img[0])
captures.append(img[1])
captures.append(img[2])

wrapped_phase_map, shadow_mask = sinus.computePhaseMap(captures)
cam_size = (img[0].shape[0], img[0].shape[1])
print("before")
unwrapped_phase_map=np.zeros([img[0].shape[0], img[0].shape[1]],np.uint8)
unwrapped_phase_map = sinus.unwrapPhaseMap(wrapped_phase_map, cam_size, shadow_mask)
print("after")

r2d3:
> ## Hi the following codings make my python kernel die.
> The problem is
> 'unwrapped_phase_map = sinus.unwrapPhaseMap(wrapped_phase_map, cam_size, shadow_mask)'
> Can you tell me how to correct it?

Which version of OpenCV are you using ?

~~~ python
import cv2
print(cv2.__version__)
~~~

You should have a recent 3.4 release (ideally 3.4.10) to get the correction

Regards

PR #2473, opened 2020-03-22
Minor fix for solvePnPRansac params mismatch 

akashsharma02:
Fixes issue #2064: Fairly straightforward fix
Apologize for the previous PR in the wrong branch. 
Tested compilation on Ubuntu 18.04
@alalek could you please review this

```
force_builders=linux,windows,docs,mac
```

### DIFF
--- a/modules/ccalib/src/ccalib.cpp
+++ b/modules/ccalib/src/ccalib.cpp
@@ -445,8 +445,13 @@ bool CustomPattern::findRtRANSAC(InputArray objectPoints, InputArray imagePoints
             OutputArray rvec, OutputArray tvec, bool useExtrinsicGuess, int iterationsCount,
             float reprojectionError, int minInliersCount, OutputArray inliers, int flags)
 {
+    int npoints = imagePoints.getMat().checkVector(2);
+    CV_Assert(npoints > 0);
+    double confidence_factor = (double)minInliersCount / (double)npoints;
+    double confidence = confidence_factor < 0.001 ? 0.001 : confidence_factor > 0.999 ? 0.999 : confidence_factor;
+
     solvePnPRansac(objectPoints, imagePoints, cameraMatrix, distCoeffs, rvec, tvec, useExtrinsicGuess,
-                    iterationsCount, reprojectionError, minInliersCount, inliers, flags);
+                    iterationsCount, reprojectionError, confidence, inliers, flags);
     return true; // for consistency with the other methods
 }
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -10,3 +10,4 @@ tags
 tegra/
 *.i
 .download*
+.ccls-cache
```
alalek: Please revert change in .gitignore file.

A general recommendation about IDE-specific files is using of global .gitignore approach: https://help.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

modules/ccalib/src/ccalib.cpp
```diff
@@ -445,8 +445,11 @@ bool CustomPattern::findRtRANSAC(InputArray objectPoints, InputArray imagePoints
             OutputArray rvec, OutputArray tvec, bool useExtrinsicGuess, int iterationsCount,
             float reprojectionError, int minInliersCount, OutputArray inliers, int flags)
 {
+    float confidence_factor = minInliersCount / (imagePoints.total()/2);
```
alalek: > imagePoints.total()/2

Why is half only?

---

Please use `checkVector()` instead of "2 coordinates trick".
Something like:

```
int npoints = std::max(1, objectPoints.getMat().checkVector(3));
```

.gitignore
akashsharma02: Fixed

modules/ccalib/src/ccalib.cpp
akashsharma02: Fixed

modules/ccalib/src/ccalib.cpp
```diff
@@ -445,8 +445,13 @@ bool CustomPattern::findRtRANSAC(InputArray objectPoints, InputArray imagePoints
             OutputArray rvec, OutputArray tvec, bool useExtrinsicGuess, int iterationsCount,
             float reprojectionError, int minInliersCount, OutputArray inliers, int flags)
 {
+    int npoints = imagePoints.getMat().checkVector(2);
+    CV_Assert(npoints != -1);
+    double confidence_factor = minInliersCount / npoints;
```
alalek: `double confidence_factor = minInliersCount / npoints;`

in divides "int" by "int" and then converts "zero" result to double.
This is why some tests are required...

So,

1. `CV_Assert(npoints > 0);` to avoid other "div by zero" error (it is not recoverable for integer arithmetic, also we don't go deeper with Inf/NaN values).
2. force conversion: `double confidence_factor = (double)minInliersCount / (double)npoints;`

akashsharma02: Thanks for your review. Fixed now. Should be all good finally.

akashsharma02: Btw, is there any way that I can take a look at MSVC warnings even when compiling on Linux? Currently even with the warnings as errors option on CMAKE GCC didn't catch the conversion warning.

akashsharma02:
Can we please get this merged if there are no more changes required?

PR #2480, opened 2020-03-24
Pythonwrap for the dnn_superres module

Saafke:
Hello!

This PR adds a python wrap for the dnn_superres module. This allows for using opencv via python to upscale images via deep neural networks. Before this was only possible in c++. The tutorial is updated to include an example of how to do this in python.


alalek:
Thank you for contribution!

Could you please add tests for Python bindings? Some simple checks for existing of required methods and/or processing of passed parameters (preferable without requirement for test data) and returning of non-empty results.
No need to add complex accuracy checks (it should be done in C++ only).

See `misc/python/test` sub-directories of other modules as reference (search for `test_*.py` files)

modules/dnn_superres/misc/python/test_dnn_superres.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python
```
alalek: Please add one more subdirectory:
```.diff
-modules/dnn_superres/misc/python/test_dnn_superres.py
+modules/dnn_superres/misc/python/test/test_dnn_superres.py
```

Saafke: Oh right! Done :)

onurberkay:
>  from cv2 import dnn_superres
> ImportError: cannot import name 'dnn_superres' from 'cv2' 
I am getting import error.
When building cuda support there is no problem with dnn_superres but python cant find module. I check python path, library path and opencv path. They are correct.

khvmaths:
> > from cv2 import dnn_superres
> > ImportError: cannot import name 'dnn_superres' from 'cv2'
> 
> I am getting import error.
> When building cuda support there is no problem with dnn_superres but python cant find module. I check python path, library path and opencv path. They are correct.

I presumed you installed the package through PyPI. Quoted from the PyPI:

>Option 1 - Main modules package: pip install opencv-python
>Option 2 - Full package (contains both main modules and contrib/extra modules): pip install opencv-contrib-python (check contrib/extra modules listing from OpenCV documentation)

I solved the problem by only install the opencv-contrib-python and uninstall the opencv-python.

PR #2483, opened 2020-03-28
Added python setters to the tracker resolves #12009

jshiwam:

- [ ] I agree to contribute to the project under OpenCV (BSD) License.


alalek:
Like C++ case, correct way is utilize `Params` structure in Python too instead of creating new one API.

jshiwam:
> Like C++ case, correct way is utilize `Params` structure in Python too instead of creating new one API.

At first I tried that method where I tried to wrap default Param() constructor with CV_WRAP but doing that showed following error:




Note: Class Feature2D has more than 1 base class (not supported by Python C extensions)
      Bases:  cv::Algorithm, cv::class, cv::Feature2D, cv::Algorithm
      Only the first base class will be used
Note: Class detail_GraphCutSeamFinder has more than 1 base class (not supported by Python C extensions)
      Bases:  cv::detail::GraphCutSeamFinderBase, cv::detail::SeamFinder
      Only the first base class will be used
  File "/home/shiwam/opencvfolde/opencv/modules/python/bindings/..//src2/gen2.py", line 1132, in <module>
    generator.gen(srcfiles, dstdir)
  File "/home/shiwam/opencvfolde/opencv/modules/python/bindings/..//src2/gen2.py", line 1021, in gen
    self.add_func(decl)
  File "/home/shiwam/opencvfolde/opencv/modules/python/bindings/..//src2/gen2.py", line 947, in add_func
    self.classes[classname].constructor = func
**KeyError: u'TrackerMedianFlow_Params'**
modules/python_bindings_generator/CMakeFiles/gen_opencv_python_source.dir/build.make:302: recipe for target 'modules/python_bindings_generator/pyopencv_generated_enums.h' failed
make[2]: *** [modules/python_bindings_generator/pyopencv_generated_enums.h] Error 1
CMakeFiles/Makefile2:7571: recipe for target 'modules/python_bindings_generator/CMakeFiles/gen_opencv_python_source.dir/all' failed
make[1]: *** [modules/python_bindings_generator/CMakeFiles/gen_opencv_python_source.dir/all] Error 2

I had no idea why that error occured so I tried this way. Please let me know whether wrapping that way is the right way or not. I followed [this](https://docs.opencv.org/3.4.9/da/d49/tutorial_py_bindings_basics.html) documentation in order to understand the bindings.

jshiwam:
> Like C++ case, correct way is utilize `Params` structure in Python too instead of creating new one API.

I tried exporting the params to python by this approach:
[tracker.hpp](https://gist.github.com/jshiwam/0c41f0be8b8d48bd3191d47145198638)
[trackerMedianFlow.cpp](https://gist.github.com/jshiwam/3be09e53ba1b28760b16f2fdf3ad4e62)

The build completes upto 100% but while generating pyopencv_generated_types_content.h it gives an error saying 
**Params was not declared in this scope**
Can you tell me what I am doing wrong here.

asenyaev:

PR #2516, opened 2020-04-27
LaSOT-based benchmark for trackers

ieliz:
Added Python3 script with benchmark for trackers.


modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,153 @@
...
+            if not init_once:
+                temp = tracker.init(frame, init_bb)
+                init_once = True
+
+            temp, new_bb = tracker.update(frame)
+            # Coordinates of points of bounding boxes
+            new_bb_xmin = new_bb[0] - new_bb[2] / 2
+            new_bb_xmax = new_bb[0] + new_bb[2] / 2
+            new_bb_ymin = new_bb[1] - new_bb[3] / 2
+            new_bb_ymax = new_bb[1] + new_bb[3] / 2
```
dkurt, line 116: `update` should return `x, y, w, h` where `x, y` are left top points

ieliz: I correct formulas for evaluation. Also I fix case if ground truth bounding box is (0, 0, 0, 0).

.gitignore
```diff
@@ -10,3 +10,6 @@ tags
 tegra/
 *.i
 .download*
+*.caffemodel
+*.prototxt
+build_dir/
```
dkurt, line 15: I think it's better not to include it to gitignore

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,187 @@
+import numpy as np
+import cv2 as cv
+import sys
+import glob
+import argparse
+
+parser = argparse.ArgumentParser(description="Run benchmark")
```
dkurt, line 7: Which benchmark? Please specify the algorithm (single object tracking) and name of the dataset.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,187 @@
...
+                    default="results.txt", help="Full path to file with results")
+args = parser.parse_args()
+
+list_of_videos = [None for _ in range(280)]
+file_with_video_names = open(args.path_to_dataset + "/testing_set.txt", "r")
+
+results = open(args.path_to_results, "w")
+
+for i in range(280):
+    list_of_videos[i] = file_with_video_names.readline()
```
dkurt, line 20: Can we avoid hardcoded 280?

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,187 @@
...
+
+trackers = [None for _ in range(7)]
+trackers[0] = "Boosting"
+trackers[1] = "MIL"
+trackers[2] = "KCF"
+trackers[3] = "MedianFlow"
+trackers[4] = "GOTURN"
+trackers[5] = "MOSSE"
+trackers[6] = "CSRT"
```
dkurt, line 29:
```python
trackers = [
  "Boosting",
  "MIL",
  "KCF",
  "MedianFlow",
  "GOTURN",
  "MOSSE",
  "CSRT"
]
```

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,187 @@
...
+                elif tracker_name == "KCF":
+                    tracker = cv.TrackerKCF_create()
+                elif tracker_name == "MedianFlow":
+                    tracker = cv.TrackerMedianFlow_create()
+                elif tracker_name == "GOTURN":
+                    tracker = cv.TrackerGOTURN_create()
+                elif tracker_name == "MOSSE":
+                    tracker = cv.TrackerMOSSE_create()
+                elif tracker_name == "CSRT":
+                    tracker = cv.TrackerCSRT_create()
```
dkurt, line 124: please remove code duplication. In example, define a method which returns tracker by name

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,187 @@
...
+            elif tracker_name == "KCF":
+                frames_before_reinit = 110
+            elif tracker_name == "MedianFlow":
+                frames_before_reinit = 20
+            elif tracker_name == "GOTURN":
+                frames_before_reinit = 20
+            elif tracker_name == "MOSSE":
+                frames_before_reinit = 20
+            elif tracker_name == "CSRT":
+                frames_before_reinit = 250
```
dkurt, line 105: please return also in the method which creates tracker by name

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,187 @@
...
+
+        iou_values += sum_iou / frame_counter
+        pr_values += sum_pr / frame_counter
+        norm_pr_values += sum_norm_pr / frame_counter
+
+    mean_iou = iou_values / 280
+    mean_pr = pr_values / 280
+    mean_norm_pr = norm_pr_values / 280
+
+    results.write(tracker_name + ":\n\tmean IoU = " + str(mean_iou) + "\n\tmean precision = " + str(mean_pr) + "\n\tmean normalized precision = " + str(mean_norm_pr) + "\n\n")
```
dkurt, line 187: I think that it's better to print results on the screen

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,187 @@
...
+
+            gt_bb = gt_file.readline().replace("\n", "").split(",")
+
+        iou_values += sum_iou / frame_counter
+        pr_values += sum_pr / frame_counter
+        norm_pr_values += sum_norm_pr / frame_counter
+
+    mean_iou = iou_values / 280
+    mean_pr = pr_values / 280
+    mean_norm_pr = norm_pr_values / 280
```
dkurt, line 185: Why not to implement a single metric? In example, `success`

modules/tracking/samples/tracker.py
```diff
@@ -2,38 +2,44 @@
 import cv2 as cv
 import sys
 
-if len(sys.argv) != 2:
-    print('Input video name is missing')
-    exit()
+#if len(sys.argv) != 2:
+#    print('Input video name is missing')
+#    exit()
```
dkurt, line 7: Is that necessary to change a sample?

ieliz: It was commented for more convenient experiments. For now it is not necessary to change it in final version.

modules/tracking/samples/lasot_benchmark.py
ieliz: I want to implement metrics from [LaSOT](https://arxiv.org/abs/1809.07845) and [TrackingNet](https://arxiv.org/abs/1803.10794) papers - it is common and generally accepted metrics in visual object tracking. It's implementation is not that hard, but implementation of appropriate averaging is more complex. In TrackingNet repository I found code and I want to try implement it without Pandas.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
ieliz: 280 is a number of videos in testing part of LaSOT dataset. Whole benchmark is created for working with certain dataset, so I do not think that hardcode is bad here.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

.gitignore
ieliz: It was added for more convenient experiments with trackers in opencv-contrib. I will remove it from gitignore.

modules/tracking/samples/lasot_benchmark.py
dkurt: But we can extract this value from the list of videos

ieliz: Right! I will set it as `len(list_of_videos)` .

modules/tracking/samples/lasot_benchmark.py
ieliz: Added new method for initialization and re-initialization of tracker, which also return number of frames for re-initialization.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,158 @@
...
+    elif tracker_name == "GOTURN":
+        tracker = cv.TrackerGOTURN_create()
+        frames_before_reinit = 20
+    elif tracker_name == "MOSSE":
+        tracker = cv.TrackerMOSSE_create()
+        frames_before_reinit = 20
+    elif tracker_name == "CSRT":
+        tracker = cv.TrackerCSRT_create()
+        frames_before_reinit = 250
+    return [tracker, frames_before_reinit]
```
dkurt, line 28: Let's use advantages of Python more:
```python
config = {
  "Boosting": (cv.TrackerBoosting_create(), 500),
  "MIL": (cv.TrackerMIL_create(), 1000),
...
}
return config[tracker_name]
```

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,158 @@
...
+    args = parser.parse_args()
+
+    list_of_videos = [None for _ in range(280)]
+    file_with_video_names = open(args.path_to_dataset + "/testing_set.txt", "r")
+
+    for i in range(len(list_of_videos)):
+        list_of_videos[i] = file_with_video_names.readline()
+
+    trackers = {'Boosting': 500, 'MIL': 1000, 'KCF': 110,
+                'MedianFlow': 20, 'GOTURN': 20, 'MOSSE': 20, 'CSRT': 250}
```
dkurt, line 43: Duplicated logic with number of iterations?

modules/tracking/samples/lasot_benchmark.py
dkurt: Can you please write in this PR description resulting metric for some subset? 3-4 videos.

modules/tracking/samples/lasot_benchmark.py
ieliz: Do you mean equivalent numbers for different trackers? If yes, all numbers were received via experiments with video sequence.

dkurt: I mean that there are already these numbers in the `init_tracker` method above

ieliz: I just forgot to remove dictionary `trackers` after adding `init_tracker` method, my bad.
Here is enough to use list for trackers loop, like you mentioned earlier: 
`
trackers = [
  "Boosting",
  "MIL",
  "KCF",
  "MedianFlow",
  "GOTURN",
  "MOSSE",
  "CSRT"
]
`

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
ieliz: Added values from current state of the benchmark for 4 videos in LaSOT test part.

dkurt:
@l-bat, please join to review

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,158 @@
...
+        frames_before_reinit = 20
+    elif tracker_name == "CSRT":
+        tracker = cv.TrackerCSRT_create()
+        frames_before_reinit = 250
+    return [tracker, frames_before_reinit]
+
+def main():
+    parser = argparse.ArgumentParser(description="Run LaSOT-based benchmark for visual object trackers")
+    parser.add_argument("--path_to_dataset", type=str,
+                        default="Test part of LaSOT dataset", help="Full path to LaSOT folder")
```
dkurt, line 33: Please add visualization option (like `--show`) so user would be able to see how tracker works.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,158 @@
...
+    args = parser.parse_args()
+
+    list_of_videos = [None for _ in range(280)]
+    file_with_video_names = open(args.path_to_dataset + "/testing_set.txt", "r")
```
l-bat, line 37: Please, use `os.path.join`

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,158 @@
...
+        frames_before_reinit = 250
+    return [tracker, frames_before_reinit]
+
+def main():
+    parser = argparse.ArgumentParser(description="Run LaSOT-based benchmark for visual object trackers")
+    parser.add_argument("--path_to_dataset", type=str,
+                        default="Test part of LaSOT dataset", help="Full path to LaSOT folder")
+    args = parser.parse_args()
+
+    list_of_videos = [None for _ in range(280)]
```
l-bat, line 36: I think is better to do: 
```
list_of_videos = []
for i in range(280):
        list_of_videos.append(file_with_video_names.readline())
```

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,158 @@
...
+    file_with_video_names = open(args.path_to_dataset + "/testing_set.txt", "r")
+
+    for i in range(len(list_of_videos)):
+        list_of_videos[i] = file_with_video_names.readline()
+
+    trackers = {'Boosting': 500, 'MIL': 1000, 'KCF': 110,
+                'MedianFlow': 20, 'GOTURN': 20, 'MOSSE': 20, 'CSRT': 250}
+
+    #For every tracker
+    for tracker_id in range(len(list(trackers.keys()))):
```
l-bat, line 46: I think is better to do:
```
for tracker_id, tracker_name in enumerate(trackers):
```

modules/tracking/samples/lasot_benchmark.py
l-bat, line 33: Invalid default value for argument

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,158 @@
...
+            print("Video: " + str(video_name))
+
+            video_name = video_name.replace("\n", "")
+            gt_file = open(
+                args.path_to_dataset + video_name + "/groundtruth.txt", "r")
+            gt_bb = gt_file.readline().replace("\n", "").split(",")
+            init_bb = gt_bb
+
+            for _ in range(len(init_bb)):
+                init_bb[_] = float(init_bb[_])
```
l-bat, line 66: Please, use i or another name to iterator. _ is used if iter is not used in the loop

modules/tracking/samples/lasot_benchmark.py
l-bat: Or 
```
for key, value in trackers:
```

modules/tracking/samples/lasot_benchmark.py
l-bat: In this case better: `init_bb = [float(b) for b in init_bb]`

modules/tracking/samples/lasot_benchmark.py
ieliz: Data set include 280 videos, all of them used for 7 trackers, are you sure that visualization is necessary?

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,158 @@
...
+                    normalized_precision_value = 0.0
+                else:
+                    normalized_precision_value = 1.0
+                sum_norm_pr += normalized_precision_value
+
+                gt_bb = gt_file.readline().replace("\n", "").split(",")
+
+            iou_values += sum_iou / frame_counter
+            pr_values += sum_pr / frame_counter
+            norm_pr_values += sum_norm_pr / frame_counter
```
l-bat, line 149: Please, init iou_values, pr_values and norm_pr_values before loop
`local variable 'iou_values' referenced before assignment`

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,158 @@
...
+
+                precision = np.sqrt((new_cx - gt_cx) * (new_cx - gt_cx) + (new_cy - gt_cy) * (new_cy - gt_cy))
+                if precision > 20.0:
+                    precision_value = 0.0
+                else:
+                    precision_value = 1.0
+                sum_pr += precision_value
+
+                normalized_precision = np.sqrt((new_cx - gt_cx) / gt_bb[2] * (new_cx - gt_cx) / gt_bb[2] + (
+                    new_cy - gt_cy) / gt_bb[3] * (new_cy - gt_cy) / gt_bb[3])
```
l-bat, line 138: Please add a zero division check

modules/tracking/samples/lasot_benchmark.py
ieliz: Fixed: as a default value used name of original dataset folder.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
ieliz: Still don`t know how to fix problem with averaging, please write me if you have any suggestions.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,179 @@
...
+            pr_values = 0
+            norm_pr_values = 0
+            frame_counter = len(video_sequence)
+
+            # For every frame in video
+            for number_of_the_frame, image in enumerate(video_sequence):
+
+                for i in range(len(gt_bb)):
+                    gt_bb[i] = float(gt_bb[i])
+                gt_bb = tuple(gt_bb)
```
l-bat, line 82: Why we need tuple?
`init_gt = tuple([float(gt) for b in gt_bb])`

ieliz: Tracker require tuple as input.

ieliz: Trackers requires tuple for input.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,179 @@
...
+
+        # Evaluating mean value for dataset
+        mean_iou = iou_values / len(list_of_videos)
+        mean_pr = pr_values / len(list_of_videos)
+        mean_norm_pr = norm_pr_values / len(list_of_videos)
+
+        print(tracker_name + ":\n\tmean IoU = " + str(
+            mean_iou) + "\n\tmean precision = " + str(
+                mean_pr) + "\n\tmean normalized precision = " + str(
+                    mean_norm_pr) + "\n\n")
```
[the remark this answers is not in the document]
ieliz: Like this:
`
results_for_print = '{}{}{}{}{}{}{}{}'.format(tracker_name, ":\n\tMean IoU = ", str(mean_iou), "\n\tMean precision = ", str(mean_pr), "\n\tMean normalized precision = ", str(mean_norm_pr), "\n\n")
`
?

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,179 @@
...
+        # Loop for every video
+        for video_name in list_of_videos:
+
+            tracker, frames_before_reinit = init_tracker(tracker_name)
+            init_once = False
+
+            print("Video: " + str(video_name))
+
+            # Open specific video and read ground truth for it
+            video_name = video_name.replace("\n", "")
```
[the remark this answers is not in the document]
ieliz: Seems like we really do not need that line anymore.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,179 @@
...
+    parser = argparse.ArgumentParser(
+        description="Run LaSOT-based benchmark for visual object trackers")
+    # As a default argument used name of 
+    # original dataset folder
+    parser.add_argument("--path_to_dataset", type=str,
+                        default="LaSOTTesting", help="Full path to LaSOT")
+    args = parser.parse_args()
+
+    # Creating list with names of videos via reading names from txt file
+    list_of_videos = [None for _ in range(280)]
```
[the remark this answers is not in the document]
ieliz: I guess no.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,179 @@
...
+    video_names = os.path.join(args.path_to_dataset, "/testing_set.txt")
```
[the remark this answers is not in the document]
ieliz: I forgot properties of this, thank you for reminder!

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,179 @@
...
+
+            tracker, frames_before_reinit = init_tracker(tracker_name)
+            init_once = False
+
+            print("Video: " + str(video_name))
+
+            # Open specific video and read ground truth for it
+            video_name = video_name.replace("\n", "")
+            gt_file = open(
+                args.path_to_dataset + video_name + "/groundtruth.txt", "r")
```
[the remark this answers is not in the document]
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,179 @@
...
+            gt_bb = gt_file.readline().replace("\n", "").split(",")
+            init_bb = gt_bb
+            init_bb = tuple([float(b) for b in init_bb])
+
+            print("Initial bounding box: ", init_bb)
+
+            # Creating blob from image sequence
+            video_sequence = sorted(glob.glob(str(args.path_to_dataset + str(
+                video_name) + "/img/*.jpg")))
```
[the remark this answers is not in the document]
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,179 @@
...
+                        new_bb_ymax - new_bb_ymin) + (gt_xmax - gt_xmin) * (
+                            gt_ymax - gt_ymin) - area_of_overlap
+                    # Zero division check
+                    if area_of_union != 0:
+                        iou = area_of_overlap / area_of_union
+                        sum_iou += iou
+
+                    # Precision
+                    precision = np.sqrt((new_cx - gt_cx) * (new_cx - gt_cx) + (
+                        new_cy - gt_cy) * (new_cy - gt_cy))
```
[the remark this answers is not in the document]
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,179 @@
...
+                    new_bb_ymin = new_bb[1]
+                    new_bb_ymax = new_bb[1] + new_bb[3] - 1.0
+                    gt_xmin = gt_bb[0]
+                    gt_xmax = gt_bb[0] + gt_bb[2] - 1.0
+                    gt_ymin = gt_bb[1]
+                    gt_ymax = gt_bb[1] + gt_bb[3] - 1.0
+                    new_cx = new_bb[0] + (new_bb[2] + 1.0) / 2
+                    new_cy = new_bb[1] + (new_bb[3] + 1.0) / 2
+                    gt_cx = gt_bb[0] + (gt_bb[2] + 1.0) / 2
+                    gt_cy = gt_bb[1] + (gt_bb[3] + 1.0) / 2
```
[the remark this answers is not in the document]
ieliz: I adding function, but do not rewriting calculations for it yet.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,179 @@
...
+                        precision_value = 1.0
+                    sum_pr += precision_value
+
+                    # Normalized precision
+                    # Zero division check
+                    if gt_bb[2] != 0 and gt_bb[3] != 0:
+                        normalized_precision = np.sqrt((new_cx - gt_cx) / \
+                            gt_bb[2] * (new_cx - gt_cx) / gt_bb[2] + \
+                                (new_cy - gt_cy) / gt_bb[3] * \
+                                    (new_cy - gt_cy) / gt_bb[3])
```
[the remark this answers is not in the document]
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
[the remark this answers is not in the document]
dkurt: Why not to use just multiple prints?
```python
print('Tracker:', tracker_name)
print('mean IoU =', mean_iou)
...
```

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,188 @@
+import numpy as np
+import cv2 as cv
+import glob
```
l-bat, line 3: `glob` is no longer used

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,188 @@
...
+    with open(video_names, 'rt') as f:
+        list_of_videos = f.read().rstrip('\n').split('\n')
+
+    trackers = [
+        'Boosting', 'MIL', 'KCF', 'MedianFlow', 'GOTURN', 'MOSSE', 'CSRT']
+
+    # Loop for every tracker
+    for tracker_id in range(len(trackers)):
+
+        tracker_name = trackers[tracker_id]
```
l-bat, line 53: Why not 

`for tracker_name in trackers:`

modules/tracking/samples/lasot_benchmark.py
dkurt: Yes, it's for debugging only. By default, should be `false`

modules/tracking/samples/lasot_benchmark.py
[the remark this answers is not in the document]
ieliz: It does not work with 
`
video_sequence = sorted(os.listdir(os.path.join(args.path_to_dataset, video_name, "img")))
`
So I want to return 
`
video_sequence = sorted(glob.glob(str(args.path_to_dataset + str(video_name) + "/img/*.jpg")))
`
Reason: `error: (-215) size.width>0 && size.height>0 in function cv::imshow`

modules/tracking/samples/lasot_benchmark.py
ieliz: For now I am using it to create glob of frames.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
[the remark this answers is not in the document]
ieliz: For now I am using multiple prints.

modules/tracking/samples/lasot_benchmark.py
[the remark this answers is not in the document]
ieliz: Function is used, code duplication removed.

modules/tracking/samples/lasot_benchmark.py
ieliz: Added flag for visualization.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,193 @@
...
+            print(gt_bb)
+            init_bb = gt_bb
+            init_bb = tuple([float(b) for b in init_bb])
+
+            print("Initial bounding box: ", init_bb)
+
+            # Creating blob from image sequence
+            #video_sequence = sorted(os.listdir(os.path.join(args.path_to_dataset, video_name, "img")))
+            video_sequence = sorted(glob.glob(str("D:/lasot/" + str(video_name) +
+                                                  "/img/*.jpg")))
```
l-bat, line 78: I think it's better to use `os.listdir` because we can delete `glob`. `os.listdir` gives file names from directory. And below you can use:

```python
for number_of_the_frame, image in enumerate(video_sequence):
    frame = cv.imread(os.path.join(args.path_to_dataset, video_name, image))
```

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,193 @@
...
+def calc_coords(bb):
+    xmin = bb[0]
+    xmax = bb[0] + bb[2] - 1.0
+    ymin = bb[1]
+    ymax = bb[1] + bb[3] - 1.0
+    cx = bb[0] + (bb[2] + 1.0) / 2
+    cy = bb[1] + (bb[3] + 1.0) / 2
+    coords = {'xmin': xmin, 'xmax': xmax, 'ymin': ymin, 'ymax': ymax,
+                   'cx': cx, 'cy': cy}
+    return coords
```
l-bat, line 18: I think we can simplify this:
```python
return xmin, xmax, ymin, ymax, cx, cy
```
And below you can get:
```python
new_xmin, new_xmax, new_ymin, new_ymax, new_cx, new_cy = calc_coords(new_bb)
```

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,193 @@
...
+        list_of_videos = f.read().rstrip('\n').split('\n')
+
+    trackers = [
+        'Boosting', 'MIL', 'KCF', 'MedianFlow', 'GOTURN', 'MOSSE', 'CSRT']
+
+    # Loop for every tracker
+    for tracker_id in trackers:
+
+        tracker_name = tracker_id
+        print(tracker_name)
```
l-bat, line 56: Why we need `tracker_id`?

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,193 @@
...
+        # Calculating mean arithmetic value for dataset
+        mean_iou = iou_values / len(list_of_videos)
+        mean_pr = pr_values / len(list_of_videos)
+        mean_norm_pr = norm_pr_values / len(list_of_videos)
+
+        print(tracker_name + ":")
+        print("\n\t" + str("Mean Intersection over Union = " + str(mean_iou)))
+        print("\n\t" + str("Mean Precision = ") + str(mean_pr))
+        print("\n\t" + str("Mean Normalized Precision = ") + str(mean_norm_pr))
+        print("\n\n")
```
l-bat, line 190: I think `np.mean` is better:
```python
        print(tracker_name, ":")
        print("Mean Intersection over Union =", np.mean(iou_values))
        print("Mean Precision =", np.mean(pr_values))
        print("Mean Normalized Precision =", np.mean(norm_pr_values))
```

modules/tracking/samples/lasot_benchmark.py
dkurt: Looks like centers are extra. Can calculate them in runtime?

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,193 @@
...
+                        sum_iou += iou
+
+                    # Precision
+                    precision = np.sqrt((new_cx - gt_cx) ** 2 + (
+                        new_cy - gt_cy) ** 2)
+                    if precision > 20.0:
+                        pr_value = 0.0
+                    else:
+                        pr_value = 1.0
+                    sum_pr += pr_value
```
l-bat, line 157: Maybe create separate functions for calculating IoU and precision metrics?

ieliz: Here is no code duplication, so I do not see a reason to do that.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
ieliz: For now, centers are calculated only once for every bounding box, so if we remove their calculation in method, we will be forced to calculate them twice:
1) for precision;
2) for normalized precision.

modules/tracking/samples/lasot_benchmark.py
ieliz: I am sorry, I do not understand that question.
I am using `tracker_id` in the cycle `for `for taking tracker`s name from the list.

modules/tracking/samples/lasot_benchmark.py
l-bat: For me, this makes main function more readable and we don't need comments like:
```
# Intersection over Union
# Precision
# Normalized precision
```

`iou = get_iou(new_coords, gt_coords)`

modules/tracking/samples/lasot_benchmark.py
l-bat: Why not
```
for tracker_name in trackers:
```

ieliz: Oh, now I get it :)
Thank you!

modules/tracking/samples/lasot_benchmark.py
ieliz: IMHO clear comments and variable names are better for readability of code. These calculations are not that hard. Hiding 7-8 lines of nonrepeating code in the separate method can cause more difficulties for the user. I do not understand why we should hide calculations like this.

dkurt: We need to separate the logic from implementation. That's why using methods we can show what's programmed.

ieliz: OK, but you want to leave normalized precision as is, am I correct?

modules/tracking/samples/lasot_benchmark.py
ieliz: Maybe I am doing something wrong, but `os.path.join` does not work here.
When I use `glob`, it creates the list of `D:\lasot\airplane-1\img\00000001.jpg`, and os.path.join creates list of `00000001.jpg`, so the script can not find any images.

l-bat: Yes, `os.path.join` creates list of *.jpg. Below you can change 
`frame = cv.imread(image)`
to 
`frame = cv.imread(os.path.join(args.path_to_dataset, video_name, image))`

Or do you have another question?

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,187 @@
...
+    cx = bb[0] + (bb[2] + 1.0) / 2
+    cy = bb[1] + (bb[3] + 1.0) / 2
+    coords = {'xmin': xmin, 'xmax': xmax, 'ymin': ymin, 'ymax': ymax,
+                   'cx': cx, 'cy': cy}
+    return coords
+
+# During calculation of intersection over union we are checking numerical value
+# of area_of_overlap, because if it is equal to 0 - we have no intersection
+def get_iou(
+    new_xmin, new_xmax, new_ymin, new_ymax, gt_xmin, gt_xmax, gt_ymin, gt_ymax):
```
l-bat, line 23: Function takes too many arguments. Better pass `new_coords` and `gt_coords`, and unpack them inside a function.

modules/tracking/samples/lasot_benchmark.py
ieliz: I think I found the mistake:
it should be `frame = cv.imread(os.path.join(args.path_to_dataset, video_name, "img", image))`.
I missed the name of the folder with pictures.
I will test it a few more times and add it to the next commit.

modules/tracking/samples/lasot_benchmark.py
ieliz: I change it for all functions and visualization part.

modules/tracking/samples/lasot_benchmark.py
ieliz: I add a method for every metric.

modules/tracking/samples/lasot_benchmark.py
ieliz: As I see  - it works. I remove `glob` and using `os.path.join` now.

modules/tracking/samples/lasot_benchmark.py
ieliz: `glob` is removed.

modules/tracking/samples/lasot_benchmark.py
ieliz: Still waiting for your ideas about averaging for metrics. Benchmark works with mean arithmetical but using averaging from TrackingNet should be better.
I provided links for original papers more than a week ago, but you can find them in this comment above and in the description.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,193 @@
...
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Run LaSOT-based benchmark for visual object trackers")
+    # As a default argument used name of 
+    # original dataset folder
+    parser.add_argument("--path_to_dataset", type=str,
+                        default="LaSOTTesting", help="Full path to LaSOT")
+    parser.add_argument("--visualization", type=str,
+                        default=False, help="Showing process of tracking on video")
```
l-bat, line 78: `parser.add_argument("--visualization", action='store_true', help="Showing process of tracking on video")`

dkurt:
Please format the metrics as a table and exclude all non related information (bounding boxes, number of videos).

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,236 @@
...
+
+        list_iou.append(np.mean(iou_values))
+        list_pr.append(np.mean(pr_values))
+        list_n_pr.append(np.mean(norm_pr_values))
+    titles = ["Names:", "IoU:", "Precision:", "N.Precision:"]
+    data = [titles] + list(zip(trackers, list_iou, list_pr, list_n_pr))
+    for number, for_tracker in enumerate(data):
+        line = '|'.join(str(x).ljust(20) for x in for_tracker)
+        print(line)
+        if i == 0:
```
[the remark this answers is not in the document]
ieliz: Yes, thank you!

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,236 @@
...
+            # Calculating mean arithmetic value for specific video
+            iou_values += sum_iou / frame_counter
+            pr_values += sum_pr / frame_counter
+            norm_pr_values += sum_norm_pr / frame_counter
+
+        list_iou.append(np.mean(iou_values))
+        list_pr.append(np.mean(pr_values))
+        list_n_pr.append(np.mean(norm_pr_values))
+    titles = ["Names:", "IoU:", "Precision:", "N.Precision:"]
+    data = [titles] + list(zip(trackers, list_iou, list_pr, list_n_pr))
```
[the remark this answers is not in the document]
ieliz: OK.

ieliz:
For now, I am trying to fix incorrect calculations of metrics.
My results for current benchmark state:
```

(base) D:\Work>python lasot_benchmark.py --path_to_dataset D:/lasot --visualization
Tracker name:  Boosting
        Video name: airplane-1
Tracker name:  MIL
        Video name: airplane-1
Tracker name:  KCF
        Video name: airplane-1
Tracker name:  MedianFlow
        Video name: airplane-1
Tracker name:  GOTURN
        Video name: airplane-1
Tracker name:  MOSSE
        Video name: airplane-1
Tracker name:  CSRT
        Video name: airplane-1
[ WARN:0] global D:\opencv_master\modules\core\src\matrix_expressions.cpp (1334) cv::MatOp_AddEx::assign OpenCV/MatExpr: processing of multi-channel arrays might be changed in the future: https://github.com/opencv/opencv/issues/16739
Names:              |IoU:                |Precision:          |N.Precision:
-----------------------------------------------------------------------------------
Boosting            |45.6842             |3098.8164           |31.7185
MIL                 |46.3047             |3032.9089           |33.7500
KCF                 |44.4952             |2438.5312           |31.5100
MedianFlow          |53.0228             |1179.4745           |24.8247
GOTURN              |22.9376             |1764.9301           |17.5049
MOSSE               |40.2134             |1788.6926           |26.6768
CSRT                |40.6743             |2597.7403           |38.4245

```

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,249 @@
...
+
+    # Loop for every tracker
+    for tracker_name in trackers:
+
+        print("Tracker name: ", tracker_name)
+
+        iou_video = np.zeros(21)
+        pr_video = np.zeros(21)
+        n_pr_video = np.zeros(21)
+        iou_thr = np.linspace(0, 1, 21)
```
dkurt, line 141: Why IoU that should be a value in range `[0, 1]` has integer threshoulds?

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,249 @@
...
+        pr_thr = np.linspace(0, 50, 21)
+        n_pr_thr = np.linspace(0, 0.5, 21)
```
dkurt, line 143: What is `21`?

modules/tracking/samples/lasot_benchmark.py
dkurt: Please replace a loop into a single line:
```python
init_gt = tuple([float(gt) for b in gt_bb])
```

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,249 @@
...
+            for number_of_the_frame, image in enumerate(video_sequence):
+                frame = cv.imread(os.path.join(
+                    args.path_to_dataset, video_name, "img", image))
+                for i in range(len(gt_bb)):
+                    gt_bb[i] = float(gt_bb[i])
+                gt_bb = tuple(gt_bb)
+
+                # Condition of tracker`s re-initialization
+                if ((number_of_the_frame + 1) % frames_for_reinit == 0) and (
+                        number_of_the_frame != 0):
```
dkurt, line 181: Is `number_of_the_frame != 0` necessary?

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,249 @@
...
+    as keys and tuple with call method and number of frames for
+    reinitialization as values
+    '''
+    config = {"Boosting": (cv.TrackerBoosting_create(), 500),
+              "MIL": (cv.TrackerMIL_create(), 1000),
+              "KCF": (cv.TrackerKCF_create(), 1000),
+              "MedianFlow": (cv.TrackerMedianFlow_create(), 1000),
+              "GOTURN": (cv.TrackerGOTURN_create(), 250),
+              "MOSSE": (cv.TrackerMOSSE_create(), 1000),
+              "CSRT": (cv.TrackerCSRT_create(), 1000)}
```
dkurt, line 107: 1000 frames is almost 30 seconds of 30FPS video. Can you please check videos framerate?

modules/tracking/samples/lasot_benchmark.py
ieliz: IoU lies in range ```[0, 1]``` , with ```np.linspace()``` we create ```np.ndarray``` with 21 ```np.float64``` values. It is used in original code, link in the description of the PR.

modules/tracking/samples/lasot_benchmark.py
ieliz: Number of thresholds, used for integration.

modules/tracking/samples/lasot_benchmark.py
ieliz: Prevent of using the first frame.

modules/tracking/samples/lasot_benchmark.py
ieliz: It is just .jpg pictures.

modules/tracking/samples/lasot_benchmark.py
dkurt: I see but seems like `(number_of_the_frame + 1) % frames_for_reinit == 0` is enough for it

modules/tracking/samples/lasot_benchmark.py
dkurt: Got it, thanks! So maybe we need to create a separate method that calculates overall metric for all the groudtruth and predicted boxes?

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,249 @@
...
+                # Setting as ground truth bounding box from next frame
+                gt_bb = gt_file.readline().replace("\n", "").split(",")
+
+            # Calculating mean arithmetic value for specific video
+            iou_video += (np.fromiter([sum(i >= thr for i in iou_list).astype(
+                float) / frame_counter for thr in iou_thr], dtype=float))
+            pr_video += (np.fromiter([sum(i <= thr for i in pr_list).astype(
+                float) / frame_counter for thr in pr_thr], dtype=float))
+            n_pr_video += (np.fromiter([sum(i <= thr for i in n_pr_list).astype(
+                float) / frame_counter for thr in n_pr_thr], dtype=float))
```
dkurt, line 225: As for me we should compute it ones for all the ground truth BB and predicted ones. Metric should be invariant for number of videos. They could be merged in example into a single one.

modules/tracking/samples/lasot_benchmark.py
dkurt: Please add a named constant for it.

modules/tracking/samples/lasot_benchmark.py
ieliz: In original papers for that was used Pandas, but I rewrite Python generators without it. In this way we can skip frames with ground truth = (0, 0, 0, 0) more easily, so I think this way is better.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
dkurt: I mean that there we need to compute it once for the outputs from videos but not for each video independently.

modules/tracking/samples/lasot_benchmark.py
dkurt: Got it, thanks!

modules/tracking/samples/lasot_benchmark.py
ieliz: Compute it separable for each video is better because like this we can see results for particular case if we need it, it is more flexible way to measure parameters of trackers.

modules/tracking/samples/lasot_benchmark.py
ieliz: Is it about comment below?
If "yes", then I answered it.

modules/tracking/samples/lasot_benchmark.py
ieliz: I think problem with metrics is solved.

modules/tracking/samples/lasot_benchmark.py
dkurt: But is that the reliable final metric?

ieliz: I think yes. Calculations from the original code were reproduced very accurate.

dkurt: Ok, that's fine. Thanks for clarification!

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

dkurt:
@ieliz, do you have thoughts why DaSiam RPN tracker has worse results comparing to others? Have you visualized it's results?

ieliz:
@dkurt , I have an idea about that.
As I mentioned in description, metrics were measured for the 1 video of the LaSOT dateset as a test for the benchmark. So this values just showing us that benchmark can correctly measure metrics. To evaluate tracker`s metrics properly we need to use all 280 videos. This specific case showing us that deep learning based trackers (as GOTURN and DaSiamRPN) are not good in cases like that. But that does not mean that they are worse than classical trackers.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,247 @@
...
+            init_once = False
+
+            print("\tVideo name: " + str(video_name))
+
+            # Open specific video and read ground truth for it
+            gt_file = open(os.path.join(args.path_to_dataset, video_name,
+                           "groundtruth.txt"), "r")
+            gt_bb = gt_file.readline().replace("\n", "").split(",")
+            init_bb = gt_bb
+            init_bb = tuple([float(b) for b in init_bb])
```
dkurt, line 159: Replace 
```python
init_bb = gt_bb
init_bb = tuple([float(b) for b in init_bb])
```
to
```python
init_bb = tuple([float(b) for b in gt_bb])
```

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,247 @@
...
+
+                    iou_list.append(get_iou(new_coords, gt_coords))
+                    pr_list.append(get_pr(new_coords, gt_coords))
+                    n_pr_list.append(get_norm_pr(
+                        new_coords, gt_coords, gt_bb[2], gt_bb[3]))
+                else:
+                    frame_counter -= 1
+
+                # Setting as ground truth bounding box from next frame
+                gt_bb = gt_file.readline().replace("\n", "").split(",")
```
dkurt, line 215: `replace("\n", "")` -> `strip('\n')`

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,247 @@
...
+    return config[tracker_name]
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Run LaSOT-based benchmark for visual object trackers")
+    # As a default argument used name of
+    # original dataset folder
+    parser.add_argument("--path_to_dataset", type=str,
+                        default="LaSOTTesting", help="Full path to LaSOT")
```
dkurt, line 117: Please make flags shorter: `--dataset` and `-v`.

dkurt:
```python
parser.add_argument('-v', dest='visualization', action='store_true',
                    help="Showing process of tracking")
```

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,246 @@
...
+def get_iou(new, gt):
+    '''
+    During the calculation of intersection over union, we are checking
+    numerical value of area_of_overlap, because if it is equal to 0,
+    we have no intersection.
+    Lists, generated by 'calc_coords', are used as input for these methods.
+
+    Respectively:
+    new[0],gt[0] -> xmin, new[1],gt[1] -> xmax, new[2],gt[2] -> ymin
+    new[3],gt[3] -> ymax, new[4],gt[4] -> cx, new[5],gt[5] -> cy
```
l-bat, line 38: I think this description is redundant

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,246 @@
...
+    '''
+    In calculations of precision and normalized precision are used thresholds
+    from original TrackingNet paper: for metric calculation TrackingNet using
+    lists of thresholds, but here we use only one numerical value of the
+    threshold for each metric
+    Lists, generated by 'calc_coords', are used as input for these methods.
+
+    Respectively:
+    new[0],gt[0] -> xmin, new[1],gt[1] -> xmax, new[2],gt[2] -> ymin
+    new[3],gt[3] -> ymax, new[4],gt[4] -> cx, new[5],gt[5] -> cy
```
l-bat, line 64: I think this description is redundant

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,246 @@
...
+    '''
+    new_xmin, new_xmax, new_ymin, new_ymax, new_cx, new_cy = new
+    gt_xmin, gt_xmax, gt_ymin, gt_ymax, gt_cx, gt_cy = gt
+    precision = np.sqrt((new_cx - gt_cx) ** 2 + (new_cy - gt_cy) ** 2)
+    return precision
+
+
+def get_norm_pr(new, gt, gt_bb_w, gt_bb_h):
```
l-bat, line 72: We can combine `get_norm_pr` with `get_pr`

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,246 @@
+import numpy as np
+import cv2 as cv
+import argparse
+import os
+
+
+def calc_coords(bb):
```
l-bat, line 7: I think we can remove `calc_coords`, because `xmax`,`ymax` are used only to calculate `IoU` and `cx, cy` are used to calculate `Precision`. We can calculate necessary coords in corresponding function.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,246 @@
...
+
+            # Creating blob from image sequence
+            video_sequence = sorted(os.listdir(os.path.join(
+                args.dataset, video_name, "img")))
+
+            # Variables for saving sum of every metric for every frame and
+            # every video respectively
+            iou_list = []
+            pr_list = []
+            n_pr_list = []
```
l-bat, line 168: We need to find better name for these variables. Above you have 
```python
    list_iou = []
    list_pr = []
    list_n_pr = []
```
If you give a suitable name for variables, comments are not needed

modules/tracking/samples/lasot_benchmark.py
ieliz: I disagree, because in with `calc_coords` we calculating corners and centers of the bounding boxes just once. If we split calculations between other functions it is gonna be more complex. Example: for now we calculate xmax, xmin, ymax, ymin once for every frame. If we will remove `calc_coords` we would be forced to calculate it three times in every method.

modules/tracking/samples/lasot_benchmark.py
ieliz: I do not think it is necessary, because it is different metrics and for flexibility of the benchmark it is better if we can work with every metric independently.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
ieliz: OK. I will change some names of the variables.

modules/tracking/samples/lasot_benchmark.py
l-bat: These metrics are very similar and easy to combine

l-bat:
```python
def get_precision(bbox, gt, is_norm=False):
    '''
    In calculations of precision and normalized precision are used thresholds
    from original TrackingNet paper: for metric calculation TrackingNet using
    lists of thresholds, but here we use only one numerical value of the
    threshold for each metric
    '''
    bb_x, bb_y, bb_w, bb_h = bbox
    gt_x, gt_y, gt_w, gt_h = gt
    def get_center(coord, size): return coord + (size + 1.0) / 2
    bb_cx, bb_cy = get_center(bb_x, bb_w), get_center(bb_y, bb_h)
    gt_cx, gt_cy = get_center(gt_x, gt_w), get_center(gt_y, gt_h)
    if is_norm:
        bb_cx, gt_cx = bb_cx / gt_w, gt_cx / gt_w
        bb_cy, gt_cy = bb_cy / gt_h, gt_cy / gt_h

    thr = 0.1 if is_norm else 20
    normalized_precision = np.sqrt((bb_cx - gt_cx)**2 + (bb_cy - gt_cy)**2)
    norm_pr_value = 0.0 if normalized_precision > thr else 1.0
    return norm_pr_value
```

ieliz: Yes, but if user want to remove one of them, just comment it is enough. If we combine these methods, user would be forced to change method inside it. It is not conveniently.

modules/tracking/samples/lasot_benchmark.py
l-bat: If we removed this function, then we calculated `xmax, ymax` in IoU and `cx, cy` in `get_precision(bbox, gt, is_norm=False)`. Thus, calculation will not be more complicated.

ieliz: By "complex" I mean the large number of calculations, sorry for inaccuracy in the previous message.
So, if we will remove `calc_coords` we would be forced to calculate `xmax, xmin, ymax, ymin` three times (in every method) and `cx, cy`  two times (for precision and normalized precision) for _every frame_. And we have 280 videos with at least 1000 frames each. I do not think it is rational.

modules/tracking/samples/lasot_benchmark.py
l-bat: I don't understand you. If user wants to get only normalized precision, he can:
`norm_pr = get_precision(bbox, gt, is_norm=True)`
or 
`pr  = get_precision(bbox, gt, is_norm=False)`

Why might the user need to remove function?

ieliz: It is about flexibility of the benchmark: I want to make the benchmark flexible for user.
If user want to add new metrics or remove old ones it is more convenient to operate these metrics via methods. Combining metrics is good for now, but it is not convenient for the future work, improvements and changes.

modules/tracking/samples/lasot_benchmark.py
l-bat: Now, we call `calc_coords `in a loop:
```python
# For every frame in video
for number_of_the_frame, image in enumerate(video_sequence):
```
calculated for each frame, right?
> three times (in every method) 
What are these methods? Why are `ymax, ymin`  needed in `get_precision(bbox, gt)`?

modules/tracking/samples/lasot_benchmark.py
l-bat: The user can simply add new metrics rather than remove old ones

modules/tracking/samples/lasot_benchmark.py
ieliz: Yes, `calc_coords` called twice for each frame: for ground truth and for prediction.
You are right: `xmax, xmin, ymax, ymin` are not needed for precision/normalized precision - my mistake.
I will think about how `get_iou`, `get_pr` and `get_norm_pr` methods can be reworked.
But, I still think combining `get_pr` and `get_norm_pr` can affect flexibility of the benchmark.

modules/tracking/samples/lasot_benchmark.py
ieliz: But I still think that 1 metric  == 1 method is more convenient.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,240 @@
...
+
+def get_iou(new, gt):
+    '''
+    During the calculation of intersection over union, we are checking
+    numerical value of area_of_overlap, because if it is equal to 0,
+    we have no intersection.
+    Lists, generated by 'calc_coords', are used as input for these methods.
+    Respectively:
+    new[0],gt[0] -> xmin, new[1],gt[1] -> xmax, new[2],gt[2] -> ymin
+    new[3],gt[3] -> ymax, new[4],gt[4] -> cx, new[5],gt[5] -> cy
```
l-bat, line 37: I think this description is redundant

ieliz: I thought that it improves understandability of the code - it should help to new user understand all calculations.

l-bat:
```
new_xmin, new_xmax, new_ymin, new_ymax, new_cx, new_cy = new
gt_xmin, gt_xmax, gt_ymin, gt_ymax, gt_cx, gt_cy = gt
```
On these lines it is better to understand all the calculations.

ieliz: OK.

modules/tracking/samples/lasot_benchmark.py
dkurt: Code duplication is not very good practice. These two metrics are actually similar so it's better to have a single implementation.

ieliz: We are talking about few lines of code - it is partially duplicated code, but it is better for scalability of the benchmark. I do not see anything bad about these functions.

l-bat: It’s more clear for the user to use one function with the `is_norm` flag for similar metrics than two different functions with a minimum difference

dkurt: @l-bat, why not
```python
def get_pr(new, gt, is_norm):
    new_xmin, new_xmax, new_ymin, new_ymax, new_cx, new_cy = new
    gt_xmin, gt_xmax, gt_ymin, gt_ymax, gt_cx, gt_cy = gt
    dx = new_cx - gt_cx
    dy = new_cy - gt_cy
    if is_norm:
       dx /= gt_bb_w
       dy /= gt_bb_h

    return np.sqrt(dx ** 2 + dy ** 2)
```

l-bat: It seems to me that `calc_coords` isn't needed and we can calculate `cx, cy` here. It’s a good idea to normalize only the difference.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,203 @@
...
+    from original TrackingNet paper. If third argument is "True" - we calculate
+    normalized precision, if it is "False" - we calculate precision without
+    normalization
+    '''
+    new_cx = new[0] + (new[2] + 1.0) / 2
+    new_cy = new[1] + (new[3] + 1.0) / 2
+    gt_cx = gt[0] + (gt[2] + 1.0) / 2
+    gt_cy = gt[1] + (gt[3] + 1.0) / 2
+    gt_bb_w = gt[2]
+    gt_bb_h = gt[3]
```
l-bat, line 45:
```python
    new_x, new_y, new_w, new_h = new
    gt_x, gt_y, gt_w, gt_h = gt
```

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,203 @@
...
+    gt_xmin = gt[0]
+    gt_xmax = gt[0] + gt[2] - 1.0
+    gt_ymin = gt[1]
+    gt_ymax = gt[1] + gt[3] - 1.0
+    dx = max(0, min(new_xmax, gt_xmax) - max(new_xmin, gt_xmin))
+    dy = max(0, min(new_ymax, gt_ymax) - max(new_ymin, gt_ymin))
+    area_of_overlap = dx * dy
+    area_of_union = (new_xmax - new_xmin) * (new_ymax - new_ymin) + (
+        gt_xmax - gt_xmin) * (gt_ymax - gt_ymin) - area_of_overlap
+    if area_of_overlap != 0 and area_of_union != 0:
```
l-bat, line 26: Why are you adding `area_of_overlap != 0`?
`iou = area_of_overlap / area_of_union if area_of_union != 0 else 0`
better for me

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,203 @@
...
+    we have no intersection.
+    '''
+    new_xmin = new[0]
+    new_xmax = new[0] + new[2] - 1.0
+    new_ymin = new[1]
+    new_ymax = new[1] + new[3] - 1.0
+    gt_xmin = gt[0]
+    gt_xmax = gt[0] + gt[2] - 1.0
+    gt_ymin = gt[1]
+    gt_ymax = gt[1] + gt[3] - 1.0
```
l-bat, line 20:
```suggestion
    new_xmin, new_ymin, new_w, new_h = new
    gt_xmin, gt_ymin, gt_w, gt_h = gt
    def get_max_coord(coord, size): return coord + size - 1.0
    new_xmax, new_ymax = get_max_coord(new_xmin, new_w), get_max_coord(new_ymin, new_h)
    gt_xmax, gt_ymax = get_max_coord(gt_xmin, gt_w), get_max_coord(gt_ymin, gt_h)
```

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,203 @@
...
+                    init_bb = gt_bb
+
+                if not init_once:
+                    init_state = tracker.init(frame, init_bb)
+                    init_once = True
+                init_state, new_bb = tracker.update(frame)
+
+                if args.visualization:
+                    cv.rectangle(frame, (int(new_bb[0]), int(new_bb[1])), (int(
+                        new_bb[0] + new_bb[2]), int(new_bb[1] + new_bb[3])), (
```
l-bat, line 156: `new_x, new_y, new_w, new_h = map(int, new_bb) `

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,203 @@
...
+                            200, 0, 0))
+                    cv.imshow("Tracking", frame)
+                    cv.waitKey(1)
+
+                iou_values.append(get_iou(new_bb, gt_bb))
+                pr_values.append(get_pr(new_bb, gt_bb, False))
+                n_pr_values.append(get_pr(new_bb, gt_bb, True))
```
l-bat, line 163: I think it is better to use keyword arg for `is_norm`:
```suggestion
                pr_values.append(get_pr(new_bb, gt_bb, is_norm=False))
                n_pr_values.append(get_pr(new_bb, gt_bb, is_norm=True))
```

modules/tracking/samples/lasot_benchmark.py
l-bat: For python3:
`new_x, new_y, new_w, new_h = list(map(int, new_bb))`

modules/tracking/samples/lasot_benchmark.py
ieliz: Why?

ieliz:
I noticed one thing when comparing the results of the benchmark on Ubuntu and Windows 10:
results are different for the same methods, trackers, and videos (tested for 1 video/1 tracker). I gonna do some more experiments, but I want to ask - can it be connected with OS or it is about something else?

dkurt:
@ieliz, first you need to check what exactly differs - before getting a final metric there are a lot of intermediate steps - dataset loading, tracker execution, metric calculation.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,198 @@
...
+    return config[tracker_name]
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Run LaSOT-based benchmark for visual object trackers")
+    # As a default argument used name of the original dataset folder
+    parser.add_argument("--dataset", type=str,
+                        default="LaSOTTesting", help="Full path to LaSOT")
+    parser.add_argument("--v", dest="visualization", action='store_true',
```
dkurt, line 75: Just `-v`

ieliz: OK

ieliz:
May I change some parts of the benchmark for adding DaSiamRPN in the list of trackers, due to changes in the DaSiamRPN tracker sample?
Like:
```python
from dasiamrpn_tracker import DaSiamRPNTracker
```

dkurt:
> May I change some parts of the benchmark for adding DaSiamRPN in the list of trackers, due to changes in the DaSiamRPN tracker sample?
Like:

from dasiamrpn_tracker import DaSiamRPNTracker

Sure! That's what we exactly wanted to do (unify both benchmarks: for DaSiam and for the rest of trackers into one)

dkurt:
Please unite all the numbers into a single table (seems to me that we need to run DaSiam again as a part of new script)

ieliz:
I ran the DaSiamRPN tracker as a part of the new benchmark and compare results with the results of the integrated benchmark (in the description of the PR). Results for 'coin-3' video from the Lasot dataset are equal.

modules/tracking/samples/lasot_benchmark.py
```diff
@@ -0,0 +1,213 @@
...
+Link to LaSOT paper: https://arxiv.org/abs/1809.07845
+Link to TrackingNet paper: https://arxiv.org/abs/1803.10794
+"""
+
+import numpy as np
+import cv2 as cv
+import argparse
+import os
+
+from opencv.samples.dnn.dasiamrpn_tracker import DaSiamRPNTracker
```
dkurt, line 15: I'd like to suggest to use this construction:
```python
import sys

path = cv.samples.findFile('samples/dnn/dasiamrpn_tracker.py')
sys.path.append(os.path.dirname(path))
from dasiamrpn_tracker import DaSiamRPNTracker

```

ieliz: OK

asenyaev:

PR #2536, opened 2020-05-18
Modified Stream support functionality.

vchiluka5:
Modified stream support functionality
Added streams to optical_flow.cpp sample app
Modified documentation of streams wrt NVIDIA Optical Flow

<cut/>

```
force_builders=Custom
buildworker:Custom=linux-4
```

### DIFF
--- a/modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
+++ b/modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
@@ -118,7 +118,9 @@ public:
     @param inputImage Input image.
     @param referenceImage Reference image of the same size and the same type as input image.
     @param flow A buffer consisting of inputImage.Size() / getGridSize() flow vectors in CV_16SC2 format.
-    @param stream Stream for the asynchronous version.
+    @param stream It is highly recommended that CUDA streams for pre and post processing of optical flow vectors should be set once per session in create() function as a part of optical flow session creation.
+                  This parameter is left here for backward compatibility and may be removed in the future.
+                  Default value is NULL stream;
     @param hint Hint buffer if client provides external hints. Must have same size as flow buffer.
                 Caller can provide flow vectors as hints for optical flow calculation.
     @param cost Cost buffer contains numbers indicating the confidence associated with each of the generated flow vectors.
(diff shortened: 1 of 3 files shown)


vchiluka5:
Hi @alalek 
Can you please review the above commit and if all seems well then can you please merge it to master?

alalek:
Please take a look on build warnings from "Custom" builder.

modules/cudaoptflow/include/opencv2/cudaoptflow.hpp
```diff
@@ -444,7 +452,9 @@ class CV_EXPORTS_W NvidiaOpticalFlow_1_0 : public NvidiaHWOpticalFlow
         bool enableTemporalHints = false,
         bool enableExternalHints = false,
         bool enableCostBuffer = false,
-        int gpuId = 0);
+        int gpuId = 0,
+        Stream inputStream = Stream::Null(),
+        Stream outputStream = Stream::Null());
```
alalek, line 457: According to other CUDA APIs this parameter is usually passed as `Stream&`.

Not sure about "const". /cc @asmorkalov

vchiluka5: > According to other CUDA APIs this parameter is usually passed as `Stream&`.
> 
> Not sure about "const". /cc @asmorkalov

When we do Stream x = Stream(), It creates a new stream and returns pointer to it right?

alalek: See [here](https://github.com/opencv/opencv/blob/4.3.0/modules/core/include/opencv2/core/cuda.hpp#L1046).

vchiluka5: > See [here](https://github.com/opencv/opencv/blob/4.3.0/modules/core/include/opencv2/core/cuda.hpp#L1046).

Thanks. Modified it accordingly.

alalek:
@vchiluka5 There is build error from this patch:
```
/build/master-contrib_cuda-lin64/opencv_contrib/modules/cudaoptflow/samples/optical_flow.cpp: In function 'int main(int, const char**)':
/build/master-contrib_cuda-lin64/opencv_contrib/modules/cudaoptflow/samples/optical_flow.cpp:185:119: error: invalid initialization of non-const reference of type 'cv::cuda::Stream&' from an rvalue of type 'cv::cuda::Stream'
         NvidiaOpticalFlow_1_0::NVIDIA_OF_PERF_LEVEL::NV_OF_PERF_LEVEL_FAST, false, false, false, 0, Stream(), Stream());
                                                                                                                       ^
In file included from /build/master-contrib_cuda-lin64/opencv_contrib/modules/cudaoptflow/samples/optical_flow.cpp:7:0:
/build/master-contrib_cuda-lin64/opencv_contrib/modules/cudaoptflow/include/opencv2/cudaoptflow.hpp:447:47: note:   initializing argument 8 of 'static cv::Ptr<cv::cuda::NvidiaOpticalFlow_1_0> cv::cuda::NvidiaOpticalFlow_1_0::create(int, int, cv::cuda::NvidiaOpticalFlow_1_0::NVIDIA_OF_PERF_LEVEL, bool, bool, bool, int, cv::cuda::Stream&, cv::cuda::Stream&)'
     CV_WRAP static Ptr<NvidiaOpticalFlow_1_0> create(
                                               ^
```

Please take a look. Perhaps "const reference" is required in API.

tomoaki0705:
please have a look in #2540 
that patch worked for me

vchiluka5:
> please have a look in #2540
> that patch worked for me

@tomoaki0705  we need to pass a valid stream and not NULL() stream. 
@alalek  Its strange since i was able to build and run the app as well on windows. Also builds didnt gave any error right before merging PR?
Let me fix it. I still dont see error on my windows machine. Il setup a linux machine and try it out. Thanks for pointing it out.

PR #2541, opened 2020-05-23
NVOF Build Error Fix

vchiluka5:
relates #2536

Please check if the below patch resolves build errors. I am able to compile and run the sample app on my machine

- [*] I agree to contribute to the project under OpenCV (BSD) License.
- [*] To the best of my knowledge, the proposed patch is not based on a code under GPL or other license that is incompatible with OpenCV
- [*] The PR is proposed to proper branch
- [*] The feature is well documented and sample code can be built with the project CMake

```
force_builders=Custom
buildworker:Custom=linux-4
```


vchiluka5:
@alalek  I removed streams from the API call for now. Its a temporary fix. This should fix the build errors but i am trying to understand more about streams and why its used as Stream& everywhere. Will submit another patch where non NULL() streams will be passed to create() API.

PR #2543, opened 2020-05-23
NVOF - Passing non null streams to create API

vchiluka5:
relates #2536

This PR is to check if build fails using non NULL streams or not. 

```
force_builders=Custom
buildworker:Custom=linux-4
```

### DIFF
--- a/modules/cudaoptflow/samples/optical_flow.cpp
+++ b/modules/cudaoptflow/samples/optical_flow.cpp
@@ -177,12 +177,14 @@ int main(int argc, const char* argv[])
 
     GpuMat d_flow(frame0.size(), CV_32FC2), d_flowxy;
 
+    Stream inputStream, outputStream;
+
     Ptr<cuda::BroxOpticalFlow> brox = cuda::BroxOpticalFlow::create(0.197f, 50.0f, 0.8f, 10, 77, 10);
     Ptr<cuda::DensePyrLKOpticalFlow> lk = cuda::DensePyrLKOpticalFlow::create(Size(7, 7));
     Ptr<cuda::FarnebackOpticalFlow> farn = cuda::FarnebackOpticalFlow::create();
     Ptr<cuda::OpticalFlowDual_TVL1> tvl1 = cuda::OpticalFlowDual_TVL1::create();
     Ptr<cuda::NvidiaOpticalFlow_1_0> nvof = cuda::NvidiaOpticalFlow_1_0::create(frame0.size().width, frame0.size().height,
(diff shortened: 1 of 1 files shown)


vchiluka5:
@alalek  Can you please add the same custom build with which build was failing with last PR to this PR?

vchiluka5:
> This PR is to check if build fails using non NULL streams or not.
> 
> ```
> force_builders=Custom
> buildworker:Custom=linux-4
> build_image:Custom=ubuntu-cuda:18.04
> ```

Thanks @alalek  
Looks like the build succeeded. Please verify and merge this PR.

modules/cudaoptflow/samples/optical_flow.cpp
```diff
@@ -177,12 +177,14 @@ int main(int argc, const char* argv[])
...
 
+    Stream inputStream, outputStream;
+
     Ptr<cuda::BroxOpticalFlow> brox = cuda::BroxOpticalFlow::create(0.197f, 50.0f, 0.8f, 10, 77, 10);
     Ptr<cuda::DensePyrLKOpticalFlow> lk = cuda::DensePyrLKOpticalFlow::create(Size(7, 7));
     Ptr<cuda::FarnebackOpticalFlow> farn = cuda::FarnebackOpticalFlow::create();
     Ptr<cuda::OpticalFlowDual_TVL1> tvl1 = cuda::OpticalFlowDual_TVL1::create();
     Ptr<cuda::NvidiaOpticalFlow_1_0> nvof = cuda::NvidiaOpticalFlow_1_0::create(frame0.size().width, frame0.size().height,
-        NvidiaOpticalFlow_1_0::NVIDIA_OF_PERF_LEVEL::NV_OF_PERF_LEVEL_FAST, false, false, false, 0);
+        NvidiaOpticalFlow_1_0::NVIDIA_OF_PERF_LEVEL::NV_OF_PERF_LEVEL_FAST, false, false, false, 0, inputStream, outputStream);
```
alalek, line 187: API behavior is confusing.
Why we provide default values for parameters (see #2536), but still require them?

How this correlates with documentation?

> If input stream is not set, the execute function will use default stream which is NULL stream;

vchiluka5: We dont require them. 
Those are optional parameters. Default values are provided so that if user dont pass it then it will consider default value.
This sample app demonstrates comparison of different optical flow algorithms and we added streams to this so as to see the benefit of it in performance compared with other algorithms.

Issue #2548, opened 2020-05-29
BRISQUE: empty don't works as expected 


Nuzhny007:
##### System information (version)
- OpenCV => 4.3
- Operating System / Platform => Windows 64 Bit
- Compiler => Visual Studio 2017
##### Detailed description

When I create QualityBRISQUE  with wrong path to the range file the QualityBRISQUE::create returns a valid pointer. Ho can I check that all data was loaded correctly? I call QualityBase::empty - but it always returns true: https://github.com/opencv/opencv_contrib/blob/master/modules/quality/include/opencv2/quality/qualitybase.hpp#L49

May be it can add overloaded method empty to the QualityBRISQUE? And if model or range was not loaded correctly than it will return true:
`CV_WRAP bool empty() const CV_OVERRIDE { return _range.empty() || _model.empty(); }`

clunietp:
Thanks @Nuzhny007 - IIRC the constructor would throw an exception if the range file was empty, but apparently either I am mistaken or that behavior has changed.  Regardless, I agree there should be a fix.  Are you still going to pursue your PR?  

My original intent was to throw an exception on invalid files (thus never leaving the brisque obj in an invalid state), but I can see no harm in implementing `empty()` as well

Nuzhny007:
Now I can set the wrong path to the range file and it will be exception in compute(). I was about to make PR, but the tests failed. Can you see what is wrong with them?

clunietp:
When you fork, you need to create a new branch as well in your repo.  You need to create a new branch in your repo, and then initiate a PR to opencv:master from that.  Make sense?

Also, in your `empty()` method, you may also want to check for `_model != nullptr` (or similar) prior to checking `_model.empty()`.   Eg:  `return _range.empty() || !_model || _model.empty();`

Nuzhny007:
Ok: https://github.com/opencv/opencv_contrib/pull/2565

PR #2565, opened 2020-06-08
Add override empty method to QualityBRISQUE

Nuzhny007:
Fixed this issue: #2548

### DIFF
--- a/modules/quality/test/test_brisque.cpp
+++ b/modules/quality/test/test_brisque.cpp
@@ -77,6 +77,18 @@ TEST(TEST_CASE_NAME, compute_features)
     EXPECT_EQ(features.cols, 36);
 }
 
+// check empty method
+TEST(TEST_CASE_NAME, check_empty)
+{
+    auto ptr1 = create_brisque();
+    EXPECT_FALSE(ptr1->empty());
+
+    const auto model = cvtest::findDataFile(MODEL_FNAME, false);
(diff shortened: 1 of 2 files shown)


modules/quality/test/test_brisque.cpp
```diff
@@ -77,6 +77,18 @@ TEST(TEST_CASE_NAME, compute_features)
...
 }
 
+// check empty method
+TEST(TEST_CASE_NAME, check_empty)
+{
+    auto ptr1 = create_brisque();
+    EXPECT_FALSE(ptr1->empty());
+
+    const auto model = cvtest::findDataFile(MODEL_FNAME, false);
+    const auto range = cvtest::findDataFile("some_wrong_range_name", false);
```
clunietp, line 87: Don't use `findDataFile` here, it will throw a `SkipTestException` and the code below this line won't run.  See https://github.com/opencv/opencv/blob/08a3eef62a378c66df5d4e7a786b57d1c68819a5/modules/ts/src/ts.cpp#L1053

Nuzhny007:
How can I restart tests with latest commit?

alalek:
You can always trigger new builds by updating your patch. For example, you can push amended commit (or rebase / squash commits).
```
git commit --am
git push --force <remote> <branch>
```

---

I rescheduled builds manually.

PR #2572, opened 2020-06-18
 rapid: add "Optimal Local Search" implementation 

paroj:

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/rapid/include/opencv2/rapid.hpp
+++ b/modules/rapid/include/opencv2/rapid.hpp
@@ -122,6 +122,33 @@ CV_EXPORTS_W void convertCorrespondencies(InputArray cols, InputArray srcLocatio
  */
 CV_EXPORTS_W float rapid(InputArray img, int num, int len, InputArray pts3d, InputArray tris, InputArray K,
                          InputOutputArray rvec, InputOutputArray tvec, CV_OUT double* rmsd = 0);
+
+/// Abstract base class for stateful silhouette trackers
+class CV_EXPORTS_W Tracker : public Algorithm
+{
+public:
+    virtual ~Tracker();
+    CV_WRAP virtual float
+    compute(InputArray img, int num, int len, InputArray K, InputOutputArray rvec, InputOutputArray tvec,
+            const TermCriteria& termcrit = TermCriteria(TermCriteria::MAX_ITER | TermCriteria::EPS, 5, 1.5)) = 0;
+    CV_WRAP virtual void clearState() = 0;
+};
+
+/// wrapper around @ref rapid function for uniform access
+class CV_EXPORTS_W Rapid : public Tracker
+{
+public:
+    CV_WRAP static Ptr<Rapid> create(InputArray pts3d, InputArray tris);
+};
+
+/** implements "Optimal local searching for fast and robust textureless 3D object tracking in highly
+ * cluttered backgrounds" @cite seo2013optimal
+ */
+class CV_EXPORTS_W OLSTracker : public Tracker
+{
+public:
+    CV_WRAP static Ptr<OLSTracker> create(InputArray pts3d, InputArray tris, int histBins = 8, uchar sobelThesh = 10);
+};
 //! @}
 } /* namespace rapid */
 } /* namespace cv */
--- a/modules/rapid/include/opencv2/rapid.hpp
+++ b/modules/rapid/include/opencv2/rapid.hpp
@@ -24,12 +24,12 @@ namespace rapid
 /**
  * Debug draw markers of matched correspondences onto a lineBundle
  * @param bundle the lineBundle
- * @param srcLocations the according source locations
- * @param newLocations matched source locations
+ * @param cols column coordinates in the line bundle
  * @param colors colors for the markers. Defaults to white.
  */
-CV_EXPORTS_W void drawCorrespondencies(InputOutputArray bundle, InputArray srcLocations,
-                                       InputArray newLocations, InputArray colors = noArray());
+CV_EXPORTS_W void drawCorrespondencies(InputOutputArray bundle, InputArray cols,
+                                       InputArray colors = noArray());
+
 /**
  * Debug draw search lines onto an image
  * @param img the output image
@@ -83,20 +83,22 @@ CV_EXPORTS_W void extractLineBundle(int len, InputArray ctl2d, InputArray img, O
  * Find corresponding image locations by searching for a maximal sobel edge along the search line (a single
  * row in the bundle)
  * @param bundle the line bundle
- * @param srcLocations the according source image location
- * @param newLocations image locations with maximal edge along the search line
+ * @param cols correspondence-position per line in line-bundle-space
  * @param response the sobel response for the selected point
  */
-CV_EXPORTS_W void findCorrespondencies(InputArray bundle, InputArray srcLocations, OutputArray newLocations,
+CV_EXPORTS_W void findCorrespondencies(InputArray bundle, OutputArray cols,
                                        OutputArray response = noArray());
 
 /**
- * Filter corresponding 2d and 3d points based on mask
+ * Collect corresponding 2d and 3d points based on correspondencies and mask
+ * @param cols correspondence-position per line in line-bundle-space
+ * @param srcLocations the source image location
  * @param pts2d 2d points
  * @param pts3d 3d points
  * @param mask mask containing non-zero values for the elements to be retained
  */
-CV_EXPORTS_W void filterCorrespondencies(InputOutputArray pts2d, InputOutputArray pts3d, InputArray mask);
+CV_EXPORTS_W void convertCorrespondencies(InputArray cols, InputArray srcLocations, OutputArray pts2d,
+                                          InputOutputArray pts3d = noArray(), InputArray mask = noArray());
 
 /**
  * High level function to execute a single rapid @cite harris1990rapid iteration
@@ -104,7 +106,7 @@ CV_EXPORTS_W void filterCorrespondencies(InputOutputArray pts2d, InputOutputArra
  * 1. @ref extractControlPoints
  * 2. @ref extractLineBundle
  * 3. @ref findCorrespondencies
- * 4. @ref filterCorrespondencies
+ * 4. @ref convertCorrespondencies
  * 5. @ref solvePnPRefineLM
  *
  * @param img the video frame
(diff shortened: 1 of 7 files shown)


paroj:
@fegorsch

modules/rapid/include/opencv2/rapid.hpp
```diff
@@ -115,10 +117,11 @@ CV_EXPORTS_W void filterCorrespondencies(InputOutputArray pts2d, InputOutputArra
  * @param K camera matrix
  * @param rvec rotation between mesh and camera. Input values are used as an initial solution.
  * @param tvec translation between mesh and camera. Input values are used as an initial solution.
+ * @param cols optional column coordinates in the line bundle
  * @return ratio of search lines that could be extracted and matched
  */
 CV_EXPORTS_W float rapid(InputArray img, int num, int len, InputArray pts3d, InputArray tris, InputArray K,
-                         InputOutputArray rvec, InputOutputArray tvec);
+                         InputOutputArray rvec, InputOutputArray tvec, OutputArray cols = noArray());
```
fegorsch, line 124: as the line bundle is an implementation detail, I think "coordinates in the line bundle" have no use outside this function and should be hidden, too. I'd rather return the RMSD directly.

modules/rapid/include/opencv2/rapid.hpp
```diff
@@ -122,6 +122,32 @@ CV_EXPORTS_W void convertCorrespondencies(InputArray cols, InputArray srcLocatio
...
                          InputOutputArray rvec, InputOutputArray tvec, OutputArray cols = noArray());
+
+/// Abstract base class for stateful silhouette trackers
+class CV_EXPORTS_W Tracker : public Algorithm
+{
+public:
+    virtual ~Tracker();
+    CV_WRAP virtual float
+    compute(InputArray img, int num, int len, InputArray K, InputOutputArray rvec, InputOutputArray tvec,
+            const TermCriteria& termcrit = TermCriteria(TermCriteria::MAX_ITER, 5, 0)) = 0;
```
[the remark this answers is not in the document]
fegorsch: also some documentation what `termCrit` means would be nice.

modules/rapid/src/rapid.cpp
```diff
@@ -355,5 +355,46 @@ float rapid(InputArray img, int num, int len, InputArray vtx, InputArray tris, I
...
+
+        Mat cols;
+        for(int i = 0; i < niter; i++)
+        {
+            ret = rapid(img, num, len, pts3d, tris, K, rvec, tvec, cols);
+
+            if(termcrit.type & TermCriteria::EPS)
+            {
+                cols -= len + 1;
+                double rmsd = std::sqrt(norm(cols, NORM_L2SQR) / cols.rows);
```
fegorsch, line 385: for clarification: this effectively measures, how far the object moved in 2D. so we always need at least 2 iterations to converge?

paroj: since we assume an initial position, one iteration is enough to measure the 2D distance from that.

modules/rapid/include/opencv2/rapid.hpp
paroj: yeah thats probably right.. OutputArray was just easier to wrap than an output double

modules/rapid/include/opencv2/rapid.hpp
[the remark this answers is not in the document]
paroj: TermCriteria::EPS would mean that we use it for termination, which we effectively dont if we set it the threshold to 0.

modules/rapid/src/rapid.cpp
fegorsch: I don't get it :)

In the following assume that moving one world unit maps to moving one pixel. The object moved 10px to the right in image space, no rotation.

Iteration one:
- init pose: t=(0, 0, 0)
- rapid does it stuff and finds a the optimal solution: t=(10, 0, 0)
- RMSD=10 => do another iteration
Iteration two:
- init pose: t=(10, 0, 0)
- rapid does it stuff and finds a the optimal solution: t=(10, 0, 0)
- RMSD=0 => finished

modules/rapid/include/opencv2/rapid.hpp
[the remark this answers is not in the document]
fegorsch: wouldn't `TermCriteria(TermCriteria::MAX_ITER | TermCriteria::EPS, 5, 0)` mean:

```
if iter == 5 or rmsd == 0:
    stop
```
?

`rmsd == 0` if rapid found an optimal solution, i.e. one where all sobel-maxima are at the centers of the search lines (leaving errors in float-to-zero-comparison aside).

modules/rapid/src/histogram.cpp
```diff
@@ -0,0 +1,226 @@
...
+            int start = 0;
+            for (int j = 0; j < bundleHSV.cols; j++)
+            {
+                if (bundleGrad.at<uchar>(i, j))
+                {
+                    // compute the histogram between last candidate point to current candidate point
+                    // as in eq. (4)
+                    hist = 0;
+                    calcHueSatHist(bundleHSV({i, i + 1}, {start, j}), hist);
+                    hist /= std::max(sum(hist), 1.0f);
```
fegorsch, line 128:
```
float histSum = cv::sum(hist)[0];
hist /= std::max(histSum, 1.0f);
```
? (Pseudocode)

Arguably a little more complicated here, but saves the custom `sum`-function.

modules/rapid/src/histogram.cpp
```diff
@@ -0,0 +1,226 @@
...
+    CV_DbgAssert(hist.isContinuous());
+    float ret = 0;
+    int N = int(hist.total());
+    const float* ptr = hist.ptr<float>();
+    for (int i = 0; i < N; i++)
+        ret += ptr[i];
+    return ret;
+}
+
+static double bhattacharyya_coeff(const Mat& a, const Mat& b)
```
fegorsch, line 57:
```suggestion
static double bhattacharyyaCoeff(const Mat& a, const Mat& b)
```

modules/rapid/src/histogram.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+static void findCorrespondenciesOLS(const cv::Mat_<float>& scores, cv::Mat_<int>& cols)
+{
+    cols.resize(scores.rows);
+    for (int i = 0; i < scores.rows; i++)
+    {
+        int pos = -1;
+        for (int j = scores.cols - 1; j >= 0; j--)
+        {
+            if (scores(i, j) >= 0.35)
```
fegorsch, line 77: I liked that you referenced where your magic numbers come from. you should do that here, too.

modules/rapid/src/histogram.cpp
```diff
@@ -0,0 +1,226 @@
...
+        vtx = _pts3d.getMat();
+        tris = _tris.getMat();
+
+        tau = 1.0; // currently does not work as intended. effectively disable
+        sobelThresh = _sobelThesh;
+
+        bgHist.create(histBins, histBins);
+    }
+
+    void computeAppearanceScore(const Mat& bundleHSV, const Mat& bundleGrad, Mat_<float>& score) const
```
fegorsch, line 111:
```suggestion
    void computeAppearanceScores(const Mat& bundleHSV, const Mat& bundleGrad, Mat_<float>& scores) const
```

modules/rapid/src/histogram.cpp
paroj: the custom sum function is due to performance reasons as the dispatch code in cv::sum becomes relevant when you are summing 8x8 px.

modules/rapid/src/histogram.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+static void calcHueSatHist(const Mat_<Vec3b>& hsv, Mat_<float>& hist)
+{
+    for (int i = 0; i < hsv.rows; i++)
+    {
+        for (int j = 0; j < hsv.cols; j++)
+        {
+            const Vec3b& c = hsv(i, j);
+            // thresholds as in sec. 4.1
+            if (c[1] > 25 && c[2] > 50)
```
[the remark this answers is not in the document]
paroj: no

modules/rapid/src/rapid.cpp
paroj: yes, you always do one iteration more then you need. But with init pose: t=(10, 0, 0), you stop after 1 iteration

modules/rapid/include/opencv2/rapid.hpp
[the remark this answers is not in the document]
paroj: yes, but rmsd == 0 does not happen in practice. What I meant that `TermCriteria::EPS -> rmsd > 0` (~3 in the single experiment that I did)

modules/rapid/src/histogram.cpp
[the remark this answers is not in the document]
fegorsch: so filtering happens in S and V, but the histogram is built for H and S?

modules/rapid/include/opencv2/rapid.hpp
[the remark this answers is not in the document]
fegorsch: > yes, but rmsd == 0 does not happen in practice

as `cols` has pixel-accuracy, that doesn't seem sooo improbable to me (as said: ignoring float-inaccuracies).

but leaving that aside, my original point is that "always do five iterations" is not a good default-behaviour in my eyes, as a "good" solution is probably found earlier in a lot of cases. I.e. I don't care about the exact default value of `epsilon`, but I think it should be considered by default.

modules/rapid/src/histogram.cpp
```diff
@@ -0,0 +1,226 @@
...
+namespace rapid
+{
+static void compute1DCanny(const cv::Mat& src, cv::Mat& dst, uchar threshold)
+{
+    compute1DSobel(src, dst);
+
+    // step2: compute 1D non-maximum suppression + threshold
+    for (int i = 0; i < dst.rows; i++)
+    {
+        for (int j = 1; j < dst.cols - 1; j++)
```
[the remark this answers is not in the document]
paroj: they are zero

modules/rapid/src/histogram.cpp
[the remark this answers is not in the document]
paroj: yes

paroj:
@alalek this PR is complete from my side

PR #2599, opened 2020-07-11
Create CUDA backend support for dnn_superres

Saafke:
Hi there,

This PR allows users to set CUDA backend in the dnn_superres module. Now users can use their GPU's for inference while upscaling images via the neural networks. I've tried a few images, and I've seen around an 6x increase in speed (GTX 1080Ti).

I'm not sure this is the best way of doing this. Perhaps it is better to also allow users to choose other types of backends? What do you think? I'll wait for the feedback before adding a tutorial (and test?) for this.

### DIFF
--- a/modules/dnn_superres/src/dnn_superres.cpp
+++ b/modules/dnn_superres/src/dnn_superres.cpp
@@ -91,6 +91,13 @@ void DnnSuperResImpl::setModel(const String& algo, int scale)
     this->alg = algo;
 }
 
+void DnnSuperResImpl::setCUDA()
+{
+    net.setPreferableBackend(cv::dnn::DNN_BACKEND_CUDA);
+    net.setPreferableTarget(cv::dnn::DNN_TARGET_CUDA);
+    CV_LOG_INFO(NULL, "Successfully set CUDA backend and target.");
+}
+
 void DnnSuperResImpl::upsample(InputArray img, OutputArray result)
 {
     if (net.empty())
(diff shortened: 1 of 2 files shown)


modules/dnn_superres/src/dnn_superres.cpp
```diff
@@ -91,6 +91,13 @@ void DnnSuperResImpl::setModel(const String& algo, int scale)
     this->alg = algo;
 }
 
+void DnnSuperResImpl::setCUDA()
+{
+    net.setPreferableBackend(cv::dnn::DNN_BACKEND_CUDA);
+    net.setPreferableTarget(cv::dnn::DNN_TARGET_CUDA);
```
alalek, line 97: More robust solution is to proxy these calls instead:
- setPreferableBackend
- setPreferableTarget

Otherwise we will get soon setOpenCL, setCUDA_FP16, etc.

Saafke: You mean like this? (See new commit)

Saafke: Okay, good. Should I add a note in the tutorials somewhere? And I don't think a test is needed (or possible) for this right?

alalek: I believe this change is straightforward and doesn't require extra documentation or tests.

tompollok:
@saafke which networks have you tried for upscaling? Are the networks publicly available?

Saafke:
> @Saafke which networks have you tried for upscaling? Are the networks publicly available?

@tompollok It works with all four available models. Check the [tutorial](https://docs.opencv.org/master/d8/df8/tutorial_table_of_content_dnn_superres.html) for how to use this module. This [blog](https://towardsdatascience.com/deep-learning-based-super-resolution-with-opencv-4fd736678066) is also helpful.

PR #2610, opened 2020-07-21
Simplified generateWindowName

EricFlorin:
Fixes https://github.com/opencv/opencv/issues/10784

Removed a few lines which prevented "Viz - " to be prepended to the names "Viz" and "", as pointed out in the original issue.

Since this issue is 2 years old, I am not sure if it was for an older version of opencv or more recent versions. I believe the 3.4 version of opencv in opencv/opencv may still have the error, since `viz` is one of modules packaged with 3.4, but I have not checked it yet.


alalek:
@EricFlorin Thank you for contribution!

I believe we should clarify documentation instead according to issue's description.

EricFlorin:
Thanks for the review @alalek! I will revert the changes I made and edit the documentation. Give me two to three days because it's been a busy week this week, but I will get it done.

alalek (approved):

PR #2626, opened 2020-08-09
[ISCAS] loadPLYSimple return Zero Mat after reading PLY file

TachikakaMin:

Function Doc: https://docs.opencv.org/master/d9/d25/group__surface__matching.html#gad893719cc47a72b24482f2bfdebb409e

This function will return a zero Mat after it reads .ply file. 

Example File: https://www.artec3d.com/3d-models/vw-bus#download


alalek:
OpenCV loader expects text format only.

> format ascii 1.0

Your file is in binary format:

> format binary_little_endian 1.0


So this patch is not complete.
- determine format properly. throw error in case of unknowns
- add dedicated loader for binary data (with proper processing of little-endian / big-endian byte orders)

TachikakaMin:
Oh, that's true. I will do it later.

mihaibujanca:
@TachikakaMin The title of the PR and the content don't match. There should be a PR for the `loadPLYSimple` problem, but the commits relate to the planar simplification project.

vpisarev:
@TachikakaMin, @mihaibujanca; let me close this pull request; the code needs a substantial rewrite. If/when it's done, feel free to reopen this PR or create a new one

PR #2629, opened 2020-08-12
TSDF Optimization

DumDereDum:
## This PR changes
/----[ WI ]----/

- Reducing memory consumption:
  * Create new type (half)
  * Replace float with new type - TsdfType(int8_t):
     + CPU 
     + GPU
     + INTRINSICS
     + HashTSDF
  * Replace int with uchar:
     + CPU 
     + GPU
     + INTRINSICS
     + HashTSDF
- TSDF optimization:
  * Speeded up the calculation of the pixNorm variable by incrementing this variable on the fixed number
- HashTSDF optimization:
  * Vectorized interpolate function
  * Reduced the call to the at() function, acceleration of the getNormalVoxel() function by 33%, and the algorithm by 10%
- Tests:
  * Create new struct "Settings" for performance test (perf_rgbd)
- Other minor fixes

### DIFF
--- a/modules/rgbd/perf/perf_tsdf.cpp
+++ b/modules/rgbd/perf/perf_tsdf.cpp
@@ -169,42 +169,40 @@ Ptr<Scene> Scene::create(Size sz, Matx33f _intr, float _depthFactor)
     return makePtr<SemisphereScene>(sz, _intr, _depthFactor);
 }
 
-std::vector<Affine3f> getPoses(Ptr<kinfu::Params> _params )
+struct Settings
 {
     Ptr<kinfu::Params> _params;
+    Ptr<kinfu::Volume> volume;
+    Ptr<Scene> scene;
+    std::vector<Affine3f> poses;
+};
+
+Settings setSettings(bool useHashTSDF)
+{
+    Settings res;
     if (useHashTSDF)
-        _params = kinfu::Params::hashTSDFParams(true);
+        res._params = kinfu::Params::hashTSDFParams(true);
     else
-        _params = kinfu::Params::coarseParams();
+        res._params = kinfu::Params::coarseParams();
 
-    Ptr<kinfu::Volume> volume = kinfu::makeVolume(_params->volumeType, _params->voxelSize, _params->volumePose.matrix,
-        _params->raycast_step_factor, _params->tsdf_trunc_dist, _params->tsdf_max_weight,
-        _params->truncateThreshold, _params->volumeDims);
+    res.volume = kinfu::makeVolume(res._params->volumeType, res._params->voxelSize, res._params->volumePose.matrix,
+                                res._params->raycast_step_factor, res._params->tsdf_trunc_dist, res._params->tsdf_max_weight,
+                                res._params->truncateThreshold, res._params->volumeDims);
 
-    Ptr<Scene> scene = Scene::create(_params->frameSize, _params->intr, _params->depthFactor);
-    std::vector<Affine3f> poses = scene->getPoses();
-
-    return poses;
+    res.scene = Scene::create(res._params->frameSize, res._params->intr, res._params->depthFactor);
+    res.poses = res.scene->getPoses();
 }
 
 PERF_TEST(Perf_TSDF, integrate)
 {
-    Ptr<kinfu::Params> _params = kinfu::Params::coarseParams();
-
-    Ptr<kinfu::Volume> volume = kinfu::makeVolume(_params->volumeType, _params->voxelSize, _params->volumePose.matrix,
-        _params->raycast_step_factor, _params->tsdf_trunc_dist, _params->tsdf_max_weight,
-        _params->truncateThreshold, _params->volumeDims);
-
-    Ptr<Scene> scene = Scene::create(_params->frameSize, _params->intr, _params->depthFactor);
-    std::vector<Affine3f> poses = scene->getPoses();
-
-    for (size_t i = 0; i < poses.size(); i++)
+    Settings settings = setSettings(false);
+    for (size_t i = 0; i < settings.poses.size(); i++)
     {
         UMat _points, _normals;
-        Matx44f pose = poses[i].matrix;
-        Mat depth = scene->depth(pose);
+        Matx44f pose = settings.poses[i].matrix;
+        Mat depth = settings.scene->depth(pose);
         startTimer();
-        volume->integrate(depth, _params->depthFactor, pose, _params->intr);
+        settings.volume->integrate(depth, settings._params->depthFactor, pose, settings._params->intr);
         stopTimer();
     }
     SANITY_CHECK_NOTHING();
@@ -212,24 +210,16 @@ PERF_TEST(Perf_TSDF, integrate)
 
 PERF_TEST(Perf_TSDF, raycast)
 {
-    Ptr<kinfu::Params> _params = kinfu::Params::coarseParams();
-
-    Ptr<kinfu::Volume> volume = kinfu::makeVolume(_params->volumeType, _params->voxelSize, _params->volumePose.matrix,
-        _params->raycast_step_factor, _params->tsdf_trunc_dist, _params->tsdf_max_weight,
-        _params->truncateThreshold, _params->volumeDims);
-
-    Ptr<Scene> scene = Scene::create(_params->frameSize, _params->intr, _params->depthFactor);
-    std::vector<Affine3f> poses = scene->getPoses();
-
-    for (size_t i = 0; i < poses.size(); i++)
+    Settings settings = setSettings(false);
+    for (size_t i = 0; i < settings.poses.size(); i++)
     {
         UMat _points, _normals;
-        Matx44f pose = poses[i].matrix;
-        Mat depth = scene->depth(pose);
+        Matx44f pose = settings.poses[i].matrix;
+        Mat depth = settings.scene->depth(pose);
 
-        volume->integrate(depth, _params->depthFactor, pose, _params->intr);
+        settings.volume->integrate(depth, settings._params->depthFactor, pose, settings._params->intr);
         startTimer();
-        volume->raycast(pose, _params->intr, _params->frameSize, _points, _normals);
+        settings.volume->raycast(pose, settings._params->intr, settings._params->frameSize, _points, _normals);
         stopTimer();
     }
     SANITY_CHECK_NOTHING();
@@ -237,22 +227,14 @@ PERF_TEST(Perf_TSDF, raycast)
 
 PERF_TEST(Perf_HashTSDF, integrate)
 {
-    Ptr<kinfu::Params> _params = kinfu::Params::hashTSDFParams(true);
-
-    Ptr<kinfu::Volume> volume = kinfu::makeVolume(_params->volumeType, _params->voxelSize, _params->volumePose.matrix,
-        _params->raycast_step_factor, _params->tsdf_trunc_dist, _params->tsdf_max_weight,
-        _params->truncateThreshold, _params->volumeDims);
-
-    Ptr<Scene> scene = Scene::create(_params->frameSize, _params->intr, _params->depthFactor);
-    std::vector<Affine3f> poses = scene->getPoses();
-
-    for (size_t i = 0; i < poses.size(); i++)
+    Settings settings = setSettings(true);
+    for (size_t i = 0; i < settings.poses.size(); i++)
     {
         UMat _points, _normals;
-        Matx44f pose = poses[i].matrix;
-        Mat depth = scene->depth(pose);
+        Matx44f pose = settings.poses[i].matrix;
+        Mat depth = settings.scene->depth(pose);
         startTimer();
-        volume->integrate(depth, _params->depthFactor, pose, _params->intr);
+        settings.volume->integrate(depth, settings._params->depthFactor, pose, settings._params->intr);
         stopTimer();
     }
     SANITY_CHECK_NOTHING();
@@ -260,24 +242,16 @@ PERF_TEST(Perf_HashTSDF, integrate)
 
 PERF_TEST(Perf_HashTSDF, raycast)
 {
-    Ptr<kinfu::Params> _params = kinfu::Params::hashTSDFParams(true);
-
-    Ptr<kinfu::Volume> volume = kinfu::makeVolume(_params->volumeType, _params->voxelSize, _params->volumePose.matrix,
-        _params->raycast_step_factor, _params->tsdf_trunc_dist, _params->tsdf_max_weight,
-        _params->truncateThreshold, _params->volumeDims);
-
-    Ptr<Scene> scene = Scene::create(_params->frameSize, _params->intr, _params->depthFactor);
-    std::vector<Affine3f> poses = scene->getPoses();
-
-    for (size_t i = 0; i < poses.size(); i++)
+    Settings settings = setSettings(true);
+    for (size_t i = 0; i < settings.poses.size(); i++)
     {
         UMat _points, _normals;
-        Matx44f pose = poses[i].matrix;
-        Mat depth = scene->depth(pose);
+        Matx44f pose = settings.poses[i].matrix;
+        Mat depth = settings.scene->depth(pose);
 
-        volume->integrate(depth, _params->depthFactor, pose, _params->intr);
+        settings.volume->integrate(depth, settings._params->depthFactor, pose, settings._params->intr);
         startTimer();
-        volume->raycast(pose, _params->intr, _params->frameSize, _points, _normals);
+        settings.volume->raycast(pose, settings._params->intr, settings._params->frameSize, _points, _normals);
         stopTimer();
     }
     SANITY_CHECK_NOTHING();
--- a/modules/rgbd/perf/perf_tsdf.cpp
+++ b/modules/rgbd/perf/perf_tsdf.cpp
@@ -191,6 +191,8 @@ Settings setSettings(bool useHashTSDF)
 
     res.scene = Scene::create(res._params->frameSize, res._params->intr, res._params->depthFactor);
     res.poses = res.scene->getPoses();
+    
+    return res;
 }
 
 PERF_TEST(Perf_TSDF, integrate)
@@ -198,7 +200,6 @@ PERF_TEST(Perf_TSDF, integrate)
     Settings settings = setSettings(false);
     for (size_t i = 0; i < settings.poses.size(); i++)
     {
-        UMat _points, _normals;
         Matx44f pose = settings.poses[i].matrix;
         Mat depth = settings.scene->depth(pose);
         startTimer();
@@ -228,9 +229,9 @@ PERF_TEST(Perf_TSDF, raycast)
 PERF_TEST(Perf_HashTSDF, integrate)
 {
     Settings settings = setSettings(true);
+
     for (size_t i = 0; i < settings.poses.size(); i++)
     {
-        UMat _points, _normals;
         Matx44f pose = settings.poses[i].matrix;
         Mat depth = settings.scene->depth(pose);
         startTimer();
--- a/modules/rgbd/perf/perf_tsdf.cpp
+++ b/modules/rgbd/perf/perf_tsdf.cpp
@@ -169,35 +169,33 @@ Ptr<Scene> Scene::create(Size sz, Matx33f _intr, float _depthFactor)
     return makePtr<SemisphereScene>(sz, _intr, _depthFactor);
 }
 
-struct Settings
+class Settings
 {
+public:
     Ptr<kinfu::Params> _params;
     Ptr<kinfu::Volume> volume;
     Ptr<Scene> scene;
     std::vector<Affine3f> poses;
+
+    Settings(bool useHashTSDF)
+    {
+        if (useHashTSDF)
+            _params = kinfu::Params::hashTSDFParams(true);
+        else
+            _params = kinfu::Params::coarseParams();
+
+        volume = kinfu::makeVolume(_params->volumeType, _params->voxelSize, _params->volumePose.matrix,
+            _params->raycast_step_factor, _params->tsdf_trunc_dist, _params->tsdf_max_weight,
+            _params->truncateThreshold, _params->volumeDims);
+
+        scene = Scene::create(_params->frameSize, _params->intr, _params->depthFactor);
+        poses = scene->getPoses();
+    }
 };
-
-Settings setSettings(bool useHashTSDF)
-{
-    Settings res;
-    if (useHashTSDF)
-        res._params = kinfu::Params::hashTSDFParams(true);
-    else
-        res._params = kinfu::Params::coarseParams();
-
-    res.volume = kinfu::makeVolume(res._params->volumeType, res._params->voxelSize, res._params->volumePose.matrix,
-                                res._params->raycast_step_factor, res._params->tsdf_trunc_dist, res._params->tsdf_max_weight,
-                                res._params->truncateThreshold, res._params->volumeDims);
-
-    res.scene = Scene::create(res._params->frameSize, res._params->intr, res._params->depthFactor);
-    res.poses = res.scene->getPoses();
-
-    return res;
-}
 
 PERF_TEST(Perf_TSDF, integrate)
 {
-    Settings settings = setSettings(false);
+    Settings settings(false);
     for (size_t i = 0; i < settings.poses.size(); i++)
     {
         Matx44f pose = settings.poses[i].matrix;
@@ -211,7 +209,7 @@ PERF_TEST(Perf_TSDF, integrate)
 
 PERF_TEST(Perf_TSDF, raycast)
 {
-    Settings settings = setSettings(false);
+    Settings settings(false);
     for (size_t i = 0; i < settings.poses.size(); i++)
     {
         UMat _points, _normals;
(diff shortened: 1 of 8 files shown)


modules/rgbd/src/tsdf.cpp
```diff
@@ -1116,7 +1209,7 @@ void TSDFVolumeCPU::fetchNormals(InputArray _points, OutputArray _normals) const
 
 ///////// GPU implementation /////////
 
-#ifdef HAVE_OPENCL
+#ifdef HAVE_OPENCL_
 TSDFVolumeGPU::TSDFVolumeGPU(float _voxelSize, cv::Matx44f _pose, float _raycastStepFactor, float _truncDist, int _maxWeight,
                              Point3i _resolution) :
     TSDFVolume(_voxelSize, _pose, _raycastStepFactor, _truncDist, _maxWeight, _resolution, false)
```
savuor, line 1215: No `CV_32FC2`, this corresponds to old voxel data types (float32 + int32), now there needed something 16-bit-sized.

modules/rgbd/src/opencl/tsdf.cl
```diff
@@ -4,10 +4,63 @@
...
+
+struct TsdfVoxel
+{
+    TsdfType tsdf;
+    WeightType weight;
+};
+
+static inline TsdfType floatToTsdf(float num)
+{
+    int8_t res = (int8_t) ( (int) (num * (-128)) );
```
savuor, line 20: extra conversion to int

DumDereDum: done

modules/rgbd/src/tsdf.cpp
DumDereDum: done

modules/rgbd/src/hash_tsdf.cpp
```diff
@@ -16,10 +16,26 @@
...
+{
+    //CV_Assert(-1 < num <= 1);
+    int8_t res = int8_t(num * (-128));
+    res = res ? res : (num < 0 ? 1 : -1);
+    return res;
+}
+
+static inline float tsdfToFloat(TsdfType num)
+{
+    return float(num) * (-0.0078125f); // * (-1/128)
```
savuor, line 36: It's not necessary to calculate the constant; keeping it as `(-1.f/128.f)` is more human-readable and compiles in the same constant.

modules/rgbd/src/hash_tsdf.cpp
```diff
@@ -16,10 +16,26 @@
...
+
 namespace cv
 {
 namespace kinfu
 {
+
+static inline TsdfType floatToTsdf(float num)
+{
+    //CV_Assert(-1 < num <= 1);
+    int8_t res = int8_t(num * (-128));
```
savuor, line 29: please use `(-128.f)` instead

modules/rgbd/src/tsdf.cpp
```diff
@@ -11,9 +11,53 @@
...
+    //CV_Assert(-1 < num <= 1);
+    int8_t res = int8_t(num * (-128));
+    res = res ? res : (num < 0 ? 1 : -1);
+    return res;
+}
+
+static inline float tsdfToFloat(TsdfType num)
+{
+    //return float(num) / (-128);
+    return float(num) * (-0.0078125f); // * (-1/128)
```
savuor, line 56: The same notion about constants as in `hash_tsdf.cpp` file

modules/rgbd/src/tsdf.cpp
```diff
@@ -501,20 +562,33 @@ inline TsdfType TSDFVolumeCPU::interpolateVoxel(Point3f p) const
...
-    for(int i = 0; i < 8; i++)
-        vx[i] = volData[neighbourCoords[i] + coordBase].tsdf;
+    float v00 = vx[0] + tz * (vx[1] - vx[0]);
+    float v01 = vx[2] + tz * (vx[3] - vx[2]);
+    float v10 = vx[4] + tz * (vx[5] - vx[4]);
+
+    float v0 = v00 + ty * (v01 - v00);
+
+    return v0 + tx * (v10 - v00);
+*/
```
savuor, line 577: Please remove commented code

modules/rgbd/src/tsdf.cpp
```diff
@@ -11,9 +11,53 @@
...
+static inline float tsdfToFloat(TsdfType num)
+{
+    if(!num) return 0.f;
+    Cv32suf u;
+    u.f = float(num);
+    u.u -= 7 << 23; // divide by 128 == sub 7 from exponent
+    u.u ^= 1u << 31; // negate
+    return u.f;
+}
+*/
```
savuor, line 35: Please remove commented code

modules/rgbd/src/tsdf.cpp
```diff
@@ -11,9 +11,53 @@
...
+
+static inline float tsdfToFloat(TsdfType num)
+{
+    //return float(num) / (-128);
+    return float(num) * (-0.0078125f); // * (-1/128)
+}
 
 TSDFVolume::TSDFVolume(float _voxelSize, Matx44f _pose, float _raycastStepFactor, float _truncDist,
-                       int _maxWeight, Point3i _resolution, bool zFirstMemOrder)
+                       WeightType _maxWeight, Point3i _resolution, bool zFirstMemOrder)
```
savuor, line 60: Since `WeightType` is `int8_t` now, we should make an assert about maxWeight - right before the place where converted down from `int` to `int8_t` (obviously not here but probably in the factory method).

modules/rgbd/src/tsdf.cpp
```diff
@@ -1121,7 +1206,7 @@ TSDFVolumeGPU::TSDFVolumeGPU(float _voxelSize, cv::Matx44f _pose, float _raycast
                              Point3i _resolution) :
     TSDFVolume(_voxelSize, _pose, _raycastStepFactor, _truncDist, _maxWeight, _resolution, false)
 {
-    volume = UMat(1, volResolution.x * volResolution.y * volResolution.z, CV_32FC2);
+    volume = UMat(1, volResolution.x * volResolution.y * volResolution.z, CV_8UC2 /*CV_32FC2*/);
```
savuor, line 1209: Please remove that extra comment, it's confusing like `CV_32FC2` is a valid alternative

modules/rgbd/src/tsdf.hpp
```diff
@@ -19,32 +19,36 @@ namespace kinfu
...
    public:
+
     Point3i volResolution;
     WeightType maxWeight;
 
     Point3f volSize;
     float truncDist;
     Vec4i volDims;
     Vec8i neighbourCoords;
+    Vec6i _neighbourCoords;
```
savuor, line 51: Do we still need this?

modules/rgbd/src/tsdf.cpp
savuor: `makeTSDFVolume` is the place to add such an assert

modules/rgbd/src/tsdf.cpp
```diff
@@ -52,6 +96,15 @@ TSDFVolume::TSDFVolume(float _voxelSize, Matx44f _pose, float _raycastStepFactor
         volDims.dot(Vec4i(1, 1, 0)),
         volDims.dot(Vec4i(1, 1, 1))
     );
+
```
savuor, line 99: The same: unused code

modules/rgbd/src/opencl/tsdf.cl
```diff
@@ -4,10 +4,63 @@
...
+
+static inline float tsdfToFloat(TsdfType num)
+{
+    if(!num) return 0.f;
+    Cv32suf u;
+    u.f = (float) num;
+    u.u -= 7 << 23; // divide by 128 == sub 7 from exponent
+    u.u ^= 1u << 31; // negate
+    return u.f;
+}
```
savuor, line 58: Extra comments

modules/rgbd/src/hash_tsdf.cpp
```diff
@@ -289,17 +305,17 @@ inline TsdfType HashTSDFVolumeCPU::interpolateVoxel(const cv::Point3f& point) co
     float ty = point.y - iy;
     float tz = point.z - iz;
 
-    TsdfType vx[8];
+    float vx[8];
     for (int i = 0; i < 8; i++)
-        vx[i] = at(neighbourCoords[i] * voxelSize + point).tsdf;
+        vx[i] = tsdfToFloat(at(neighbourCoords[i] * voxelSize + point).tsdf);
```
savuor, line 310: The function `at` performs several calls to hash table. Maybe manual inlining `at` + pushing hash table calls before the loop would give perf boost.

modules/rgbd/src/hash_tsdf.cpp
```diff
@@ -317,19 +333,116 @@ inline Point3f HashTSDFVolumeCPU::getNormalVoxel(Point3f point) const
...
+        if (i % 4 > 1)
+            c1_n_vx[i] = tsdfToFloat(at(neighbourCoords[i] + c1_pointNext).tsdf);
+        else 
+            c1_p_vx[i] = tsdfToFloat(at(neighbourCoords[i] + c1_pointPrev).tsdf);
+
+    for (int i = 0; i < 8; i++)
+        if (i % 2 == 1)
+            c2_n_vx[i] = tsdfToFloat(at(neighbourCoords[i] + c2_pointNext).tsdf);
+        else
+            c2_p_vx[i] = tsdfToFloat(at(neighbourCoords[i] + c2_pointPrev).tsdf);
```
savuor, line 416: 1. Can be put into one 0 to 8 loop
2. The same about hash table calls

modules/rgbd/perf/perf_tsdf.cpp
```diff
@@ -219,4 +219,54 @@ PERF_TEST(Perf_TSDF, raycast)
...
+        Matx44f pose = poses[i].matrix;
+        Mat depth = scene->depth(pose);
+
+        volume->integrate(depth, _params->depthFactor, pose, _params->intr);
+        startTimer();
+        volume->raycast(pose, _params->intr, _params->frameSize, _points, _normals);
+        stopTimer();
+    }
+    SANITY_CHECK_NOTHING();
+}
```
savuor, line 270: Can these 2 functions be joined together as it was done in accuracy tests?

modules/rgbd/src/hash_tsdf.cpp
```diff
@@ -317,19 +333,116 @@ inline Point3f HashTSDFVolumeCPU::getNormalVoxel(Point3f point) const
...
+
```
savuor, line 417: The same about at inlining and hash table calls

modules/rgbd/src/hash_tsdf.cpp
DumDereDum: done

modules/rgbd/src/hash_tsdf.cpp
DumDereDum: done

modules/rgbd/src/tsdf.cpp
DumDereDum: done

modules/rgbd/src/tsdf.cpp
DumDereDum: done

modules/rgbd/src/tsdf.cpp
DumDereDum: done

modules/rgbd/src/tsdf.cpp
DumDereDum: done

modules/rgbd/src/tsdf.cpp
DumDereDum: done

modules/rgbd/src/tsdf.hpp
DumDereDum: done

modules/rgbd/src/tsdf.cpp
DumDereDum: done

DumDereDum: done

modules/rgbd/src/opencl/tsdf.cl
DumDereDum: done

modules/rgbd/perf/perf_tsdf.cpp
DumDereDum: no, i tried :(

modules/rgbd/src/tsdf.cpp
```diff
@@ -394,21 +422,20 @@ struct IntegrateInvoker : ParallelLoopBody
...
 
                         TsdfVoxel& voxel = volDataY[z*volume.volDims[2]];
                         WeightType& weight      = voxel.weight;
                         TsdfType& value  = voxel.tsdf;
 
                         // update TSDF
-                        value  = (value*weight+tsdf) / (weight + 1);
-                        weight = min(weight + 1, volume.maxWeight);
+                        value  = floatToTsdf((tsdfToFloat(value)*weight+ tsdfToFloat(tsdf)) / (weight + 1));
+                        weight = min(int(weight + 1), int(volume.maxWeight));
```
akashsharma02, line 438: Need to make sure that `tsdf_max_weight` < 255 always, and document it.

savuor: See line 40, there was a special assert added for that. And yes, this should be documented.

modules/rgbd/src/hash_tsdf.cpp
savuor: I think there are 4 classes of points:
* Most popular case: all neighbors are inside 1 volume unit
* Neighbors share the same face of 2 volume units
* Neighbors share the same edge of 4 volume units
* Neighbors share the same vertex of 8 volume units
I don't know a flexible way to process all 4 cases, but if we cover at least the 1st one, a lot of hash table calls can be saved.

modules/rgbd/perf/perf_tsdf.cpp
DumDereDum: done, but partially

modules/rgbd/src/hash_tsdf.cpp
DumDereDum: done

modules/rgbd/perf/perf_tsdf.cpp
```diff
@@ -169,51 +169,90 @@ Ptr<Scene> Scene::create(Size sz, Matx33f _intr, float _depthFactor)
...
+struct Settings
 {
     Ptr<kinfu::Params> _params;
-    _params = kinfu::Params::coarseParams();
+    Ptr<kinfu::Volume> volume;
+    Ptr<Scene> scene;
+    std::vector<Affine3f> poses;
+};
+
+Settings setSettings(bool useHashTSDF)
```
savuor, line 180: Can be implemented as a constructor for `Settings` class

modules/rgbd/src/tsdf.cpp
```diff
@@ -388,27 +420,34 @@ struct IntegrateInvoker : ParallelLoopBody
...
+                    }
+                    if (!pixNorm_flag && z==endZ-1)
+                        {
+                            pixNorm_o = (pixNorm - pixNorm_s) / (z - startZ);
+                            pixNorm_flag = true;
+                        }
                     // norm(camPixVec) produces double which is too slow
-                    float pixNorm = sqrt(camPixVec.dot(camPixVec));
+                    if (pixNorm_flag)
+                        pixNorm += pixNorm_o;
```
savuor, line 433: The fact that this approximation works is not more than a happy coincidence. Changing a pose more radically would fail this. Here's how `pixNorm` function looks: https://www.geogebra.org/3d/pretpsdt where the typical x, y are between -1 and +1, i.e. in the area of maximal non-linearity. Imagine a straight line projected from the Oxy plane onto that figure: it would be far from a straight line.
The optimization I was talking about is pre-calculation of `pixNorm` for all pixels and accessing it instead of calculation.

modules/rgbd/src/hash_tsdf.cpp
```diff
@@ -317,19 +335,222 @@ inline Point3f HashTSDFVolumeCPU::getNormalVoxel(Point3f point) const
...
+    c2_n_vx[2] = c0_n_vx[1];
+    c2_p_vx[3] = c0_n_vx[3];
+    c2_n_vx[4] = c0_p_vx[4];
+    c2_p_vx[5] = c0_p_vx[6];
+    c2_n_vx[6] = c0_n_vx[0];
+    c2_p_vx[7] = c0_n_vx[2];
+
+    normal.x = interpolate(c0_pointNext, c0_n_vx) - interpolate(c0_pointPrev, c0_p_vx);
+    normal.y = interpolate(c1_pointNext, c1_n_vx) - interpolate(c1_pointPrev, c1_p_vx);
+    normal.z = interpolate(c2_pointNext, c2_n_vx) - interpolate(c2_pointPrev, c2_p_vx);
```
savuor, line 438: Here each subtraction of 2 `interpolate`-s can be shrunk to just 1 `interpolate`:
* that way is used in `tsdf.cpp` in `getNormalVoxel`, please refer there for details
* this works because `c{0,1,2}_pointNext` and `c{0,1,2}_pointPrev` differ by 1 in each dimension, therefore they share the same fractional part

savuor: Like this:
```
float c0_vx[8], c1_vx[8], c2_vx[8];
for(int i = 0; i < 8; i++)
{
    c0_vx[i] = c0_n_vx[i] - c0_p_vx[i];
}
normal.x = interpolate(c0_pointNext, c0_vx);
...
	```

File: modules/rgbd/src/hash_tsdf.cpp [hunk 21]
```diff
@@ -317,19 +335,222 @@ inline Point3f HashTSDFVolumeCPU::getNormalVoxel(Point3f point) const
...
+            c1_n_vx[i] = tsdfToFloat(_at_1arg(neighbourCoords[i] + c1_pointNext).tsdf);
+        else
+            c1_p_vx[i] = tsdfToFloat(_at_1arg(neighbourCoords[i] + c1_pointPrev).tsdf);
+
+    for (int i = 0; i < 8; i++)
+        if (i % 2 == 1)
+            c2_n_vx[i] = tsdfToFloat(_at_1arg(neighbourCoords[i] + c2_pointNext).tsdf);
+        else
+            c2_p_vx[i] = tsdfToFloat(_at_1arg(neighbourCoords[i] + c2_pointPrev).tsdf);
+
```
savuor, line 417: I think the code can be simplified & boosted in performance sense.
1. We can precalculate coordinate offsets of all 32 points we need instead of these complex for loops
1. We should check if **all** points we need are in the same volume unit just by checking point coordinate remainder.
2. If they are not, then the following procedure may happen:
* We can make a small array which works like a dictionary from volume unit index to corresponding iterator. We base it on the following:
 * there can't be more than 8 adjacent volume units
 * their indices can be easily mapped to 0 to 7 using coordinate parity:
  - `(x % 2) + (y % 2)*2 + (z % 2)*4`
* For each point we:
 * get its volume unit index
 * check if it's in our dictionary
 * if it's not, we make a hash table query and write a result into a dictionary
 * if it's in, we use it with that `at` function which takes an iterator
* then we rearrange these values into vectors and pass them into `interpolate()` calls

modules/rgbd/src/hash_tsdf.cpp
savuor: In fact, all interpolation coeffs are the same at each interpolation

DumDereDum: done

modules/rgbd/src/hash_tsdf.cpp
DumDereDum: done

modules/rgbd/perf/perf_tsdf.cpp
DumDereDum: done

modules/rgbd/src/tsdf.hpp
```diff
@@ -47,12 +50,14 @@ class TSDFVolume : public Volume
     Vec8i neighbourCoords;
 };
 
+cv::Mat preCalculationPixNorm(Depth depth, const Intr& intrinsics);
```
savuor, line 53: Please make this function static and do not put its declaration into the header

modules/rgbd/src/tsdf.cpp
```diff
@@ -434,20 +495,22 @@ void TSDFVolumeCPU::integrate(InputArray _depth, float depthFactor, const cv::Ma
     CV_Assert(_depth.type() == DEPTH_TYPE);
     CV_Assert(!_depth.empty());
     Depth depth = _depth.getMat();
-    IntegrateInvoker ii(*this, depth, intrinsics, cameraPose, depthFactor);
+    Mat pixNorms = preCalculationPixNorm(depth, intrinsics);
+    IntegrateInvoker ii(*this, depth, intrinsics, cameraPose, depthFactor, pixNorms);
     Range range(0, volResolution.x);
-    parallel_for_(range, ii);
+    //parallel_for_(range, ii);
+    ii(range);
```
savuor, line 502: I hope this is a debugging piece of code

modules/rgbd/src/tsdf.cpp
```diff
@@ -434,20 +495,22 @@ void TSDFVolumeCPU::integrate(InputArray _depth, float depthFactor, const cv::Ma
     CV_Assert(_depth.type() == DEPTH_TYPE);
     CV_Assert(!_depth.empty());
     Depth depth = _depth.getMat();
-    IntegrateInvoker ii(*this, depth, intrinsics, cameraPose, depthFactor);
+    Mat pixNorms = preCalculationPixNorm(depth, intrinsics);
```
savuor, line 498: Since `pixNorm` depends on image size only, it shouldn't be updated at each call. We can keep the precalculated `pixNorm` inside a class with intrinsics and frame size and update it only if these parameters are changed.

modules/rgbd/src/tsdf.cpp
```diff
@@ -473,7 +536,8 @@ inline TsdfType TSDFVolumeCPU::interpolateVoxel(const v_float32x4& p) const
     for(int i = 0; i < 8; i++)
         vx[i] = volData[neighbourCoords[i] + coordBase].tsdf;
 
-    v_float32x4 v0246(vx[0], vx[2], vx[4], vx[6]), v1357(vx[1], vx[3], vx[5], vx[7]);
+    v_float32x4 v0246 = tsdfToFloat_INTR(v_int32x4(vx[0], vx[2], vx[4], vx[6]));
+    v_float32x4 v1357 = tsdfToFloat_INTR(v_int32x4(vx[1], vx[3], vx[5], vx[7]));
```
savuor, line 540: Please replace these places by `v_load_deinterleave` as in `hash_tsdf.cpp` file

modules/rgbd/src/tsdf.cpp
```diff
@@ -567,7 +632,8 @@ inline v_float32x4 TSDFVolumeCPU::getNormalVoxel(const v_float32x4& p) const
             vx[i] = volData[neighbourCoords[i] + coordBase + 1*dim].tsdf -
                     volData[neighbourCoords[i] + coordBase - 1*dim].tsdf;
 
-        v_float32x4 v0246(vx[0], vx[2], vx[4], vx[6]), v1357(vx[1], vx[3], vx[5], vx[7]);
+        v_float32x4 v0246 = tsdfToFloat_INTR(v_int32x4(vx[0], vx[2], vx[4], vx[6]));
+        v_float32x4 v1357 = tsdfToFloat_INTR(v_int32x4(vx[1], vx[3], vx[5], vx[7]));
```
savuor, line 636: The same about `v_load_deinterleave`

modules/rgbd/src/tsdf.cpp
```diff
@@ -388,27 +436,38 @@ struct IntegrateInvoker : ParallelLoopBody
...
                     depthType v = bilinearDepth(depth, projected);
-                    if(v == 0)
+                    if (v == 0) {
                         continue;
-
-                    // norm(camPixVec) produces double which is too slow
-                    float pixNorm = sqrt(camPixVec.dot(camPixVec));
+                    }
+                    float pixNorm;
+                    if (false)
```
savuor, line 443: Please replace this `false` by a meaningful const bool

DumDereDum: done

modules/rgbd/src/tsdf.cpp
DumDereDum: done

modules/rgbd/src/tsdf.cpp
DumDereDum: done

modules/rgbd/src/tsdf.cpp
DumDereDum: We  have float at start in `hash_tsdf.cpp` , but in `tsdf.cpp` we have int_8t

modules/rgbd/src/tsdf.hpp
DumDereDum: done

savuor (approved):

savuor:
👍

alalek:
Looks like there are failed tests in Debug build: #2675

PR #2630, opened 2020-08-13
feat: Modernized python code using `futurize`

ckaran:
[futurize](https://www.python-future.org/futurize.html) is a tool that is designed to make python 2 code compatible with both python 2 and python 3.  This commit contains the output of that tool, which I've inspected and manually adjusted to remove spurious fixes (e.g., print statements that were written as functions in the python 2 code would get an extra set of parenthesis which needed to be manually removed).

- [ ] The PR is proposed to proper branch - **UNKNOWN, SEE COMMENTS BELOW**
- [ ] There is reference to original bug report and related work - **I didn't create a bug for this, I just made the updates directly.**
      Patch to opencv_extra has the same branch name. - **See comments below**
- [ ] The feature is well documented and sample code can be built with the project CMake - **Not applicable**

### Comments

I based this off of `master`, but I don't know if that is the correct branch to base this off of.  

**I CANNOT TEST THIS CODE!** For reasons I don't understand, I cannot get opencv to compile on my system natively.  One of my coworkers has it working under docker, and that image seems to work, but I don't know enough about docker to modify it to use this repository directly.  Please test this before merging it in.

### DIFF
--- a/modules/aruco/misc/pattern_generator/MarkerPrinter.py
+++ b/modules/aruco/misc/pattern_generator/MarkerPrinter.py
@@ -4,6 +4,12 @@
 #
 # Copyright (c) 2019, Josh Chien. All rights reserved.
 
+from __future__ import print_function
+from __future__ import division
+from builtins import str
+from builtins import range
+from builtins import object
+from past.utils import old_div
 from argparse import ArgumentParser
 import numpy as np
 from PIL import Image
@@ -51,7 +57,7 @@ def SaveArucoDictBytesList(filePath = "arucoDictBytesList.npz"):
 
     return None
 
-class MarkerPrinter:
+class MarkerPrinter(object):
 
     debugMode = None # "LINE" "BLOCK"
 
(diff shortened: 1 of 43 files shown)


alalek:
Good PR:
- must fix actual problem (you should have evidences which part don't work)
- should contain minimal patch which is easy for review (nobody want to recheck 40+ files)
- avoid unnecessary changes (just tool run is not enough)

ckaran:
@alalek 

> Good PR:
> 
>     * must fix actual problem (you should have evidences which part don't work)

Understood.  I'll work to retest this under a stock Ubuntu 20.04 system, which doesn't have python 2 on it.  If there are problems, I'll work on them directly.

>     * should contain minimal patch which is easy for review (nobody want to recheck 40+ files)

Would you prefer that I update individual files and make PRs for each of them instead?  I can do that, although it will be slower to do so.

>     * avoid unnecessary changes (just tool run is not enough)

Understood, I'll review the tool runs and remove any changes that aren't strictly needed to get everything working under Python 3.

Thanks for the review!

PR #2634, opened 2020-08-15
Fix mctkernel issues

pemmanuelviel:

### DIFF
--- a/modules/stereo/src/descriptor.hpp
+++ b/modules/stereo/src/descriptor.hpp
@@ -78,14 +78,14 @@ namespace cv
                 {
                     if (image[i][rrWidth + jj] > image[i][rWidth + j])
                     {
-                        c[i] = c[i] + 1;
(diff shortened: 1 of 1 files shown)


pemmanuelviel:
Previous code was not making sense as, in case
l.114:  if (image[i][rrWidth + jj] > image[i][rWidth + j] - t)
condition was not fulfilled, we were checking
l.119:  else if (image[i][rWidth + j] - t < image[i][rrWidth + jj] && image[i][rWidth + j] + t >= image[i][rrWidth + jj])
what is equivalent to:
else if (   image[i][rrWidth + jj] > image[i][rWidth + j] - t
         && image[i][rrWidth + jj] <= image[i][rWidth + j] + t )
Obviously we would never get this condition true if condition at l.114 is false.

pemmanuelviel:
Checked with the author Muresan Mircea Paul that confirms the change

PR #2642, opened 2020-08-17
Fix mistake between 0x11 and 0b11 and change |=1 for +=1, increasing …

pemmanuelviel:
…chance compiler will use ++


pemmanuelviel:
Indeed...

Le lun. 17 août 2020 à 12:44, Maksim Shabunin <notifications@github.com> a
écrit :

PR #2644, opened 2020-08-19
Added some helper function in 'mcc'

AjitPant:
Earlier the CChecker class only stores the four corners of the chart, others were to be calculated by the user using perspective transform. Now the centers of each squares in the chart, along with a crop of their center is stored, also the loss per patch is also stored, which can be used by user. Also removed some code duplication in transform_points_forward.

- [ X ] I agree to contribute to the project under OpenCV (BSD) License.
- [ X ] To the best of my knowledge, the proposed patch is not based on a code under GPL or other license that is incompatible with OpenCV

### DIFF
--- a/modules/mcc/src/checker_model.cpp
+++ b/modules/mcc/src/checker_model.cpp
@@ -31,6 +31,9 @@
 #include "checker_model.hpp"
 #include "dictionary.hpp"
 
+#include<iostream>
+using namespace std;
+
 namespace cv
 {
 namespace mcc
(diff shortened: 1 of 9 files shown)


modules/mcc/src/checker_model.cpp
```diff
@@ -420,7 +420,7 @@ void CCheckerImpl::setPatchCenters(std::vector<Point2f> _patchCenters)
 }
 void CCheckerImpl::setPatchBoundingBoxes(std::vector<Point2f> _patchBoundingBoxes)
 {
-    //Use of this function is not recommended as it can cause contradictory information in patBoundingBoxes and bo
+    //Use of this function is not recommended as it can cause contradictory information in patBoundingBoxes and box
```
garybradski, line 427: I don't really know the context, but is this function then only meant to be used internally. Why allow a function that can make the state fo the program contradictory? Make it private instead?

AjitPant: Currently, all the geometric details about the chart are calculated entirely from the position of corners (calculate() function does this). But sometimes giving the users the ability to correct some of the details might be better. Like if the user does not want to extract the colour from the centre but from some corner part or maybe if some prediction is wrong, they can change the location of bounding boxes to fit their need. Also perhaps in future, finding corners using centre positions of each patch can be added, not sure if it will be useful though.

wezeyf (approved):

wezeyf (approved):

PR #2657, opened 2020-09-04
ximgproc: add python wrappers for disparity functions

berak:

questions:
* should the license be changed ?
* does it need a python test ? (there is already a [c++ one here](https://github.com/opencv/opencv_contrib/blob/master/modules/ximgproc/test/test_disparity_wls_filter.cpp#L19))

### DIFF
--- a/modules/ximgproc/misc/python/test/test_disparity.py
+++ b/modules/ximgproc/misc/python/test/test_disparity.py
@@ -1,0 +1,24 @@
+#!/usr/bin/env python
+import cv2 as cv
+import numpy as np
+
+from tests_common import NewOpenCVTests
+
+class disparity_test(NewOpenCVTests):
+    def test_disp(self):
+        # readGT
+        ret,GT = cv.ximgproc.readGT(self.find_file("cv/disparityfilter/GT.png"))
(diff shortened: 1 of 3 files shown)


alalek:
3.4 code is under BSD license. 4.5+ is under Apache 2.0 license with BSD code snippets.
Python test should check binding interface only (input type, output type and output size is enough). No need to validate function accuracy.

berak:
> 3.4 code is under BSD license.

right, i forgot ;)

modules/ximgproc/misc/python/test/test_disparity.py
```diff
@@ -0,0 +1,24 @@
...
+import cv2 as cv
+import numpy as np
+
+from tests_common import NewOpenCVTests
+
+class disparity_test(NewOpenCVTests):
+    def test_disp(self):
+        # readGT
+        ret,GT = cv.ximgproc.readGT(self.find_file("cv/disparityfilter/GT.png"))
+        self.assertFalse(ret) # returns 0 on success!
```
alalek, line 11:
```
self.assertEqual(ret, 0)
```
?

alalek (approved):
Well done! Thank you :+1:

PR #2687, opened 2020-09-27
xfeatures2d: fix MSVS2019 build

alalek:
- MSVC 19.27.29111.0

Error message (points to the meaningless end of function `}`, looks like a compiler bug):
```
opencv_contrib\modules\xfeatures2d\src\stardetector.cpp(350,1): error C2109: subscript requires array or pointer type
```

```
force_builders=Win64 OpenCL
buildworker:Win64 OpenCL=windows-2
build_image:Win64 OpenCL=msvs2019
```


alalek:
:+1:

Issue #2667, opened 2020-09-11
The Fast Line Detector should be able to input an edge image


tsukada-cs:
Hi,

I want to input the edge image directly into FastLineDetector, but the current implementation has to use the Canny method internally.

The important function of FastLineDetector is to extract line segments from the edge image, not the creation of the edge image, which is not the essence of the FastLineDetector.

It would be easy to just not use the Canny method when an edge image is input, but I'm not confident I can write that in c++.

If this feature is added, I'd appreciate it.

sturkmen72:
Please close down.

PR #2716, opened 2020-10-16
Added edge input feature to fast_line_detector

tsukada-cs:
Fixed issue #2667 in opencv/opencv_contrib

Added the `is_edge` option to the `detect` method in `fast_line_detector`.

The Fast Line Detector used to use the canny method for edge detection internally, but `is_edge=true` makes it possible to input the edge image directly. Parameters for the canny method are ignored. The default value is `false`, so it won't break anyone's code.

This feature will provide users with additional flexibility and versatility in FastLineDetector.

### DIFF
--- a/modules/ximgproc/include/opencv2/ximgproc/fast_line_detector.hpp
+++ b/modules/ximgproc/include/opencv2/ximgproc/fast_line_detector.hpp
@@ -42,7 +42,8 @@ public:
       brighter side is on their left.
       @param is_edge If true, image will be considerd as edge and negrect the canny parameters.
       */
-    CV_WRAP virtual void detect(InputArray _image, OutputArray _lines, bool is_edge) = 0;
+    CV_WRAP virtual void detect(InputArray _image, OutputArray _lines,
+            bool is_edge = false) = 0;
 
     /** @brief Draws the line segments on a given image.
       @param _image The image, where the lines will be drawn. Should be bigger
--- a/modules/ximgproc/include/opencv2/ximgproc/fast_line_detector.hpp
+++ b/modules/ximgproc/include/opencv2/ximgproc/fast_line_detector.hpp
@@ -40,8 +40,9 @@ public:
       and ending point of a line.  Where Vec4f is (x1, y1, x2, y2), point
       1 is the start, point 2 - end. Returned lines are directed so that the
       brighter side is on their left.
+      @param is_edge If true, image will be considerd as edge and negrect the canny parameters.
       */
-    CV_WRAP virtual void detect(InputArray _image, OutputArray _lines) = 0;
+    CV_WRAP virtual void detect(InputArray _image, OutputArray _lines, bool is_edge) = 0;
 
     /** @brief Draws the line segments on a given image.
       @param _image The image, where the lines will be drawn. Should be bigger
--- a/modules/ximgproc/include/opencv2/ximgproc/fast_line_detector.hpp
+++ b/modules/ximgproc/include/opencv2/ximgproc/fast_line_detector.hpp
@@ -65,17 +65,16 @@ public:
                                          hysteresis procedure in Canny()
 @param _canny_th2           50         - Second threshold for
                                          hysteresis procedure in Canny()
-@param _canny_aperture_size 3          - Aperturesize for the sobel
-                                         operator in Canny()
+@param _canny_aperture_size 3          - Aperturesize for the sobel operator in Canny().
+                                         If zero, Canny() is not applied and the input
+                                         image is taken as an edge image.
 @param _do_merge            false      - If true, incremental merging of segments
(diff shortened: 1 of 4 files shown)


modules/ximgproc/include/opencv2/ximgproc/fast_line_detector.hpp
```diff
@@ -40,8 +40,10 @@ class CV_EXPORTS_W FastLineDetector : public Algorithm
       and ending point of a line.  Where Vec4f is (x1, y1, x2, y2), point
       1 is the start, point 2 - end. Returned lines are directed so that the
       brighter side is on their left.
+      @param is_edge If true, image will be considerd as edge and negrect the canny parameters.
       */
-    CV_WRAP virtual void detect(InputArray _image, OutputArray _lines) = 0;
+    CV_WRAP virtual void detect(InputArray _image, OutputArray _lines,
+            bool is_edge = false) = 0;
```
alalek, line 46: `detect()` is not a right place for algorithm parameters tuning. See `createFastLineDetector()`

Can we check some canny parameters to handle that without touching API?
For example, `canny_aperture_size == 0`

tsukada-cs:
@alalek 
Thanks for the comment.
I tried changing the internal assert statement and then `canny_aperture_size = 0`, `canny_th1=0`, `canny_th2=0`, etc., but it didn't work.
So I removed the `is_edge` option of `detect()` and added the `input_edge` option to `createFastLineDetector()`.

alalek:
> but it didn't work as expected

Try to split this assertion check first.

BTW, There is no sense to define `canny_*` parameters if we don't use that.

tsukada-cs:
@alalek 

> Try to split this assertion check first.

I failed to state the situation correctly.
I split the assertion check and tried `canny_appature_size=0`, but the canny method was applied as a result.
Also, the documentation says that the `ksize` of `Sobel()` must be 1, 3, 5 or 7, so we are not supposed to enter `canny_appature_size = 0` (`ksize=0`).

> BTW, There is no sense to define canny_* parameters if we don't use that.

I agree.
Should we change the implementation?

alalek:
> and tried `canny_appature_size=0`, but the canny method was applied as a result

You should add this check before `Canny()` call to bypass it.

tsukada-cs:
@alalek 

Thanks for the kind comments.
I changed the implementation to use the input image as an edge image when `canny_aperture_size=0`.

modules/ximgproc/src/fast_line_detector.cpp
```diff
@@ -544,7 +545,8 @@ void FastLineDetectorImpl::lineDetection(const Mat& src, std::vector<SEGMENT>& s
     std::vector<Point2i> points;
     std::vector<SEGMENT> segments, segments_tmp;
     Mat canny;
-    Canny(src, canny, canny_th1, canny_th2, canny_aperture_size);
+    if (canny_aperture_size == 0) canny = src;
+    else Canny(src, canny, canny_th1, canny_th2, canny_aperture_size);
```
alalek, line 549: Consider using dedicated lines for if/else bodies.
It is necessary to see real picture from "code coverage" line-based reports ([example of weekly report](http://pullrequest.opencv.org/buildbot/export/opencv_releases/master-contrib_coverage-lin64-debug/20201017-080737_10545/coverage_html/index.html)).

modules/ximgproc/src/fast_line_detector.cpp
```diff
@@ -29,10 +29,11 @@ class FastLineDetectorImpl : public FastLineDetector
...
          * @param _canny_th2           50         - Second threshold for
          *        _                                 hysteresis procedure in Canny()
-         * @param _canny_aperture_size 3          - Aperturesize for the sobel
-         *        _                                 operator in Canny()
+         * @param _canny_aperture_size 3          - Aperturesize for the sobel operator in Canny().
+         *                                          If zero, Canny() is not applied and the input
+         *                                          image is taken as an edge image.
          * @param _do_merge            false      - If true, incremental merging of segments
-                                                   will be perfomred
+         *                                          will be perfomred
```
sturkmen72, line 36: will be performed

tsukada-cs:
@alalek 

> Consider using dedicated lines for if/else bodies.
> It is necessary to see real picture from "code coverage" line-based reports (example of weekly report).

I tried, but is this what you mean?

modules/ximgproc/src/fast_line_detector.cpp
tsukada-cs: Thanks for pointing that out. Fixed it.

alalek (approved):
Well done! Thank you :+1:

tsukada-cs:
Thank you for accepting it.  Thanks for all the kind comments.

PR #2719, opened 2020-10-17
DynamicFusion cost function reworked

savuor:
**WIP**

DynaFu is being reworked to support true Dual Quaternion Blending as in the article.

TODOs:

- [ ] Write real TODOs and PR description
- [x] Use sparse solver from `rgbd` module instead of the one in the code
- [ ] Replace internal dual quaternions by the ones from OpenCV-next
- [ ] Transform debug code into tests
- [ ] Fix everything marked as TODO URGENT before debugging
- [ ] Debug the code
- [ ] Strip any extra options
- [ ] Document it properly


savuor:
This PR still does not work and diverged with the 5.x too much.
It'd be easier to recreate it in the future than to make it work.
Useful things from this PR like dual quaternions or LevMarq were integrated in 5.x.

PR #2757, opened 2020-11-19
1D Barcode support

Certseeds:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/860

[opencv/opencv#18668](https://github.com/opencv/opencv/issues/18668)

<cut/>

```
test_modules=barcode

opencv_extra=Barcode
```


Certseeds:
The building bot is waiting for builds for a few days.
Before it builds successfully, would anyone like to Review our class interface in [barcode.hpp](https://github.com/opencv/opencv_contrib/pull/2757/files#diff-6a09a0b19533bb15392ebebdc4137e6da119619682711301878ca639579dc0d0)?

modules/barcode/include/opencv2/barcode.hpp
```diff
@@ -0,0 +1,69 @@
...
+    CV_WRAP BarcodeDetector();
+
+    ~BarcodeDetector();
+
+    /** @brief Detects Barcode in image and returns the rectangle(s) containing the code.
+
+    @param img grayscale or color (BGR) image containing (or not) Barcode.
+    @param rects Output vector of rotated rectangle(s) containing the barcode.
+     */
+    CV_WRAP bool detect(InputArray img, CV_OUT std::vector <RotatedRect> &rects) const;
```
alalek, line 36: > RotatedRect

RotatedRect is not a best option to handle perspective transformations.

Consider switching on quadrangle result:
- 4 points through `std::vector<cv::Point>` (or `std::vector<cv::Point2f>`).
    Note: `std::array` is not supported by binding generator.
- fix order of these points (in meaning of bl/tl/tr/br) to avoid unexpected 90/180 degree rotation

These 4 points can be used with `getPerspectiveTransform()` + `warpPerspective()` to restore "plain" barcode for decoding.

modules/barcode/include/opencv2/barcode.hpp
```diff
@@ -0,0 +1,69 @@
...
+     */
+    CV_WRAP bool detectAndDecode(InputArray img, CV_OUT std::vector <std::string> &decoded_info,
+                                 CV_OUT std::vector <RotatedRect> &rects) const;
+
+    /** @brief Decode without detects
+
+    @param img grayscale or color (BGR) image containing barcode.
+    @param decoded_info UTF8-encoded output vector of string(s) or empty vector of string if the codes do not contain barcode.
+    */
+    CV_WRAP bool docodeDirectly(InputArray img, CV_OUT std::vector <std::string> &decoded_info) const;
```
alalek, line 62: std::vector ?

modules/barcode/include/opencv2/barcode.hpp
Certseeds: We had replaced the `std::vector <RotatedRect>` by `vector<vector<Point2f>>` with Specific order of points (bl,tl,tr,br) in the commit [ad603e0](https://github.com/opencv/opencv_contrib/pull/2757/commits/ad603e0fe5c557380e8a36d1ae2f24f365150f88)

modules/barcode/include/opencv2/barcode.hpp
Certseeds: `decodeDirectly()` should only one output string, we had repaired this design problem in [ad603e0](https://github.com/opencv/opencv_contrib/pull/2757/commits/ad603e0fe5c557380e8a36d1ae2f24f365150f88)

darkliang:
Now, we pushed the implementation code that was tested in linux based on arm64 and windows 10 based on amd64.  Some minor changes on ports caused the python bindings fail to be generated.

JinhengZhang:
@alalek @vpisarev  Please, have a review. Thank you!

darkliang:
@alalek Hi, I thought our interface code was written in correct format, but there is no our class(BarcodeDetector) in the generated Python bindings. The Python interpreter shows: 
```shell
PS C:\WINDOWS\System32> python
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import cv2.barcode as barcode
>>> barcode.BarcodeDetector()
  File "<stdin>", line 1, in <module>
AttributeError: module 'cv2.barcode' has no attribute 'BarcodeDetector'
>>> dir(barcode)
['EAN_13', 'EAN_8', 'NONE', 'UPC_A', 'UPC_E', 'UPC_EAN_EXTENSION', '__doc__', '__loader__', '__name__', '__package__', '__spec__']
```
Can you help us for this?

vpisarev:
thank you for the contribution! Could you please implement some regression tests and add a sample?

darkliang:
@alalek Hi, we added the tests and sample, but tests failed because test data was not merged into opencv_extra. Can you fix it?

alalek:
opencv_extra PR must be created from the same "user" GitHub account and use the same source branch name as "main" PR (to run validation).

darkliang:
@alalek Hi, it seems we pass all the building checks but the patch size of opencv_extra.

alalek:
Hi!
OpenCV tests are not intended to test all possible cases. So lets keep number of testcases small enough with some common representatives (such tests detect patches where algorithm is broken completely).
In general, avoid adding new test cases if this doesn't increase coverage of existed code.

BTW, we think about adding bigger algorithms "tests" which can use external larger public datasets through "opencv-benchmarks", but this activity is currently "on hold".

---

"multiple" test images are very large (12MP, 4+Mb).
In general we don't use images larger than 1-2MP (HD: 1280x720 or FHD: 1920x1080 - common resolution for popular cameras)
Also could we keep one case only (e.g, with 4 codes)?

darkliang:
@alalek Hi, we think our PR is ready to review. As mentioned [above](https://github.com/opencv/opencv_contrib/pull/2757#issuecomment-759321012), we still have not resolved this issue. Can you look into this?

modules/barcode/src/detector/bardetect.cpp
```diff
@@ -0,0 +1,422 @@
+//
+//
+#include "../precomp.hpp"
```
alalek, line 3: Please add OpenCV license header + extra copyright/author statements if necessary

modules/barcode/tutorials/barcode_detect_and_decode/barcode_detect_and_decode.markdown
```diff
@@ -0,0 +1,129 @@
...
+@code{.cpp}
+#include "opencv2/barcode.hpp"
+#include "opencv2/imgproc.hpp"
+
+using namespace cv;
+barcode::BarcodeDetector bardet;
+Mat frame = imread("bar_code.jpg");
+std::vector<Point> corners;
+bool ok = bardet.detect(frame, corners);
+@endcode
```
alalek, line 42: Please do not put code into the documentation directly. This code may not be updated on changes (it is not checked for compilation).

Use `@snippet` / `@include` doxygen statements for embedding code from .cpp files (see corresponding examples in other OpenCV modules).

alalek (commented):
Thank you for the contribution!

darkliang:
> Thank you for the contribution!

Thanks for review, we have updated the code again. By the way, we figured out the usage of generated python code, it should be `cv.barcode_BarcodeDetector()`.

vpisarev:
:+1:

modules/barcode/samples/barcode.cpp
```diff
@@ -0,0 +1,302 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+// Copyright (c) 2020-2021 darkliang wangberlinT Certseeds
```
alalek, line 4: OpenCV **samples** are widely used in documentation, so these files don't have license headers

> except for the samples

https://github.com/opencv/opencv/wiki/Coding_Style_Guide#file-structure

modules/barcode/misc/java/src/cpp/barcode_converters.hpp
```diff
@@ -0,0 +1,20 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html
+
+// Author: darkliang
+
+#ifndef BARCODE_CONVERTERS_HPP
+#define	BARCODE_CONVERTERS_HPP
```
alalek, line 8: please remove tabs from source code files

modules/barcode/samples/barcode.cpp
```diff
@@ -0,0 +1,302 @@
...
+    //! [initialize]
+    try{
+        bardet = makePtr<barcode::BarcodeDetector>(sr_prototxt, sr_model);
+    } catch (const std::exception& e)
+    {
+        cout <<
+             "\n---------------------------------------------------------------\n"
+             "Failed to initialize super resolution.\n"
+             "Please, download 'sr.*' from\n"
+             "https://github.com/WeChatCV/opencv_3rdparty/tree/wechat_qrcode\n"
```
alalek, line 77: > wechat_qrcode

1. is this a correct message?
2. is this a mandatory prerequisite? (in QR codes it doesn't)

modules/barcode/src/decoder/upcean_decoder.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+namespace cv {
+namespace barcode {
+
+static constexpr int DIVIDE_PART = 15;
+static constexpr int BIAS_PART = 2;
+//void UPCEANDecoder::drawDebugLine(Mat &debug_img, const Point2i &begin, const Point2i &end) const
+//{
+//    Result result;
+//    std::vector<uchar> middle;
```
alalek, line 18: > `//`

Instead of using of multi-line comments, consider using of `#if 0` to disable debug parts in the code.

modules/barcode/src/decoder/upcean_decoder.cpp
```diff
@@ -0,0 +1,285 @@
...
+{
+    Result result = this->decode(line);
+    if (result.format == BarcodeType::NONE)
+    {
+        result = this->decode(std::vector<uchar>(line.crbegin(), line.crend()));
+    }
+    return result;
+}
+
+bool UPCEANDecoder::isValid(std::string result) const
```
alalek, line 194: const reference?

modules/barcode/src/barcode.cpp
```diff
@@ -0,0 +1,293 @@
...
+};
+
+void BarDecode::init(const vector<Mat> &bar_imgs_)
+{
+    bar_imgs = bar_imgs_;
+}
+
+bool BarDecode::decodeMultiplyProcess()
+{
+    class ParallelBarCodeDecodeProcess : public ParallelLoopBody
```
alalek, line 87: It makes sense to move classes of out the functions.

Moreover, there is C++11 is available. Consider using C++ lambdas with parallel_for instead of creation processing structs.

modules/barcode/src/barcode.cpp
```diff
@@ -0,0 +1,293 @@
...
+    return !result_info.empty();
+}
+
+
+struct BarcodeDetector::Impl
+{
+public:
+    Impl() = default;;
+
+    ~Impl() = default;;
```
alalek, line 148: > `;;`

one symbol is enough

modules/barcode/src/barcode.cpp
```diff
@@ -0,0 +1,293 @@
...
+
+    std::shared_ptr<SuperScale> sr;
+    bool use_nn_sr = false;
+};
+
+// return cropped and scaled bar img
+vector<Mat> BarcodeDetector::Impl::initDecode(const Mat &src, const vector<Point2f> &points) const
+{
+    vector<vector<Point2f>> src_points;
+    //CV_TRACE_FUNCTION();
```
alalek, line 160: > `//CV_TRACE_FUNCTION();`

is it needed or not?

If yes, then it should be the first statement of the function/method.

modules/barcode/src/decoder/abs_decoder.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+void fillCounter(const std::vector<uchar> &row, uint start, Counter &counter)
+{
+    size_t counter_length = counter.pattern.size();
+    std::fill(counter.pattern.begin(), counter.pattern.end(), 0);
+    counter.sum = 0;
+    size_t end = row.size();
+    if (start >= end)
+    {
+    }
```
alalek, line 40: Static analyzers dislikes empty if bodies (code defects).

modules/barcode/src/decoder/abs_decoder.cpp
```diff
@@ -0,0 +1,124 @@
...
+                counter.pattern[counterPosition] = 1;
+                counter.sum++;
+                color = 255 - color;
+            }
+        }
+        ++start;
+    }
+    if (!(counterPosition == counter_length || (counterPosition == counter_length - 1 && start == end)))
+    {
+    }
```
alalek, line 68: one more empty `if` body

modules/barcode/src/decoder/abs_decoder.cpp
```diff
@@ -0,0 +1,124 @@
...
+{
+    size_t numCounters = counter.pattern.size();
+    int total = static_cast<int>(counter.sum);
+    int patternLength = std::accumulate(pattern.cbegin(), pattern.cend(), 0);
+    if (total < patternLength)
+    {
+        // If we don't even have one pixel per unit of bar width, assume this is too small
+        // to reliably match, so fail:
+        // and use constexpr functions
+        return 2147483647;// max
```
alalek, line 82: Avoid using of magic numbers.

For example, `std::numeric_limits<int32_t>::max()` is used few lines below.

modules/barcode/src/decoder/abs_decoder.hpp
```diff
@@ -0,0 +1,73 @@
...
+};
+
+struct Counter
+{
+    std::vector<int> pattern;
+    uint sum;
+
+    explicit Counter(vector<int> _pattern)
+    {
+        pattern = std::move(_pattern);
```
alalek, line 41: const reference (to keep external interfaces clear) + assign operator without std::move?

modules/barcode/src/decoder/abs_decoder.hpp
```diff
@@ -0,0 +1,73 @@
...
+{
+public:
+    virtual std::pair<Result, float> decodeROI(const Mat &bar_img) const = 0;
+
+    virtual ~AbsDecoder() = default;
+
+protected:
+    virtual Result decode(const vector<uchar> &data) const = 0;
+
+    virtual bool isValid(string result) const = 0;
```
alalek, line 56: const reference

modules/barcode/src/decoder/common/hybrid_binarizer.cpp
```diff
@@ -0,0 +1,195 @@
...
+// Licensed under the Apache License, Version 2.0 (the "License").
+
+#include "../../precomp.hpp"
+#include "hybrid_binarizer.hpp"
+
+namespace cv {
+namespace barcode {
+
+
+#define CAP(x, x1, x2) x < (x1) ? (x1) : ((x) > (x2) ? (x2) : (x))
```
alalek, line 14: > CAP

`CLAMP`?

BTW, [`std::clamp`](https://en.cppreference.com/w/cpp/algorithm/clamp) can't be used because it is C++17 only

modules/barcode/src/decoder/common/hybrid_binarizer.cpp
```diff
@@ -0,0 +1,195 @@
...
+// So this is the smallest dimension in each axis we can accept.
+constexpr static int BLOCK_SIZE_POWER = 3;
+constexpr static int BLOCK_SIZE = 1 << BLOCK_SIZE_POWER; // ...0100...00
+constexpr static int BLOCK_SIZE_MASK = BLOCK_SIZE - 1;   // ...0011...11
+constexpr static int MINIMUM_DIMENSION = BLOCK_SIZE * 5;
+constexpr static int MIN_DYNAMIC_RANGE = 24;
+
+void
+calculateThresholdForBlock(const std::vector<uchar> &luminances, int sub_width, int sub_height, int width, int height,
+                           Mat black_points, Mat &dst)
```
alalek, line 26: > Mat black_points

const reference

modules/barcode/src/decoder/common/super_scale.cpp
```diff
@@ -0,0 +1,76 @@
...
+// Modified by darkliang wangberlinT
+
+#include "../../precomp.hpp"
+#include "opencv2/dnn.hpp"
+#include "super_scale.hpp"
+
+
+namespace cv {
+namespace barcode {
+#define CLIP(x, x1, x2) max(x1, min(x, x2))
```
alalek, line 16: > CLIP

One more name and implementation for `CLAMP()`

modules/barcode/src/decoder/common/super_scale.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+    dst = Mat(prob.size[2], prob.size[3], CV_8UC1);
+
+    for (int row = 0; row < prob.size[2]; row++)
+    {
+        const float *prob_score = prob.ptr<float>(0, 0, row);
+        for (int col = 0; col < prob.size[3]; col++)
+        {
+            float pixel = prob_score[col] * 255.0f;
+            dst.at<uint8_t>(row, col) = static_cast<uint8_t>(CLIP(pixel, 0.0f, 255.0f));
```
alalek, line 70: `cv::saturate_cast<uchar>()` ?

modules/barcode/src/detector/bardetect.cpp
```diff
@@ -0,0 +1,422 @@
...
+    ptr += left_col + integral_cols * top_row;\
+    top_left = static_cast<float>(*(ptr));\
+    (ptr) += right_col - left_col;\
+    top_right = static_cast<float>(*(ptr));\
+    (ptr) += (bottom_row - top_row) * integral_cols;\
+    bottom_right = static_cast<float>(*(ptr));\
+    (ptr) -= right_col - left_col;\
+    bottom_left = static_cast<float>(*(ptr));\
+    (ptr) -= (bottom_row - top_row) * integral_cols;\
+    (ptr) -= left_col + integral_cols * top_row;\
```
alalek, line 25: > `(ptr) += ...`
> `(ptr) -= ...`

It doesn't make sense to modify `ptr` itself.
Modern compilers are smart enough to detect and optimize common sub-expressions.

(also such modification may block many optimizations in compilers, e.g `ptr` is a class field, so it must be update with all intermediate results)

modules/barcode/src/detector/bardetect.cpp
```diff
@@ -0,0 +1,422 @@
...
+{
+    if (x >= 0)
+    {
+        return atan(y / x);
+    }
+    if (y >= 0)
+    {
+        return (atan(y / x) + PI);
+    }
+    return (atan(y / x) - PI);
```
alalek, line 54: why is not [atan2](http://www.cplusplus.com/reference/cmath/atan2/)?

Also there is `cv::fastAtan()` (but result is in degrees instead of radians and probably modified/shifted range)

modules/barcode/src/detector/bardetect.cpp
```diff
@@ -0,0 +1,422 @@
...
+        transformation_points.emplace_back(vector<Point2f>{temp[0], temp[1], temp[2], temp[3]});
+    }
+
+    return !transformation_points.empty();
+}
+
+
+void Detect::preprocess()
+{
+    Mat scharr_x(resized_barcode.size(), CV_32F), scharr_y(resized_barcode.size(), CV_32F), temp;
```
alalek, line 151: No need to pre-allocate destination `cv::Mat`.

modules/barcode/src/detector/bardetect.cpp
```diff
@@ -0,0 +1,422 @@
...
+    Scharr(resized_barcode, scharr_x, CV_32F, 1, 0);
+    Scharr(resized_barcode, scharr_y, CV_32F, 0, 1);
+    // calculate magnitude of gradient, normalize and threshold
+    magnitude(scharr_x, scharr_y, gradient_magnitude);
+    threshold(gradient_magnitude, gradient_magnitude, 56, 1, THRESH_BINARY);
+    gradient_magnitude.convertTo(gradient_magnitude, CV_8U);
```
alalek, line 157: It is better to avoid in-placing of in/out parameters.
Especially with different types.

modules/barcode/src/detector/bardetect.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+
+void Detect::preprocess()
+{
+    Mat scharr_x(resized_barcode.size(), CV_32F), scharr_y(resized_barcode.size(), CV_32F), temp;
+    Scharr(resized_barcode, scharr_x, CV_32F, 1, 0);
+    Scharr(resized_barcode, scharr_y, CV_32F, 0, 1);
+    // calculate magnitude of gradient, normalize and threshold
+    magnitude(scharr_x, scharr_y, gradient_magnitude);
+    threshold(gradient_magnitude, gradient_magnitude, 56, 1, THRESH_BINARY);
```
alalek, line 156: > 56

Please avoid using of magic numbers.
Use dedicated const properly named variable for that.

modules/barcode/src/detector/bardetect.cpp
```diff
@@ -0,0 +1,422 @@
...
+            }
+            // flag
+            coherence_row[x] = 0;
+            growingPoints.clear();
+            growingImgPoints.clear();
+
+            pt = Point(x, y);
+            cur_value = orientation.at<float_t>(pt);
+            sin_sum = sin(2 * cur_value);
+            cos_sum = cos(2 * cur_value);
```
alalek, line 298: BTW, in general `sin`/`cos` computation in nested loops is not very cheap.

modules/barcode/samples/barcode.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+static void drawBarcodeResults(Mat &frame, const vector<Point> &corners, const vector<cv::String> &decode_info,
+                               const vector<cv::barcode::BarcodeType> &decode_type, double fps)
+{
+    if (!corners.empty())
+    {
+        for (size_t i = 0; i < corners.size(); i += 4)
+        {
+            size_t bar_idx = i / 4;
+            vector<Point> qrcode_contour(corners.begin() + i, corners.begin() + i + 4);
```
alalek, line 136: > qrcode_contour

misnamed

alalek (commented):
Great contribution!

modules/barcode/src/detector/bardetect.cpp
darkliang: Do you have some suggestion for this? I analyzed the performance by Visual Studio, found that 
image
It seems like preprocess() is much more time-consuming than this.

modules/barcode/src/detector/bardetect.cpp
darkliang: I did not really understand what's your point? I rewrote the code like this:
```cpp
    magnitude(scharr_x, scharr_y, temp);
    threshold(temp, temp, THRESHOLD_MAGNITUDE, 1, THRESH_BINARY);
    temp.convertTo(gradient_magnitude, CV_8U);
```

modules/barcode/samples/barcode.cpp
darkliang: 1. I added that sentence:

> Or you can leave sr_prototxt and sr_model unspecified.

2. It is not a mandatory prerequisite.

darkliang:
Based on your comments, I refined the code. Please check again. @alalek @vpisarev

modules/barcode/src/detector/bardetect.cpp
alalek: Thank you for checking this.
You are right, this is not critical part of computations for now.

Possible optimization are usually done through using of pre-computed sin/cos tables (with input value quantization).

alalek (approved):
Well done! Thank you :+1:

darkliang:
> Well done! Thank you 👍

Thanks for  your review. I fix a bug about java binding, please check. 😆

PR #2776, opened 2020-12-06
freetype2: Fix comments in freetype.hpp and README.md

Kumataro:
Fix https://github.com/opencv/opencv_contrib/issues/2771 and related small problems about documentation.

1.  freetype.hpp is fixed.
  - putText() is supported for only 8UC3 image with only 8UC3 color.

2. README.md is fixed.
  - Use correct markdown format.
  - Change document title from "freetype module" to "freetype wrapper module". 
    - This module title was "freetype **module**". 
    - FText primitive in GAPI uses `freetype **library**". 
    - Generally  "module" and "library" in the computer world are similar.
    - So that, I think this module title should be changed to not confuse it.
  - `Installation` is replaced as `Requested external libraries`
  - `Usage` is update to it can be compile in real. 
  - `Future work` is removed. 

### DIFF
--- a/modules/freetype/README.md
+++ b/modules/freetype/README.md
@@ -1,10 +1,11 @@
 FreeType2 Wrapper Module
-==========================
+========================
 
 This FreeType2 wrapper module allows to draw strings with outlines and bitmaps.
 
-Requested external libraries.
-------------------------------
+Requested external libraries
+----------------------------
+
 harfbuzz is requested to convert UTF8 to gid(GlyphID).
 
 freetype library is requested to rasterize given gid.
@@ -13,7 +14,7 @@ freetype library is requested to rasterize given gid.
 - freetype https://www.freetype.org/
 
 Usage
------------
+-----
 
 ```
 cv::Ptr<cv::freetype::FreeType2> ft2;
@@ -24,7 +25,7 @@ ft2->putText(mat, "hello world", cv::Point(20, 200),
 ```
 
 Option
-------------
+------
 - 2nd argument of loadFontData is used if font file has many font data.
 - 3 drawing mode is available.
     - outline mode is used if lineWidth is larger than 0. (like original putText)
--- a/modules/freetype/README.md
+++ b/modules/freetype/README.md
@@ -1,34 +1,34 @@
-FreeType Module
-===========
+FreeType2 Wrapper Module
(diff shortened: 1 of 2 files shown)


modules/freetype/include/opencv2/freetype.hpp
```diff
@@ -99,11 +99,11 @@ If you want to draw small glyph, small is better.
 
 The function putText renders the specified text string in the image. Symbols that cannot be rendered using the specified font are replaced by "Tofu" or non-drawn.
 
-@param img Image.
+@param img Image. (Only 8UC3 image is supported.)
 @param text Text string to be drawn.
 @param org Bottom-left/Top-left corner of the text string in the image.
 @param fontHeight Drawing font size by pixel unit.
-@param color Text color.
+@param color Text color.(Only 3ch 8bit Scalar is supported.)
```
alalek, line 106: > 8bit

Scalar values are `double`s by design.
I believe we should mention used range, like 0-255, here instead.

modules/freetype/README.md
```diff
@@ -1,34 +1,33 @@
...
+FreeType2 Wrapper Module
+==========================
 
-This FreeType module allows you to draw strings with outlines and bitmaps.
+This FreeType2 wrapper module allows to draw strings with outlines and bitmaps.
 
-Installation
------------
+Requested external libraries.
+------------------------------
```
alalek, line 7: Please remove dots from the `Headers`

Please align number of `===` / `---` symbols on the next line (as markdown requested)

alalek (commented):
Thank you for contribution.

modules/freetype/README.md
Kumataro: Thank you very much ! I'll fix it.

modules/freetype/include/opencv2/freetype.hpp
Kumataro: Thank you very much! 

Because freetype2 wrapper module requests that img is  8UC3, I reconsider this limitation of color is very clear. 
  - first 3 channels of 'color' are used, 4th channel of it is not. 
  - The range of element are 0-255. 

This change is not necessary.  I cancel adding comment.

alalek (approved):

PR #2778, opened 2020-12-07
surface_matching/PPF3DDetector::match : fix memory leak

txpaper:
I think the define guards are wrong, the accumulator is allocated with `calloc` regardless of  `_OPENMP`, so it should be freed in any case.

### DIFF
--- a/modules/surface_matching/src/ppf_match_3d.cpp
+++ b/modules/surface_matching/src/ppf_match_3d.cpp
@@ -583,9 +583,7 @@ void PPF3DDetector::match(const Mat& pc, std::vector<Pose3DPtr>& results, const 
       poseList.push_back(pose);
     }
 
-#if defined (_OPENMP)
     free(accumulator);
-#endif
   }
 
   // TODO : Make the parameters relative if not arguments.


modules/surface_matching/src/ppf_match_3d.cpp
```diff
@@ -583,9 +583,7 @@ void PPF3DDetector::match(const Mat& pc, std::vector<Pose3DPtr>& results, const
       poseList.push_back(pose);
     }
 
-#if defined (_OPENMP)
     free(accumulator);
```
alalek, line 586: > uint* accumulator = (uint*)calloc(numAngles*n, sizeof(uint));

BTW, it is better to avoid manual memory management and use `cv::AutoBuffer` or `std::vector`.

alalek (commented):
Thank you for contribution!

As a bugfix this patch should go into [3.4 branch](https://github.com/opencv/opencv/wiki/Branches) first.
We will merge changes from 3.4 into master regularly (weekly/bi-weekly).

Please:

modules/surface_matching/src/ppf_match_3d.cpp
txpaper: Totally agree.

Unfortunately I found this bug while helping a friend, not using this opencv module myself I do not feel confident enough to refactor the code to a proper "C++ style" memory management.

alalek (approved):

PR #2795, opened 2020-12-21
Update cudacodec to work with Nvidia Video Codec SDK 11.0

cudawarped:
Nvidia only allows users to download the most recent version of the Nvidia Video Codec SDK.  As a result cudacodec has been broken since the release of v 11.0 when the AV1 codec was added.

This small fix should allow cudacodec to pass the inbuilt tests again.

```
buildworker:Custom=linux-4
```


alalek (approved):
LGTM :+1:

PR #2800, opened 2020-12-29
[xfeatures2d:GMS]Solve out of range

snail0614:

### DIFF
--- a/modules/xfeatures2d/src/gms.cpp
+++ b/modules/xfeatures2d/src/gms.cpp
@@ -370,6 +370,9 @@ int GMSMatcher::run(const int rotationType)
         // Mark inliers
         for (size_t i = 0; i < mNumberMatches; i++)
         {
+            if (mvMatchPairs[i].first < 0)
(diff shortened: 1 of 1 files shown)


modules/xfeatures2d/src/gms.cpp
```diff
@@ -370,6 +370,9 @@ int GMSMatcher::run(const int rotationType)
         // Mark inliers
         for (size_t i = 0; i < mNumberMatches; i++)
         {
+            if (mvMatchPairs[i].first < 0)
+                continue;
+                
             if (mvMatchPairs[i].first >= 0 && mCellPairs[mvMatchPairs[i].first] == mvMatchPairs[i].second)
                 mvbInlierMask[i] = true;
         }
```
alalek, line 378: This patch changes nothing in this code behavior.

alalek (commented):
Please start from the stable complete minimal reproducer for the problem.
Otherwise you can't check the "fix"

PR #2801, opened 2020-12-29
Added stream support on hough circles, lines and segments

Atlas42:
- Passed the stream to the different cuda, OpenCV and thurst library calls
- Replace all device by cuda synchronizes
- Added extra synchronize calls after device to host transfers
- Replaced the cuda globals by allocated values

<cut/>

```
force_builders=Custom
buildworker:Custom=linux-4
Xbuild_image:Custom=ubuntu-cuda-cc52:18.04
```

### DIFF
--- a/modules/cudaimgproc/src/cuda/hough_lines.cu
+++ b/modules/cudaimgproc/src/cuda/hough_lines.cu
@@ -44,6 +44,7 @@
 
 #include <thrust/device_ptr.h>
 #include <thrust/sort.h>
+#include <thrust/system/cuda/execution_policy.h>
 
 #include "opencv2/core/cuda/common.hpp"
 #include "opencv2/core/cuda/emulation.hpp"
--- a/modules/cudaimgproc/src/cuda/hough_lines.cu
+++ b/modules/cudaimgproc/src/cuda/hough_lines.cu
@@ -53,8 +53,6 @@ namespace cv { namespace cuda { namespace device
 {
     namespace hough_lines
     {
-        __device__ int g_counter;
-
         ////////////////////////////////////////////////////////////////////////
         // linesAccum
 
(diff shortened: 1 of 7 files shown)


modules/cudaimgproc/src/cuda/hough_lines.cu
```diff
@@ -174,33 +171,30 @@ namespace cv { namespace cuda { namespace device
...
+            cudaSafeCall( cudaStreamSynchronize(stream) );
 
             totalCount = ::min(totalCount, maxSize);
 
             if (doSort && totalCount > 0)
             {
                 thrust::device_ptr<float2> outPtr(out);
                 thrust::device_ptr<int> votesPtr(votes);
-                thrust::sort_by_key(votesPtr, votesPtr + totalCount, outPtr, thrust::greater<int>());
+                thrust::sort_by_key(thrust::cuda::par.on(stream), votesPtr, votesPtr + totalCount, outPtr, thrust::greater<int>());
```
alalek, line 197: Could you please take a look if there is an easy way to guard this code for CUDA 8.x?

alalek (commented):
Thank you for contribution!

modules/cudaimgproc/src/cuda/hough_lines.cu
Atlas42: I do not understand your question here.
Is there a problem specific to CUDA 8 ? I have checked the API and it seems fine.
As Cuda 8 is not supported anymore on Ubuntu 18.04 or Visual Studio 19 (my dev tools),
I took some time today to setup an Ubuntu 16.04/CUDA 8 and it builds properly with OpenCV 4.5.0 (master did not on some unrelated problem).

alalek: Please take a look on build issues of "Custom" CUDA builder [here](http://pullrequest.opencv.org/#/summary/contrib).

There are several ways:
- try to fix the build using `#if` guards with CUDA version checks (requested above)
- try to fix the build through disabling this algorithm for CUDA 8.0
- open a discussion about deprecation of old CUDA versions in next OpenCV releases (will take some time)

Atlas42: It should be fixed on my last commit.
I managed to reproduce the build problem, the instruction on the NVIDIA website to install cuda 8 were wrong, I should have checked.
Turns out that the thrust::cuda::par did exist in CUDA 8, but was not included by default.

alalek (approved):

PR #2803, opened 2021-01-02
Implement cv::cuda::inRange (Fixes OpenCV #6295)

amiller27:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/834

resolves https://github.com/opencv/opencv/issues/6295

Implementation of `cv::cuda::inRange`, as requested in [this issue](https://github.com/opencv/opencv/issues/6295).  It's not a complete implementation of `cv::inRange`; my implementation supports `cv::Scalar` for the upper and lower bounds, which seems like the most common use case by far, but the CPU version does also support `Mat`s for the bounds as well, which mine does not.  It seemed like more of a challenge to support that as well, but if you don't want to merge this without full feature parity I might be able to spend more time on it to figure that out.

Some questions -
1. Wasn't sure about whether the base branch should be `master` or `3.4`, happy to rebase to `3.4` if desired.  I do use `std::array`, but I think that's the only C++11 feature I'm using and it'd be easily removed.
2. Wasn't sure if I need to do anything special for performance test data - I followed the instructions [here](https://github.com/opencv/opencv/wiki/HowToUsePerfTests) to update the values for my new test only ([PR in opencv_extra](https://github.com/opencv/opencv_extra/pull/834)), but I didn't know how your performance test system deals with the fact that this should perform very differently depending on the GPU (I tested on a GeForce 750M).
3. I added Doxygen comments in the code; if there's more documentation or sample code needed I can do that as well
4. I had to use some recursive templates in `functional.hpp` to copy and compare CUDA vectors; if CUDA or OpenCV already has utilities somewhere to do this I can switch to make that simpler

<cut/>

- [ ] The PR is proposed to proper branch - see comment below
- [x] The feature is well documented and sample code can be built with the project CMake - see comment below

```
force_builders=Custom
buildworker:Custom=linux-5
```

### DIFF
--- a/modules/cudaarithm/include/opencv2/cudaarithm.hpp
+++ b/modules/cudaarithm/include/opencv2/cudaarithm.hpp
@@ -358,6 +358,31 @@ threshold types are not supported.
  */
 CV_EXPORTS_W double threshold(InputArray src, OutputArray dst, double thresh, double maxval, int type, Stream& stream = Stream::Null());
 
+/** @brief  Checks if array elements lie between two scalars.
+
+The function checks the range as follows:
+-   For every element of a single-channel input array:
+    \f[\texttt{dst} (I)= \texttt{lowerb}_0  \leq \texttt{src} (I)_0 \leq  \texttt{upperb}_0\f]
+-   For two-channel arrays:
+    \f[\texttt{dst} (I)= \texttt{lowerb}_0  \leq \texttt{src} (I)_0 \leq  \texttt{upperb}_0  \land \texttt{lowerb}_1  \leq \texttt{src} (I)_1 \leq  \texttt{upperb}_1\f]
+-   and so forth.
+
+That is, dst (I) is set to 255 (all 1 -bits) if src (I) is within the
+specified 1D, 2D, 3D, ... box and 0 otherwise.
+
+Note that unlike the CPU inRange, this does NOT accept an array for lowerb or
+upperb, only a cv::Scalar.
+
+@param src first input array.
+@param lowerb inclusive lower boundary cv::Scalar.
+@param upperb inclusive upper boundary cv::Scalar.
+@param dst output array of the same size as src and CV_8U type.
+@param stream Stream for the asynchronous version.
+
+@sa cv::inRange
+ */
+CV_EXPORTS_W void inRange(InputArray src, const Scalar& lowerb, const Scalar& upperb, OutputArray dst, Stream& stream = Stream::Null());
+
 /** @brief Computes magnitudes of complex matrix elements.
 
 @param xy Source complex matrix in the interleaved format ( CV_32FC2 ).
--- a/modules/cudaarithm/src/cuda/in_range.cu
+++ b/modules/cudaarithm/src/cuda/in_range.cu
@@ -1,0 +1,99 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#include "opencv2/opencv_modules.hpp"
+
+#ifndef HAVE_OPENCV_CUDEV
+
+#error "opencv_cudev is required"
+
+#else
+
+#include "opencv2/core/private.cuda.hpp"
+#include "opencv2/cudaarithm.hpp"
+#include "opencv2/cudev.hpp"
+
+using namespace cv;
+using namespace cv::cuda;
+using namespace cv::cudev;
+
+namespace {
+
+template <typename T, int cn>
+void inRangeImpl(const GpuMat& src,
+                 const Scalar& lowerb,
+                 const Scalar& upperb,
+                 GpuMat& dst,
+                 Stream& stream) {
+    gridTransformUnary(globPtr<typename MakeVec<T, cn>::type>(src),
+                       globPtr<uchar>(dst),
+                       InRangeFunc<T, cn>(lowerb, upperb),
+                       stream);
+}
+
+}  // namespace
+
+void cv::cuda::inRange(InputArray _src,
+                       const Scalar& _lowerb,
+                       const Scalar& _upperb,
+                       OutputArray _dst,
+                       Stream& stream) {
+    const GpuMat src = getInputMat(_src, stream);
+
+    typedef void (*func_t)(const GpuMat& src,
+                           const Scalar& lowerb,
+                           const Scalar& upperb,
+                           GpuMat& dst,
+                           Stream& stream);
+
+    // Note: We cannot support 16F with the current implementation because we
+    // use a CUDA vector (e.g. int3) to store the bounds, and there is no CUDA
+    // vector type for float16
+    static constexpr const int MAX_CHANNELS = 4;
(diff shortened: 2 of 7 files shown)


modules/cudaarithm/src/cuda/in_range.cu
```diff
@@ -0,0 +1,153 @@
...
+//  IMPORTANT: READ BEFORE DOWNLOADING, COPYING, INSTALLING OR USING.
+//
+//  By downloading, copying, installing or using the software you agree to this
+license.
+//  If you do not agree to this license, do not download, install,
+//  copy or use the software.
+//
+//
+//                           License Agreement
+//                For Open Source Computer Vision Library
```
alalek, line 12: Please use short license header: https://github.com/opencv/opencv/wiki/Coding_Style_Guide#file-structure

```
```

modules/cudaarithm/src/cuda/in_range.cu
```diff
@@ -0,0 +1,153 @@
...
+            std::array<func_t, 7>{inRangeImpl<uchar, 4>,
+                                  inRangeImpl<schar, 4>,
+                                  inRangeImpl<ushort, 4>,
+                                  inRangeImpl<short, 4>,
+                                  inRangeImpl<int, 4>,
+                                  inRangeImpl<float, 4>,
+                                  inRangeImpl<double, 4>}};
+
+    const func_t func = funcs.at(channels - 1).at(depth);
+    func(src, lb, ub, dst, stream);
```
alalek, line 148: Please use `CV_Check*()` macro to validate `channels` / `depth` values.

modules/cudaarithm/include/opencv2/cudaarithm.hpp
```diff
@@ -358,6 +358,31 @@ threshold types are not supported.
...
+Note that unlike the CPU inRange, this does NOT accept an array for lowerb or
+upperb, only a cv::Scalar.
+
+@param src first input array.
+@param lowerb inclusive lower boundary cv::Scalar.
+@param upperb inclusive upper boundary cv::Scalar.
+@param dst output array of the same size as src and CV_8U type.
+@param stream Stream for the asynchronous version.
+
+@sa inRange
```
alalek, line 382: There is [self-reference loop](http://pullrequest.opencv.org/buildbot/export/pr_contrib/2803/docs/d8/d34/group__cudaarithm__elem.html#ga448e39f50eb0fc894239c6f4c1066f11).

Try this one instead:
`@sa cv::inRange`

modules/cudaarithm/include/opencv2/cudaarithm.hpp
```diff
@@ -358,6 +358,31 @@ threshold types are not supported.
...
+ */
+CV_EXPORTS_W void inRange(InputArray src, InputArray lowerb, InputArray upperb, OutputArray dst, Stream& stream = Stream::Null());
```
alalek, line 384: > CV_EXPORTS_W

Did you try to call this from Python binding? Are `lowerb` / `upperb` handled properly?

alalek (commented):
Thank you for contribution!

modules/cudaarithm/src/cuda/in_range.cu
amiller27: Fixed

modules/cudaarithm/src/cuda/in_range.cu
amiller27: Fixed

modules/cudaarithm/include/opencv2/cudaarithm.hpp
amiller27: Fixed

modules/cudaarithm/include/opencv2/cudaarithm.hpp
amiller27: I can make calls like
```python
src = (np.random.random((1920, 1080, 4)) * 256).astype(np.uint8)
lowerb = (0, 0, 0, 0)
upperb = (255, 255, 255, 255)
dst = cv2.cuda.inRange(src, lowerb, upperb)
```
with different types and numbers of channels for `src` (up to 4 channels).  The `CV_Check` fires correctly if I pass the wrong shape for `lowerb` or `upperb`, or too many channels for `src`.  It doesn't work if I pass a `cv2.cuda_GpuMat` (I get a `TypeError: Expected Ptr<cv::UMat> for argument 'src'`) - I'm really not familiar with using OpenCV CUDA from Python so I don't know if that's the intended behavior.

alalek: > I get a TypeError

Current messages from Python bindings are quite useless. You can set `OPENCV_PYTHON_DEBUG=1` env variable to investigate conversion errors.
The "GPU" overload assumes that all inputs must be a `GpuMat` including `lowerb/upperb` parameters (because they a `InputArray`). But wrapping 4 scalars into `GpuMat` has significant overhead.

> InputArray lowerb, InputArray upperb

Try to replace type to `const Scalar&` to eliminate unnecessary complexity.

modules/cudaarithm/src/cuda/in_range.cu
alalek: > src.depth()

where is validated?

modules/cudaarithm/include/opencv2/cudaarithm.hpp
alalek: Please add simple Python test into this [file](https://github.com/opencv/opencv_contrib/blob/master/modules/cudaarithm/misc/python/test/test_cudaarithm.py).

(please note, that public OpenCV CI doesn't run CUDA code/tests)

modules/cudaarithm/src/cuda/in_range.cu
amiller27: Previously wasn't; I was trying to cover all depths.  I've now realized that it doesn't work for float16, because CUDA doesn't have a float16 vector type.  So I added a CV_Check that the depth is CV_64F or lower

modules/cudaarithm/include/opencv2/cudaarithm.hpp
amiller27: Done.  I switched to `const Scalar&` for `lowerb` and `upperb`, and added a Python test to that file which passes on my machine, it now works if you pass a `Mat` or a `cuda_GpuMat`

amiller27:
I think everything has been addressed, but the docs build is failing on CI with
```
From git://code.ocv/opencv/opencv-ci
 * branch            master     -> FETCH_HEAD
HEAD is now at 5258c2f CMAKE_CXX_FLAGS: removed duplicate -Winit-self
FATAL: Build image for 'docs--18.04' is missing on the current build worker
program finished with exit code 1
```
It passes on my machine and this seems unrelated to anything I've done, not sure if there's anything I can do to fix it

alalek:
@amiller27 It is CI issue, I will fix that soon.

Custom (CUDA) builder is failed due to changes from #2807 (should be fixed soon in separate PR)

alalek (approved):

PR #2804, opened 2021-01-04
rlof optflow: fix crash when no match are found

MrKepzie:

- [X ] I agree to contribute to the project under Apache 2 License.
- [ X] To the best of my knowledge, the proposed patch is not based on a code under GPL or other license that is incompatible with OpenCV
- [ X] The PR is proposed to proper branch
- [X ] The feature is well documented and sample code can be built with the project CMake

### DIFF
--- a/modules/optflow/src/rlofflow.cpp
+++ b/modules/optflow/src/rlofflow.cpp
@@ -203,7 +203,11 @@ public:
             filtered_prevPoints = prevPoints;
             filtered_currPoints = currPoints;
         }
-
+        // Interpolators below expect non empty matches
+        if (filtered_prevPoints.empty()) {
+            flow.setTo(0);
+            return;
+        }
         if (interp_type == InterpolationType::INTERP_EPIC)
         {
             Ptr<ximgproc::EdgeAwareInterpolator> gd = ximgproc::createEdgeAwareInterpolator();
--- a/modules/optflow/include/opencv2/optflow/rlofflow.hpp
+++ b/modules/optflow/include/opencv2/optflow/rlofflow.hpp
@@ -227,6 +227,7 @@ public:
  * @note If the grid size is set to (1,1) and the forward backward threshold <= 0 than pixelwise dense optical flow field is
  * computed by RLOF without using interpolation.
  *
+ * @note Note that in output, if no correspondences are found between \a I0 and \a I1, the \a flow is set to 0.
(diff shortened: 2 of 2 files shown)


alalek:
Thank you for contribution!

Please fix the whitespace issue (and amend commit in this PR: `git commit -a --am`).
Also please double check commit e-mails (they should be consistent with emails in [your GitHub account](https://github.com/settings/emails)). You may need to reset commit's author: `git commit -a --am --reset-author`

modules/optflow/src/rlofflow.cpp
```diff
@@ -203,7 +203,10 @@ class DenseOpticalFlowRLOFImpl : public DenseRLOFOpticalFlow
             filtered_prevPoints = prevPoints;
             filtered_currPoints = currPoints;
         }
-
+        // Interpolators below expect non empty matches
+        if (filtered_prevPoints.empty())
+            return;
```
alalek, line 208: /cc @tsenst Could you please take a look on this change? Should we "reset" some fields/parameters to return consistent result?

MrKepzie: I ammended the commit with correct author, i think this should be fixed

tsenst: I can take a look during the next days

tsenst: @MrKepzie thank you for this bugfix :+1: . Here the optical flow could not be computed because of the algorithm behaviour. It would be great if it would be possible to indicate this. Unfortunately the DenseOpticalFlow class has no interface for the error case. Maybe throwing an exception would be the correct way. If not the flow should be at least set to 0 ( e.g. flow.setTo(0) ). In both cases please update the documentation.

alalek: `CV_Error(Error::StsNoConv, ...)` may be used to indicate issue (throwing an exception).

MrKepzie: I think it's better to return a vector field with 0 everywhere is this is not really an error but just 2 images without any correspondences, I'm going to ammend commit and do as suggested

tsenst: And please add a note in the docfile that when no correspondences could be found for interpolation the optical flow field would be 0

MrKepzie: Just updated the commit

tsenst (approved):
Well done.

alalek (approved):

PR #2809, opened 2021-01-07
WeChat QRCode module open source.

dddzg:

There is an open-source version of the WeChat QR code detector.

Check #2821  for more information.

```
force_builders=linux,docs
```


dddzg:
There are four primary features of the WeChat QR code detector:

1. CNN-based QR code detector. Different from the traditional detector, we introduce a tiny CNN model for multiple code detection. The detector is based on SSD architecture with a MobileNetV2-like backbone, which is run on caffe inference framework.
 
2. CNN-based QR code enhancement. To improve the performance of tiny QR code, we design a lighten super-resolution CNN model for QR code, called QRSR. Depth-wise convolution, DenseNet concat and deconvolution are the core techniques in the QRSR model.
 
3. More robust finder pattern detection. Besides traditional horizontal line searching, we propose an area size based finder pattern detection method. we calculate the area size of black and white blocks to locate the finder pattern by the pre-computed connected cells.

4. Massive engineering optimization. Based on [zing-cpp](https://github.com/glassechidna/zxing-cpp), we conduct massive engineering optimization to boost the decoding success rate, such as trying more binarization methods, supporting N:1:3:1:1 finder pattern detection, finding more alignment pattern, clustering similar size finder pattern, and etc.

garybradski:
zing-cpp is based on Apache 2 which is good, but the file is called COPYING. It would be more standard if that was changed to LICENSE.

dddzg:
Hi, Gary.
We guess that there is no distinct difference between COPYING and LICENSE. For example, we find that [git](https://github.com/git/git) also uses COPYING.

vpisarev:
@dddzg, thank you a lot for the excellent contribution! Overall, the code is an excellent shape. Please, see some comments below

modules/wechat_qrcode/include/opencv2/wechat_qrcode.hpp
```diff
@@ -0,0 +1,89 @@
...
+#include "opencv2/core.hpp"
+#include "opencv2/wechat_qrcode/align.hpp"
+#include "opencv2/wechat_qrcode/ssd_detector.hpp"
+#include "opencv2/wechat_qrcode/super_scale.hpp"
+
+/** @defgroup wechat_qrcode WeChat QR code detector for detecting and parsing QR code.
+*/
+namespace cv {
+namespace wechat_qrcode {
+using std::string;
```
vpisarev, line 19: please, do not define such global aliases for std types in public headers. use explicit `std::`. Such aliases can be moved to precomp.hpp

modules/wechat_qrcode/include/opencv2/wechat_qrcode.hpp
```diff
@@ -0,0 +1,89 @@
...
+     * @brief Initialize the QRCodeDetector.
+     * Two models are packaged with caffe format.
+     * Therefore, there are prototxt and caffe model two files.
+     *
+     * @param detector_prototxt_path prototxt file path for the detector
+     * @param detector_caffe_model_path caffe model file path for the detector
+     * @param super_resolution_prototxt_path prototxt file path for the super resolution model
+     * @param super_resolution_caffe_model_path caffe file path for the super resolution model
+     */
+    CV_WRAP QRCodeDetector(const String& detector_prototxt_path = "",
```
vpisarev, line 43: `String` is obsolete; please, use `std::string`

modules/wechat_qrcode/include/opencv2/wechat_qrcode.hpp
```diff
@@ -0,0 +1,89 @@
...
+     * To simplify the usage, there is a only API: detectAndDecode
+     *
+     * @param img supports grayscale or color (BGR) image.
+     * @param points optional output array of vertices of the found QR code quadrangle. Will be
+     * empty if not found.
+     * @return list of decoded string.
+     */
+    CV_WRAP vector<string> detectAndDecode(InputArray img, OutputArrayOfArrays points = noArray());
+
+private:
```
vpisarev, line 60: Please, refactor the class to use either `class Impl; Ptr<Impl>/std::shared_ptr<Impl> pImpl;` Please, see https://github.com/opencv/opencv/wiki/Coding_Style_Guide#high-level-c-interface-algorithms for details

modules/wechat_qrcode/include/opencv2/wechat_qrcode.hpp
```diff
@@ -0,0 +1,89 @@
...
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Tencent is pleased to support the open source community by making WeChat QRCode available.
+// Copyright (C) 2020 THL A29 Limited, a Tencent company. All rights reserved.
+
+#ifndef __OPENCV_WECHAT_QRCODE_HPP__
+#define __OPENCV_WECHAT_QRCODE_HPP__
+#include "opencv2/core.hpp"
+#include "opencv2/wechat_qrcode/align.hpp"
```
vpisarev, line 11: can we move all those headers: align, ssd_detector and super_scale inside `src`? It looks like they are not used for QR code detection and recognition. We can always expose more functionality later, but it's much more difficult to hide functionality that we do not need anymore, because some users may rely on it

dddzg:
Hi @vpisarev , thanks a lot for your suggestions! We have refactored the code. And there is a  new pull request #2821 (not from `master`). Please, check it out.

vpisarev:
thank you! yes, let me close this one

Issue #1301, opened 2017-07-28
fatal error: boostdesc_bgm.i: No such file or directory


jinfagang:
Using this cmake config:
```
cmake -D CMAKE_BUILD_TYPE=RELEASE -D CMAKE_INSTALL_PREFIX=/usr/local -D INSTALL_PYTHON_EXAMPLES=ON 
-D OPENCV_EXTRA_MODULES_PATH=~/opencv/opencv_contrib/modules -D BUILD_opencv_dnns_easily_fooled=OFF 
-D BUILD_opencv_dnn_modern=OFF -D BUILD_opencv_cnn_3dobj=OFF 
-D PYTHON_EXECUTABLE=/usr/bin/python
 -D WITH_CUDA=OFF -D BUILD_SHARED_LIBS=OFF
 -D BUILD_EXAMPLES=ON  ..

```
Just disable some dnn module which need caffe, since I am not installed caffe. But I need xfeatures module which give me those errors. It seems boostdesc.cpp not able to found, but I installed libboost-all-dev.
Any idea about this would be very appreciated, because the exactly same process I have build succeed many times except this. My script build from latest source so I think it **might be a recently bug**!
```
[ 80%] Building CXX object modules/xfeatures2d/CMakeFiles/opencv_xfeatures2d.dir/src/boostdesc.cpp.o

/root/opencv/opencv_contrib/modules/xfeatures2d/src/boostdesc.cpp:646:37: fatal error: boostdesc_bgm.i: No such file or directory

compilation terminated.

modules/xfeatures2d/CMakeFiles/opencv_xfeatures2d.dir/build.make:94: recipe for target 'modules/xfeatures2d/CMakeFiles/opencv_xfeatures2d.dir/src/boostdesc.cpp.o' failed

```

alalek:
Check CMake stdout log - it should warn about failed files downloading. Check CmakeDownloadLog.txt from build folder for details.

Currently `xfeatures2d` modules doesn't support conditional compilation with missing external files.

jinfagang:
Many thanks, indeed cmake out shows that all things required by xfeature2d are all download fail:
```
CMake Warning at cmake/OpenCVDownload.cmake:188 (message):
  xfeatures2d/boostdesc: Download failed: 1;"Unsupported protocol"

  For details please refer to the download log file:

  /root/opencv/build/CMakeDownloadLog.txt

Call Stack (most recent call first):
[...]

PR #2815, opened 2021-01-13
Properly disable xfeatures2d if extra files cannot be downloaded

asmorkalov:

### DIFF
--- a/modules/xfeatures2d/CMakeLists.txt
+++ b/modules/xfeatures2d/CMakeLists.txt
@@ -3,7 +3,6 @@ set(the_description "Contributed/Experimental Algorithms for Salient 2D Features
 if(HAVE_CUDA)
   ocv_warnings_disable(CMAKE_CXX_FLAGS -Wundef)
 endif()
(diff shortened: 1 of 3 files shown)


alalek:
Check the issue #1301 and [this comment](https://github.com/opencv/opencv_contrib/issues/1131#issuecomment-296158328).

Root cause of configuration failure (broken compilation) is that `ocv_module_disable()` doesn't work properly on the second modules pass (too late).

There are several workarounds:
- update `.cache` manually
- disable `-DBUILD_opencv_xfeatures2d=OFF`

Proper fix should look like:
- add "optional" support for these blobs (`#ifdef DATA_BLOBS_DOWNLOADED` + `Error::StsNotImplemented` otherwise) instead of whole module disabling
- try to fix `ocv_module_disable()`


Current proposal just hides the problem (which is not a problem in standard normal build environment). With the same success this can be done through `-DBUILD_opencv_xfeatures2d=OFF` CMake flag without changes in codebase.

asmorkalov:
@alalek I added extra definitions and guards in code. The module is always enabled, but some classes can be not implemented (exception).

alalek:
#2820

PR #2821, opened 2021-01-15
WeChat QRCode module open source.

dddzg:
**Merge with 3rdparty**: https://github.com/opencv/opencv_3rdparty/pull/59

There is an open-source version of the WeChat QR code detector.

Check #2809  for more information.

```
force_builders=ARMv7,Custom
build_image:Custom=centos:7
buildworker:Custom=linux-1

build_image:Docs=docs-js
```

```
allow_multiple_commits=1
```


modules/wechat_qrcode/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/src)
```
alalek, line 1: Not allowed. Use relative paths instead.

See mess with precomp headers:

```
In file included from /Volumes/build-storage/build/precommit-contrib_ios/opencv/modules/gapi/src/api/gframe.cpp:8:
In file included from /Volumes/build-storage/build/precommit-contrib_ios/opencv_contrib/modules/wechat_qrcode/src/precomp.hpp:25:
```

dddzg: Thanks a lot! I was really confused yesterday about the compile failure.

alalek: Relative paths should be used in .cpp/.hpp files for `#include` directives.

dddzg: I know your meaning. I tried to save time by modifying CMakeLists instead of `#include` (As you can see, there are many files under the zxing module). However, I failed. Now, I have changed them with `#include` relative paths. Thanks!

dddzg:
Hi @alalek, sorry for bothering you again. Now, the code successfully supports different platforms. However, there are two issues that we meet:
1. As we can see, the patch size exceeds 1MB, could we ignore this warning?
2. during debugging on iOS, we find that we can't name the API class as `QRCodeDetector` because there is a QRCodeDetector in OpenCV `objdetect` module. Even we are under different modules (one is under `objdetect` and the other is under `wechat_qrcode`, check logs below), there is still an error. Maybe it is a feature, not a bug. So we rename the `QRCodeDetector` class as `WeChatQRCode` temporarily. What are your ideas?

logs:
`error: Multiple commands produce '/Volumes/build-storage/build/precommit-contrib_ios/build/build_ios_contrib/build/build-armv7-iphoneos/lib/Release/opencv2.framework/Headers/QRCodeDetector.h': `
1) Target 'opencv2' (project 'opencv2') has copy command from '/Volumes/build-storage/build/precommit-contrib_ios/build/build_ios_contrib/build/build-armv7-iphoneos/modules/objc_bindings_generator/ios/gen/objc/**objdetect/QRCodeDetector.h**' to '/Volumes/build-storage/build/precommit-contrib_ios/build/build_ios_contrib/build/build-armv7-iphoneos/lib/Release/opencv2.framework/Headers/QRCodeDetector.h' `
 2) Target 'opencv2' (project 'opencv2') has copy command from '/Volumes/build-storage/build/precommit-contrib_ios/build/build_ios_contrib/build/build-armv7-iphoneos/modules/objc_bindings_generator/ios/gen/objc/**wechat_qrcode/QRCodeDetector.h**' to '/Volumes/build-storage/build/precommit-contrib_ios/build/build_ios_contrib/build/build-armv7-iphoneos/lib/Release/opencv2.framework/Headers/QRCodeDetector.h'

alalek:
1.

This repository should store source code only. Models and other data blobs should be stored externally.
(There is an exception for images for samples / documentation only. Images for tests are stored in opencv_extra).

There are several options:
- [opencv_3rdparty](https://github.com/opencv/opencv_3rdparty) repository with data on branches ([example for face detector DNN data](https://github.com/opencv/opencv_3rdparty/branches/all?query=dnn_samples_)). Usually used with C++ code (downloaded by CMake scripts). This data is downloaded through `https://raw.githubusercontent.com` (AFAIK, this may cause issues for devs from China due to GFW)
- GoogleDrive, DropBox, etc (need to check if CMake is able to download data from these locations (to use in C++ code) - there are several limitations). There are no OpenCV accounts for that. Authors use their own.

Take a look `ocv_download()` function in CMake scripts.

---

2.
There is `CV_WRAP_AS(WeChatQRCode)` / `CV_EXPORTS_AS(WeChatQRCode)` macros. Need to check if Obj-C binding generator support that.

> Headers/QRCodeDetector.h

In general binding generator should handle nested namespaces (target name may look like `<namespace>_<className>`). Perhaps we need to improve generator first. Consider disabling `objc` wrapper for now.

Just renaming to `WeChatQRCode` should be OK too.

dddzg:
Hi, @alalek . I have added `ocv_download` related code. There is a new pull request for opencv_3rdparty https://github.com/opencv/opencv_3rdparty/pull/59. Please check it out.  I can't start a pull request to `wechat_qrcode` branch, so I commit to `readme` for `opencv_3rdparty`. Could you help me?

BTW, there are still some issues:
1. I have moved the models with `ocv_download`. However, the code is too much. It still exceeds the 1MB patch limit. Could we ignore it?
2. For `ocv_download`, I didn't find a standard requirement about `DESTINATION_DIR`. Now, it is set to `"${CMAKE_BINARY_DIR}/downloads/wechat_qrcode"`. What are your ideas?
3. As you mentioned `<namespace>_<className>`, it reminds me that, for the python binding generator, it also follows `<namespace>_<className>` template. So, it leads to a little weird API result likes `cv2.wechat_qrcode_WeChatQRCode` (can we adjust it to `cv2.wechat_qrcode.WeChatQRCode` or `cv2.WeChatQRCode`). However, For C++, it looks well for `cv::wechat_qrcode::WeChatQRCode`

alalek:
Thank you for update!

> I can't start a pull request to the wechat_qrcode branch

Please prepare 2 commits in this PR:
- base branch is `root` (currently it is `readme` which is not correct), you can change target branch in proposed PR using the "Edit" button.
- 1st commit with readme file (I will create `wechat_qrcode` branch on this commit before the merge)
- 2nd commit with models itself (I will create `wechat_qrcode_20210115` branch during the merge)

(no need to re-create PR, use `--force` push option)

---

1.
> However, the code is too much. It still exceeds the 1MB patch limit.
> dddzg wants to merge 59 commits

Need to squash commits into one (all binaries is still in git history)

---

2.
> I didn't find a standard requirement about DESTINATION_DIR.
Now, it is set to "${CMAKE_BINARY_DIR}/downloads/wechat_qrcode". What are your ideas?

This is the right path.

---

3.
> it leads to a little weird API result likes cv2.wechat_qrcode_WeChatQRCode

Agreed, we need to take a look on this.

> cv2.wechat_qrcode.WeChatQRCode

Submodules are not supported yet in Python bindings. However it makes sense (but we need to preserve existed Python API too to avoid breaking of Users code).

@vpisarev Any suggestions about bindings?

vpisarev:
@dddzg, @alalek, in my opinion, the patch looks great. I suggest to merge it in and adjust Python bindings/fix ObjC bindings etc. later if needed. It's usable, it works really well and that's enough for now.

@dddzg, may I ask you to do 2 things:

1. take the updated C++ example from here: https://github.com/vpisarev/opencv_contrib/blob/wechat_qrcode/modules/wechat_qrcode/samples/qrcode_example.cpp. It would be nice if the Python example is updated in a similar way.

2. make sure all the builders are green (except for the doc builder which will report warning about the patch size). Then squash commits into one:

```
$ git reset --soft HEAD~number_of_commits_in_PR  # now it would be 59, but after updating qrcode_example.cpp (and optionally qrcode.py the number will increase)
$ git commit
$ git push --force origin
```

after it the pull request will contain just 1 commit and the doc builder warning will gone, since the patch will not include model files anymore.

dddzg:
Sorry for the delay.
Hi @alalek, I have updated the `opencv_3rdparty` with 2 commits, https://github.com/opencv/opencv_3rdparty/pull/59. 
Hi @vpisarev , thanks for your suggestions, I have updated the C++ and python examples and squashed commits into one.

modules/wechat_qrcode/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+set(the_description "WeChat QR code Detector")
+ocv_define_module(wechat_qrcode opencv_core opencv_imgproc opencv_dnn WRAP java objc python js)
+
+# need to change
+set(wechat_qrcode_commit_hash "c05918e91a282728acd85b8c43281c2d7a3e63ab")
```
alalek, line 5: Please use `a8b69ccc738421293254aec5ddb38bd523503252` here
(from https://github.com/opencv/opencv_3rdparty/pull/59)

alalek (commented):
> I have updated the opencv_3rdparty with 2 commits

Update looks good. Thank you!

vpisarev:
@alalek, all the tests pass now; in my opinion, PR is ready to be merged :+1:

modules/wechat_qrcode/src/binarizermgr.cpp
```diff
@@ -0,0 +1,70 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Tencent is pleased to support the open source community by making WeChat QRCode available.
+// Copyright (C) 2020 THL A29 Limited, a Tencent company. All rights reserved.
+
+#include "binarizermgr.hpp"
+#include "imgsource.hpp"
+#include "precomp.hpp"
```
alalek, line 10: > `#include "precomp.hpp"`

should go first (for each `src/*.cpp` file).

modules/wechat_qrcode/src/detector/align.hpp
```diff
@@ -0,0 +1,42 @@
...
+//
+// Tencent is pleased to support the open source community by making WeChat QRCode available.
+// Copyright (C) 2020 THL A29 Limited, a Tencent company. All rights reserved.
+
+#ifndef __DETECTOR_ALIGN_HPP_
+#define __DETECTOR_ALIGN_HPP_
+
+#include <stdio.h>
+#include <fstream>
+#include "../precomp.hpp"
```
alalek, line 13: > #include "../precomp.hpp"

Relative path is valid, but `precomp.hpp` should not be used in .hpp files (the first include of `.cpp` files).

vpisarev (approved):

vpisarev:
@alalek, if you do not have any objections, I suggest @dddzg to squash commits into one once again; then it can be merged using our automated tool

modules/wechat_qrcode/samples/qrcode.py
```diff
@@ -0,0 +1,60 @@
+# This file is part of OpenCV project.
+# It is subject to the license terms in the LICENSE file found in the top-level directory
+# of this distribution and at http://opencv.org/license.html.
+#
+# Tencent is pleased to support the open source community by making WeChat QRCode available.
+# Copyright (C) 2020 THL A29 Limited, a Tencent company. All rights reserved.
```
alalek, line 6: Sample files in OpenCV don't include license headers (they are often used in documentation).

modules/wechat_qrcode/src/zxing/common/kmeans.cpp
```diff
@@ -0,0 +1,96 @@
...
+            val += pow((a[i] - b[i]), 2);
+    }
+    return pow(val, 0.5);
+}
+
+/*
+ * maxepoches max iteration epochs
+ * minchanged min central change times
+ */
+vector<Cluster> k_means(vector<vector<double> > trainX, uint k, uint maxepoches, uint minchanged) {
```
alalek, line 42: Future TODO: try to reuse `cv::kmeans()`

modules/wechat_qrcode/src/zxing/common/mathutils.hpp
```diff
@@ -0,0 +1,131 @@
...
+private:
+    MathUtils();
+    ~MathUtils();
+
+public:
+    /**
+     * Ends up being a bit faster than {@link Math#round(float)}. This merely
+     * rounds its argument to the nearest int, where x.5 rounds up to x+1.
+     */
+    static inline int round(double value) {
```
alalek, line 37: Future TODO: `cvRound()`

alalek (commented):
Great contribution!

modules/wechat_qrcode/LICENSE
```diff
@@ -0,0 +1,298 @@
...
+
+9. Accepting Warranty or Additional Liability. While redistributing the Work or Derivative Works thereof, You may choose to offer, and charge a fee for, acceptance of support, warranty, indemnity, or other liability obligations and/or rights consistent with this License. However, in accepting such obligations, You may act only on Your own behalf and on Your sole responsibility, not on behalf of any other Contributor, and only if You agree to indemnify, defend, and hold each Contributor harmless for any liability incurred by, or claims asserted against, such Contributor by reason of your accepting any such warranty or additional liability.
+
+END OF TERMS AND CONDITIONS
+
+
+========================================================================
+jai-imageio
```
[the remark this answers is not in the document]
dddzg: it is copied from https://github.com/zxing/zxing/blob/master/LICENSE.
We found that it is a zxing java dependency. We shall remove it.

modules/wechat_qrcode/src/wechat_qrcode.cpp
```diff
@@ -0,0 +1,210 @@
...
+WeChatQRCode::WeChatQRCode(const String& detector_prototxt_path,
+                           const String& detector_caffe_model_path,
+                           const String& super_resolution_prototxt_path,
+                           const String& super_resolution_caffe_model_path) {
+    p = makePtr<WeChatQRCode::Impl>();
+    if (!detector_caffe_model_path.empty() && !detector_prototxt_path.empty()) {
+        // initialize detector model (caffe)
+        p->use_nn_detector_ = true;
+        CV_CheckEQ(utils::fs::exists(detector_prototxt_path), true,
+                   "fail to find detector caffe prototxt file");
```
alalek, line 58: `bool` values are not supported for now in `CV_Check` macros (see `ARMv7` builder warnings).
Lets use `CV_Assert()` instead for now:

```
CV_Assert(utils::fs::exists(detector_prototxt_path));
```

(similar cases are below)

modules/wechat_qrcode/src/wechat_qrcode.cpp
```diff
@@ -0,0 +1,210 @@
...
+        p->use_nn_detector_ = false;
+        p->detector_ = NULL;
+    }
+    // initialize super_resolution_model
+    // it could also support non model weights by cubic resizing
+    // so, we initialize it first.
+    p->super_resolution_model_ = make_shared<SuperScale>();
+    if (!super_resolution_prototxt_path.empty() && !super_resolution_caffe_model_path.empty()) {
+        p->use_nn_sr_ = true;
+        // initialize dnn model (onnx format)
```
alalek, line 74: onnx?

modules/wechat_qrcode/src/wechat_qrcode.cpp
```diff
@@ -0,0 +1,210 @@
...
+#include "precomp.hpp"
+#include "opencv2/wechat_qrcode.hpp"
+#include "decodermgr.hpp"
+#include "detector/align.hpp"
+#include "detector/ssd_detector.hpp"
+#include "opencv2/core.hpp"
+#include "opencv2/core/utils/filesystem.hpp"
+#include "scale/super_scale.hpp"
+#include "zxing/result.hpp"
+using cv::InputArray;
```
alalek, line 16: > using cv::InputArray;

Please remove this line. It is not necessary.

modules/wechat_qrcode/src/wechat_qrcode.cpp
```diff
@@ -0,0 +1,210 @@
...
+};
+
+vector<string> WeChatQRCode::Impl::decode(const Mat& img, vector<Mat>& candidate_points,
+                                          vector<Mat>& points) {
+    if (candidate_points.size() == 0) {
+        return vector<string>();
+    }
+    vector<string> decode_results;
+    for (auto& point : candidate_points) {
+        cv::Mat cropped_img;
```
alalek, line 129: > cv::

Please avoid using of `cv::` prefix in OpenCV sources (under namespace cv)

modules/wechat_qrcode/src/wechat_qrcode.cpp
```diff
@@ -0,0 +1,210 @@
...
+    int incn = img.channels();
+    CV_Check(incn, incn == 1 || incn == 3 || incn == 4, "");
+    if (incn == 3 || incn == 4) {
+        cv::cvtColor(img, input_img, cv::COLOR_BGR2GRAY);
+    } else {
+        input_img = img.getMat();
+    }
+    auto candidate_points = p->detect(input_img);
+    auto res_points = vector<Mat>();
+    auto ret = p->decode(input_img, candidate_points, res_points);
```
alalek, line 104: Are points fixed size? 4 points per instance?

if so, then it make sense to return 2D Mat instead of vector:
- rows are instances
- cols(=4) are corners points of `CV_32FC2` elements (C2 means 2 channels for x/y components)

dddzg: The size of points is not fixed. Because our detector supports detecting multiple QRcode in an image. Moreover, not all detected QRCode can be decoded. So, there are `candidate_points` detected by the detector. Then we try to decode them one by one. For the successfully decoded qrcode, we will append their points to the `res_points`.

So, I prefer to use the `vector`, which is more convenient to append. Could we keep it?

alalek: > multiple QRcode in an image

So, what is represented by each vector element? 4 points corners?

---

This question is about external API representation (`points` output argument). Internal implementation can be any (we don't worry about that).

vpisarev: @alalek, I suggest to merge the current PR as-is.

alalek:
@vpisarev Do you want to merge opencv_3rdparty?

Why do you violate existed rules? ("fixup" commits should not go to upstream repository)

modules/wechat_qrcode/src/wechat_qrcode.cpp
alalek: @vpisarev Please make careful reviews. For example, `.fixedType()` check below doesn't make any sense. Are you going to fix that?

dddzg: @alalek  Yes, each Mat in `vector` represents 4 points corners, it can also be represented as `Mat(4, 2, CV_32FC1)`.

dddzg: Hi @alalek , the code `fixedType()` comes from https://github.com/opencv/opencv/blob/7a790d0d353bcbe48f1d62fb917c787ab4185846/modules/objdetect/src/qrcode.cpp#L49  Because I try to convert `vector<Mat>` to `OutputArrayOfArrays`. Do you have any good ideas?

modules/wechat_qrcode/src/wechat_qrcode.cpp
```diff
@@ -0,0 +1,205 @@
...
+    // opencv type convert
+    vector<Mat> tmp_points;
+    if (points.needed()) {
+        for (size_t i = 0; i < res_points.size(); i++) {
+            Mat tmp_point;
+            tmp_points.push_back(tmp_point);
+            res_points[i].convertTo(((OutputArray)tmp_points[i]),
+                                    ((OutputArray)tmp_points[i]).fixedType()
+                                        ? ((OutputArray)tmp_points[i]).type()
+                                        : CV_32FC2);
```
alalek, line 109: `.fixedType()` is used with external arguments only (mentioned file from `objdetect` follows this rule).
No sense to use it with "local" variables.

---

> good ideas

Reuse conventions from external API of objdetect/QRCodeDetector

dddzg: OK. I will remove it.

dddzg: @alalek  Could we reopen this PR?

vpisarev:
@alalek, it looks like our tool does not merge opencv_3rdparty, and I do not have permissions to merge into it manually. So, could you please do it?

@dddzg, open the new PR, this one is merged already

modules/wechat_qrcode/src/wechat_qrcode.cpp
CloseServer: @dddzg 
Hello, are there any steps for use and deployment? I want to try to use it in Golang, but I can't run the project of wechat_qrcode. Cmake is always running unsuccessfully, and I meet all kinds of errors, which makes me very upset!

CloseServer: @VladX

dddzg: Could you list the errors with more details?

PR #2807, opened 2021-01-06
NVIDIA_OPTICAL_FLOW_2_0_INTEGRATION

vchiluka5:
This PR integrates NVIDIA'S Optical Flow SDK 2.0.

- [*] I agree to contribute to the project under Apache 2 License.
- [*] To the best of my knowledge, the proposed patch is not based on a code under GPL or other license that is incompatible with OpenCV
- [*] The PR is proposed to proper branch
- [*] There is reference to original bug report and related work
- [*] There is accuracy test, performance test and test data in opencv_extra repository, if applicable
      Patch to opencv_extra has the same branch name.
- [*] The feature is well documented and sample code can be built with the project CMake

```
force_builders=Custom
buildworker:Custom=linux-4
build_image:Custom=ubuntu-cuda:18.04
```

PR #2824, opened 2021-01-18
cudaoptflow: disable Optical Flow SDK when CUDA version is insufficient

tomoaki0705:
relates #2807 

Just modifying this line will let the build pass for both CUDA 8.0 and CUDA 10.0

/cc @vchiluka5 
I understand that Optical flow API only supports CUDA 10.0 and later. Thank you for the comment.
Though, I'd like to propose this fix as a quick fix of build.

Eventually, I can add to disable Optical Flow API when CUDA is less than 10.0

 - Before
```
[ 81%] Building NVCC (Device) object modules/cudaoptflow/CMakeFiles/cuda_compile_1.dir/src/cuda/cuda_compile_1_generated_nvidiaOpticalFlow.cu.o
/opencv_contrib/modules/cudaoptflow/src/cuda/nvidiaOpticalFlow.cu(76): error: no instance of overloaded function "surf2Dwrite" matches the argument list
            argument types are: (short2, cudaSurfaceObject_t, unsigned long, int, cudaSurfaceBoundaryMode)

1 error detected in the compilation of "/tmp/tmpxft_00007a2a_00000000-7_nvidiaOpticalFlow.cpp1.ii".
CMake Error at cuda_compile_1_generated_nvidiaOpticalFlow.cu.o.Release.cmake:279 (message):
  Error generating file
  /opencv/build/modules/cudaoptflow/CMakeFiles/cuda_compile_1.dir/src/cuda/./cuda_compile_1_generated_nvidiaOpticalFlow.cu.o
```
 - After
```
[ 82%] Building NVCC (Device) object modules/cudaoptflow/CMakeFiles/cuda_compile_1.dir/src/cuda/cuda_compile_1_generated_nvidiaOpticalFlow.cu.o
 : 
[ 86%] Built target opencv_cudaoptflow
```

```
force_builders=Custom
buildworker:Custom=linux-4
```

### DIFF
--- a/modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
+++ b/modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
@@ -9,26 +9,39 @@
 #if !defined HAVE_CUDA || defined(CUDA_DISABLER)
 
 cv::Ptr<cv::cuda::NvidiaOpticalFlow_1_0> cv::cuda::NvidiaOpticalFlow_1_0::create
-    (int, int, NVIDIA_OF_PERF_LEVEL, bool, bool, bool, int, Stream&, Stream&) {
+    (cv::Size, NVIDIA_OF_PERF_LEVEL, bool, bool, bool, int, Stream&, Stream&) {
     throw_no_cuda(); return cv::Ptr<cv::cuda::NvidiaOpticalFlow_1_0>(); }
 
 cv::Ptr<cv::cuda::NvidiaOpticalFlow_2_0> cv::cuda::NvidiaOpticalFlow_2_0::create(
-    int, int, NVIDIA_OF_PERF_LEVEL, NVIDIA_OF_OUTPUT_VECTOR_GRID_SIZE, NVIDIA_OF_HINT_VECTOR_GRID_SIZE,
-    bool, int, cv::Rect*, bool, bool, bool, int, Stream&, Stream&) {
+    cv::Size, NVIDIA_OF_PERF_LEVEL, NVIDIA_OF_OUTPUT_VECTOR_GRID_SIZE, NVIDIA_OF_HINT_VECTOR_GRID_SIZE,
+    bool, bool, bool, int, Stream&, Stream&) {
+    throw_no_cuda(); return cv::Ptr<cv::cuda::NvidiaOpticalFlow_2_0>();
+}
+
+cv::Ptr<cv::cuda::NvidiaOpticalFlow_2_0> cv::cuda::NvidiaOpticalFlow_2_0::create(
+    cv::Size, std::vector<Rect>, NVIDIA_OF_PERF_LEVEL, NVIDIA_OF_OUTPUT_VECTOR_GRID_SIZE, NVIDIA_OF_HINT_VECTOR_GRID_SIZE,
+    bool, bool, bool, int, Stream&, Stream&) {
     throw_no_cuda(); return cv::Ptr<cv::cuda::NvidiaOpticalFlow_2_0>();
 }
 
 #elif !defined HAVE_NVIDIA_OPTFLOW
 
 cv::Ptr<cv::cuda::NvidiaOpticalFlow_1_0> cv::cuda::NvidiaOpticalFlow_1_0::create(
-    int, int, NVIDIA_OF_PERF_LEVEL, bool, bool, bool, int, Stream&, Stream&)
+    cv::Size, NVIDIA_OF_PERF_LEVEL, bool, bool, bool, int, Stream&, Stream&)
 {
     CV_Error(cv::Error::HeaderIsNull, "OpenCV was build without NVIDIA OpticalFlow support");
 }
 
 cv::Ptr<cv::cuda::NvidiaOpticalFlow_2_0> cv::cuda::NvidiaOpticalFlow_2_0::create(
-    int, int, NVIDIA_OF_PERF_LEVEL, NVIDIA_OF_OUTPUT_VECTOR_GRID_SIZE, NVIDIA_OF_HINT_VECTOR_GRID_SIZE,
-    bool, int, cv::Rect*, bool, bool, bool, int, , Stream&, Stream&)
+    cv::Size, NVIDIA_OF_PERF_LEVEL, NVIDIA_OF_OUTPUT_VECTOR_GRID_SIZE, NVIDIA_OF_HINT_VECTOR_GRID_SIZE,
+    bool, bool, bool, int, Stream&, Stream&)
+{
+    CV_Error(cv::Error::HeaderIsNull, "OpenCV was build without NVIDIA OpticalFlow support");
+}
+
+cv::Ptr<cv::cuda::NvidiaOpticalFlow_2_0> cv::cuda::NvidiaOpticalFlow_2_0::create(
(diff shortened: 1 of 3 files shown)


vchiluka5:
Sure @tomoaki0705. 
Thanks for the quick fix. I'll need to verify this fix with Optical flow outputs. 
Il get back once i verify and let you know if this is a valid fix. 
Please **DO NOT SUBMIT** till then. ETA 1/2 days

vchiluka5:
> Sure @tomoaki0705.
> Thanks for the quick fix. I'll need to verify this fix with Optical flow outputs.
> Il get back once i verify and let you know if this is a valid fix.
> Please **DO NOT SUBMIT** till then. ETA 1/2 days

Hi @tomoaki0705 
The change you suggested breaks the code. Optical flow vector outputs are empty with your change.
surf2dWrite requires to be called with short2 type so that proper memory allocation is done and final flow vectors are generated.

Please DO NOT SUBMIT this change instead
you can upgrade to cuda 10 FROM CUDA 8 
OR
Add the execption to not include NVIDIA_OPTFLOW because anyway code wont run with CUDA 8.

Thanks,
Vishal

alalek:
> only supports CUDA 10.0 and later

@vchiluka5 Perhaps this should be handled in CMake scripts through CUDA SDK version check:
- disable v2.0 for CUDA<10.0
- or fallback on previous v1.0

tomoaki0705:
Thank you @alalek @vchiluka5 
I understand that this PR is helpless, so I want to shift the point of this PR to @alalek point

> > only supports CUDA 10.0 and later
> 
> @vchiluka5 Perhaps this should be handled in CMake scripts through CUDA SDK version check:
> 
>     * disable v2.0 for CUDA<10.0
> 
>     * or fallback on previous v1.0

I'd like to propose to fallback to v1.0 in master branch (4.5 series)
To migrate totally to NVIDIA Optical Flow SDK **v2.0**, this should happen in next branch (5.0 series), at least in my opinion.
For falling back to v1.0, I'm working on this on another branch.

@vchiluka5 , what I'd like to double check is that is Optical Flow SDK v1.0 totally deprecated ?
Does falling back to v1.0 makes sense to you ?

vchiluka5:
> Thank you @alalek @vchiluka5
> I understand that this PR is helpless, so I want to shift the point of this PR to @alalek point
> 
> > > only supports CUDA 10.0 and later
> > 
> > 
> > @vchiluka5 Perhaps this should be handled in CMake scripts through CUDA SDK version check:
> > ```
> > * disable v2.0 for CUDA<10.0
> > 
> > * or fallback on previous v1.0
> > ```
> 
> I'd like to propose to fallback to v1.0 in master branch (4.5 series)
> To migrate totally to NVIDIA Optical Flow SDK **v2.0**, this should happen in next branch (5.0 series), at least in my opinion.
> For falling back to v1.0, I'm working on this on another branch.
> 
> @vchiluka5 , what I'd like to double check is that is Optical Flow SDK v1.0 totally deprecated ?
> Does falling back to v1.0 makes sense to you ?

il suggest to disable the v2.0 for CUDA < 10.0. Because CUDA 8 doesnt have proper compatibility with turing architecture and optical flow sdk 1.0 works from turing architecture and above. 
Optical flow sdk 1.0 is NOT deprecated. Users can use both sdk's but not with CUDA 8. So better we disable both 2.0 and 1.0 SDK if its CUDA <10.
@tomoaki0705  if you can work on those changes then Great since i am not aware about cmake changes.
Otherwise @alalek  please guide me on making those changes

tomoaki0705:
Thank you @vchiluka5 !
Your explanation makes things clear.
For CUDA less than 10.0, it makes sense totally disabling optical flow sdk.
Let me work on that cmake part.

vchiluka5:
> Thank you @vchiluka5 !
> Your explanation makes things clear.
> For CUDA less than 10.0, it makes sense totally disabling optical flow sdk.
> Let me work on that cmake part.

Thanks @tomoaki0705 . That will be of great help :-)

modules/cudaoptflow/samples/nvidia_optical_flow.cpp
```diff
@@ -198,6 +198,7 @@ bool parseROI(std::string ROIFileName, std::vector<Rect>& roiData)
 
 int main(int argc, char **argv)
 {
+#if defined HAVE_NVIDIA_OPTFLOW
```
alalek, line 201: Samples should not depend on such defines.
API calls will throw an exception.

Please revert compilation guards from samples.

modules/cudaoptflow/perf/perf_optflow.cpp
```diff
@@ -326,56 +326,7 @@ PERF_TEST_P(ImagePair, OpticalFlowDual_TVL1,
     }
 }
 
-//////////////////////////////////////////////////////
-// NvidiaOpticalFlow_1_0
```
alalek, line 330: `cv::cuda::NvidiaOpticalFlow_1_0` is not removed from OpenCV public API, it is just disabled (throws exceptions).

Please restore this test.

modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
```diff
@@ -249,437 +239,6 @@ class LoadNvidiaModules
     PFNNvOFAPICreateInstanceCuda GetOFLibraryFunctionPtr() { return m_NvOFAPICreateInstanceCuda; }
 };
 
-class NvidiaOpticalFlowImpl : public cv::cuda::NvidiaOpticalFlow_1_0
```
alalek, line 252: I believe `1_0` implementation can work with `2.0` SDK.
Perhaps there is no reason to remove that.

alalek (commented):
Expected changes in `CMake` (looks good) and `src/nvidiaOpticalFlow.cpp` (just add guard for `HAVE_NVIDIA_OPTFLOW`, no code removal is expected)

tomoaki0705:
Thanks for the review @alalek.
Current my recognition is that 1.0 and 2.0 doesn't have compatibility, and that's why I dropped the 1.0 implementation.
Let me get back by checking the compatibility between 1.0 and 2.0 carefully.

modules/cudaoptflow/samples/nvidia_optical_flow.cpp
tomoaki0705: Removed. thank you

modules/cudaoptflow/perf/perf_optflow.cpp
tomoaki0705: Restored. Thank you

modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
tomoaki0705: Yes, you were right.
I could restore the `1_0` implementation.

There was another build failure but that was due to the wrong function call in the disabled part.

tomoaki0705:
So, I've restored the `1_0` implementation.
Basic modification is cmake part only.
I also fixed some wrong function calls.

vchiluka5:
Just for clarification. SDK 2.0 works with 1.0 as well and hence we have 2 classes exposed.
we pull SDK 2.0 headers which are compatible with SDK 1.0.
Thanks @alalek  for reviewing.
Thanks @tomoaki0705, Final PR looks Good to me.

alalek:
@tomoaki0705 Build is failing on .cu file compilation (check "Custom" builder with CUDA 8.0).
Could you please add `#ifdef HAVE_NVIDIA_OPTFLOW` guard there?

alalek (approved):

PR #2845, opened 2021-01-29
deltaCIEDE2000_ fix into the MCC module

kevineor:
# DeltaE CIEDE2000 function fix into MCC module

This function did not pass the test case 17 and 19 of [Gaurav Sharma's test data](http://www2.ece.rochester.edu/~gsharma/ciede2000/).

This was due to a mistake into the "RT" value, it was computed with the value image instead of image.

Sources :
 - [Wikipedia](https://en.wikipedia.org/wiki/Color_difference) ;
 - [The CIEDE2000 Color-Difference Formula : Implementation Notes, Supplementary Test Data, and Mathematical Observations](http://www2.ece.rochester.edu/~gsharma/ciede2000/ciede2000noteCRNA.pdf).

And now i have a little question about this module, why color conversion and color distances classes are not exposed to the API ?
They are really useful because, at the moment, OpenCV can't convert sRGB D65 2° to CIELAB D50 2° or perform any CIE color comparison.


kevineor:
Where does the test data come from? Should i change it?

alalek (approved):
Thank you!

PR #2849, opened 2021-02-01
Wechat qrcode module refinement

dddzg:

```
force_builders=Custom
build_image:Custom=ubuntu:20.04
```


dddzg:
As mentioned in #2821 , there are some Future TODO (e.g. using cv::kmeans(),cvRound()). We will refine the code in this pr.

dddzg:
Hi,@alalek . May I have your advice about the code before it megerd into official OpenCV?

alalek (commented):
Thank you! In general changes look good to me.

> cv::kmeans()

Do you observe some issues with OpenCV implementation?

dddzg:
The kmeans from OpenCV implementation is well. However, It has different parameters and different data structures with our implementation. As we can see, our code of kmeans is short. So, it may be complicated that we convert our `vector<vector<double>>` into OpenCV InputArray, defining distance function and apply OpenCV kmeans. Then, we convert back OutputArray into our `vector<Cluster>`. Could we preserve our kmeans?

alalek (approved):
Looks good to me! Thank you :+1:

philp123:
There's a bug in unicomblock.cpp  "void UnicomBlock::Bfs(int y, int x)"  method. There's no bound check in 
112                  m_vcIndex[iPosition] = m_iNowIdx;

And when I test the image below, it causes the wechat qrcode engine to crash.

Miracle2333:
> There's a bug in unicomblock.cpp "void UnicomBlock::Bfs(int y, int x)" method. There's no bound check in line 113
> 112 m_vcIndex[iPosition] = m_iNowIdx;
> 113 m_vcQueue[iTail++] = iNextY << 16 | iNextX;
> Which means when I debug the image below, it turns out that iTail > m_vcQueue.size(), and this out of bound error causes the
> program crash.
> 
> 

any refinement on this issue? i guess i got the same "segment fault" problem

Miracle2333:
> There's a bug in unicomblock.cpp "void UnicomBlock::Bfs(int y, int x)" method. There's no bound check in line 113
> 112 m_vcIndex[iPosition] = m_iNowIdx;
> 113 m_vcQueue[iTail++] = iNextY << 16 | iNextX;
> Which means when I debug the image below, it turns out that iTail > m_vcQueue.size(), and this out of bound error causes the
> program crash.
> 
> 

i solved this problem temporarily by making the input smaller. Just resize the input image to a smaller dim (800 * 800). The detection results could be re-projected to the original image dim based on the resizing ratio and pading size.

From my side, this method could get right results of your image.

philp123:
> > There's a bug in unicomblock.cpp "void UnicomBlock::Bfs(int y, int x)" method. There's no bound check in line 113
> > 112 m_vcIndex[iPosition] = m_iNowIdx;
> > 113 m_vcQueue[iTail++] = iNextY << 16 | iNextX;
> > Which means when I debug the image below, it turns out that iTail > m_vcQueue.size(), and this out of bound error causes the
> > program crash.
> > 
> 
> i solved this problem temporarily by making the input smaller. Just resize the input image to a smaller dim (800 * 800). The detection results could be re-projected to the original image dim based on the resizing ratio and pading size.
> 
> From my side, this method could get right results of your image.

Code has been updated, please check https://github.com/opencv/opencv_contrib/pull/2972

PR #2854, opened 2021-02-02
RGBD tests refactoring

DumDereDum:


tomoaki0705:
If you are intended of refactoring RGBD module, I really appreciate if you could merge those 2 implementations.
i.e. merge `TSDFVolumeGPU` class and `TSDFVolumeCPU` class and just keep `TSDFVolume`
`TSDFVolumeGPU` class triggers several kernels under the hood, but single kernel build failure leads directly to runtime error.
The difference can be only decided at the beginning.

It will be way more convenient for user if any kernel build/run failure fall back to CPU version during the function call.

More precisely, use following cascaded function call everywhere
```
CV_OCL_RUN(ocl_integrate(....))

// followed by CPU version
cpu_integrate();
```

Putting `HAVE_OPENCL` in the test code is really the opposite way to go.

That's why it's called "T-API" (Transparent API) [1](https://software.intel.com/content/www/us/en/develop/documentation/openvino-ovx-guide/top/interoperability-with-other-apis/interoperability-with-opencv-transparent-api.html) [2](https://learnopencv.com/opencv-transparent-api/)

savuor:
@tomoaki0705, the way to organize code could be more consistent but would involve constant copying data from/to host and GPU memory, which would involve huge stalls during code execution.

savuor (changes requested):
We don't use namespace indentation, please fix this.
Also when running TSDF GPU tests with `display=true` flag I see quite glitchy image during rendering. Using viz I could find out that rendered points are OK but normals have wrong direction. Now investigating this.

savuor:
@DumDereDum Found out what's the reason of the bug:
https://github.com/opencv/opencv_contrib/blob/master/modules/rgbd/src/opencl/tsdf.cl#L230

Here you subtract compressed values then convert them to float. This is incorrect: the values should be converted to float, then subtracted, this solves the bug.
Please fix this in this PR and add the corresponding mention to PR description.

modules/rgbd/test/ocl/test_tsdf.cpp
```diff
@@ -0,0 +1,470 @@
...
+                count++;
+            }
+        }
+    }
+    return count;
+}
+
+void normal_test(bool isHashTSDF, bool isRaycast, bool isFetchPointsNormals, bool isFetchNormals)
+{
+    auto normalCheck = [](Vec4f& vector, const int*)
```
savuor, line 355: Please add explicit `setUseOpenCL()` calls here and in `normal_test` function

modules/rgbd/test/test_tsdf.cpp
```diff
@@ -288,230 +321,202 @@ void normalsCheck(Mat normals)
...
+TEST(TSDF, fetch_points_normals) { normal_test(false, false, true, false); }
+TEST(TSDF, fetch_normals) { normal_test(false, false, false, true); }
+TEST(TSDF, valid_points) { valid_points_test(false); }
+
+TEST(HashTSDF, raycast_normals) { normal_test(true, true, false, false); }
+TEST(HashTSDF, fetch_points_normals) { normal_test(true, false, true, false); }
+TEST(HashTSDF, fetch_normals) { normal_test(true, false, false, true); }
+TEST(HashTSDF, valid_points) { valid_points_test(true); }
+#else
+TEST(TSDF_CPU, raycast_normals)
```
savuor, line 465: Please remove OCL tests from this file, we have them in `ocl/` folder

DumDereDum: No, if I remove them there would be two equal GPU tests

modules/rgbd/test/ocl/test_tsdf.cpp
DumDereDum: If we have OpenCL we don't need to use `setUseOpenCL`

savuor:
👍

PR #2860, opened 2021-02-07
fixes #2662

### LINKED ISSUE #2662: Usage of uninitialized memory in module alphamat leads to unexpected behavior.

- OpenCV => 4.4
- Operating System / Platform => Linux (Kubuntu 18.04.5) 64 Bit
- Compiler => g++ 8.4.0

##### Detailed description

In module alphamat: Usage of uninitialized Mat object leads to undefined behavior (segfault) in cm.cpp function lle

Memory allocation without initialization:
https://github.com/opencv/opencv_contrib/blob/e320d6aad04eb511c116c22661275f1be18cfe6a/modules/alphamat/src/cm.cpp#L91

First usage after declaration and allocation:
https://github.com/opencv/opencv_contrib/blob/e320d6aad04eb511c116c22661275f1be18cfe6a/modules/alphamat/src/cm.cpp#L93

Next usage of uninitialized Mat object:
https://github.com/opencv/opencv_contrib/blob/e320d6aad04eb511c116c22661275f1be18cfe6a/modules/alphamat/src/cm.cpp#L126

##### Probable solution

Change 
https://github.com/opencv/opencv_contrib/blob/e320d6aad04eb511c116c22661275f1be18cfe6a/modules/alphamat/src/cm.cpp#L93
to
```.cpp
Cones = 1;
```

##### Remark
I've tested the fix on 23 images and it works for me.

##### Steps to reproduce
A look at the source code or a run of the example code in valgrind --track-origins=yes.

       answers.opencv.org, Stack Overflow, etc and have not found solution
   
 - [x] I updated to latest OpenCV version and the issue is still there
   
Sintun:
fixes #2662

This PR fixes the usage of uninitialized memory that can lead to throws and segfaults.

I think the original author wanted to add ones to a zero-initialized Mat object.
I could have used another Constructor for the Mat object, but i chose to make the minimal amount of changes to the code to work as intended.

My proposed change does not change the used algorithm.

### DIFF
--- a/modules/alphamat/src/cm.cpp
+++ b/modules/alphamat/src/cm.cpp
@@ -90,7 +90,7 @@ void lle(my_vector_of_vectors_t& indm, my_vector_of_vectors_t& samples, float ep
     Mat ptDotN(20, 1, DataType<float>::type), imd(20, 1, DataType<float>::type);
     Mat Cones(20, 1, DataType<float>::type), Cinv(20, 1, DataType<float>::type);
     float alpha, beta, lagrangeMult;
-    Cones += 1;
+    Cones = 1;
 
     C = 0;
     rhs = 1;


alalek (approved):
Thank you for contribution!

PR #1941, opened 2018-12-10
Implementation of Quasi Dense Stereo algorithm.

dimitrisPs:
I have implemented a class that establishes dense pairwise correspondences from a stereo image pair, and I wish to contribute it to the official repository. The class extracts a sparse set of correspondences from a stereo image pair using pyramidal _Lucas-Kanade_ Flow algorithm based on Shi and Tomasi features, this way the class can process non-rectified images. To interpolate a dense set of correspondences the class uses the _Quasi Dense Stereo_ algorithm proposed by _M. Lhuillier_ and _Long Quan_ et. al. [https://ieeexplore.ieee.org/document/905620].

Please let me know if you have any further questions or comments.

Thank You

PR #2869, opened 2021-02-13
Remove normalization in cv::stereo::QuasiDenseStereo getDisparity().

dimitrisPs:
This addresses user comments made in #1941 and #2819. The getDisparity function, of QuasiDenseStereo, originally returned a disparity image normalized to span the full range between 0 and 255, which was improper and understandably confused users.

I've made the following changes:
- getDisparity now returns proper CV_32F disparity maps with nan values for unknown matches. 
- included an accuracy test.

### DIFF
--- a/modules/stereo/test/test_qds_matching.cpp
+++ b/modules/stereo/test/test_qds_matching.cpp
@@ -1,77 +1,27 @@
-/*M///////////////////////////////////////////////////////////////////////////////////////
-//
-//  IMPORTANT: READ BEFORE DOWNLOADING, COPYING, INSTALLING OR USING.
-//
-//  By downloading, copying, installing or using the software you agree to this license.
-//  If you do not agree to this license, do not download, install,
-//  copy or use the software.
-//
-//
-//                        Intel License Agreement
-//                For Open Source Computer Vision Library
-//
-// Copyright (C) 2000, Intel Corporation, all rights reserved.
-// Third party copyrights are property of their respective owners.
-//
-// Redistribution and use in source and binary forms, with or without modification,
-// are permitted provided that the following conditions are met:
-//
-//   * Redistribution's of source code must retain the above copyright notice,
-//     this list of conditions and the following disclaimer.
-//
-//   * Redistribution's in binary form must reproduce the above copyright notice,
-//     this list of conditions and the following disclaimer in the documentation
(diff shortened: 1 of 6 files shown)


modules/stereo/test/test_qds_matching.cpp
```diff
@@ -0,0 +1,144 @@
...
+//
+//  IMPORTANT: READ BEFORE DOWNLOADING, COPYING, INSTALLING OR USING.
+//
+//  By downloading, copying, installing or using the software you agree to this license.
+//  If you do not agree to this license, do not download, install,
+//  copy or use the software.
+//
+//
+//                        Intel License Agreement
+//                For Open Source Computer Vision Library
```
alalek, line 11: Please use short license header: https://github.com/opencv/opencv/wiki/Coding_Style_Guide#file-structure

```
```

modules/stereo/test/test_qds_matching.cpp
```diff
@@ -0,0 +1,144 @@
...
+// or tort (including negligence or otherwise) arising in any way out of
+// the use of this software, even if advised of the possibility of such damage.
+//
+//M*/
+
+#include "test_precomp.hpp"
+
+namespace opencv_test { namespace {
+
+class CV_QdsMatchingTest : public cvtest::BaseTest
```
alalek, line 46: > cvtest::BaseTest

BTW, It makes sence to avoid using of legacy OpenCV testing infrastructure.
Prefer to use GoogleTest directly.

Start from this and then replace legacy calls (like printf => `EXPECT_*()` / `ASSERT_*()`):

```.diff
-class
 ...

-void CV_QdsMatchingTest::run(int)
+TEST(qds_matching_simple_test, accuracy)
 ...

-TEST(qds_matching_simple_test, accuracy) { CV_QdsMatchingTest test; test.safe_run(); }
```

alalek (commented):
Thank you for contribution!

modules/stereo/test/test_qds_matching.cpp
dimitrisPs: Sorry for that, I used the test files from the other stereo modules as templates, when I was writing qds' test, assuming that those follow the guidelines. I will make changes later this week, thanks for your help.

alalek (approved):

PR #2875, opened 2021-02-18
Fix omnidir::undistortPoints

rgolovanov:
The relevant bug was reported in https://github.com/opencv/opencv_contrib/issues/1612

The _xi was erroneously applied at points re-projection to camera plane.
_xi parameter was already taken in use while projection of points to unit sphere.


llschloesser:
Is there any value in treating xi as a 5th distortion parameter? It would simplify the interface and allow one to store a pinhole, fisheye, or omni calibration model all in the same data structure (without exposing xi to models that don't care about it).

rgolovanov:
@llschloesser I would consider this suggestion detached from this fix. Could you create separate issue about this?

llschloesser:
@rgolovanov , thanks, yes, I will do that.

alalek (approved):

PR #2878, opened 2021-02-24
Colored Kinect Fusion

DumDereDum:


modules/rgbd/src/tsdf_functions.cpp
```diff
@@ -418,6 +418,309 @@ void integrateVolumeUnit(
...
+                    v_float32x4 projectedRGB = v_muladd(camPixVec, rgb_vfxy, rgb_vcxy);
+                    // leave only first 2 lanes
+                    projectedRGB = v_reinterpret_as_f32(v_reinterpret_as_u32(projected) &
+                        v_uint32x4(0xFFFFFFFF, 0xFFFFFFFF, 0, 0));
+
+                    // norm(camPixVec) produces double which is too slow
+                    int _u = (int)projected.get0();
+                    int _v = (int)v_rotate_right<1>(projected).get0();
+                    int rgb_u = (int)projected.get0();
+                    int rgb_v = (int)v_rotate_right<1>(projected).get0();
```
savuor, line 571: Vectorized code diverges with scalar one in what coordinate to use for RGB

modules/rgbd/src/tsdf_functions.cpp
```diff
@@ -418,6 +418,309 @@ void integrateVolumeUnit(
...
+
+                        RGBTsdfVoxel& voxel = volDataY[z * volStrides[2]];
+                        WeightType& weight = voxel.weight;
+                        TsdfType& value = voxel.tsdf;
+                        ColorType& r = voxel.r;
+                        ColorType& g = voxel.g;
+                        ColorType& b = voxel.b;
+
+                        // update RGB
+                        if (abs(((float)(r + g + b)) - (colorRGB[0] + colorRGB[1] + colorRGB[2])) < 1000 || weight < 1)
```
savuor, line 595: What's the purpose of this extra comparison and why 1000?

DumDereDum: it is the previous version of code
will be removed

savuor:
While there are things to do (speed up using OpenCL, rewrite raycasting to get better color filtering, fix demo to allow unregistered depth-rgb pairs) they have more sense to be implemented in upcoming PRs than in this one.
Here we have demo working and tests passing which is going to be enough for now.

savuor:
👍

PR #2892, opened 2021-03-14
Pose Graph rewritten without Ceres

savuor:
Pose graph from Large KinFu rewritten using manual analytic differentiation and custom LevMarq solver based on naive block sparse matrix implementation solved by Eigen library.

Connected test data PR: https://github.com/opencv/opencv_extra/pull/869

### DIFF
--- a/modules/rgbd/CMakeLists.txt
+++ b/modules/rgbd/CMakeLists.txt
@@ -1,15 +1,3 @@
 set(the_description "RGBD algorithms")
 
-find_package(Ceres QUIET)
 ocv_define_module(rgbd opencv_core opencv_calib3d opencv_imgproc OPTIONAL opencv_viz WRAP python)
-
-if(Ceres_FOUND)
-  ocv_target_compile_definitions(${the_module} PUBLIC CERES_FOUND)
-  ocv_target_link_libraries(${the_module} ${CERES_LIBRARIES})
-  if(Ceres_VERSION VERSION_LESS 2.0.0)
-    ocv_include_directories("${CERES_INCLUDE_DIRS}")
-  endif()
-  add_definitions(/DGLOG_NO_ABBREVIATED_SEVERITIES)  # avoid ERROR macro conflict in glog (ceres dependency)
-else()
-  message(STATUS "rgbd: CERES support is disabled. Ceres Solver is Required for Posegraph optimization")
-endif()
--- a/modules/rgbd/CMakeLists.txt
+++ b/modules/rgbd/CMakeLists.txt
@@ -1,3 +1,7 @@
 set(the_description "RGBD algorithms")
 
 ocv_define_module(rgbd opencv_core opencv_calib3d opencv_imgproc OPTIONAL opencv_viz WRAP python)
+
+if(NOT HAVE_EIGEN)
+  message(STATUS "rgbd: Eigen support is disabled. Eigen is Required for Posegraph optimization")
+endif()
--- a/modules/rgbd/include/opencv2/rgbd/detail/pose_graph.hpp
+++ b/modules/rgbd/include/opencv2/rgbd/detail/pose_graph.hpp
@@ -21,169 +21,47 @@ namespace kinfu
 namespace detail
 {
 
-// TODO: text about unstable API
-class PoseGraph
+// ATTENTION! This class is used internally in Large KinFu.
+// It has been pushed to publicly available headers for tests only.
+// Source compatibility of this API is not guaranteed in the future.
+
+// This class provides tools to solve so-called pose graph problem often arisen in SLAM problems
+// The pose graph format, cost function and optimization techniques
+// repeat the ones used in Ceres 3D Pose Graph Optimization:
+// http://ceres-solver.org/nnls_tutorial.html#other-examples, pose_graph_3d.cc bullet
+class CV_EXPORTS_W PoseGraph
 {
 public:
-    struct Pose3d
-    {
-        Vec3d t;
-        Quatd q;
+    CV_WRAP static Ptr<PoseGraph> create();
+    virtual ~PoseGraph() = 0;
 
-        Pose3d() : t(), q(1, 0, 0, 0) { }
+    // Node may have any id >= 0
+    virtual void addNode(size_t _nodeId, const Affine3d& _pose, bool fixed) = 0;
+    virtual bool isNodeExist(size_t nodeId) const = 0;
+    virtual bool setNodeFixed(size_t nodeId, bool fixed) = 0;
+    virtual bool isNodeFixed(size_t nodeId) = 0;
+    virtual Affine3d getNodePose(size_t nodeId) = 0;
+    virtual std::vector<size_t> getNodesIds() const = 0;
+    virtual size_t getNumNodes() const = 0;
 
-        Pose3d(const Matx33d& rotation, const Vec3d& translation)
-            : t(translation), q(Quatd::createFromRotMat(rotation).normalize())
-        { }
+    // Edges have consequent indices starting from 0
+    virtual void addEdge(size_t _sourceNodeId, size_t _targetNodeId, const Affine3f& _transformation,
+                         const Matx66f& _information = Matx66f::eye()) = 0;
+    virtual size_t getEdgeStart(size_t i) const = 0;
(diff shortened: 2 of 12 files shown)


savuor:
@alalek Is there a good way to avoid unreachable code warnings? I disabled them with pragma for cases when there's no Eigen enabled.

modules/rgbd/src/pose_graph.cpp
```diff
@@ -4,44 +4,75 @@
...
 #include <iostream>
 #include <limits>
 #include <unordered_set>
 #include <vector>
 
-#if defined(CERES_FOUND)
-#include <ceres/ceres.h>
+// a very stupid workaround against unreachable code warning
+#if defined(_MSC_VER)
+#pragma warning(disable : 4702)
```
alalek, line 15: It makes sense to put `PoseGraph::optimize()` under `#ifdef HAVE_EIGEN` condition if Eigen is mandatory here.

BTW, Accurate condition should have `&& !defined(HAVE_EIGEN)` too.

savuor:
👍

modules/rgbd/src/pose_graph.cpp
```diff
@@ -81,89 +105,703 @@ bool PoseGraph::isValid() const
...
+const Matx44d M_Im{ 0, 0, 0, 0,
+                    0, 1, 0, 0,
+                    0, 0, 1, 0,
+                    0, 0, 0, 1 };
+
+// matrix form of conjugation
+const Matx44d M_Conj{ 1,  0,  0,  0,
+                      0, -1,  0,  0,
+                      0,  0, -1,  0,
+                      0,  0,  0, -1 };
```
alalek, line 235: > `M_Im`

used?

It make sense to move out any vars from the global scope.

modules/rgbd/src/pose_graph.cpp
```diff
@@ -4,44 +4,70 @@
 
 #include "pose_graph.hpp"
 
+#include <fstream>
 #include <iostream>
```
alalek, line 8: `iostream` should not be used in OpenCV modules.

> `std::cout`

Use `CV_LOG_*` macro instead.

modules/rgbd/CMakeLists.txt
```diff
@@ -1,15 +1,7 @@
 set(the_description "RGBD algorithms")
 
-find_package(Ceres QUIET)
```
alalek, line 3: Does it make sense to keep both implementations (to compare results) for some timeframe (e.g., until we have robust tests)?

modules/rgbd/src/pose_graph.cpp
```diff
@@ -81,89 +105,703 @@ bool PoseGraph::isValid() const
...
 
-#if defined(CERES_FOUND) && defined(HAVE_EIGEN)
-void Optimizer::createOptimizationProblem(PoseGraph& poseGraph, ceres::Problem& problem)
+PoseGraph::PoseGraph(const std::string& g2oFileName) :
+    nodes(), edges()
 {
-    int numEdges = poseGraph.getNumEdges();
-    int numNodes = poseGraph.getNumNodes();
-    if (numEdges == 0)
+    readG2OFile(g2oFileName);
```
alalek, line 111: Do we really need to work with files?

alalek (commented):
Which tests run this code?

savuor:
@alalek,
On tests, keeping Ceres backend and reading files:

1. The initial Ceres-based pose graph code wasn't covered by tests, the only way to check it works was to run demo app long enough until there are enough submaps generated to build pose graph from them.
2. The purpose of this PR was nothing more than to get rid of Ceres dependency in pose graph for further integration into 3d module.
3. Indeed, we need regression tests for pose graph, e.g. using [this dataset](https://lucacarlone.mit.edu/datasets/) I used to compare Ceres-based pose graph optimization to this new one. The function `readG2OFile` was used exactly for that purpose.
4. But (correct me if I'm wrong) making such tests requires to expose pose graph into public API. I'm not sure that we need such functionality exposed (at least in its current form, not universal enough for other SLAM pipelines), especially considering that the compatibility should be supported in the future.

Can you tell then is there a possibility to make such tests anyway?

savuor:
@alalek 
I added test for pose graph + testing data in opencv/opencv_extra#869, so we don't have to keep old Ceres implementation along with the current one.
Everything but CN builders works fine. Looks like sometimes CN builders fail to fetch test data from the corresponding branch from opencv_extra.
How this can be solved?

modules/rgbd/test/test_pose_graph.cpp
```diff
@@ -0,0 +1,159 @@
...
+    // Initialization Techniques for 3D SLAM : a Survey on Rotation Estimation and its Use in Pose Graph Optimization.
+    // In IEEE Intl.Conf.on Robotics and Automation(ICRA), pages 4597 - 4604, 2015.
+
+    std::string filename = cvtest::TS::ptr()->get_data_path() + "rgbd/sphere_bignoise_vertex3.g2o";
+    Ptr<kinfu::detail::PoseGraph> pg = readG2OFile(filename);
+
+    if (verbose)
+    {
+        cv::utils::logging::setLogLevel(cv::utils::logging::LogLevel::LOG_LEVEL_INFO);
+    }
```
alalek, line 115: Please do not touch logging level in tests.
It may be configured externally, e.g. `OPENCV_LOG_LEVEL=INFO` environment (so even `verbose` parameter is not needed and should be removed)

modules/rgbd/test/test_pose_graph.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+    return pg;
+}
+
+// Turn it on if you want to see resulting pose graph nodes
+static const bool writeToObjFile = false;
+// Turn if off if you don't need log messages
+static const bool verbose = true;
+
+#if !defined(_DEBUG) && defined(HAVE_EIGEN)
```
alalek, line 100: Use this for test filtering:
- Debug
```
applyTestTag(
    CV_TEST_TAG_LONG,  // 15sec in release
    CV_TEST_TAG_DEBUG_VERYLONG  // 400 sec in debug
)
```
- Eigen
```
#ifdef HAVE_EIGEN
    ...
#else
    throw SkipTestException("Build with Eigen is required");
#endif
```

modules/rgbd/test/test_pose_graph.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+    ASSERT_GE(iters, 0);
+    ASSERT_LE(iters, 36); // should converge in 36 iterations
+
+    double energy = pg->calcEnergy();
+
+    ASSERT_LE(energy, 1.47723e+06); // should converge to 1.47722e+06 or less
+
+    if (writeToObjFile)
+    {
```
alalek, line 127: Use `if (cvtest::debugLevel > 0)` instead (to emit extra files).

See https://github.com/opencv/opencv/pull/18955 for details.

modules/rgbd/include/opencv2/rgbd/detail/pose_graph.hpp
```diff
@@ -0,0 +1,58 @@
+#ifndef OPENCV_RGBD_GRAPH_NODE_H
```
alalek, line 1: License header.

modules/rgbd/test/test_pose_graph.cpp
```diff
@@ -0,0 +1,154 @@
...
+    // The dataset was taken from here: https://lucacarlone.mit.edu/datasets/
+    // Connected paper:
+    // L.Carlone, R.Tron, K.Daniilidis, and F.Dellaert.
+    // Initialization Techniques for 3D SLAM : a Survey on Rotation Estimation and its Use in Pose Graph Optimization.
+    // In IEEE Intl.Conf.on Robotics and Automation(ICRA), pages 4597 - 4604, 2015.
+
+    std::string filename = cvtest::TS::ptr()->get_data_path() + "rgbd/sphere_bignoise_vertex3.g2o";
+    Ptr<kinfu::detail::PoseGraph> pg = readG2OFile(filename);
+
+    // You may change logging level to view detailed optimization report
```
alalek, line 111: Move `#ifdef HAVE_EIGEN` here to remove `(void)(&readG2OFile);`

alalek (approved):
Looks good to me :+1:

PR #2894, opened 2021-03-17
ximgproc: add an assert in thinning()

### LINKED ISSUE #2887: Sizes of input arguments do not match exception in ximgproc::thinning
I'm using OpenCV v4.5.1-dev and v4.2 on Ubuntu 18.04 and 20.04 with gcc-10.1 and clang-10.0. 

This code snippet:
```
  Mat img = cv::imread("6.png");
  ximgproc::thinning(img, img, ximgproc::THINNING_ZHANGSUEN);
```
crashes with this message:
```
terminate called after throwing an instance of 'cv::Exception'
  what():  OpenCV(4.5.1-dev) /home/paul/opencv/opencv/modules/core/src/arithm.cpp:232: error: (-209:Sizes of input arguments do not match) The operation is neither 'array op array' (where arrays have the same size and type), nor 'array op scalar', nor 'scalar op array' in function 'binary_op'
```
Specifically `Sizes of input arguments do not match` doesn't make any sense to me in this context. Both input arguments are identical and in-place thinning is supported according to documentation at https://docs.opencv.org/master/df/d2d/group__ximgproc.html. Behavior is the same with `THINNING_GUOHALL` argument.

Is this a problem of wrong image type? Running `print("{} {}\n", img.depth(), img.type());` produces `0 16`. I can't find image type `16` in the documentation.

 Attached is `6.png` file.

berak:

resolves #2887

### DIFF
--- a/modules/ximgproc/src/thinning.cpp
+++ b/modules/ximgproc/src/thinning.cpp
@@ -69,6 +69,7 @@ static void thinningIteration(Mat img, int iter, int thinningType){
 // Apply the thinning procedure to a given image
 void thinning(InputArray input, OutputArray output, int thinningType){
     Mat processed = input.getMat().clone();
+    CV_CheckTypeEQ(processed.type(), CV_8UC1, "");
     // Enforce the range of the input image to be in between 0 - 255
     processed /= 255;
 


modules/ximgproc/src/thinning.cpp
```diff
@@ -69,6 +69,7 @@ static void thinningIteration(Mat img, int iter, int thinningType){
 // Apply the thinning procedure to a given image
 void thinning(InputArray input, OutputArray output, int thinningType){
     Mat processed = input.getMat().clone();
+    CV_Assert(processed.type == CV_8UC1);
```
alalek, line 72: Consider using `CV_CheckTypeEQ(input.type(), CV_8UC1, "");` for clear diagnostic messages.

alalek (approved):
Thank you!

Issue #2811, opened 2021-01-12
Incorrect ARUCO detection with bigger markers


Chris-Bee:
##### System information (version)
- OpenCV 4.2.* and 4.5.1
- Operating System: Linux Ubuntu 64 Bit

- Tag family is: DICT_7X7_250
- Image resolution: 2056x2542
- Images are grayscale

##### Detailed description

This issue concerns the implementation of the ARUCO marker detection. I observed, that the corner-detection / detection of squares in the image shows incorrect results. This issue is especially visible with bigger tag sizes (e.g. 0.3-0.5 meters). It appears that the corners or the squares are detected 2/3 into the outer border of the marker which results in incorrect height information. This issue does not occur in all conditions and seems especially related to the orientation of the tags. If the sides of the tag are aligned with the image frame (which basically means that the yaw rotation of the tag is a multiple of 90°) then this issue does not occur (compare attached image zero_yaw_rot.png). However, if the orientation is ~30-60°, then the size of the tag is incorrect (see image non_zero_yaw_rot.png below).

The sizes of the tags in image are [id, size in meter]:
"97: 0.15, 98: 0.15, 99: 0.3, 100: 0.1, 101: 0.15, 102: 0.5, 103: 0.15, 104: 0.1, 105: 0.3, 106: 0.15, 107: 0.15"

I verified that this is not only a problem in visualization, the height information does actually change if the issue occurs. A quick workaround was to transform the grayscale images into black and white images which results in correct detections.

I'm using a ROS package (https://github.com/UbiquityRobotics/fiducials) which implements the aruco detection with OpenCV. However, the attached python script will produce similar results. 

[tag_issue_images.zip](https://github.com/opencv/opencv/files/5803022/tag_issue.zip)
[tag_detect_issue.py.zip](https://github.com/opencv/opencv/files/5803141/tag_detect_issue.py.zip)

**ID 102, the middle tag is detected incorrect:**
![non_zero_yaw_rot](https://user-images.githubusercontent.com/12284463/104329606-609fcd80-54ed-11eb-91bd-0658a1ebec92.png)
**ID105, the tag to the right, is detected incorrect:**
![zero_yaw_rot](https://user-images.githubusercontent.com/12284463/104337290-9c3e9580-54f5-11eb-9e88-3d1c2e39816d.png)

paroj:
#2900 maybe?

Chris-Bee:
[...]

PR #2900, opened 2021-03-24
Fix aruco _filterTooCloseCandidates

spirraw:
_filterTooCloseCandidates was occasionally not choosing the biggest
contour of a candidate group.

Firstly, if the biggest contour of a candidate group happened to be
the first candidate of that group, it would never be chosen due to the
comparison loop skipping it.

Secondly, contour sizes were being compared by their circumference
(via counting the number of points in a contour) rather than area.
In some cases it is possible for a smaller candidate to have more
points than a bigger one. This was fixed by comparing the contour
area as defined by the contour's corners.

<cut/>

```
force_builders=linux,docs
```

### DIFF
--- a/modules/aruco/src/aruco.cpp
+++ b/modules/aruco/src/aruco.cpp
@@ -306,34 +306,36 @@ static void _filterTooCloseCandidates(const vector< vector< Point2f > > &candida
     vector< vector< Point > > smallerContours;
 
     // save possible candidates
-    for( unsigned int i = 0; i < groupedCandidates.size(); i++ ) {
-        int smallerIdx = groupedCandidates[i][0];
-        int biggerIdx = -1;
+    for(unsigned int i = 0; i < groupedCandidates.size(); i++) {
+        unsigned int smallerIdx = groupedCandidates[i][0];
+        unsigned int biggerIdx = smallerIdx;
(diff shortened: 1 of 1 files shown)


spirraw:
Here is an image with which the issue should be reproducible.
raw
Here is a visualization of the contours that are passed to _filterTooCloseCandidates:
candidates
They are grouped correctly, however, because the inner candidate has more points than the outer one, it is incorrectly chosen as the biggest contour in the group.

Chris-Bee:
This might be related to the incorrect settings of the threshold windows described here #2811 . Using the settings described there might solver your issue.

spirraw:
Prior to making this patch, I was also able to work around the problem by adjusting the adaptive thresholding parameters. Doing so eliminates the inner contours entirely, and as a result the bugs in _filterTooCloseCandidates are mitigated. While it is possible to to work around specific cases by adjusting the adaptive thresholding parameters, doing so may adversely affect performance in other cases. This patch is about fixing two underlying issues in _filterTooCloseCandidates, the function's entire purpose is to filter out these "inner candidates".

szk1509:
hi @spirraw  :)
it all looks pretty reasonable, I haven't had the time to do the testing I normally do, but in broad strokes a big :+1:

alalek (approved):
Thank you :+1:

Issue #2314, opened 2019-10-26
EDGE DRAWING LIBRARY


sturkmen72:
recently i found https://github.com/CihanTopal/ED_Lib and i see it contains useful algorithms valuable to be included in OpenCV

firstly i contacted with the author and take his permission.
i want to work on code adaptation and rewriting according to OpenCV code style

sturkmen72:
Algorithms in the original repo and their estimated completion states

- [x] ED (Edge Drawing) ...%100 - Tested Well - Same output with the original
- [x] EDLines .........................%100 - Tested Well - Same output with the original
- [ ] EDPF ..............................%90 - Tests in progress
- [ ] EDCircles .....................%95 - Tests in progress
- [ ] EDColor .......................%0

geroldmeisinger:
@sturkmen72 Thank you so much for your integration of the Edge Drawing algorithms in OpenCV and subsequently for opencv-contrib-python!

I'm using [Stable Diffusion](https://stability.ai/stable-diffusion), a generative artificial intelligence software, which can turn a simple text prompt into a full blown image. Together with the [ControlNet](https://github.com/lllyasviel/ControlNet) extension it is possible to "guide" the generation process based on edge maps into a specific direction. At the moment I'm training [a new ControlNet based on Edge Drawing](https://github.com/lllyasviel/ControlNet/discussions/318#discussioncomment-7058540) and the parameter free version already helped me a lot to increase quality of the guidance process. The first version is available on [Civitai - ControlNet MySee - Edge Drawing Parameter Free](https://civitai.com/models/149740) and [a web-demo for Edge Drawing Parameter Free](https://huggingface.co/spaces/GeroldMeisinger/edpf), thanks to your integration. This provides a lot of value to users because it doesn't require parameter tuning like canny.

This is what it looks like to most users using the Automatic1111 UI for Stable Diffusion. Left is the "ground truth" image, middle is the processed "edge map" and right is the new "generated image" based on the prompt "photo of a woman in the street, fashine, anime" and guided by the edge map to get a similar pose:
![image](https://github.com/opencv/opencv_contrib/assets/112266044/0bf35051-9c55-4d66-a0ea-ddb3676735d3)

[...]

PR #2907, opened 2021-04-02
Edge Drawing Improvements

sturkmen72:
#2314

[documentation preview](https://pullrequest.opencv.org/buildbot/export/pr_contrib/2907/docs/d4/d8b/group__ximgproc__edge__drawing.html)

```
force_builders=Linux x64,docs
```

### DIFF
--- a/modules/ximgproc/src/edge_drawing.cpp
+++ b/modules/ximgproc/src/edge_drawing.cpp
@@ -12,6 +12,86 @@ namespace cv
 namespace ximgproc
 {
 
+struct ComputeGradientBody : ParallelLoopBody
+{
+void operator() (const Range& range) const CV_OVERRIDE;
+
+Mat_<uchar> src;
+mutable Mat_<ushort> gradImage;
+mutable Mat_<uchar> dirImage;
+int gradThresh;
+int op;
+bool SumFlag;
+int* grads;
+bool PFmode;
+};
+
+void ComputeGradientBody::operator() (const Range& range) const
+{
+    const int last_col = src.cols - 1;
+    int gx = 0;
+    int gy = 0;
+    int sum;
+
+    for (int y = range.start; y < range.end; ++y)
+    {
+        const uchar* srcPrevRow = src[y - 1];
+        const uchar* srcCurRow = src[y];
+        const uchar* srcNextRow = src[y + 1];
+
+        ushort* gradRow = gradImage[y];
+        uchar* dirRow = dirImage[y];
+
+        for (int x = 1; x < last_col; ++x)
+        {
+            int com1 = srcNextRow[x + 1] - srcPrevRow[x - 1];
+            int com2 = srcPrevRow[x + 1] - srcNextRow[x - 1];
+
+            switch (op)
+            {
+            case EdgeDrawing::PREWITT:
+                gx = abs(com1 + com2 + srcCurRow[x + 1] - srcCurRow[x - 1]);
+                gy = abs(com1 - com2 + srcNextRow[x] - srcPrevRow[x]);
+                break;
+            case EdgeDrawing::SOBEL:
+                gx = abs(com1 + com2 + 2 * (srcCurRow[x + 1] - srcCurRow[x - 1]));
+                gy = abs(com1 - com2 + 2 * (srcNextRow[x] - srcPrevRow[x]));
+                break;
+            case EdgeDrawing::SCHARR:
+                gx = abs(3 * (com1 + com2) + 10 * (srcCurRow[x + 1] - srcCurRow[x - 1]));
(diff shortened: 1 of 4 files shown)


asenyaev:

sturkmen72:
@alalek please review

modules/ximgproc/src/edge_drawing.cpp
```diff
@@ -251,6 +332,9 @@ void EdgeDrawingImpl::write(cv::FileStorage& fs) const
 EdgeDrawingImpl::EdgeDrawingImpl()
 {
     params = EdgeDrawing::Params();
+    nfa = new NFALUT(1, 1/2, 1, 1);
+    dH = new double[MAX_GRAD_VALUE];
+    grads = new int[MAX_GRAD_VALUE];
```
alalek, line 337: > new

Where are these objects destroyed?

---

> dH, grads

Consider using `std::vector` for these fields.

modules/ximgproc/src/edge_drawing.cpp
```diff
@@ -337,6 +338,13 @@ EdgeDrawingImpl::EdgeDrawingImpl()
     grads = new int[MAX_GRAD_VALUE];
 }
 
+EdgeDrawingImpl::~EdgeDrawingImpl()
```
sturkmen72, line 341: is it enough for now ? i am planning to work on  optimizing all code later.

modules/ximgproc/perf/perf_edge_drawing.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+/* 1. Define parameter type and test fixture */
+typedef tuple<int, int> EDTestParam;
+typedef TestBaseWithParam<EDTestParam> EdgeDrawing_Detect_Edges;
+typedef TestBaseWithParam<EDTestParam> EdgeDrawing_Detect_Lines;
+typedef TestBaseWithParam<EDTestParam> EdgeDrawing_Detect_Ellipses;
+
+/* 2. Declare the testsuite */
+PERF_TEST_P(EdgeDrawing_Detect_Edges, perf,
+            Combine(Values(0, 1, 2, 3), Values(10, 20, 30, 50, 100, 150, 200, 250)))
```
alalek, line 19: Do we really need to track performance of all these cases? Could we use subset of them here (and below)?

You can create separate full set (e.g, for "development"), but lets keep it disabled by default (see `DISABLED_FULL, OCL_GaussianBlurFixture` from `imgproc`)

modules/ximgproc/perf/perf_edge_drawing.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+    /* 3. Get actual test parameters */
+    EDTestParam params = GetParam();
+    int EdgeDetectionOperator = get<0>(params);
+    int GradientThresholdValue = get<1>(params);
+
+    /* 4. Allocate and initialize arguments for tested function */
+    std::string filename = getDataPath("perf/1680x1050.png");
+    Mat src = imread(filename, 0);
```
alalek, line 28: > 0

Please avoid using of magic numbers. Use `IMREAD_GRAYSCALE` instead.

modules/ximgproc/src/edge_drawing.cpp
```diff
@@ -12,6 +12,87 @@ namespace cv
...
+        uchar* dirRow = dirImage[y];
+
+        for (int x = 1; x < last_col; ++x)
+        {
+            int com1 = srcNextRow[x + 1] - srcPrevRow[x - 1];
+            int com2 = srcPrevRow[x + 1] - srcNextRow[x - 1];
+            int gx = 0;
+            int gy = 0;
+
+            switch (op)
```
alalek, line 49: BTW, "constant" mode switch in inner body of two loops may cause some performance loss.

(in the current patch it is just code move, so we can keep it "as is" in this PR)

modules/ximgproc/src/edge_drawing.cpp
```diff
@@ -12,6 +12,87 @@ namespace cv
...
+                com2 = srcCurRow[x + 1] - srcNextRow[x];
+
+                gx = abs(com1 + com2);
+                gy = abs(com1 - com2);
+                break;
+            }
+
+            int sum;
+
+            if (SumFlag)
```
alalek, line 75: > if

similar to "constant" mode switch.

alalek (commented):
Thank you for updates!

Could you please reduce amount of default performance test cases?

modules/ximgproc/perf/perf_edge_drawing.cpp
sturkmen72: i will provide performans test later

alalek (approved):
Well done :+1:

PR #2923, opened 2021-04-14
sfm: link with Glog_LIBS

shr-project:
* in 4.5.0 there was explicit linkage with GLOG_LIBRARY, but since 4.5.1 with:
  https://github.com/opencv/opencv_contrib/commit/23ee62a19b7a3e50d6dbf295359d8b1aff2e03fd

  it's gone, probably because Glog_FOUND is already set from Ceres,
  but then GLOG_LIBRARIES is empty in LIBMV_LIGHT_LIBS and build with gold fails:

FAILED: bin/example_tutorial_perspective_correction
[...]

PR #2924, opened 2021-04-14
cmake: completelly disable cnn_3dobj

alalek:
- used FindGlog.cmake has buggy implementation (doesn't follow CMake policies / variable names)

relates #2923


alalek:
:+1:

PR #2931, opened 2021-04-23
Link wechat_qrcode with libiconv on MinGW

### LINKED ISSUE #2862: errors on building opencv with MinGW on windows without VisualStudio

- OpenCV => 4.2
- Operating System / Platform => Windows 10 64 Bit
- Compiler => MinGW64

##### Detailed description

 I am trying build latest version openCV library via MinGW on Window10 64 bit machine without Microsoft Visual Studio. I know that is a little bit tricky method but I don't like use Microsoft build tools.  I handled the errors I have encountered so far except latest one.

>In addition, I checked out [Microsoft VS Docs](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/isnan-isnan-isnanf?view=msvc-160). That line confused me : " **The _isnan and _isnanf functions are Microsoft-specific.** ".  I'm not sure if the method I use is wrong for my operating system. I also checked out openCV Docs and FAQ, I didn't find any solution for my problem. 

The error I get is below :

[ 45%] Building CXX object modules/wechat_qrcode/CMakeFiles/opencv_wechat_qrcode.dir/src/binarizermgr.cpp.obj
In file included from D:\opencvInstall\test3\opencv_contrib\modules\wechat_qrcode\src\zxing/common/bitarray.hpp:15,
                 from D:\opencvInstall\test3\opencv_contrib\modules\wechat_qrcode\src\zxing/common/bytematrix.hpp:13,
                 from D:\opencvInstall\test3\opencv_contrib\modules\wechat_qrcode\src\imgsource.hpp:13,
                 from D:\opencvInstall\test3\opencv_contrib\modules\wechat_qrcode\src\precomp.hpp:25,
                 from D:\opencvInstall\test3\opencv_contrib\modules\wechat_qrcode\src\binarizermgr.cpp:7:
D:/opencvInstall/test3/opencv_contrib/modules/wechat_qrcode/src/zxing/zxing.hpp: In function 'bool zxing::isnan(float)':
D:/opencvInstall/test3/opencv_contrib/modules/wechat_qrcode/src/zxing/zxing.hpp:64:37: error: '_isnan' was not declared in this scope; did you mean 'isnan'?
   64 | inline bool isnan(float v) { return _isnan(v) != 0; }
   ......|....................................................................^~~~~~
   ......|....................................................................isnan
D:/opencvInstall/test3/opencv_contrib/modules/wechat_qrcode/src/zxing/zxing.hpp: In function 'bool zxing::isnan(double)':
D:/opencvInstall/test3/opencv_contrib/modules/wechat_qrcode/src/zxing/zxing.hpp:65:38: error: '_isnan' was not declared in this scope; did you mean 'isnan'?
   65 | inline bool isnan(double v) { return _isnan(v) != 0; }
   ......|....................................................................^~~~~~
   ......|....................................................................isnan
mingw32-make[2]: *** [modules\wechat_qrcode\CMakeFiles\opencv_wechat_qrcode.dir\build.make:82: modules/wechat_qrcode/CMakeFiles/opencv_wechat_qrcode.dir/src/binarizermgr.cpp.obj] Error 1
mingw32-make[1]: *** [CMakeFiles\Makefile2:7468: modules/wechat_qrcode/CMakeFiles/opencv_wechat_qrcode.dir/all] Error 2

######  That is my first issue report on github. If there is any mistake in format so sorry.

ferdnyc:
This PR is an alternate approach that fixes #2862, and partially reverts #2916 (with original submitter @berak's approval).

Instead of disabling `wechat_qrcode`'s use of `iconv` in MinGW environments, the module simply needs to be linked with the external library `libiconv`, which is packaged and available as part of the MSYS2 collection.

(On POSIX systems `iconv` is built into libc, but on others it's provided by an external library.)

Fortunately, since CMake 3.11 the standard module `FindIconv.cmake` has been included in the distribution that can handle linking with libiconv if necessary on the target system.

The use of `FindIconv.cmake` is safe on POSIX systems; on my Fedora box it simply noted that iconv is built in to the C library and fell through harmlessly:
```text
-- Performing Test Iconv_IS_BUILT_IN
-- Performing Test Iconv_IS_BUILT_IN - Success
-- wechat_qrcode: Download: detect.caffemodel
-- wechat_qrcode: Download: detect.prototxt
-- wechat_qrcode: Download: sr.caffemodel
-- wechat_qrcode: Download: sr.prototxt
```

#### N/A:

### DIFF
--- a/modules/wechat_qrcode/CMakeLists.txt
+++ b/modules/wechat_qrcode/CMakeLists.txt
@@ -1,6 +1,14 @@
 set(the_description "WeChat QR code Detector")
 ocv_define_module(wechat_qrcode opencv_core opencv_imgproc opencv_dnn WRAP java objc python js)
 
+# iconv support isn't automatic on some systems
+if(CMAKE_VERSION VERSION_GREATER "3.11")
+  find_package(Iconv QUIET)
+  if(Iconv_FOUND)
+    ocv_target_link_libraries(${the_module} Iconv::Iconv)
+  endif()
+endif()
+
 # need to change
 set(wechat_qrcode_commit_hash "a8b69ccc738421293254aec5ddb38bd523503252")
 set(hash_detect_caffemodel "238e2b2d6f3c18d6c3a30de0c31e23cf")
--- a/modules/wechat_qrcode/CMakeLists.txt
+++ b/modules/wechat_qrcode/CMakeLists.txt
@@ -6,6 +6,8 @@ if(CMAKE_VERSION VERSION_GREATER "3.11")
   find_package(Iconv QUIET)
   if(Iconv_FOUND)
     ocv_target_link_libraries(${the_module} Iconv::Iconv)
+  else()
+    ocv_target_compile_definitions(${the_module} PRIVATE "NO_ICONV=1")
   endif()
 endif()
 
--- a/modules/wechat_qrcode/src/zxing/zxing.hpp
+++ b/modules/wechat_qrcode/src/zxing/zxing.hpp
@@ -27,7 +27,7 @@
 #define USE_ONED_WRITER 1
 #endif
 
-#if defined(__ANDROID_API__) || defined(_MSC_VER) || defined(__MINGW32__) || defined(__MINGW64__)
+#if defined(__ANDROID_API__) || defined(_MSC_VER)
 
 #ifndef NO_ICONV
 #define NO_ICONV


berak:
unfortunately, it does not work "out-of-the-box" here ;(
(reversing the  ifdef, and adding above to cmakelists.txt, rerun cmake) :

no iconv

(from cmake-gui, configure)

i do have libiconv.a in C:\prog\mingw-w64\i686-8.1.0-posix-dwarf-rt_v6-rev0\mingw32\i686-w64-mingw32

cmake is cmake-3.14.6-win32-x86

does it look for an "external package" ? any idea, how to "debug" it ?
and honestly, no idea about msys, does your proposal *only* work with the msys distribution (instead of plain mingw-w64) ?

ferdnyc:
@berak 

libiconv is [`mingw-w64-libiconv`](https://packages.msys2.org/base/mingw-w64-libiconv), packaged from https://packages.msys2.org/base/mingw-w64-libiconv and installed before building — binary packages are `mingw-w64-{i686,x86_64,ucrt,clang}-libiconv` which I guess would become a new dependency. It's not present in vanilla MinGW out of the box, no.

I suppose the question is, what to do if iconv is not available? We could just disable iconv in the module, as you'd previously been doing... since I guess it _can_ be built without it (though what effect that has on functionality isn't clear to me)... or, we could disable the entire module in the build, as not being usable due to the missing dependency.

ferdnyc:
Urg, I meant packaged from https://www.gnu.org/software/libiconv/ — sorry, wrong URL.

berak:
>  I guess would become a new dependency.

i don't like that idea (esp. given, that it works fine without)

>  It's not present in vanilla MinGW out of the box, no.

hmm, both 7.2.0 and 8.1.0 versions have it here, 
the cmake script just did not find them (different location / package, i guess)
(kinda) working approach now:
* run cmake configure (won't find the files)
* specify them manually, run again.
* cmake generate

but this is painful. we cannot expect noobs to dig 3 levels deep into their compiler installation to find those things
(i mean like : `C:/Program Files/mingw-w64/x86_64-8.1.0-posix-seh-rt_v6-rev0/mingw64/x86_64-w64-mingw32/lib/libiconv.a` - seriously ?)
(cmake also made a blunder with `\\` seperators in the 1st attempt)

ferdnyc:
Yeah, I agree the need to dig iconv out from that path is a bridge too far, when installed via the MSYS2 package `libiconv.dll` lives in `C:\msys64\mingw64\bin\libiconv.dll` and its import lib is `C:\msys64\mingw64\lib\libiconv.dll.a`, same as any other packaged DLL. (`libiconv.a` is also installed in `...\lib\`.) I dunno _what_ that file you're unearthing is, or where it comes from. :rofl: 

> (cmake also made a blunder with `\\` seperators in the 1st attempt, another pitfall)

That's interesting, actually — I wonder if that means it's finding that (weirdly buried) library using `pkg-config`? I never thought to check whether there was a `.pc` file for it, but that might explain the `\\` paths.

...Huh, there **is** a `C:\msys64\mingw64\lib\pkgconfig\iconv.pc` installed with the MSYS2 package! But looking over `FindIconv.cmake`... it never tries `pkg-config`, So, that's not it.

I thought maybe there was some implicit automatic fallback to `pkg-config`, for CMake Find modules... but if there is, it's undocumented.

Well, all of this is sort of tangential. As you say, obviously we can't be expecting users to direct CMake to `C:/Program Files/mingw-w64/x86_64-8.1.0-posix-seh-rt_v6-rev0/mingw64/x86_64-w64-mingw32/lib/libiconv.a` just to build the `wechat_qrcode` module. And since you say it works fine without iconv (I've never tried, my sole goal was to get it to _build_ so that it was no longer blocking me from submitting the 4.5.2 upgrade to MSYS) then I suppose the best thing to do is to just add a `target_compile_definition` to the `CMakeLists.txt` that sets `NO_ICONV_INSIDE`, if iconv support can't be discovered during configuration.

I'll update the PR with that change.

ferdnyc:
(Sorry, forgot the @berak there.)

ferdnyc:
That should do it. Though, turns out I can't easily _test_ that change... the MSYS2 `cmake` itself is linked with libiconv, and it breaks if I uninstall the DLL. :laughing:

berak:
> That should do it. 

indeed, it does, all fine here :)

@deadprogram , maybe you want to check, too ?

alalek (approved):
@ferdnyc Thank you for contribution :+1:

@berak Thank you for reviewing!

sergiomb2:
Hello ,  to add this to opencv Fedora package, I need clarification on license of content on https://github.com/WeChatCV/opencv_3rdparty . These models are patent free ? and what license they have ? 

Thank you

ferdnyc:
@sergiomb2 Given that there's no license statement at that repo, I don't think that's a question anyone but @dddzg (the author of nearly all commits related to WeChat + OpenCV, and uploader of the model files) can answer.

The checked-in **code** is all licensed under Apache 2.0, so maybe the intent is that the models would be licensed the same way... but no way that's explicit enough to satisfy legal. Without a license file _in_ that repo, it's pretty much in limbo. Even a clarification statement here would be unlikely to fly.

I suppose someone could submit a PR to [WeChatCV/opencv_3rdparty](https://github.com/WeChatCV/opencv_3rdparty), adding an Apache 2.0 license file to the repo, and see if it gets merged. That would presumably count as clarification/agreement.

Otherwise, perhaps the package can contain a download script the user can use to retrieve the models themselves? Has legal ever ruled on whether they'd be OK with that?

sergiomb2:
what user download in his computer, is not a concern . 
The concern is distribute one content that might not be free (as freedom) on Fedora and ReahHat products .
I liked your ideia and will propose join the License File 
Thank you

alalek:
Licensing discussion should be in initial patch: #2821

Whole repo license for opencv_3rdparty is not possible due to different content (repo is a collection of 3rdparty dependencies or files or models).
License is applicable on "branches" only.

alalek:
> and not a single one included a license file, OR any license statement in the README.

It depends: https://github.com/opencv/opencv_3rdparty/tree/ffmpeg/4.x

---

Anyway, we should go to #2821 or create a new issue and ask @dddzg (contribution author) for clarification with model licenses.

PR #2933, opened 2021-04-27
KinFu impl without ICP

DumDereDum:

### DIFF
--- a/modules/rgbd/samples/kinfu_demo.cpp
+++ b/modules/rgbd/samples/kinfu_demo.cpp
@@ -175,16 +175,16 @@ int main(int argc, char **argv)
     int64 prevTime = getTickCount();
 
     if (haveQS)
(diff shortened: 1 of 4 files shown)


modules/rgbd/samples/kinfu_demo.cpp
```diff
@@ -208,7 +225,12 @@ int main(int argc, char **argv)
             {
                 imshow("depth", cvt8);
 
-                if(!kf->update(frame))
+                bool isCorr;
+                if (haveQS)
+                    isCorr = !kf->update(frame, pose);
+                else
+                    isCorr = !kf->update(frame, pose);
```
alvoron, line 232: both branches look identical - is it copy-paste issue?

DumDereDum: Fixed
I didn't notice, thx

DumDereDum:
It will be rewritten in future

Issue #2129, opened 2019-05-30
I think it is a bug, happened when using `FacemarkLBF` in the `face` module. Please check it. 


seanwen86:
From my point of view, I think it's a bug. :(. So I would like to report to you guys for checking that.

##### Error message
terminate called after throwing an instance of 'cv::Exception'
  what():  OpenCV(3.4.6) .../opencv346/opencv_contrib-3.4.6/modules/face/src/facemarkLBF.cpp:332: error: (-5:Bad argument) Training data is not provided. Consider to add using addTrainingSample() function! in function 'training'

After checking the source code, find it means NO FACE is detected at all.

##### System information (version)
- OpenCV => 3.4.6, **but find it also exists in *master* branch**
- Operating System / Platform =>  Ubuntu desktop 64bit
- Compiler => g++

I was trying to use the `FacemarkLBF` method in the `face` module to train a facial landmark detection model. I used my own face detector `MyFaceDetection`.
```cpp
   cv::Ptr<cv::face::FacemarkLBF> facemark = cv::face::FacemarkLBF::create(params);
   facemark->setFaceDetector((cv::face::FN_FaceDetector)MyFaceDetection, nullptr);
   //... some code
   facemark->training();
```
My own face detector NEEDS a **RGB** image as the input, but the following code in file `facemarkLBF.cpp` located in `face` module converts **RGB into GRAY** image before calling my face detector which actually happens in `getBBox(img, _shape);`. In my opinion, the conversion operation should be delayed to the actual face detector to decide, just like the `FacemarkLBFImpl::defaultFaceDetector` does. It's really redundant and a bug to leave it here from point of view.

So please check it whether I am right, guys.

The detailed snippet in the file `facemarkLBF.cpp` is as follows, and the three lines should be line 505-507:
```.cpp
void FacemarkLBFImpl::prepareTrainingData(Mat img, std::vector<Point2f> facePoints,
    std::vector<Mat> & cropped, std::vector<Mat> & shapes, std::vector<BBox> &boxes)
{
    if(img.channels()>1){
        cvtColor(img,img,COLOR_BGR2GRAY);
    }

    Mat shape;
    Mat _shape = Mat(facePoints).reshape(1);
    Rect box = getBBox(img, _shape);
 ```

PR #2953, opened 2021-05-15
Do conversion to grayscale after face detection.

iclemens:
#2129

The proposed change would convert the training image to grayscale AFTER running the face detector. The default face detector already converts the images to grayscale and a custom detector should be able to decide if it want to use grayscale/color images.

I've kept the conversion itself as I'm not sure if the actual training requires grayscale images.

### DIFF
--- a/modules/face/src/facemarkLBF.cpp
+++ b/modules/face/src/facemarkLBF.cpp
@@ -535,13 +535,14 @@ Rect FacemarkLBFImpl::getBBox(Mat &img, const Mat_<double> shape) {
 void FacemarkLBFImpl::prepareTrainingData(Mat img, std::vector<Point2f> facePoints,
     std::vector<Mat> & cropped, std::vector<Mat> & shapes, std::vector<BBox> &boxes)
 {
+    Mat shape;
+    Mat _shape = Mat(facePoints).reshape(1);
+    Rect box = getBBox(img, _shape);
+
     if(img.channels()>1){
         cvtColor(img,img,COLOR_BGR2GRAY);
     }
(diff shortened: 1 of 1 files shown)


modules/face/src/facemarkLBF.cpp
```diff
@@ -535,13 +535,14 @@ Rect FacemarkLBFImpl::getBBox(Mat &img, const Mat_<double> shape) {
 void FacemarkLBFImpl::prepareTrainingData(Mat img, std::vector<Point2f> facePoints,
     std::vector<Mat> & cropped, std::vector<Mat> & shapes, std::vector<BBox> &boxes)
 {
+    Mat shape;
+    Mat _shape = Mat(facePoints).reshape(1);
+    Rect box = getBBox(img, _shape);
+
     if(img.channels()>1){
         cvtColor(img,img,COLOR_BGR2GRAY);
```
alalek, line 543: Perhaps, it makes sense to convert color of cropped image only. Update the code below:

```
Mat crop = img(Rect((int)min_x, (int)min_y, (int)w, (int)h)).clone();
```

(BTW, avoid using of in-place src/dst parameters - they cause performance regressions)

iclemens: Oh, I completely agree... as you can see I just moved some existing code around (i.e. tried to change as little as possible).

alalek: I believe it is OK to make API more friendly for Python or other bindings.

alalek (approved):

PR #2957, opened 2021-05-25
further work on Ptcloud module

berak:

trying to take over, where #2584 left it.

added sprt stopping, preemptive segmentation, python bindings and tests


berak:
**notes / todo / rfc**

RTSP stopping is very useful here, and the USAC implementation from calib3d works nice.
however, i could not use the interfaces (needed to copy the code) because:
* they're neither public (header is in calib3d src) nor are they exported (win dll) [minor]
* they don't let me collect inlier indices (needed for local optimization later) [major]
* the rtsp verifier run its own per-point loop, and neither collects inliers, nor errors (for ransac or sac), also we cannot use a index mask for already segmented points here

i finally mimicked the Error interface from USAC, a distance function *per point*. sadly this is hard to low-level optimize (simd,parallel,..), like this is done in pcl.

some stats (timing, inlier counts) in the interface would be nice, too

berak:
superseeded by https://github.com/opencv/opencv/pull/21168 
(which has much better USAC integration)

PR #2964, opened 2021-06-03
ICP Docs

DumDereDum:

Add info about ICP algorithm in RGBD

```
force_builders_only=docs
```


mshabunin:
Why is this documentation added to the _rgbd_ module? I can see that ICP algorithm is implemented in the _surface_matching_ module: https://docs.opencv.org/master/dc/d9b/classcv_1_1ppf__match__3d_1_1ICP.html

I propose extracting this section to separate page and referring it in the _rgbd_ and _ICP_ descriptions. In order to do it you have to add a page to the `<module>/doc` folder and use `@ref` Doxygen references to link to it (e.g. https://github.com/opencv/opencv_contrib/tree/master/modules/bioinspired/doc).

modules/rgbd/doc/kinfu_icp.markdown
```diff
@@ -0,0 +1,111 @@
...
+
+<img src="https://render.githubusercontent.com/render/math?math=T_{p2q}(p) = (R \cdot  p %2B t)">
+
+Where ***R*** - rotation, ***t*** - translation.
+
+We use the Gauss-Newton method for the minimization of function.
+In the beginning, we will perform some mathematical operations:
+
+<img src="https://render.githubusercontent.com/render/math?math=E=\sum\left\|[(R\cdot p %2B t)-q]^{T}\cdot n\right\|_{2}">
+
```
savuor, line 35: It's better to add something like this:
In Gauss-Newton method we do sequential steps by changing R and t in the direction of the function E decrease, i.e. in the direction of its gradient:
1. At each step we approximate the function E linearly as its current value plus Jacobian matrix multiplied by delta x which is concatenated delta R and delta t vectors.
1. We find delta R and delta t by solving the equation E_approx(delta_x) = 0
1. We apply delta R and delta t to current Rt transform and proceed to next iteration

To linearize E, let's approximate it in infinitesimal neighborhood.

modules/rgbd/doc/kinfu_icp.markdown
```diff
@@ -0,0 +1,111 @@
...
+<img src="https://render.githubusercontent.com/render/math?math=E = \sum \left \| [skew(R_{shift}) \cdot  p %2B t %2B p- q]^{T}  \cdot n \right \|_{2} ">
+
+Let a new function:
+
+<img src="https://render.githubusercontent.com/render/math?math=f(x, p) = skew(R_{shift}) \cdot  p %2B t">
+
+<img src="https://render.githubusercontent.com/render/math?math=E = \sum \left \| [f(x, p) %2B p- q]^{T}  \cdot n \right \|_{2}">
+
+Let's find out differential of ***E***:
```
savuor, line 65: *E is mininal when its differential is zero:

savuor (approved):

PR #2973, opened 2021-06-11
Make aruco board ids field accessible from python

rogday:
Fixing https://github.com/opencv/opencv-python/issues/451.


alalek:
BTW, it makes sense to create a simple test for this functionality (use modules/aruco/mics/python/test_aruco.py file, see other modules for reference)

alalek (approved):
Well done! Thank you :+1:

PR #2980, opened 2021-06-17
Update chessboard_corners.jl

yakir12:
Just added a download so that the user doesn't need to manually do it.

### DIFF
--- a/modules/julia/samples/chessboard_corners.jl
+++ b/modules/julia/samples/chessboard_corners.jl
@@ -2,9 +2,8 @@ using OpenCV
 
 const cv = OpenCV
 
-
-# chess1.png is at https://raw.githubusercontent.com/opencv/opencv_extra/master/testdata/cv/cameracalibration/chess1.png
-img = cv.imread("chess1.png",cv.IMREAD_GRAYSCALE)
+file = download("https://raw.githubusercontent.com/opencv/opencv_extra/master/testdata/cv/cameracalibration/chess1.png")
+img = cv.imread(file,cv.IMREAD_GRAYSCALE)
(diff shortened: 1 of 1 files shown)


archit120:
Add a check if the file already exists?

yakir12:
How would you check that? By name? This is a downloaded file, it could have been named to whatever, indeed, in this PR I download it to a temp file (I don't even know what it's named to). 
Adding a known md5sum would be a nice touch, but the identification of the content of the file is implied by the link, and in any case, that would seem like an overkill for this kind of MWE. But if you disagree, go right ahead. The main intention with this PR has only been to facilitate running this example, so a user can include this file and have it just work (rather than manually downloading that file, naming it to something specific, etc).

modules/julia/samples/chessboard_corners.jl
```diff
@@ -2,9 +2,8 @@ using OpenCV
 
 const cv = OpenCV
 
-
-# chess1.png is at https://raw.githubusercontent.com/opencv/opencv_extra/master/testdata/cv/cameracalibration/chess1.png
-img = cv.imread("chess1.png",cv.IMREAD_GRAYSCALE)
+file = download("https://raw.githubusercontent.com/opencv/opencv_extra/master/testdata/cv/cameracalibration/chess1.png")
```
alalek, line 5: In general samples should not force downloading of external content.
We should have pre-check if file already exists in current working directory (e.g., downloaded by user before)

yakir12: Honestly, the best way would be to use an Artifact. That would take care of everyone's needs.

PR #2981, opened 2021-06-18
KinFu python samples

DumDereDum:


savuor (approved):

modules/rgbd/samples/kinfu_demo.py
```diff
@@ -0,0 +1,60 @@
...
+from argparse import ArgumentParser
+
+def get_depth_list(folder):
+    if sys.platform.find('linux')+1:
+        f = open(folder + '/depth.txt', 'r')
+        rgb = [folder + '/' + s.replace('/', '/') for s in f.read().split() if '.png' in s]
+    else:
+        f = open(folder + '\\depth.txt', 'r')
+        rgb = [folder + '\\' + s.replace('/', '\\') for s in f.read().split() if '.png' in s]
+    return rgb
```
alalek, line 13: Do we really need this branching?

Windows should work properly with forward slashes (/) too.

---

> `if '.png' in s`

`if s.endswith('.png')`

modules/rgbd/samples/kinfu_demo.py
```diff
@@ -0,0 +1,60 @@
+import cv2
+import numpy as np
```
alalek, line 2: numpy import should go before cv2

modules/rgbd/samples/kinfu_demo.py
```diff
@@ -0,0 +1,60 @@
...
+    else:
+        raise ValueError("Incorrect kinfu type name")
+
+    for path in depth_list:
+
+        image = cv2.imread(path, cv2.IMREAD_ANYDEPTH)
+        (height, width) = image.shape
+
+        cv2.imshow('input', image)
+        cv2.waitKey(1)
```
alalek, line 42: > waitKey(1)


There is `pollKey()` without any timeouts.

modules/rgbd/samples/kinfu_demo.py
```diff
@@ -0,0 +1,60 @@
...
+        if not kf.update(image):
+            kf.reset()
+        else:
+            kf.render(cvt8)
+            cv2.imshow('render', cvt8)
+
+
+if __name__ == '__main__':
+    print(__doc__)
+    cv2.setUseOptimized(True)
```
alalek, line 58: > `cv2.setUseOptimized(True)`

Useless call. Optimized code path is run by default.

modules/rgbd/samples/kinfu_demo.py
```diff
@@ -0,0 +1,60 @@
+import cv2
```
alalek, line 1: import cv2 as cv

Replace `cv2.` => `cv.`

modules/rgbd/samples/kinfu_demo.py
DumDereDum: It added for the case when you need to off ocl

modules/rgbd/samples/kinfu_demo.py
DumDereDum: I cann't find this function in cv2

modules/rgbd/samples/kinfu_demo.py
DumDereDum: done

modules/rgbd/samples/kinfu_demo.py
DumDereDum: done

modules/rgbd/samples/kinfu_demo.py
alalek: => `ArgumentParser` is for that case.

Anyway it turns off too many parts.
For OpenCL there is cv.ocl.setUseOpenCL() call

modules/rgbd/samples/kinfu_demo.py
alalek: https://docs.opencv.org/master/d7/dfc/group__highgui.html#ga6d20fbd3100ec3badc1eaa653aff99d7

modules/rgbd/samples/kinfu_demo.py
DumDereDum: done

modules/rgbd/samples/kinfu_demo.py
DumDereDum: done, thx

modules/rgbd/samples/kinfu_demo.py
DumDereDum: done

savuor (approved):

modules/rgbd/samples/kinfu_demo.py
```diff
@@ -0,0 +1,57 @@
...
+
+        size = height, width, 4
+        cvt8 = np.zeros(size, dtype=np.uint8)
+
+        if not kf.update(image):
+            kf.reset()
+        else:
+            kf.render(cvt8)
+            cv.imshow('render', cvt8)
+
```
alalek, line 52: Perhaps we should have `waitKey(0)` after the loop before exit.
Otherwise app will just close in the end. User can't see the final 'render' frame.

DumDereDum: do you min replacing `pollKey()` after the loop

alalek: pollKey() must be called after each imshow() call to update GUI window content (handle UI redraw events). See `imshow()` documentation.

We need `waitKey(0)` for infinite wait after loop and before calling `cv.destroyAllWindows()`.

DumDereDum: done

alalek:
Why do we have 2 `cv.waitKey(0)` calls?

DumDereDum:
@alalek , which one i need to leave?

alalek:
The first one should be enough.

DumDereDum:
@alalek done)

modules/rgbd/samples/kinfu_demo.py
```diff
@@ -0,0 +1,59 @@
...
+        raise ValueError("Incorrect kinfu type name")
+
+    depth_list = get_depth_list(args.input)
+    for path in depth_list:
+
+        image = cv.imread(path, cv.IMREAD_ANYDEPTH)
+        (height, width) = image.shape
+
+        cv.imshow('input', image)
+        cv.pollKey()
```
savuor, line 42: Don't need that `pollKey()` call

modules/rgbd/samples/kinfu_demo.py
```diff
@@ -0,0 +1,59 @@
...
+
+        size = height, width, 4
+        cvt8 = np.zeros(size, dtype=np.uint8)
+
+        if not kf.update(image):
+            kf.reset()
+        else:
+            kf.render(cvt8)
+            cv.imshow('render', cvt8)
+            cv.pollKey()
```
savuor, line 52: This `pollKey()` call should be done after the if block

savuor (changes requested):
2 pollKey changes and it's fine

modules/rgbd/samples/kinfu_demo.py
DumDereDum: done

modules/rgbd/samples/kinfu_demo.py
DumDereDum: done

savuor (approved):

PR #2988, opened 2021-06-28
Do not use = 0 for a cv::Mat.

vrabaud:
There are several operator= overloads and some compilers can be confused.

This is similar to opencv/opencv#20304 and https://github.com/opencv/opencv_contrib/pull/2987 but for opencv_contrib master.


modules/alphamat/src/cm.cpp
```diff
@@ -92,7 +92,7 @@ void lle(my_vector_of_vectors_t& indm, my_vector_of_vectors_t& samples, float ep
     float alpha, beta, lagrangeMult;
     Cones = 1;
 
-    C = 0;
+    C.setTo(cv::Scalar::all(0));
```
alalek, line 95: `C.setTo(0);` should do the same (it should work with "0").

Also, there is `Cones = 1;` which may be critical for that. `rhs =1` do the same.

BTW, Anyway, it is better to initialize data in ctor instead of having dedicated `.setTo()` call.

vrabaud: I fixed it for Cones too. My compiler did not complain about it though, probably because it does not understand 1 as a pointer.

alalek (approved):

Issue #2553, opened 2020-06-01
An average quaternion cannot just be obtained by taking a Euclidean weighted mean


SoapJN:
https://github.com/opencv/opencv_contrib/blob/8fc6067b5987316d0586bc5db45c10aadad8c97f/modules/surface_matching/src/ppf_match_3d.cpp#L387-L401

The unit quaternions are no longer units when averaged.
We have two unit quaternions q1(w1, x1, y1, z1)、q2(w2, x2, y2, z2)，
||q1|| = sqr(w1) + sqr(x1) + sqr(y1) + sqr(z1) = 1,
||q2|| = sqr(w2) + sqr(x2) + sqr(y2) + sqr(z2) = 1.
Result of averaged the quaternions: q3((w1+w2) / 2, (x1+x2) / 2, (y1+y2) / 2, (z1+z2) / 2)
||q3|| = sqr((w1+w2) / 2) +  sqr((x1+x2) / 2) + sqr((y1+y2) / 2) + sqr((z1+z2) / 2) 
= 1/2 + w1*w2/2 +  x1*x2/2 + y1*y2/2 + z1*z2/2 
Only when q1 is equal to q2, the q3 is unit.
quaternions averaged cannot represent rigid transform matrix averaged

SoapJN:
I studied one paper(**Averages of Rotations and Orientations in 3-space**) about this problem.  in the paper, it explicitly states that the method in the code is wrong

catree:
[Averaging Quaternions](https://github.com/tolgabirdal/averaging_quaternions)?

SoapJN:
> [Averaging Quaternions](https://github.com/tolgabirdal/averaging_quaternions)?

Your reference is correct ，This code have the problem

SoapJN:
Your reference is correct ，This code can be optimized for higher accuracy

> [Averaging Quaternions](https://github.com/tolgabirdal/averaging_quaternions)?

Your reference is correct ，this code can be optimized for higher accuracy

SoapJN:
> [Averaging Quaternions](https://github.com/tolgabirdal/averaging_quaternions)?

Your reference is correct ，This code can be optimized for higher accuracy

savuor:
Yes, here's a mistake, a quaternion should be of unit norm, it's expected to be unit norm. This can be changed to eigenvalue-based method in the article you mentioned.
[...]

PR #2995, opened 2021-07-06
surface matching

hengguan:
### normalize quaternion
After weighted average quaternion of poses in the same cluster, the weighted average quaternion should be normalized.


savuor:
This is not the only (and not the most accurate) pose averaging method, see my comments to issue #2553
But it is accurate enough if the poses are close to each other. The only way to check this is to run the corresponding tests.

savuor:
:+1:

PR #2997, opened 2021-07-06
Add read/write functions to xfeatures2d and normalize naming convention

augustinmanecy:

### DIFF
--- a/modules/xfeatures2d/src/surf.hpp
+++ b/modules/xfeatures2d/src/surf.hpp
@@ -25,16 +25,24 @@ public:
 
     void read( const FileNode& fn)
     {
-      fn["extended"] >> extended;
-      fn["upright"] >> upright;
-      fn["nOctaves"] >> nOctaves;
-      fn["nOctaveLayers"] >> nOctaveLayers;
+      // if node is empty, keep previous value
+      if (!fn["hessianThreshold"].empty())
+        fn["hessianThreshold"] >> hessianThreshold;
+      if (!fn["extended"].empty())
+        fn["extended"] >> extended;
+      if (!fn["upright"].empty())
+        fn["upright"] >> upright;
+      if (!fn["nOctaves"].empty())
+        fn["nOctaves"] >> nOctaves;
+      if (!fn["nOctaveLayers"].empty())
+        fn["nOctaveLayers"] >> nOctaveLayers;
     }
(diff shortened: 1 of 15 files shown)


modules/xfeatures2d/src/surf.hpp
```diff
@@ -23,6 +23,25 @@ class SURF_Impl : public SURF
                                int nOctaves = 4, int nOctaveLayers = 2,
                                bool extended = true, bool upright = false);
 
+    void read( const FileNode& fn)
+    {
+      fn["extended"] >> extended;
```
berak, line 28: what about `hessianThreshold` ?

berak:
you'll have to [fix the java tests](https://pullrequest.opencv.org/buildbot/builders/precommit-contrib_linux64/builds/15360/steps/test_java/logs/junit-html#org.opencv.test.features2d) which currently expect either [empty](https://github.com/opencv/opencv/blob/master/modules/features2d/misc/java/test/SURFFeatureDetectorTest.java#L161) files, or a [different layout](https://github.com/opencv/opencv/blob/master/modules/features2d/misc/java/test/SURFFeatureDetectorTest.java#L146)

and there's a small general problem. 
loading an xml from a previous version (no params in it) will lead to resetting the algorithm's params to 0, example:

```
<?xml version="1.0"?>
<opencv_storage>
</opencv_storage>
```

```
FileStorage fs2("my.xml",0);
int nOctaves=4; 
fs2["nOctaves"] >> nOctaves;
// nOctaves is 0 now
```

augustinmanecy:
@berak I think I corrected the java tests which failed in your link (TestCase Features2dTest.TestPTOD, and tests for BRIEF and SURF) 
However, as I never dealt with java before and I do not know how to run all of this tests before committing a validated fix ;p Can you give me some instructions to be able to execute the test on my side before committing again?

alalek:
Build target: `opencv_test_java`
From build dir: `python <opencv_src_path>/modules/ts/misc/run.py -v -a -t java --package=xfeatures2d`
(package parameter is optional)

augustinmanecy:
Hi, I just pushed a new commit which fix Java tests. I also added additionnal check in read methods in order to avoid erasing current value of parameters which are not in the configuration file (if a paramater was missing in the configuration file, it was replaced by 0). 
Normally tests should pass now.

PR #2998, opened 2021-07-07
Remove non-fatal error if there was no ROI support and fix some tests.

tolysz:
ROI support looks optional and the code already was handling the case, the `CV_Error` was breaking the action.
Test been failing if there is more than one device, they been using gpu0 to load the library but try to compute on all.
`OpticalFlowDual_TVL1` requires more changes to use the correct GPU, forcing the first for now. Keeping it enabled breaks further tests.

### DIFF
--- a/modules/ximgproc/src/seeds.cpp
+++ b/modules/ximgproc/src/seeds.cpp
@@ -136,8 +136,8 @@ private:
     //compute initial label for sublevels: level <= seeds_top_level
     //this is an equally sized grid with size nr_h[level]*nr_w[level]
     int computeLabel(int level, int x, int y) {
-        return std::min(y / (height / nr_wh[2 * level + 1]), nr_wh[2 * level + 1] - 1) * nr_wh[2 * level]
-                + std::min((x / (width / nr_wh[2 * level])), nr_wh[2 * level] - 1);
+        return std::min(y * nr_wh[2 * level + 1] / height, nr_wh[2 * level + 1] - 1) * nr_wh[2 * level]
+            + std::min(x * nr_wh[2 * level] / width, nr_wh[2 * level] - 1);
     }
     inline int nrLabels(int level) const {
         return nr_wh[2 * level + 1] * nr_wh[2 * level];


modules/cudaoptflow/test/test_optflow.cpp
```diff
@@ -468,7 +468,7 @@ CUDA_TEST_P(OpticalFlowDual_TVL1, Async)
 }
 
 INSTANTIATE_TEST_CASE_P(CUDA_OptFlow, OpticalFlowDual_TVL1, testing::Combine(
-    ALL_DEVICES,
+    FIRST_DEVICE,
```
alalek, line 471: I believe there should be a `ts` option to "reconfigure" CUDA `ALL_DEVICES` instead of changing of each tests.

/cc @asmorkalov

modules/cudaoptflow/src/nvidiaOpticalFlow.cpp
```diff
@@ -845,7 +845,7 @@ NvidiaOpticalFlowImpl_2::NvidiaOpticalFlowImpl_2(
         {
             m_enableROI = false;
             m_roiData = nullptr;
-            CV_Error(Error::StsBadFunc, "ROI not supported on this GPU");
+            // CV_Error(Error::StsBadFunc, "ROI not supported on this GPU");
```
alalek, line 848: As I understand the final result is different (on supported and on unsupported GPUs), so this unsupported code path can't be just ignored.

/cc @vchiluka5

tolysz: https://docs.nvidia.com/video-technologies/optical-flow-sdk/nvofa-application-note/
I see. But then the application dies at this point. 
I would imagine that this check could be made as a function outside, so user could have advance notice and decide to solve it in some other way, or accept this degradation.

modules/cudaoptflow/test/test_optflow.cpp
tolysz: I would imagine that we could fix the funcion to use the actual device not just gpu0 for all tests.

PR #3001, opened 2021-07-09
Correct width/height returned by cudacodec::VideoReader::FormatInfo()

cudawarped:
When decoding certain resolutions (e.g. 1080p), the size of the frame returned from `nextFrame()` (1920x**1088**) can be larger than the usable area (1920x**1080**) due to the coded sizes being multiples of n bytes, for efficient coding/decoding.

This causes a problem in cudacodec because currently the values of width and height returned by `cudaCodec::Format()` are different depending on which video source is used.  The two possibilities are FFmpeg and cuvid, with FFmpeg returning the usable width and height (1920x**1080**) and cuvid returning the coded width and height (1920x**1088**).

I think it would make sense for both video sources to return the coded width and height if possible because that corresponds to the dimensions of the frame returned by `nextFrame()`.  Unfortunately in the case of the FFmpeg backend this won't be known for certain until after the first call to `nextFrame()` where the decoder may or may not reconfigure itself based on the coded bit stream.  To avoid changing the internals of cudacodec too much (it needs a re-write to mirror the existing Nvidia sample code anyway) I have included an ugly hack using a flag to signal if the dimensions of width and height can be relied upon.

The accuracy test case has been updated and should fail without this fix.

Additionally I have included the display area so that the frame can be cropped to get the usable area (e.g. 1920x1080 from 1920x1088).

```
force_builders=Custom
```

### DIFF
--- a/modules/cudacodec/src/ffmpeg_video_source.cpp
+++ b/modules/cudacodec/src/ffmpeg_video_source.cpp
@@ -131,6 +131,8 @@ cv::cudacodec::detail::FFmpegVideoSource::FFmpegVideoSource(const String& fname)
     format_.codec = FourccToCodec(codec);
     format_.height = cap.get(CAP_PROP_FRAME_HEIGHT);
     format_.width = cap.get(CAP_PROP_FRAME_WIDTH);
+    format_.displayArea = Rect(0, 0, cap.get(CAP_PROP_FRAME_WIDTH), cap.get(CAP_PROP_FRAME_HEIGHT));
+    format_.valid = false;
     FourccToChromaFormat(pixelFormat, format_.chromaFormat, format_.nBitDepthMinus8);
 }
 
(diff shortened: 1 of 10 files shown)


modules/cudacodec/src/ffmpeg_video_source.cpp
```diff
@@ -131,6 +131,8 @@ cv::cudacodec::detail::FFmpegVideoSource::FFmpegVideoSource(const String& fname)
     format_.codec = FourccToCodec(codec);
     format_.height = cap.get(CAP_PROP_FRAME_HEIGHT);
     format_.width = cap.get(CAP_PROP_FRAME_WIDTH);
+    format_.displayArea = Rect(0, 0, cap.get(CAP_PROP_FRAME_WIDTH), cap.get(CAP_PROP_FRAME_HEIGHT));
```
alalek, line 134: It makes sense to reuse already fetched `.get()` results instead of querying a new one.

alalek (commented):

modules/cudacodec/src/ffmpeg_video_source.cpp
cudawarped: Agreed :thumbsup:

alalek (approved):
Thank you :+1:

PR #3002, opened 2021-07-11
[GSoC] Loop closure detection.

zihaomu:
## [GSoC] Loop closure detection.
Hi, this is a draft PR and only contains some implementation ideas.
DNN model (**HF-Net**) is used as a feature extractor in my project. I try to make the model support all platforms, but unfortunately, only the OpenVINO backend is currently supported.  And more backend models will be supported in the future. 

### DIFF
--- a/modules/rgbd/src/large_kinfu.cpp
+++ b/modules/rgbd/src/large_kinfu.cpp
@@ -193,25 +193,107 @@ bool LargeKinfuImpl<Mat>::update(InputArray _depth)
 }
 
 template<>
-bool LargeKinfuImpl<UMat>::update(InputArray _depth)
+bool LargeKinfuImpl<UMat>::update(InputArray _depth, InputArray _rgb)
 {
     CV_Assert(!_depth.empty() && _depth.size() == params.frameSize);
 
-    UMat depth;
-    if (!_depth.isUMat())
+    UMat depth, rgb;
+    if (!(_depth.isUMat() && _rgb.isUMat()))
     {
-        _depth.copyTo(depth);
-        return updateT(depth);
+        if (!_depth.isUMat()) _depth.copyTo(depth);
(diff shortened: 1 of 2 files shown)


modules/rgbd/src/submap.hpp
```diff
@@ -119,6 +121,24 @@ void Submap<MatType>::updatePyrPointsNormals(const int pyramidLevels)
...
+    if(iter == constraints.end())
+    {
+        // Frome pose to tarPose transformation
+        Affine3f estimatePose = tarPose * pose.inv();
+
+        // Create new Edge.
+        PoseConstraint& preConstrain = getConstraint(tarSubmapID);
+        preConstrain->accumulatePose(estimatePose, 1);
+    }
+}
```
zihaomu, line 140: Hi, @savuor. When the LOOP is detected, I need to add new edges and constraints. Is this implementation right?

modules/rgbd/src/large_kinfu.cpp
```diff
@@ -136,6 +143,10 @@ class LargeKinfuImpl : public LargeKinfu
 
     int frameCounter;
     Affine3f pose;
+
+    // Loop Closure
+    bool useLCD;
```
savuor, line 148: Do we need `useLCD` field? We can check `lcd` pointer instead

zihaomu: Thanks for code reviewing. it will be updated in the next version.

alalek:
Use this link: http://pullrequest.opencv.org/#/summary/contrib

zihaomu:
Hi @alalek. Thank you, the code has been updated, but it still fails on CI on ARM platform. I don't know how to deal with that..

alalek:
@zihaomu Required builds are green now.

PR #3006, opened 2021-07-15
[GSOC] Deltille Tag Addition

jakeywatson:
This is a draft PR with limited scope.

The code is largely taken from [here](https://github.com/gedikli/detector), as part of my GSOC 21 project on fiducial tags. 

[I] I agree to contribute to the project under Apache 2 License.
[I] To the best of my knowledge, the proposed patch is not based on a code under GPL or other license that is incompatible with OpenCV
[I] The PR is proposed to proper branch
[I] There is reference to original bug report and related work
[X] There is accuracy test, performance test and test data in opencv_extra repository, if applicable
[X] Patch to opencv_extra has the same branch name.
[X] The feature is well documented and sample code can be built with the project CMake


alalek (changes requested):
Please keep only your commits in the PR.
Properly select target and the name of the source branch (use descriptive name, avoid using upstream names).

> from here

LGPL or derived work is not allowed for contribution into OpenCV. Please contact with your mentor and GSoC admins.

PR #3011, opened 2021-07-29
Add wrapper to drawCharucoDiamond

gokberkgul:
`drawCharucoBoard` function can be exported since `Vec4i` is wrapped in the core

### DIFF
--- a/modules/aruco/misc/python/test/test_aruco.py
+++ b/modules/aruco/misc/python/test/test_aruco.py
@@ -30,5 +30,10 @@ class aruco_test(NewOpenCVTests):
         with self.assertRaises(cv.error):
             board.setIds(np.array([0]))
 
+    def test_drawCharucoDiamond(self):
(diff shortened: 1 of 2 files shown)


alalek (commented):
It would be nice to have simple Python test for this case (check binding wrapper only, no need for accuracy checks - non-empty result check would be enough).

modules/aruco/misc/python/test/test_aruco.py
```diff
@@ -30,5 +30,10 @@ def test_idsAccessibility(self):
         with self.assertRaises(cv.error):
             board.setIds(np.array([0]))
 
+    def test_drawCharucoDiamond(self):
+        aruco_dict = cv.aruco.Dictionary_get(0)
```
andy-held, line 34: `cv2.aruco.getPredefinedDictionary(cv.aruco.DICT_4X4_50)` is more common.

andy-held (changes requested):

andy-held (approved):

alalek (approved):

PR #3012, opened 2021-08-02
Fix: undefined reference to `cv::imread

garfeng:
F:/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/10.3.0/../../../../i686-w64-mingw32/bin/ld.exe: ../../lib/libcorrespondence.a(nRobustViewMatching.cc.obj):nRobustViewMatching.cc:(.text$_ZN5libmv14correspondence19nRobustViewMatching11computeDataERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0x3b): undefined reference to `cv::imread(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int)'

The sfm uses imread, but opencv_imgcodecs was not added to it's CMakeLists.txt

### DIFF
--- a/modules/sfm/CMakeLists.txt
+++ b/modules/sfm/CMakeLists.txt
@@ -154,7 +154,7 @@ endif()
 #include_directories(${OCV_TARGET_INCLUDE_DIRS_${the_module}})
 add_subdirectory("${CMAKE_CURRENT_LIST_DIR}/src/libmv_light" "${CMAKE_CURRENT_BINARY_DIR}/src/libmv")
 
-ocv_target_link_libraries(${the_module} ${LIBMV_LIGHT_LIBS})
+ocv_target_link_libraries(${the_module} ${LIBMV_LIGHT_LIBS} opencv_imgcodecs)
 
 
 ### CREATE OPENCV SFM TESTS ###


alalek:
imread usage should be removed instead.
Generic OpenCV modules should not depend on I/O modules.

garfeng:
Will anyone  fix the issue ?

`nRobustViewMatching.cc` in sfm still uses imread.

opencv_contrib-4.5.3\modules\sfm\src\libmv_light\libmv\correspondence\nRobustViewMatching.cc

Line: 53

Use param of cv::Mat instead ?

``` cpp
/**
 * Compute the data and store it in the class map<string,T>
 *
 * \param[in] filename   The file from which the data will be extracted.
 *
 * \return True if success.
 */
bool nRobustViewMatching::computeData(const string & filename)
{
  cv::Mat im_cv = cv::imread(filename, 0);
  if (im_cv.empty()) {
    LOG(FATAL) << "Failed loading image: " << filename;
    return false;
  }
  else
  {
    libmv::vector<libmv::Feature *> features;
    std::vector<cv::KeyPoint> features_cv;
    m_pDetector->detect( im_cv, features_cv );
    features.resize(features_cv.size());
    for(size_t i=0; i<features_cv.size(); ++i)
      features[i] = new libmv::PointFeature(features_cv[i]);

    cv::Mat descriptors;
    m_pDescriber->compute(im_cv, features_cv, descriptors);

    // Copy data.
    m_ViewData.insert( make_pair(filename,FeatureSet()) );
    FeatureSet & KeypointData = m_ViewData[filename];
    KeypointData.features.resize(descriptors.rows);
    for(int i = 0;i < descriptors.rows; ++i)
    {
      KeypointFeature & feat = KeypointData.features[i];
      descriptors.row(i).copyTo(feat.descriptor);
      *(PointFeature*)(&feat) = *(PointFeature*)features[i];
    }

    DeleteElements(&features);

    return true;
  }
}
```

modules/sfm/CMakeLists.txt
```diff
@@ -154,7 +154,7 @@ endif()
 #include_directories(${OCV_TARGET_INCLUDE_DIRS_${the_module}})
 add_subdirectory("${CMAKE_CURRENT_LIST_DIR}/src/libmv_light" "${CMAKE_CURRENT_BINARY_DIR}/src/libmv")
 
-ocv_target_link_libraries(${the_module} ${LIBMV_LIGHT_LIBS})
+ocv_target_link_libraries(${the_module} ${LIBMV_LIGHT_LIBS} opencv_imgcodecs)
```
alalek, line 157: On the line 114 there is already included `opencv_imgcodecs` dependency of the module. No need to duplicate that.

More details are required:
- failed command line via `make VERBOSE=1` (without any `-j` option)

alalek:
> Use param of cv::Mat instead

Right.
Computer Vision library should not be limited on work with files only.
Usually data comes from cameras or other sources. `cv::Mat` is a generic type to handle any input.

---

But it is a separate big enough task ("string" is a part of public API for now so it can't be easily changed for now - breaking change).

modules/sfm/CMakeLists.txt
alalek: Perhaps we need dependency for libmv library instead of opencv sfm module.

Try to modify this line instead:
https://github.com/opencv/opencv_contrib/blob/4.5.3/modules/sfm/CMakeLists.txt#L98

PR #3024, opened 2021-08-23
aruco: fix subpixel coordinates in interpolateCornersCharuco

andy-held:
The two possible paths of `interpolateCornersCharuco`, `_interpolateCornersCharucoApproxCalib` and `_interpolateCornersCharucoLocalHom`, assume that the top-left corner of the top-left pixel of an image has the coordinate (0,0) whereas for `cornerSubpix` that is the center of that pixel and the top-left corner is actually (-0.5, -0.5). To check this I rendered a ChArUco board in a way that has the board corners (not the marker corners) exactly at integer pixel coordinates. Both `_interpolateCornersCharucoApproxCalib` and `_interpolateCornersCharucoLocalHom` return exactly these coordinates already, however, `cornerSubpix` then optimizes that towards the top-left.

In the test `interpolateCornersCharuco` is once called with the `cameraMatrix` parameter, once without, so both paths (`_interpolateCornersCharucoApproxCalib` and `_interpolateCornersCharucoLocalHom`) are tested.

This seems to be an incarnation of undocumented subpixel coodinates, see https://github.com/opencv/opencv/issues/10130

The test also needs the data contributed here: https://github.com/opencv/opencv_extra/pull/900

### DIFF
--- a/modules/aruco/src/charuco.cpp
+++ b/modules/aruco/src/charuco.cpp
@@ -315,7 +315,7 @@ static int _selectAndRefineChessboardCorners(InputArray _allCorners, InputArray 
 
         for (int i = begin; i < end; i++) {
             vector<Point2f> in;
-            in.push_back(filteredChessboardImgPoints[i]);
+            in.push_back(filteredChessboardImgPoints[i] - Point2f(0.5, 0.5)); // adjust sub-pixel coordinates for cornerSubPix
             Size winSize = filteredWinSizes[i];
             if (winSize.height == -1 || winSize.width == -1)
                 winSize = Size(params->cornerRefinementWinSize, params->cornerRefinementWinSize);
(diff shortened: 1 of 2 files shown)


andy-held:
I could just take corners & ids and discard the actual image, that would make this pull a little less complex.

paroj:
one should probably note that `_interpolateCornersCharucoApproxCalib` generates candidates using `projectPoints`.

> projectPoints and cornerSubPix disagree on subpixel coordinates

it would help, if you could mention what the disagreement actually is.

also, did you check, whether this affects the `_interpolateCornersCharucoLocalHom` path?

andy-held:
`_interpolateCornersCharucoApproxCalib` and `_interpolateCornersCharucoLocalHom` assume that the top-left corner of the top-left pixel of an image has the coordinate (0,0) whereas for `cornerSubpix` that is the center of that pixel and the top-left corner is actually (-0.5, -0.5). To check this I rendered a ChArUco board in a way that has the board corners (not the marker corners) exactly at integer pixel coordinates. Both `_interpolateCornersCharucoApproxCalib` and `_interpolateCornersCharucoLocalHom` return exactly these coordinates already, however, `cornerSubpix` then optimizen that towards the top-left.

In the test `interpolateCornersCharuco` is once called with an intrinsic, once without, so both paths (`_interpolateCornersCharucoApproxCalib` and `_interpolateCornersCharucoLocalHom`) are tested.

paroj (approved):

PR #3081, opened 2021-10-18
Uniqueness ratio support for cuda::StereoBM

asmorkalov:
Merge with https://github.com/opencv/opencv_extra/pull/924

```
force_builders=Custom
```

### DIFF
--- a/modules/cudastereo/src/cuda/stereobm.cu
+++ b/modules/cudastereo/src/cuda/stereobm.cu
@@ -43,6 +43,7 @@
 #if !defined CUDA_DISABLER
 
 #include "opencv2/core/cuda/common.hpp"
+#include <limits.h>
 
 namespace cv { namespace cuda { namespace device
 {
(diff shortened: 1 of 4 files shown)


modules/cudastereo/src/cuda/stereobm.cu
```diff
@@ -313,29 +360,90 @@ namespace cv { namespace cuda { namespace device
...
                         //
                         // From CUDA 9, using "__syncwarp" with proper mask value instead of using "__syncthreads"
                         // could be an option, but the shared memory access pattern does not allow this option,
                         // resulting in race condition. (Checked via "cuda-memcheck --tool racecheck")
 
                         if (X < cwidth - RADIUS)
                         {
-                            int idx = row * cminSSD_step;
-                            if (minSSD.x < minSSDImage[idx])
+                            unsigned int last_opt = line_ssd_tails[3*row + 0];
```
JulieBar, line 406: could it be moved to a separate function to decrease duplicated code? see line 300 with row = 0

modules/cudastereo/src/cuda/stereobm.cu
```diff
@@ -393,14 +527,8 @@ namespace cv { namespace cuda { namespace device
...
 
-            cudaSafeCall( cudaMemcpyToSymbol( cwidth, &left.cols, sizeof(left.cols) ) );
-            cudaSafeCall( cudaMemcpyToSymbol( cheight, &left.rows, sizeof(left.rows) ) );
-            cudaSafeCall( cudaMemcpyToSymbol( cminSSDImage, &minSSD_buf.data, sizeof(minSSD_buf.data) ) );
-
             size_t minssd_step = minSSD_buf.step/minSSD_buf.elemSize();
-            cudaSafeCall( cudaMemcpyToSymbol( cminSSD_step,  &minssd_step, sizeof(minssd_step) ) );
-
-            callers[winsz2](left, right, disp, maxdisp, stream);
+            callers[winsz2](left, right, disp, maxdisp, uniquenessRatio, minSSD_buf.data, minssd_step, left.cols, left.rows, stream);
```
JulieBar, line 531: maybe we can use cudaMemset2DAsync instead of cudaMemset2D?

modules/cudastereo/src/cuda/stereobm.cu
asmorkalov: Theoretically yes. I tried to extract the function and faced the following issues:
- The function has very large context that is re-used on caller side. I do not want to re-compute some values several times.
- There are some simplifications for `row=0`.
- - The code become even less obvious, because most of variables are not function local and should be read "inline" to the caller side.

modules/cudastereo/src/cuda/stereobm.cu
```diff
@@ -313,29 +360,90 @@ namespace cv { namespace cuda { namespace device
...
+                        batch_ssds[4] = CalcSSD<RADIUS>(col_ssd_cache + threadIdx.x, col_ssd + 4 * (BLOCK_W + 2 * RADIUS), X, cwidth);
+                        __syncthreads();
+                        batch_ssds[5] = CalcSSD<RADIUS>(col_ssd_cache + threadIdx.x, col_ssd + 5 * (BLOCK_W + 2 * RADIUS), X, cwidth);
+                        __syncthreads();
+                        batch_ssds[6] = CalcSSD<RADIUS>(col_ssd_cache + threadIdx.x, col_ssd + 6 * (BLOCK_W + 2 * RADIUS), X, cwidth);
+                        __syncthreads();
+                        batch_ssds[7] = CalcSSD<RADIUS>(col_ssd_cache + threadIdx.x, col_ssd + 7 * (BLOCK_W + 2 * RADIUS), X, cwidth);
+
+                        uint mssd = ::min(::min(::min(batch_ssds[0], batch_ssds[1]), ::min(batch_ssds[4], batch_ssds[5])),
+                                         ::min(::min(batch_ssds[2], batch_ssds[3]), ::min(batch_ssds[6], batch_ssds[7])));
```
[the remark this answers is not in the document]
asmorkalov: Restored function as it was in original code.

modules/cudastereo/src/cuda/stereobm.cu
asmorkalov: Addressed.

alalek (approved):

PR #3090, opened 2021-10-26
Add Radon transform function to ximgproc

kuloPo:
This function calculates the Hough Space of a given image in any range.
Hough Transform is a discrete implementation of Radon Transform.


AleksandrPanov:
I compared the current version of Radon Transform with the [scikit-image radon version](https://scikit-image.org/docs/dev/api/skimage.transform.html#skimage.transform.radon) based on [scikit-image example.](https://scikit-image.org/docs/dev/auto_examples/transform/plot_radon_transform.html)

- The current version always crops the image when it is rotated.
- The current version gives a slightly different result compared to the scikit-image version (this may be due to different interpolation or different rounding.). Need add more tests to check the meaning of the result (for example tests from radon.py.txt).
- It might be helpful to add CV_32FC1 for src/dst.

[radon.py.txt](https://github.com/opencv/opencv_contrib/files/7455701/radon.py.txt)
radon

kuloPo:
Updates:

- Change function name to RadonTransform
- Fix always cropping corner when rotate
- transpose the output - now each column is the integral
- add support for float number

The small difference may caused by:
- rounding from rad to deg in scikit
- different rotation method
- copying image to a bigger background to capture the rotated image

AleksandrPanov:
Thank you, I think it got better after your update
radon2
:

AleksandrPanov:
LGTM + minor comments:
- this patch should go into 4.x branch (only bug fixes get into 3.4 branch).
- rebase your commits from 3.4 onto 4.x branch. For example:

Note: no need to re-open PR, apply changes "inplace".

kuloPo:
The branch is rebased to 4.x

modules/ximgproc/perf/perf_radon_transform.cpp
```diff
@@ -0,0 +1,36 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#include "perf_precomp.hpp"
+
+namespace opencv_test {
+    namespace {
+
+        typedef tuple<Size, MatType> RadonTransformPerfTestParam;
```
alalek, line 10: Please avoid indentation in namespaces

modules/ximgproc/samples/radon_transform_demo.cpp
```diff
@@ -0,0 +1,16 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#include <opencv2/highgui.hpp>
+#include <opencv2/ximgproc/radon_transform.hpp>
+
+int main() {
+	cv::Mat src = cv::imread("peilin_plane.png", cv::IMREAD_GRAYSCALE);
+	cv::Mat radon;
```
alalek, line 10: Please use 4 spaces for indentation (no tabs)

modules/ximgproc/samples/radon_transform_demo.cpp
```diff
@@ -0,0 +1,16 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#include <opencv2/highgui.hpp>
+#include <opencv2/ximgproc/radon_transform.hpp>
+
+int main() {
+	cv::Mat src = cv::imread("peilin_plane.png", cv::IMREAD_GRAYSCALE);
```
alalek, line 9: > cv::

In OpenCV samples we add `using namespace cv;` after includes and removing all `cv::` prefixes.

modules/ximgproc/test/test_radon_transform.cpp
```diff
@@ -0,0 +1,79 @@
...
+namespace opencv_test {namespace {
+
+TEST(RadonTransformTest, output_size)
+{
+    Mat src(Size(256, 256), CV_8U, Scalar(0));
+    circle(src, Point(128, 128), 64, Scalar(255), FILLED);
+    Mat radon;
+    cv::ximgproc::RadonTransform(src, radon);
+
+    ASSERT_EQ(radon.rows, 363);
```
alalek, line 16: > _EQ

uses different order of parameters: `(expected, actual)`.

Signature for `_EQ` is `ASSERT_EQ(expected_reference, actual_result);`.
Correct order of parameters are required to emit valid error messages.

Reference: https://github.com/opencv/opencv/blob/4.0.0/modules/ts/include/opencv2/ts/ts_gtest.h#L8196-L8200

```
GTEST_API_ AssertionResult EqFailure(const char* expected_expression,
                                     const char* actual_expression,
                                     const std::string& expected_value,
                                     const std::string& actual_value,
                                     bool ignoring_case);
```

modules/ximgproc/perf/perf_radon_transform.cpp
```diff
@@ -0,0 +1,36 @@
...
+        {
+            Size srcSize = get<0>(GetParam());
+            int  srcType = get<1>(GetParam());
+
+            Mat src(srcSize, srcType);
+            Mat radon;
+
+            declare.in(src, WARMUP_RNG);
+
+            TEST_CYCLE_N(3)
```
alalek, line 28: > TEST_CYCLE_N

is misused here. No need to specify amount of inner samples here.

Please use:
- `TEST_CYCLE()`
- or `PERF_SAMPLE_BEGIN()` + `PERF_SAMPLE_END()`

modules/ximgproc/test/test_radon_transform.cpp
```diff
@@ -0,0 +1,79 @@
...
+}
+
+TEST(RadonTransformTest, accuracy_by_pixel)
+{
+    Mat src(Size(256, 256), CV_8U, Scalar(0));
+    circle(src, Point(128, 128), 64, Scalar(255), FILLED);
+    Mat radon;
+    cv::ximgproc::RadonTransform(src, radon);
+
+    ASSERT_EQ(radon.at<int>(0, 0), 0);
```
alalek, line 54: Before using `.at<int>` it makes sense to validate type.

```
ASSERT_EQ(CV_32SC1, radon.type());

EXPECT_EQ(0, radon.at<int>(0, 0));
```

(prefer to use `EXPECT_` for non-critical failures - e.g. code below would not crash)

alalek (commented):
@kuloPo Thank you for updates!

kuloPo:
All the issues are fixed.

AleksandrPanov (approved):
👍

PR #3091, opened 2021-10-28
Pass warping matrix as parameter instead of const memory

asmorkalov:
- To exclude data raises if warping is done in different streams
- Reduce amount of syscalls and memory transactions

```
force_builders=Custom
```

### DIFF
--- a/modules/cudawarping/src/cuda/warp.cu
+++ b/modules/cudawarping/src/cuda/warp.cu
@@ -53,55 +53,75 @@ namespace cv { namespace cuda { namespace device
 {
     namespace imgproc
     {
-        __constant__ float c_warpMat[3 * 3];
-
         struct AffineTransform
         {
-            static __device__ __forceinline__ float2 calcCoord(int x, int y)
+            static const int rows = 2;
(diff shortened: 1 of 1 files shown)


modules/cudawarping/src/cuda/warp.cu
```diff
@@ -53,55 +53,75 @@ namespace cv { namespace cuda { namespace device
...
+                const float coeff = 1.0f / (warpMat[6] * x + warpMat[7] * y + warpMat[8]);
 
-                const float xcoo = coeff * (c_warpMat[0] * x + c_warpMat[1] * y + c_warpMat[2]);
-                const float ycoo = coeff * (c_warpMat[3] * x + c_warpMat[4] * y + c_warpMat[5]);
+                const float xcoo = coeff * (warpMat[0] * x + warpMat[1] * y + warpMat[2]);
+                const float ycoo = coeff * (warpMat[3] * x + warpMat[4] * y + warpMat[5]);
 
                 return make_float2(xcoo, ycoo);
             }
+            struct Coefficients
```
JulieBar, line 90: Maybe we can re-use common Coefficients for both structs?

JulieBar (approved):
LGTM

PR #3092, opened 2021-10-28
fixed /modules/text/textdetection.py sample

aimbot6120:

- [ Y] I agree to contribute to the project under Apache 2 License.
- [ Y] To the best of my knowledge, the proposed patch is not based on a code under GPL or other license that is incompatible with OpenCV
- [ Y] The PR is proposed to proper branch
- [ N] There is reference to original bug report and related work
- [ N] There is accuracy test, performance test and test data in opencv_extra repository, if applicable
- [ N] The feature is well documented and sample code can be built with the project CMake

The sample python code `opencv/opencv_contrib/modules/text/sample/textdetection.py` contained errors and could not be be run. Moreover, the code was not optimized. This pull request fixes both issues.

What caused errors:
In line (old) 27, cv.text.computeNMChannels(img) returns a tuple containing extracted image channels. There is an attempt to append negative channels which results in error as tuples are not mutable.

Fix
converted the returned tuple to list
Removed extra bracket pair in line 30 which could also be causing errors.
Also changed the loading path to `trained_classifierNM1.xml` for classifiers because my system was not able to detect the older path
````
pathname+'/trained_classifierNM1.xml'
````

Performance Optimization

Issue:
From line 38 to 43 in the old code, the classifiers are loaded inside a for loop repeatedly. They need to be loaded only once to provide same results. 

Fix:
Moved the classifiers loading outside the for block (line 33-39 in PR)

I have tested that the python file is now working

### DIFF
--- a/modules/text/samples/textdetection.py
+++ b/modules/text/samples/textdetection.py
@@ -17,29 +17,29 @@ if (len(sys.argv) < 2):
 
 pathname = os.path.dirname(sys.argv[0])
 
-
 img      = cv.imread(str(sys.argv[1]))
 # for visualization
 vis      = img.copy()
 
 
 # Extract channels to be processed individually
-channels = cv.text.computeNMChannels(img)
+channels = list(cv.text.computeNMChannels(img))
 # Append negative channels to detect ER- (bright regions over dark background)
 cn = len(channels)-1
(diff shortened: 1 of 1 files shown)


alalek (approved):

PR #3093, opened 2021-10-31
New superpixel algorithm (F-DBSCAN)

scloke:
Implementation of a new superpixel algorithm, "Accelerated superpixel image segmentation with a parallelized DBSCAN algorithm".

```
force_builders=Linux x64,Win64,Mac,Android armeabi-v7a,Docs,iOS,Win32,ARMv7,ARMv8,Linux x64 Debug
```


modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
```diff
@@ -0,0 +1,127 @@
...
+#define __OPENCV_SCANSEGMENT_HPP__
+#ifdef __cplusplus
+
+#include <opencv2/core.hpp>
+
+namespace cv
+{
+	namespace ximgproc
+	{
+/** @brief Class implementing the F-DBSCAN (Accelerated superpixel image segmentation with a parallelized DBSCAN algorithm) superpixels
```
sturkmen72, line 65: please add
//! @addtogroup ximgproc_superpixel
//! @{

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
```diff
@@ -0,0 +1,127 @@
...
+		@param threads Number of processing threads for parallelisation. Default -1 uses the maximum number
+		of threads. In practice, four threads is enough for smaller images and eight threads for larger ones.
+		@param merge_small merge small segments to give the desired number of superpixels. Processing is
+		much faster without merging, but many small segments will be left in the image.
+		The function initializes a ScanSegment object for the input image. It stores the parameters of
+		the image: image_width and image_height. It also sets the parameters of the F-DBSCAN superpixel
+		algorithm, which are: num_superpixels, threads, and merge_small.
+		*/
+		CV_EXPORTS_W cv::Ptr<ScanSegment> createScanSegment(int image_width, int image_height, int num_superpixels, int threads = -1, bool merge_small = true);
+	}
```
sturkmen72, line 124: please add
//! @}

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
```diff
@@ -0,0 +1,127 @@
...
+		get the actual number.
+		@param threads Number of processing threads for parallelisation. Default -1 uses the maximum number
+		of threads. In practice, four threads is enough for smaller images and eight threads for larger ones.
+		@param merge_small merge small segments to give the desired number of superpixels. Processing is
+		much faster without merging, but many small segments will be left in the image.
+		The function initializes a ScanSegment object for the input image. It stores the parameters of
+		the image: image_width and image_height. It also sets the parameters of the F-DBSCAN superpixel
+		algorithm, which are: num_superpixels, threads, and merge_small.
+		*/
+		CV_EXPORTS_W cv::Ptr<ScanSegment> createScanSegment(int image_width, int image_height, int num_superpixels, int threads = -1, bool merge_small = true);
```
sturkmen72, line 123: please add
//! @}

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
```diff
@@ -0,0 +1,127 @@
...
+#ifndef __OPENCV_SCANSEGMENT_HPP__
+#define __OPENCV_SCANSEGMENT_HPP__
+#ifdef __cplusplus
+
+#include <opencv2/core.hpp>
+
+namespace cv
+{
+	namespace ximgproc
+	{
```
sturkmen72, line 64: please add
//! @addtogroup ximgproc_superpixel
//! @{

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
scloke: inserted the @addtogroup block as requested

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
scloke: inserted the @addtogroup block as requested

sturkmen72: you should put in `scansegment.hpp`

scloke: my apologies. fixed the files.

sturkmen72:
@scloke first of all thank you for the contribution
as a common OpenCV user, after compilation, I tested using the following python code
and want to share my experience.

code :
```
img = cv.imread('d:/test/hajandrade.jpg')
ss = cv.ximgproc.createScanSegment(img.shape[1],img.shape[0],500,1,True)
tm = cv.TickMeter()
tm.start()
ss.iterate(img)
res = ss.getLabelContourMask(True)
tm.stop()
print(tm.getTimeMilli())
res = cv.cvtColor(res,cv.COLOR_GRAY2BGR)
res = cv.add(res,img)
cv.imshow("Output", res);
cv.waitKey()
cv.destroyAllWindows()
```

1. is it possible to create one instance of ScanSegment and use it for different-sized images?
2. giving `createScanSegment `different values to `threads` produces different results.

threads = 1

image

threads = 4 

image

scloke:
Hi,

Thanks for the comments. Appreciate it+++ Am a bit of a newbie at open source, so I will try and explain the background of this contribution in a bit of detail.

For this algorithm, it was developed to try and speed-up superpixel segmentation. The underlying principle is straightforward. What it does is to parallelise two important processes, which are: 1) the actual segmentation process which is DBSCAN based, and 2) the merging of small segments.

Normally DBSCAN is considered too slow for real-time image work, and quite hard to parallelise since many small segments are created and large segments formed by different processes will overlap. The innovation here is to limit the cluster size and convert the colour difference function to simple integer-based arithmetic. This can be processed very fast and segments are limited in size and hence don't overlap.

Similarly, merging of small segments is quite slow as I am using an adaptation of OpenCV's watershed algorithm. I managed to parallelise it efficiently by dividing it in the same way as the segmentation, and using a window with a surrounding margin of 1 pixel.

This explains a couple of things which you have noticed. 1) The inverted-T pattern flows from the way that the DBSCAN clustering works. If you have a large homogenous textureless image, this pattern is repeated throughout. 2) The use of 4 threads shows a clear horizontal and vertical division, and this comes from the merging algorithm. 3) the output is deterministic when the number of threads is fixed (this is in the comments of the .hpp file), but different threads counts will give different segmentation results. 4) using the initialisation function will only allocated the buffers. Each run of the iterate function will segment a new image using the allocated buffers. No problems.

This algorithm has an O(n2) complexity (quadratic), so it is better for smaller images. When tested on the Berkeley Segmentation Dataset (smaller images), it is about six times faster than SEEDS or any of the other OpenCV algorithms (the new algorithm is F-DBSCAN on the chart).

Fig6

The segmentation accuracy is about the same as the OpenCV routines.
Fig9

When tested on this random 2 megapixel image on my hard drive with 200 superpixels, it gave 199 segments. In comparison, SEEDS gave 144 segments using the default settings. Runtime was about 32.9 s for 1000 iterations for F-DBSCAN and 58.3 s for 1000 iterations for SEEDS on a 10-core i9 Windows machine.

**Original**
7bda7bd38a24f2d02c8e80acf8e966d2

**F-DBSCAN**
7bda7bd38a24f2d02c8e80acf8e966d2_Segmented_199

**SEEDS**
7bda7bd38a24f2d02c8e80acf8e966d2_Seeds_144

The original algorithm was written in ISO C++ 17.0, so I had to adapt it to C++ 11.0 for OpenCV and use the inbuilt parallelisation routines. The speed is different from the original, but the segmentation accuracy should be the same, so I intend to submit this first and if it is approved, then I will use the finalised code to build a short comparison sample program to test performance against the other OpenCV algorithms. I will also write a proper guide on usage with examples.

SC

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
```diff
@@ -0,0 +1,132 @@
...
+		//! @addtogroup ximgproc_superpixel
+		//! @{
```
alalek, line 66: Please avoid indentation in namespaces.

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
```diff
@@ -0,0 +1,132 @@
...
+//
+//  IMPORTANT: READ BEFORE DOWNLOADING, COPYING, INSTALLING OR USING.
+//
+//  By downloading, copying, installing or using the software you agree to this license.
+//  If you do not agree to this license, do not download, install,
+//  copy or use the software.
+//
+//
+//                           License Agreement
+//                For Open Source Computer Vision Library
```
alalek, line 11: Please use short license header: https://github.com/opencv/opencv/wiki/Coding_Style_Guide#file-structure

```
//
// Copyright (C) 2021, Dr Seng Cheong Loke (lokesengcheong@gmail.com)
```

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
```diff
@@ -0,0 +1,132 @@
...
+// loss of use, data, or profits; or business interruption) however caused
+// and on any theory of liability, whether in contract, strict liability,
+// or tort (including negligence or otherwise) arising in any way out of
+// the use of this software, even if advised of the possibility of such damage.
+//
+//
+
+/*
+ * BibTeX reference
+@article{loke2021accelerated,
```
alalek, line 44: Please move this into this file: https://github.com/opencv/opencv_contrib/blob/4.x/modules/ximgproc/doc/ximgproc.bib

(preserve alpha-numeric order)

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
```diff
@@ -0,0 +1,132 @@
...
+  pages={1--16},
+  year={2021},
+  publisher={Springer}
+}
+*/
+
+
+#ifndef __OPENCV_SCANSEGMENT_HPP__
+#define __OPENCV_SCANSEGMENT_HPP__
+#ifdef __cplusplus
```
alalek, line 57: 1. `OPENCV_XIMGPROC_SCANSEGMENT_HPP`
2. remove `#ifdef __cplusplus` as C++ (even C++11) is prerequisite.

modules/ximgproc/src/scansegment.cpp
```diff
@@ -0,0 +1,926 @@
...
+		{
+			return makePtr<ScanSegmentImpl>(image_width, image_height, num_superpixels, threads, merge_small);
+		}
+
+		ScanSegmentImpl::ScanSegmentImpl(int image_width, int image_height, int num_superpixels, int threads, bool merge_small)
+		{
+			// set the number of process threads
+			processthreads = std::thread::hardware_concurrency();
+			if (threads > 0) {
+				processthreads = MIN(processthreads, threads);
```
alalek, line 229: As different "threads" values provide different results it is better to avoid changing this parameter depending on runtime environment.

It makes sense to rename this parameter to "slices" or similar.

modules/ximgproc/src/scansegment.cpp
```diff
@@ -0,0 +1,926 @@
...
+			indexProcessVec[processthreads - 1] = std::make_pair(processCurrent, indexSize);
+
+			// create buffers and initialise
+			std::vector<cv::Point> tempLoc{ cv::Point(-1, -1), cv::Point(0, -1), cv::Point(1, -1), cv::Point(-1, 0), cv::Point(1, 0), cv::Point(-1, 1), cv::Point(0, 1), cv::Point(1, 1) };
+			neighbourLoc = static_cast<cv::Point*>(malloc(8 * sizeof(cv::Point)));
+			memcpy(neighbourLoc, tempLoc.data(), 8 * sizeof(cv::Point));
+
+			labelsBuffer = static_cast<int*>(malloc(indexSize * sizeof(int)));
+			clusterBuffer = static_cast<int*>(malloc(indexSize * sizeof(int)));
+			pixelBuffer = static_cast<uchar*>(malloc(indexSize));
```
alalek, line 309: Manual memory management should be avoided (malloc/new).
Use `std::vector` or `cv::AutoBuffer` instead.

modules/ximgproc/src/scansegment.cpp
```diff
@@ -0,0 +1,926 @@
...
+
+			// set labels buffer to UNCLASSIFIED
+			std::fill(labelsBuffer, labelsBuffer + indexSize, UNCLASSIFIED);
+
+			// apply light blur
+			cv::medianBlur(src, src, 3);
+
+			// start at the center of the rect, then run through the remainder
+			labBuffer = reinterpret_cast<cv::Vec3b*>(src.data);
+			cv::parallel_for_(cv::Range(0, (int)indexNeighbourVec.size()), PP1(reinterpret_cast<ScanSegmentImpl*>(this)));
```
alalek, line 433: BTW, C++11 allows to use lambdas here instead of manual ParallelLoopBody classes.

modules/ximgproc/src/scansegment.cpp
```diff
@@ -0,0 +1,926 @@
...
+					clusterBuffer[currentClusterIndex + 1] = offsetEnd;
+				}
+			}
+		}
+
+		void ScanSegmentImpl::calculateCluster(int* offsetBuffer, int* offsetEnd, int pointIndex, int currentClusterID)
+		{
+			for (int i = 0; i < neighbourCount; i++) {
+				if (*offsetEnd < clusterSize) {
+					int intoffset2 = pointIndex + neighbourLocBuffer[i];
```
alalek, line 568: indentation is 4 spaces (no tabs)

alalek (commented):
Thank you for contribution!

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
scloke: extra indents removed on both the hpp and cpp files

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
scloke: license agreement updated

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
scloke: reference moved to ximgproc.bib. Bib file not in alpha-numeric order, so new reference added to the end.

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
scloke: c++ def removed

modules/ximgproc/src/scansegment.cpp
scloke: parameter changed to slices. default slices set to 8. However, setting to -1 will use the maximum number of processing threads. If the number of cpu cores from std::thread::hardware_concurrency() is less than this setting, it will use the smaller of the two to set the number of processing threads.

modules/ximgproc/src/scansegment.cpp
scloke: tab indents replaced with 4 spaces

modules/ximgproc/src/scansegment.cpp
scloke: I would prefer to leave this unchanged as my editor (Visual Studio Community) can't compile C++ 11 directly. I can only set it to ISO C++ 14 or 17, so changing to lambdas may introduce errors which I can't check. Unless someone else can help :)

modules/ximgproc/src/scansegment.cpp
scloke: I did a test of malloc vs vector performance previously, and manual memory management is much more efficient.

Ran the following short test with a buffer size of 100000 over 100000 iterations. Results were malloc 1477646 us, vector 2696077 us, which is 45.2% faster.

void testMalloc(int iterate, int buffersize, int* malloctime, int* vectortime)
{
    auto tstart1 = std::chrono::high_resolution_clock::now();

    for (int i = 0; i < iterate; i++) {
        int* testBuffer = static_cast<int*>(malloc(buffersize * sizeof(int)));
        std::fill(testBuffer, testBuffer + buffersize, 0);
        free(testBuffer);
    }

    auto tend1 = std::chrono::high_resolution_clock::now();
    *malloctime = (int)std::chrono::duration_cast<std::chrono::microseconds>(tend1 - tstart1).count();

    auto tstart2 = std::chrono::high_resolution_clock::now();

    for (int i = 0; i < iterate; i++) {
        std::vector<int> testVector(buffersize);
        std::fill(testVector.begin(), testVector.end(), 0);
        testVector.resize(0);
    }

    auto tend2 = std::chrono::high_resolution_clock::now();
    *vectortime = (int)std::chrono::duration_cast<std::chrono::microseconds>(tend2 - tstart2).count();
}

alalek: >    std::vector<int> testVector(buffersize);
>    std::fill(testVector.begin(), testVector.end(), 0);

`std::vector` already initializes (with zeros) in constructor.
Your micro benchmark dooes this twice.

There is `cv::AutoBuffer<int, 0>` which doesn't initialize values on creation.

---

Manual memory management is not exception safe code.

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
alalek: header guard should be renamed too (to include XIMGPROC in the name - see coding style at Wiki)

modules/ximgproc/src/scansegment.cpp
scloke: Thanks for the info. I used the autobuffer in the same way as in the CLAHE module, and the speed is the same as for malloc (approx). So I have changed the code to reflect this.

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
scloke: sorry, missed this. Changed in code.

modules/ximgproc/src/scansegment.cpp
scloke: Hi, could I check if the removal of malloc and replacement with autobuffer is sufficient to ensure exception-safe code?

modules/ximgproc/src/scansegment.cpp
scloke: I had a think about this and set the number of processing threads to be equal to the default slices. This should ensure consistent results. The user is still able to change this value to another number or even -1, and this will offer the option of speeding the algorithm if consistent results is not a priority.

modules/ximgproc/src/scansegment.cpp
```diff
@@ -0,0 +1,880 @@
...
+
+    // set labels buffer to UNCLASSIFIED
+    std::fill(labelsBuffer, labelsBuffer + indexSize, UNCLASSIFIED);
+
+    // apply light blur
+    cv::medianBlur(src, src, 3);
+
+    // start at the center of the rect, then run through the remainder
+    labBuffer = reinterpret_cast<cv::Vec3b*>(src.data);
+    cv::parallel_for_(cv::Range(0, (int)indexNeighbourVec.size()), PP1(reinterpret_cast<ScanSegmentImpl*>(this)));
```
alalek, line 385: BTW, using of C++11 lambdas with parallel_for looks like:

```
    parallel_for_(Range(0, (int)indexNeighbourVec.size()), [&](const Range& range) {
        for (int i = range.start; i < range.end; i++) {
            OP1(i);
        }
    });
```


OP4 case:

```
    // copy back to labels mat
    parallel_for_(Range(0, (int)indexProcessVec.size()), [&](const Range& range) {
        for (int i = range.start; i < range.end; i++) {
            OP4(indexProcessVec[i]);
        }
    });
```

modules/ximgproc/src/scansegment.cpp
```diff
@@ -0,0 +1,880 @@
...
+    int smallClusters;                      // clusters below this pixel count are considered small for merging
+
+    cv::AutoBuffer<cv::Rect> _seedRects;    // autobuffer of seed rectangles
+    cv::AutoBuffer<cv::Rect> _seedRectsExt; // autobuffer of extended seed rectangles
+    cv::AutoBuffer<cv::Rect> _offsetRects;  // autobuffer of offset rectangles
+    cv::AutoBuffer<cv::Point> _neighbourLoc;// autobuffer of neighbour locations
+    cv::Rect* seedRects;					// array of seed rectangles
+    cv::Rect* seedRectsExt;					// array of extended seed rectangles
+    cv::Rect* offsetRects;					// array of offset rectangles
+    cv::Point* neighbourLoc;				// neighbour locations
```
alalek, line 70: In general it is dangerous to store dedicated RAW pointers (RAW pointers unable to control lifetime of allocated buffer).
Also it doesn't make sense as `.data()` method is fast as RAW pointer.
Moreover `operator [](size_t i)` would check index for valid range in debug builds.

alalek (commented):
Thank you for updates!

modules/ximgproc/src/scansegment.cpp
```diff
@@ -0,0 +1,880 @@
...
+
+// Get highest absolute channel difference in diff
+#define c_diff(ptr1,ptr2,diff)           \
+{                                        \
+db = std::abs((ptr1)[0] - (ptr2)[0]); \
+dg = std::abs((ptr1)[1] - (ptr2)[1]); \
+dr = std::abs((ptr1)[2] - (ptr2)[2]); \
+diff = ws_max(db, dg);                \
+diff = ws_max(diff, dr);              \
+assert(0 <= diff && diff <= 255);  \
```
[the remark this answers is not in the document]
scloke: substituted with CV_Assert()

modules/ximgproc/src/scansegment.cpp
```diff
@@ -0,0 +1,880 @@
...
+#include <numeric>
+#include <atomic>
+#include <thread>
+
+namespace cv {
+namespace ximgproc {
+//! @addtogroup ximgproc_superpixel
+//! @{
+
+class ScanSegmentImpl : public ScanSegment
```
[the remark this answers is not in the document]
scloke: done

modules/ximgproc/src/scansegment.cpp
```diff
@@ -0,0 +1,880 @@
...
+node = free_node;                   \
+free_node = storage[free_node].next; \
+storage[node].next = 0;             \
+storage[node].mask_ofs = mofs;      \
+storage[node].img_ofs = iofs;       \
+if (q[idx].last)                   \
+storage[q[idx].last].next = node; \
+else                                \
+q[idx].first = node;            \
+q[idx].last = node;                 \
```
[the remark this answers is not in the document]
scloke: indentation added

modules/ximgproc/src/scansegment.cpp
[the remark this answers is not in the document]
scloke: all C-style assert replaced with CV_Assert()

modules/ximgproc/src/scansegment.cpp
```diff
@@ -0,0 +1,880 @@
...
+
+CV_EXPORTS Ptr<ScanSegment> createScanSegment(int image_width, int image_height, int num_superpixels, int slices, bool merge_small)
+{
+    return makePtr<ScanSegmentImpl>(image_width, image_height, num_superpixels, slices, merge_small);
+}
+
+ScanSegmentImpl::ScanSegmentImpl(int image_width, int image_height, int num_superpixels, int slices, bool merge_small)
+{
+    // set the number of process threads
+    processthreads = std::thread::hardware_concurrency();
```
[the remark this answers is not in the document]
scloke: replaced

modules/ximgproc/src/scansegment.cpp
scloke: Thanks+++ Appreciate the help. I compiled on my system on C++ 14 and no problems.

modules/ximgproc/src/scansegment.cpp
scloke: The .data() method is new to me, so I used the CLAHE implementation code in OpenCV as a template to follow. The code that was used there converted the .data() to a RAW pointer:
    cv::AutoBuffer<int> _tileHist(histSize);
    int* tileHist = _tileHist.data();

There is a significant speed difference when converting to .data() without dedicated RAW pointers. I ran with and without dedicated RAW pointers over a thousand cycles, twice, and the speed difference was 31.1s (with RAW) vs 37.3s (without RAW). I managed to improve the speed a bit by converting the neighbourLoc to a pre-initialised buffer, but otherwise I left it unchanged.

I think that the use of RAW pointers in this case should be safe enough since they are sourced from the AutoBuffers which are initialised as class variables that are allocated and deallocated based on the lifetime of the class. The evidence of this is:
1) running several thousand cycles in both debug and release builds showed no instability / memory leaks / buffer overruns.
2) the CLAHE module in OpenCV uses the same method and there has be no report of instability

alalek: I really have several questions for such micro-benchmarks... No idea that they measure.

https://github.com/opencv/opencv/blob/ac4b592b4e550a0ced1977e9aa19e8059a796e3c/modules/core/include/opencv2/core/utility.hpp#L127-L142

scloke: From the code you quoted, .data() should give a straight reference to the aligned pointer of the data in the AutoBuffer. The operator should also function similarly. Hence, there should not be any speed difference as you said.

Previously, I ran the entire code to process the 2 MP image I described earlier 1000 times. Now I have written a microbenchmark to test the AutoBuffer specifically.

// test 10000 autobuffer with raw vs without raw
void testIterate(int iterate, int buffersize, int* rawtime, int* norawtime)
{
    cv::AutoBuffer<int> _testBuffer = cv::AutoBuffer<int>(buffersize);
    int* testBuffer = _testBuffer.data();
    std::fill(testBuffer, testBuffer + buffersize, 0);

    auto tstart1 = std::chrono::high_resolution_clock::now();

    for (int i = 0; i < iterate; i++) {
        for (int j = 0; j < buffersize; j++) {
            testBuffer[j] = 0;
            testBuffer[j] = testBuffer[j] + 1;
        }
    }

    auto tend1 = std::chrono::high_resolution_clock::now();
    *rawtime = (int)std::chrono::duration_cast<std::chrono::microseconds>(tend1 - tstart1).count();


    cv::AutoBuffer<int> testBuffer2 = cv::AutoBuffer<int>(buffersize);
    std::fill(testBuffer2.data(), testBuffer2.data() + buffersize, 0);

    auto tstart2 = std::chrono::high_resolution_clock::now();

    for (int i = 0; i < iterate; i++) {
        for (int j = 0; j < buffersize; j++) {
            testBuffer2.data()[j] = 0;
            testBuffer2.data()[j] = testBuffer2.data()[j] + 1;
        }
    }

    auto tend2 = std::chrono::high_resolution_clock::now();
    *norawtime = (int)std::chrono::duration_cast<std::chrono::microseconds>(tend2 - tstart2).count();
}

These were the results in Debug and Release mode respectively, with the numbers in microseconds, iterating 50000 times, with a buffer size of 50000.

DEBUG
Debug

RELEASE
Release

There is a large difference between the use of RAWs and without, so much so that I am wondering if this is a compiler optimisation effect we are looking at.

If this is so, then the use of RAW pointers may be easier for the compiler to optimise, hence the speed difference we are seeing.

What do you think?

scloke: I tried again using XCode on iOS, and here are the new values

DEBUG
iOS Debug

RELEASE
iOS Release

This time the release numbers are looking more like expected. Once again, compiler differences? I am open to suggestions and have kept both versions on my system. If you think that it's better to go completely without RAWs, then I will upload the new version. Do let me know.

alalek: Release difference is about 0.1% - measurement accuracy. I would say we don't loose performance here.
Debug difference is expected - `.data()` method and other functions are not inlined by default, more checks are involved to validate code assumptions (e.g, CV_DbgAssert). Usually debug builds are not tracked for performance.

It is better to replace RAW pointers from code safety/security perspective.

scloke: I have replaced the RAW pointers in the code except for labBuffer since this is taken from a Mat.data rather than an Autobuffer. This should be safe since the lifetime of the pointer is short (only used in OP1 and read just before invocation), and is read-only for operator values.

modules/ximgproc/src/scansegment.cpp
```diff
@@ -0,0 +1,772 @@
...
+        for (int x = seedRect.x; x < seedRect.x + seedRect.width; x++) {
+            expandCluster(offsetVec[v].data(), cv::Point(x, y));
+        }
+    }
+}
+
+void ScanSegmentImpl::OP2(std::pair<int, int> const& p)
+{
+    std::pair<int, int>& q = const_cast<std::pair<int, int>&>(p);
+    for (int i = q.first; i < q.second; i++) {
```
alalek, line 340: Why do we need this alias?

> `std::pair<int, int>& q = const_cast<std::pair<int, int>&>(p);`


---

The same note is about OP4

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
```diff
@@ -0,0 +1,83 @@
...
+};
+
+/** @brief Initializes a ScanSegment object.
+
+The function initializes a ScanSegment object for the input image. It stores the parameters of
+the image: image_width and image_height. It also sets the parameters of the F-DBSCAN superpixel
+algorithm, which are: num_superpixels, threads, and merge_small.
+
+@param image_width Image width.
+@param image_height Image height.
```
alalek, line 71: BTW, prefer to use `Size image_size` instead of 2 dedicated values

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
```diff
@@ -0,0 +1,83 @@
...
+namespace cv { namespace ximgproc {
+
+/** @brief Class implementing the F-DBSCAN (Accelerated superpixel image segmentation with a parallelized DBSCAN algorithm) superpixels
+algorithm by Loke SC, et al.
+
+The algorithm uses a parallelised DBSCAN cluster search that is resistant to noise, competitive in segmentation quality, and faster than
+existing superpixel segmentation methods. When tested on the Berkeley Segmentation Dataset, the average processing speed is 175 frames/s
+with a Boundary Recall of 0.797 and an Achievable Segmentation Accuracy of 0.944. The computational complexity is quadratic O(n2) and
+more suited to smaller images, but can still process a 2MP colour image faster than the SEEDS algorithm in OpenCV. The output is deterministic
+when the number of processing threads is fixed, and requires the source image to be in Lab colour format.
```
alalek, line 21: Please add `@cite loke2021accelerated` in this documentation section.

alalek (commented):
Pushed update with fixed coding style, added smoke test.

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
scloke: I was thinking of this also, but in the end I went with following the pattern in createSuperpixelSEEDS which used two dedicated values. If you prefer cv::Size, let me know and I will change.

modules/ximgproc/include/opencv2/ximgproc/scansegment.hpp
scloke: added.

modules/ximgproc/src/scansegment.cpp
scloke: My apologies. This was some old code that got carried over. Have updated it.

alalek (approved):
Great job! Thank you for contribution :+1:

scloke:
My pleasure 😊

 

SC

scloke:
My pleasure :) SC

PR #3095, opened 2021-11-01
Add regression test for remap in case of overflow

JulieBar:
Related issue: https://github.com/opencv/opencv/issues/18224
Related fix in OpenCV: https://github.com/opencv/opencv/pull/20977

Fix as well as the issue addressed only BORDER_CONSTANT case; maybe later we should check issues for other borders too;

```
force_builders=Custom
Xbuild_image:Custom=ubuntu-cuda:16.04
```

### DIFF
--- a/modules/cudawarping/test/test_remap.cpp
+++ b/modules/cudawarping/test/test_remap.cpp
@@ -41,6 +41,8 @@
 //M*/
 
 #include "test_precomp.hpp"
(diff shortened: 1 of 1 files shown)


alalek (approved):
Looks good to me :+1:

alalek:
Could you please take a look on new build warnings from "Custom" (CUDA) builder?

PR #3105, opened 2021-11-11
fixed bad shape of markers (1x4) in several cases and added tests

AleksandrPanov:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/935

Fixes https://github.com/opencv/opencv/issues/14014
Problem:

- `detectMarkers()` and `refineDetectedMarkers()` returns `OutputArrayOfArrays corners` and `rejectedCorners` with different dimensions (for similar cases).
- This problem is only relevant for vector<Mat> input.
- In this case `detectMarkers()` **always** returns `corners` and `rejectedCorners` as `vector<Mat>`, with Mat dimension **4x1**.
- In this case `refineDetectedMarkers()` works **differently**, if it finds a new marker. Then the dimension of Mats inside the `vector<Mat>` **changes** from **4x1** to **1x4**.
- The reason for the problem is that `detectMarkers()` and `refineDetectedMarkers()` create `corners` and `rejectedCorners` with different ways.
- Current tests show no errors because they use `vector<vector<Point2f> >` for  `corners` and `rejectedCorners`. Also, the current tests only check the number of refined corners (dimension and value are not checked).

In this PR `refineDetectedMarkers()` for create `OutputArrayOfArrays corners` and `rejectedCorners` uses `_copyVector2Output()`, just like `detectMarkers()`.

`opencv_extra=fix_refineDetectedMarkers_shape`

### DIFF
--- a/modules/aruco/src/aruco.cpp
+++ b/modules/aruco/src/aruco.cpp
@@ -1486,7 +1486,7 @@ void refineDetectedMarkers(InputArray _image, const Ptr<Board> &_board,
     _convertToGrey(_image, grey);
 
     // vector of final detected marker corners and ids
-    vector< Mat > finalAcceptedCorners;
+    vector<vector<Point2f> > finalAcceptedCorners;
     vector< int > finalAcceptedIds;
     // fill with the current markers
     finalAcceptedCorners.resize(_detectedCorners.total());
@@ -1597,38 +1597,18 @@ void refineDetectedMarkers(InputArray _image, const Ptr<Board> &_board,
 
     // parse output
     if(finalAcceptedIds.size() != _detectedIds.total()) {
-        _detectedCorners.clear();
-        _detectedIds.clear();
-
         // parse output
         Mat(finalAcceptedIds).copyTo(_detectedIds);
-
-        _detectedCorners.create((int)finalAcceptedCorners.size(), 1, CV_32FC2);
-        for(unsigned int i = 0; i < finalAcceptedCorners.size(); i++) {
-            _detectedCorners.create(4, 1, CV_32FC2, i, true);
-            for(int j = 0; j < 4; j++) {
-                _detectedCorners.getMat(i).ptr< Point2f >()[j] =
-                    finalAcceptedCorners[i].ptr< Point2f >()[j];
-            }
-        }
+        _copyVector2Output(finalAcceptedCorners, _detectedCorners);
 
         // recalculate _rejectedCorners based on alreadyIdentified
-        vector< Mat > finalRejected;
+        vector<vector<Point2f> > finalRejected;
         for(unsigned int i = 0; i < alreadyIdentified.size(); i++) {
             if(!alreadyIdentified[i]) {
                 finalRejected.push_back(_rejectedCorners.getMat(i).clone());
             }
         }
-
-        _rejectedCorners.clear();
-        _rejectedCorners.create((int)finalRejected.size(), 1, CV_32FC2);
-        for(unsigned int i = 0; i < finalRejected.size(); i++) {
-            _rejectedCorners.create(4, 1, CV_32FC2, i, true);
-            for(int j = 0; j < 4; j++) {
-                _rejectedCorners.getMat(i).ptr< Point2f >()[j] =
-                    finalRejected[i].ptr< Point2f >()[j];
-            }
-        }
+        _copyVector2Output(finalRejected, _rejectedCorners);
 
         if(_recoveredIdxs.needed()) {
             Mat(recoveredIdxs).copyTo(_recoveredIdxs);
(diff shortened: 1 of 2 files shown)


modules/aruco/src/aruco.cpp
```diff
@@ -1563,7 +1563,7 @@ void refineDetectedMarkers(InputArray _image, const Ptr<Board> &_board,
 
         _detectedCorners.create((int)finalAcceptedCorners.size(), 1, CV_32FC2);
         for(unsigned int i = 0; i < finalAcceptedCorners.size(); i++) {
-            _detectedCorners.create(4, 1, CV_32FC2, i, true);
+            _detectedCorners.create(1, 4, CV_32FC2, i, true);
```
alalek, line 1566: Why?

Below we have 4,1 again: `_rejectedCorners.create(4, 1, ...)`

modules/aruco/test/test_charucodetection.cpp
```diff
@@ -677,4 +677,38 @@ TEST(Charuco, testCharucoCornersCollinear_false)
...
+
+    std::string testImagePath = cvtest::TS::ptr()->get_data_path() + "aruco/";
+    Mat img = imread(testImagePath + "nonrecover.png");
+    vector<vector<Point2f> > rejectedPoints;
+    vector<Mat> corners;
+    vector<int> ids;
+
+    aruco::detectMarkers(img, dico, corners, ids, detectorParams, rejectedPoints);
+    ASSERT_FALSE(corners.empty());
+    EXPECT_EQ(Size(4, 1), corners[0].size());
```
alalek, line 696: > `EXPECT_EQ(Size(4, 1), corners[0].size());`

Why do you think that `Size(1, 4)` is wrong?

modules/aruco/src/aruco.cpp
AleksandrPanov: It is surprising that the dimension of the matrices `out[i], _objPoints[i], _detectedCorners[i], _rejectedCorners[i], _diamondCorners[i]` is: (1 (rows) , 4 (cols)). But `create` with (4 (rows), 1 (cols)) was used.
why did m's size change?
Flag `allowTransposed` is ignored in this case (same values for true/false).
Looks like a bug in `void _OutputArray::create(int _rows, int _cols, int mtype, int i, bool allowTransposed, int fixedDepthMask) const`.

modules/aruco/test/test_charucodetection.cpp
AleksandrPanov: Size(1, 4) is not wrong (because it's just 4 2d-points: x1, y1, x2, y2, x3, y3, x4, y4) , but `corners[i]` must be the same size in all cases to avoid undefined behavior.=
I also think that `InputOutputArrayOfArrays` should not be used to pass array of pointers. You suggested replacing the `InputOutputArrayOfArrays` with a `InputOutputArray` with cv::Mat with Size(4, npoints) of CV_32FC2. I'll try to do that.

AleksandrPanov: Let me clarify:
In all tests and all examples in OpenCV detectMarkers return `corners` with `Size(4, 1)` (for `corners[i]`). It is strange if this condition is violated for no reason.

modules/aruco/src/aruco.cpp
alalek: Such changes affecting API are considered as a "breaking change".

So proper and complete change should have:
- clear justification
- consistent solution (fix 1 place only and left 6 others is not a good way)
- adapt coding rules to avoid such/similar problems in the future.

---

> surprising

There is `allowTransposed` parameter on this line.

AleksandrPanov: This strange code change is removed. Now `refineDetectedMarkers()` for create `OutputArrayOfArrays corners` and `rejectedCorners` uses `_copyVector2Output()`, as `detectMarkers()`.

It's strange that before `detectMarkers()` and `refineDetectedMarkers()` created `OutputArrayOfArrays corners` and `rejectedCorners` in different ways.

AleksandrPanov:
@alalek, could you recheck PR? I changed the description of the PR. Code changes made more clear. PR description has been improved.

modules/aruco/src/aruco.cpp
```diff
@@ -1597,38 +1597,18 @@ void refineDetectedMarkers(InputArray _image, const Ptr<Board> &_board,
...
-        _rejectedCorners.clear();
-        _rejectedCorners.create((int)finalRejected.size(), 1, CV_32FC2);
-        for(unsigned int i = 0; i < finalRejected.size(); i++) {
-            _rejectedCorners.create(4, 1, CV_32FC2, i, true);
-            for(int j = 0; j < 4; j++) {
-                _rejectedCorners.getMat(i).ptr< Point2f >()[j] =
-                    finalRejected[i].ptr< Point2f >()[j];
-            }
-        }
+        _copyVector2Output(finalRejected, _rejectedCorners);
```
AleksandrPanov, line 1611: Now, `refineDetectedMarkers()` uses `_copyVector2Output()` to copy, just as `detectMarkers()` does.

alalek (approved):
Well done :+1:

PR #3113, opened 2021-11-18
Update color_cvt.hpp, fix a potential overflow bug in Lab2BGR func

TracelessLe:
fix a potential overflow bug in cv::cuda::cvtColor(cv::COLOR_Lab2BGR) func, which need to clamp on B/G/R.

```
force_builders=Custom
Xbuild_image:Custom=ubuntu-cuda:18.04
```

### DIFF
--- a/modules/cudev/include/opencv2/cudev/functional/detail/color_cvt.hpp
+++ b/modules/cudev/include/opencv2/cudev/functional/detail/color_cvt.hpp
@@ -1083,9 +1083,9 @@ namespace color_cvt_detail
             else
                 Z = Z * Z * Z;
 
-            float B = 0.052891f * X - 0.204043f * Y + 1.151152f * Z;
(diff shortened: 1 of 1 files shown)


alalek:
Did you run CUDA tests locally?

TracelessLe:
> Did you run CUDA tests locally?

Sure, I did local tests in my project and using the "opencv/build/bin/opencv_test_cudaimgproc" executable file to verify both. This PR aims to fix the potential overflow bug in cv::cuda::cvtColor(cv::COLOR_Lab2BGR) function, which may cause some abnormal color if not clamp the float B/G/R value to range (0.0, 1.0).

The abnormal situation is such as attached below, where channel Green is overflow from float -0.0...1 (a very small negative number) to uint 255:




Some reference codes use cpu to do this color conversion are listed [here](https://github.com/opencv/opencv/blob/394e640909/modules/imgproc/src/opencl/color_lab.cl#L356).

alalek (approved):

PR #3121, opened 2021-11-23
rgbd: colored kinfu - allow fetching colors for pointcloud

paroj:

### DIFF
--- a/modules/rgbd/src/colored_kinfu.cpp
+++ b/modules/rgbd/src/colored_kinfu.cpp
@@ -144,7 +144,7 @@ public:
     void render(OutputArray image) const CV_OVERRIDE;
     void render(OutputArray image, const Matx44f& cameraPose) const CV_OVERRIDE;
 
-    virtual void getCloud(OutputArray points, OutputArray normals) const CV_OVERRIDE;
+    virtual void getCloud(OutputArray points, OutputArray normals, OutputArray colors) const CV_OVERRIDE;
     void getPoints(OutputArray points) const CV_OVERRIDE;
     void getNormals(InputArray points, OutputArray normals) const CV_OVERRIDE;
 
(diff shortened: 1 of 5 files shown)


modules/rgbd/src/colored_kinfu.cpp
```diff
@@ -344,9 +344,9 @@ void ColoredKinFuImpl<MatType>::render(OutputArray image, const Matx44f& _camera
 
 
 template< typename MatType >
-void ColoredKinFuImpl<MatType>::getCloud(OutputArray p, OutputArray n) const
+void ColoredKinFuImpl<MatType>::getCloud(OutputArray p, OutputArray n, OutputArray c) const
 {
-    volume->fetchPointsNormals(p, n);
+    dynamic_cast<ColoredTSDFVolume&>(*volume).fetchPointsNormalsColors(p, n, c);
```
DumDereDum, line 349: It will be better to add the function `fetchPointsNormalsColors` to Volume class.
Check [volume.hpp](https://github.com/opencv/opencv_contrib/blob/4.x/modules/rgbd/include/opencv2/rgbd/volume.hpp#L36-L39) and [colored_tsdf.cpp](https://github.com/opencv/opencv_contrib/blob/4.x/modules/rgbd/src/colored_tsdf.cpp#L73-L74) 
Shortly, add this function for all classes, and add `CV_Error` for not colored_tsdf classes

paroj: changed. Note, that the previous code had similar effect, as the dynamic_cast would throw a std::bad_cast if the volume was not of the desired type.

savuor (approved):
Looks good to me

DumDereDum (approved):
me too

alalek:
No way to automatically merge this to 5.x branch.
Dedicated PRs with porting this to 5.x branch is needed.

PR #3157, opened 2022-01-24
https://github.com/opencv/opencv_contrib/issues/1578

### LINKED ISSUE #1578: cv::xphoto::inpaint crashed in KDTreeComparator::operator()(int const&, int const&) const

- OpenCV => 3.4.1
- Operating System / Platform => iOS 11
inpaint-crash

danopdev:
resolves #1578

It's not THE SOLUTION because it does not fix the root cause (I don't know the algorithm) but it's a WORKAROUND to fix indexes out of bounds.
Using this changed I was able to do the inpaint (SHIFTMAP) on the example files.

output_shiftmap

- [ x ] I agree to contribute to the project under Apache 2 License.
- [ x ] To the best of my knowledge, the proposed patch is not based on a code under GPL or other license that is incompatible with OpenCV
- [ x ] The PR is proposed to proper branch
- [ x ] There is reference to original bug report and related work
- [ x ] There is accuracy test, performance test and test data in opencv_extra repository, if applicable
- [ x ] The feature is well documented and sample code can be built with the project CMake

### DIFF
--- a/modules/xphoto/src/inpainting.cpp
+++ b/modules/xphoto/src/inpainting.cpp
@@ -281,6 +281,9 @@ namespace xphoto
         /** Writing result **/
         for (size_t i = 0; i < labelSeq.size(); ++i)
         {
+            if (pPath[i].x >= img.cols || pPath[i].y >= img.rows)
+                continue;
+
             cv::Vec <float, cn> val = pointSeq[i][labelSeq[i]];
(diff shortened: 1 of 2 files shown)


doprea:
My bad. Fix tailing spaces.

alalek (approved):
LGTM :+1:

Issue #3163, opened 2022-01-28
opencv cuda filter lacks step(stride) information


yl070360:
##### System information (version)

- OpenCV => 4.2 - 4.5
- Operating System / Platform => linux
- Compiler => gcc/g++

##### Steps to reproduce

    Hi，I found that the processing of GPUMat in filtering.cpp(line:284) lacks step(stride) information, 
    And tried the following modifications, it seems to work normally

    filtering.cpp
    // C++ code example
    void LinearFilter::apply(InputArray _src, OutputArray _dst, Stream& _stream)
    {
        GpuMat src = _src.getGpuMat();
        ...
        Point ofs;
        Size wholeSize;
        src.locateROI(wholeSize, ofs);

        // GpuMat srcWhole(wholeSize, src.type(), src.datastart);
        // line:284, this is my modifications test
        GpuMat srcWhole(wholeSize, src.type(), src.datastart, src.step);
        

        func_(srcWhole, ofs.x, ofs.y, dst, kernel_.ptr<float>(),
              kernel_.cols, kernel_.rows, anchor_.x, anchor_.y,
              borderMode_, borderVal_.val, StreamAccessor::getStream(_stream));
    }

##### Issue submission checklist

 - [ ] I report the issue, it's not a question
   
 - [ ] I checked the problem with documentation, FAQ, open issues,
       forum.opencv.org, Stack Overflow, etc and have not found any solution
   
 - [x] I updated to the latest OpenCV version and the issue is still there
   
 - [ ] There is reproducer code and related data files: videos, images, onnx, etc

yl070360:
@cudawarped  can you help me

cudawarped:
> @cudawarped can you help me

Can you give me an example of where it previously failed?
[...]

PR #3167, opened 2022-02-02
Add missing stride in cuda::LinearFilter::apply - #3163

cudawarped:
This aplies the fix identified by @yl070360 in  https://github.com/opencv/opencv_contrib/issues/3163.

The stride information is currently missing inside `cuda::LinearFilter::apply()` producing incorrect results when `GpuMat`'s are constructed using pointers to pre-allocated device memory.

The current tests should to be sufficient to check this because they are wrongly passing at the moment due to another bug (see below) in the construction of `GpuMat`'s which can be fixed once this is merged.

**GpuMat Bug**
`GpuMat::dataend` wrongly points to the last element of an [alocated row ](https://github.com/opencv/opencv/blob/415a42f327104653604fc99314eb215cd938d6d7/modules/core/src/cuda/gpu_mat.cu#L191) instead of the last useable [element of the ROI](https://github.com/opencv/opencv/blob/415a42f327104653604fc99314eb215cd938d6d7/modules/core/src/cuda_gpu_mat.cpp#L100) unless the `GpuMat` is constructed from a pointer to pre-allocated device memory.
This means that `locateROI(wholeSize,ofs)` currently returns the stride in `wholeSize` instead of the width of the ROI.

```
force_builders=Custom
```

### DIFF
--- a/modules/cudafilters/src/filtering.cpp
+++ b/modules/cudafilters/src/filtering.cpp
@@ -281,7 +281,7 @@ namespace
         Size wholeSize;
         src.locateROI(wholeSize, ofs);
 
-        GpuMat srcWhole(wholeSize, src.type(), src.datastart);
+        GpuMat srcWhole(wholeSize, src.type(), src.datastart, src.step);
 
         func_(srcWhole, ofs.x, ofs.y, dst, kernel_.ptr<float>(),
               kernel_.cols, kernel_.rows, anchor_.x, anchor_.y,


alalek:
> Merge branch 'master' of ...

Could you please rebase on upstream/4.x to eliminate unrelated commits? (keep 1 commit in PR)

(where `upstream` is configured by following [this GitHub guide](https://help.github.com/articles/configuring-a-remote-for-a-fork/) and fetched (`git fetch upstream`))

asmorkalov (approved):

PR #3169, opened 2022-02-03
cmake(sfm): fix sfm static target dependencies

alalek:
replaces #3012

Error is triggered by `-Wl,--no-undefined` compiler option.


alalek:
:+1:

PR #3185, opened 2022-02-28
Use texture Object to make cv::cuda::HoughSegmentDetectorImpl::detect() thread-safe

### LINKED ISSUE #3184: cv::cuda::HoughSegmentDetector crashes when used in multiple threads
`cv::cuda::HoughSegmentDetector` crashes when used in multiple threads because it uses texture reference.

A texture reference is similar to global memory which is not thread-safe.

The current implementation of HoughSegmentDetector uses texture reference. See here : https://github.com/opencv/opencv_contrib/blob/a5b2490a3f20c9a8d86e8567176e8108e29daf98/modules/cudaimgproc/src/cuda/hough_segments.cu#L52

For `CV_CUDEV_ARCH > =300`, CUDA implementation should instead use texture objects. This already has been implemented in `opencv_contrib` but for a different algorithm. Specifically, cv::cuda::createCannyEdgeDetector in this commit : https://github.com/opencv/opencv_contrib/commit/6ca24c818f8266c3a7e554e7a583dcdf5dc2010f

I already have an implementation which solves this problem for `cv::cuda::HoughSegmentDetector()` based on `cv::cuda::createCannyEdgeDetector()` implementation.

I will submit a PR with this solution shortly.

- OpenCV => 4.5.5-dev
- Operating System / Platform => Ubuntu 18.04
- Compiler => gcc

##### Detailed description

See above.

##### Code to reproduce

This code launches 5 threads and each executes probabilistic hough segment detector on a sample image 1000 times.
When two threads are trying to access the texture reference (which behaves similar to global memory), the code crashes.

```.cpp
#include <iostream>
#include <opencv2/opencv.hpp>
#include "opencv2/cudaimgproc.hpp"
#include <pthread.h>

#define NUM_THREADS 5
#define NUM_IMG_PER_THREAD 1000

cv::Mat testImg;

void *findHough(void *threadid) {
    uint64_t threadNo = (uint64_t)threadid;

    cv::Ptr<cv::cuda::HoughSegmentDetector> hough = cv::cuda::createHoughSegmentDetector(1.0f, (float) (CV_PI / 180.0f), 50, 5);

    for(int i=0; i<NUM_IMG_PER_THREAD; i++) {
        cv::cuda::Stream stream;

        cv::cuda::GpuMat d_lines;
        cv::cuda::GpuMat srcImgGpu;

        cv::Mat houghLines;
        srcImgGpu.upload(testImg, stream);
        hough->detect(srcImgGpu, d_lines, stream);
        d_lines.download(houghLines, stream);
        stream.waitForCompletion();

        std::cout << "Thread, " << threadNo << ", index, " << i << ",Found #lines, " << houghLines.cols << std::endl;
    }

    return NULL;
}

int main()
{
    std::cout << "Test case for hough segmentation cuda crash..." << std::endl;

    std::cout << "Reading test image..." << std::endl;
    testImg = cv::imread("./sample_img/test_img_4k.jpg", cv::IMREAD_GRAYSCALE);

    if (testImg.empty()) {
        std::cout << "Test image empty." << std::endl;
        return -1;
    } else {
        std::cout << "Test image reading successful." << std::endl;
    }

    pthread_t threads[NUM_THREADS];

    int rc;
    uint64_t i;

    for( i = 0; i < NUM_THREADS; i++ ) {
      rc = pthread_create(&threads[i], NULL, findHough, (void *) i);
      
      if (rc) {
         std::cout << "Error:unable to create thread," << rc << std::endl;
         exit(-1);
      }
   }

   for (int i = 0; i < NUM_THREADS; i++) {
       pthread_join(threads[i], NULL);
   }
    
    return 0;
}
    
```

Here's the crash log.
```
root@df29f8235b3a:/src/testdata# ./hough_crash_reproduce 
Test case for hough segmentation cuda crash...
Reading test image...
Test image reading successful.
Thread, 4, index, 0,Found #lines, 4096
Thread, 3, index, 0,Found #lines, 4096
Thread, 1, index, 0,Found #lines, 4096
Thread, 0, index, 0,Found #lines, 4096
...
Thread, 4, index, 21,Found #lines, 4096
terminate called after throwing an instance of 'cv::Exception'
  what():  OpenCV(4.5.3) /opencv_contrib-4.5.3/modules/cudaimgproc/src/cuda/hough_segments.cu:234: error: (-217:Gpu API call) an illegal memory access was encountered in function 'houghLinesProbabilistic_gpu'
terminate called recursively

terminate called recursively
Aborted (core dumped)
```

I'm attaching the test image that was used to reproduce the crash. However, any image can be used to reproduce this bug.
test_img_4k

mitul93:
This PR fixes #3184 by using texture Object to make `cv::cuda::HoughSegmentDetectorImpl::detect()` thread-safe.

This PR also includes unit test for probabilistic hough segment detector. Previously, there was no unit test for this function.

Also note that, this solution is only thread-safe for `CV_CUDEV_ARCH>=300`. This is the same solution as existing code for canny algorithm that is already part of `opencv_contrib`. See here : https://github.com/opencv/opencv_contrib/commit/6ca24c818f8266c3a7e554e7a583dcdf5dc2010f

<cut/>

```
force_builders=Custom
```

### DIFF
--- a/modules/cudaimgproc/test/test_hough.cpp
+++ b/modules/cudaimgproc/test/test_hough.cpp
@@ -156,7 +156,7 @@ CUDA_TEST_P(HoughLinesProbabilistic, Accuracy)
     Ptr<cv::cuda::HoughSegmentDetector> hough = cv::cuda::createHoughSegmentDetector(rho, theta, minLineLength, maxLineGap);
 
     cv::cuda::GpuMat d_lines;
-    hough->detect(loadMat(src, false), d_lines);
+    hough->detect(loadMat(src, useRoi), d_lines);
 
     std::vector<cv::Vec4i> linesSegment;
     std::vector<cv::Vec2f> lines;
--- a/modules/cudaimgproc/test/test_hough.cpp
+++ b/modules/cudaimgproc/test/test_hough.cpp
@@ -114,6 +114,68 @@ INSTANTIATE_TEST_CASE_P(CUDA_ImgProc, HoughLines, testing::Combine(
     WHOLE_SUBMAT));
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////////
+// HoughLines Probabilistic
+PARAM_TEST_CASE(HoughLinesProbabilistic, cv::cuda::DeviceInfo, cv::Size, UseRoi)
+{
+    static void generateLines(cv::Mat& img)
+    {
+        img.setTo(cv::Scalar::all(0));
+
+        cv::line(img, cv::Point(10, 0), cv::Point(10, 20), cv::Scalar::all(255));
+        cv::line(img, cv::Point(20, 50), cv::Point(40, 50), cv::Scalar::all(255));
+    }
+
+    static void drawLines(cv::Mat& dst, const std::vector<cv::Vec4i>& linesSegment)
+    {
+        dst.setTo(cv::Scalar::all(0));
+
+        for (size_t i = 0; i < linesSegment.size(); ++i)
+        {
+            Vec4i l = linesSegment[i];
+            cv::line(dst, cv::Point(l[0], l[1]), cv::Point(l[2], l[3]), cv::Scalar::all(255));
+        }
+
+    }
+};
+
+CUDA_TEST_P(HoughLinesProbabilistic, Accuracy)
+{
+    const cv::cuda::DeviceInfo devInfo = GET_PARAM(0);
+    cv::cuda::setDevice(devInfo.deviceID());
+    const cv::Size size = GET_PARAM(1);
+    const bool useRoi = GET_PARAM(2);
+
+    const float rho = 1.0f;
+    const float theta = (float) (1.0 * CV_PI / 180.0);
+    const int minLineLength = 15;
+    const int maxLineGap = 8;
+
+    cv::Mat src(size, CV_8UC1);
+    generateLines(src);
+
+    Ptr<cv::cuda::HoughSegmentDetector> hough = cv::cuda::createHoughSegmentDetector(rho, theta, minLineLength, maxLineGap);
+
+    cv::cuda::GpuMat d_lines;
+    hough->detect(loadMat(src, false), d_lines);
+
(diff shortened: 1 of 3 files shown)


deadprogram:
The macOS test build that fails does not appear to be related to any modules in this change. Or am I missing something here?

deadprogram:
@alalek or any other maintainers, can you please help us get the CI passing? It is not clear how the change from this PR would impact the test from a different module, and then only on macOS.

asmorkalov (approved):
:+1:

asmorkalov:
I re-triggered CI build on Mac. Let's wait for status.

deadprogram:
Looks like CI now all passing thanks for re-triggering the build @asmorkalov

modules/cudaimgproc/test/test_hough.cpp
```diff
@@ -113,6 +113,68 @@ INSTANTIATE_TEST_CASE_P(CUDA_ImgProc, HoughLines, testing::Combine(
...
+
+    }
+};
+
+CUDA_TEST_P(HoughLinesProbabilistic, Accuracy)
+{
+    const cv::cuda::DeviceInfo devInfo = GET_PARAM(0);
+    cv::cuda::setDevice(devInfo.deviceID());
+    const cv::Size size = GET_PARAM(1);
+    const bool useRoi = GET_PARAM(2);
```
alalek, line 146: > useRoi

According to [build](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_custom_linux64/builds/468) report this variable is not used below (but it should as it is a test parameter).

deadprogram: I have just added commit https://github.com/opencv/opencv_contrib/pull/3185/commits/4d05fb56276a9aa6ce32cd1a421a57b4d69e91ce to correctly utilize the test parameter.

asmorkalov (approved):

deadprogram (approved):
I think all requested changes have now been made, and also all CI tests are now passing green.

Issue #2738, opened 2020-11-05
cv::aruco::detectMarkers consumes all available memory when CORNER_REFINE_CONTOUR


sebasth:
##### System information (version)

- OpenCV => 4.2.0 and 4.5.0
- Operating System / Platform => Ubuntu 18.04.5
- Compiler => GCC 7.5.0

##### Detailed description

`cv::aruco::detectMarkers` consumes all available memory when ` CORNER_REFINE_CONTOUR` is used and camera distortion parameter are provided. This does not happen if no distortion parameters are provided. This happens in both C++ and in Python. Example program below which demonstrates the issue.

I have tested this with with OpenCV versions 4.2.0 and 4.5.0.

##### Steps to reproduce

Example creates aruco marker and places it in the middle of the image. Dummy camera matrix (focal length is image width) is used, but the issue also present when real images calibrated camera values are used.

```.py
import cv2 as cv
import numpy as np

aruco_dict = cv.aruco.getPredefinedDictionary(cv.aruco.DICT_4X4_100)
# create 100x100 aruco marker in center of 1280x720 image
sz = 100
w, h = 1280, 720

marker = cv.aruco.drawMarker(aruco_dict, 0, sz)
im = np.zeros([h, w, 3], dtype=np.uint8)
im += 255
im[((h-sz)//2):((h+sz)//2), ((w-sz)//2):((w+sz)//2)] = marker[..., np.newaxis]

camera_matrix = np.identity(3, dtype=np.float)
camera_matrix[0, 0] = w
camera_matrix[1, 1] = w
camera_matrix[0, 2] = w/2
camera_matrix[1, 2] = h/2
dist_coeffs = None

print("camera matrix: \n", camera_matrix)
print("distortion coefficients: ", dist_coeffs)

params = cv.aruco.DetectorParameters_create()
params.cornerRefinementMethod = cv.aruco.CORNER_REFINE_CONTOUR

# works
corners, ids, _ = cv.aruco.detectMarkers(im, aruco_dict, None, None,
                                         params, None, camera_matrix, dist_coeffs)
print("corners:\n", corners)

print("\nwith distortion cofficients")

# consumes all available memory (OOM)
dist_coeffs = np.zeros([5], dtype=np.float)
print("distortion cofficients: ", dist_coeffs)
corners, ids, _ = cv.aruco.detectMarkers(im, aruco_dict, None, None,
                                         params, None, camera_matrix, dist_coeffs)
```
Output
```
[...]

PR #3186, opened 2022-02-28
fix CORNER REFINE CONTOUR

AleksandrPanov:
fix [cv::aruco::detectMarkers consumes all available memory when CORNER_REFINE_CONTOUR #2738](https://github.com/opencv/opencv_contrib/issues/2738)

The problem is that `undistortPoints()` is missing for `nCorners`. Therefore, the algorithm does not find any corner points:

```
	for ( unsigned int i =0; i < nContours.size(); i++ ) {
		for(unsigned int j=0; j<4; j++){
			if ( nCorners[j] == contour2f[i] ){ // nCorners[j] != contour2f[I] after `undistortPoints()`
				cornerIndex[j] = i;
				group=j;
			}
		}
		cntPts[group].push_back(contour2f[i]);
	}
```
The most logical solution is to make `undistort()` for the original image and look for ArUco markers on an undistorted image.

Also added a check that each of the four corners is found to avoid an infinite loop.

### DIFF
--- a/modules/aruco/include/opencv2/aruco.hpp
+++ b/modules/aruco/include/opencv2/aruco.hpp
@@ -206,21 +206,19 @@ struct CV_EXPORTS_W DetectorParameters {
  * @param parameters marker detection parameters
  * @param rejectedImgPoints contains the imgPoints of those squares whose inner code has not a
  * correct codification. Useful for debugging purposes.
- * @param cameraMatrix optional input 3x3 floating-point camera matrix
- * \f$A = \vecthreethree{f_x}{0}{c_x}{0}{f_y}{c_y}{0}{0}{1}\f$
- * @param distCoeff optional vector of distortion coefficients
- * \f$(k_1, k_2, p_1, p_2[, k_3[, k_4, k_5, k_6],[s_1, s_2, s_3, s_4]])\f$ of 4, 5, 8 or 12 elements
  *
  * Performs marker detection in the input image. Only markers included in the specific dictionary
  * are searched. For each detected marker, it returns the 2D position of its corner in the image
  * and its corresponding identifier.
  * Note that this function does not perform pose estimation.
- * @sa estimatePoseSingleMarkers,  estimatePoseBoard
+ * @note The function does not correct lens distortion or takes it into account. It's recommended to undistort
+ * input image with corresponging camera model, if camera parameters are known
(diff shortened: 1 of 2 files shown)


modules/aruco/src/aruco.cpp
```diff
@@ -953,10 +949,10 @@ static void _refineCandidateLines(std::vector<Point>& nContours, std::vector<Poi
 		}
 		cntPts[group].push_back(contour2f[i]);
 	}
-
+    for (int i = 0; i < 4; i++)
+        CV_Assert(cornerIndex[i] != -1);
```
AleksandrPanov, line 953: Added check that each of the four corners is found.

modules/aruco/src/aruco.cpp
```diff
@@ -953,10 +949,10 @@ static void _refineCandidateLines(std::vector<Point>& nContours, std::vector<Poi
...
 	// saves extra group into corresponding
 	if( !cntPts[4].empty() ){
-            CV_CheckLT(group, 4, "FIXIT: avoiding infinite loop: implementation should be revised: https://github.com/opencv/opencv_contrib/issues/2738");
```
AleksandrPanov, line 959: If each (or at least one) of the 4 corners is found, then `group` will not be equal to 4.

modules/aruco/include/opencv2/aruco.hpp
```diff
@@ -206,10 +206,6 @@ struct CV_EXPORTS_W DetectorParameters {
  * @param parameters marker detection parameters
  * @param rejectedImgPoints contains the imgPoints of those squares whose inner code has not a
  * correct codification. Useful for debugging purposes.
- * @param cameraMatrix optional input 3x3 floating-point camera matrix
- * \f$A = \vecthreethree{f_x}{0}{c_x}{0}{f_y}{c_y}{0}{0}{1}\f$
- * @param distCoeff optional vector of distortion coefficients
- * \f$(k_1, k_2, p_1, p_2[, k_3[, k_4, k_5, k_6],[s_1, s_2, s_3, s_4]])\f$ of 4, 5, 8 or 12 elements
```
asmorkalov, line 212: I propose to add reference to `undistortPoints` and `fisheye::undistortPoints` to `@sa`. And also add note like "The function does not correct lens distortion or takes it into account. It's recommended to undistort input image with corresponging camera model, if camera parameters are known."

alalek (approved):

PR #3198, opened 2022-03-16
cudacodec::VideoReader add colour format selection functionality

cudawarped:
As mentioned in https://github.com/opencv/opencv_contrib/issues/3195 the frames returned from cudacodec::VideoReader::nextFrame/retrieve are always BGRA with a zero alpha channel.  Regardless of the reason for this (likely memory optimization for previous generation gpu's and/or current OpenCV CUDA functions), GPU memory is at a premium and it seems like a reasonable idea to allow this to be changed to BGR and Gray if desired by the user.

It may impact performance of some CUDA routines if a BGR frame is used instead of BGRA but that would be up to the user to experiment with.

The default is left as BGRA for compatibility and the test has been updated.


alalek (approved):
Thank you :+1:

Issue #3192, opened 2022-03-07
Unable to detect Aruco marker under some circumstances


pla66:
##### System information (version)

- OpenCV => : 4.5.5
- Operating System / Platform => : Ubuntu 18.04.5
- Compiler => : CLion 2020.3

##### Detailed description

I have found that the function `cv::aruco::detectMarkers()` can not detect ArUco marker in some cases using latest OpenCV 4.5.5, but the function works well in older version of OpenCV (3.2.0). This happens only when the input image is blurred. If the input image has high resolution, both functions in 4.5.5 and 3.2.0 can detect it correctly.

So I read and compare the source codes in 4.5.5 and 3.2.0 and found some differences, I think it is maybe a bug, or maybe I was wrong if there are deeper reasons for this.

The source codes of function `cv::aruco::detectMarkers()` in 4.5.5 and 3.2.0 both has a pretreatment step of input image: `_filterTooCloseCandidates()`,  which aims to "check candidates that are too close to each other" because maybe two or more boxes are drawn near the same place and are almost coincident. Both codes in 3.2.0 and 4.5.5 marked these candidates and keep only one of them. However, usually there has one case that some of detected candidates are not close to each other, which should also be retained for subsequent identification. The codes in 3.2.0 keeps these candidates, but codes in 4.5.5 does not. That is to say, codes in 4.5.5 only keep candidates that detected twice or more, and remove candidates that detected only once, which may lead to missed detection.

I wonder whether the author deliberately designed it like this in latest version. I think whether a candidate is an ArUco can be identified in the following function `_identifyCandidates()` rather than delete it in function `_filterTooCloseCandidates()` in advance. So there is maybe a mistake.

##### Steps to reproduce

Use this image can reproduce the mistake. Parameters `DetectorParameters` are all set to default.
![0](https://user-images.githubusercontent.com/54338961/156989654-3f0f4eea-4794-4105-8729-5055f40f6d52.png)

Using OpenCV 4.5.5, the detect result like this:
![111](https://user-images.githubusercontent.com/54338961/156990154-b0719a16-158b-405e-879d-bb0f1c82406d.png)

Using OpenCV 3.2.0, the detect result like this:
![222](https://user-images.githubusercontent.com/54338961/156990819-ea264e8b-2be0-4d4b-ae59-03307ea0e70f.png)

[...]

Issue #2492, opened 2020-04-07
Aruco detection is worsened since OpenCV 4.1.2


stefanopini:
##### System information (version)

- OpenCV => 4.2.0.34 (opencv-contrib-python)
- Operating System / Platform => Windows 10 64bit / Python 3.6.4
- Compiler => Package installed with pip

##### Detailed description

The last version of the Aruco detection algorithm ( see pull request #2236 ), which is included in OpenCV 4.1.2 and above, has worse performance than the previous one when the outer border of black markers is small.  
The algorithm completely fails to find black markers whose outer white border is quite small (just few pixels) with both real and synthetic images. The old algorithm finds them.  
I tried to change some `DetectorParameters` (e.g. `minMarkerPerimeterRate`, `minCornerDistanceRate` and `markerBorderBits`) without success.

##### Steps to reproduce

Please see the attached reference Python code and image.
Code is tested with OpenCV 4.1.1.26 (8/8 markers found) and 4.2.0.34 (2/8 markers found) with Python 3.6.4. Corresponding results are reported.

[aruco-detection-issue.zip](https://github.com/opencv/opencv_contrib/files/4446970/aruco-detection-issue.zip)

##### Issue submission checklist

 - [x] I report the issue, it's not a question
   
 - [x] I checked the problem with documentation, FAQ, open issues,
       answers.opencv.org, Stack Overflow, etc and have not found solution
   
 - [x] I updated to latest OpenCV version (available through PyPI) and the issue is still there
   
 - [x] There is reproducer code and related data files: videos, images, onnx, etc

paroj:
@szk1509

szk1509:
Hi @stefanopini, you can reduce the minimum marker distance rate (in this case to 0.025), which allows to recognize all the markers at the bottom of the image provided, i.e. the synthetic markers.

stefanopini:
Hi @szk1509 , thank you for spotting it.
However, it seems to be just a workaround: the issue doesn't seem related to the distance between markers (it seems to be caused by the small/missing outer white border only from OpenCV 4.1.2) and it is still unable to detect half of the markers.

I looked again at your pull request (#2236 improving the accuracy with white markers), but I didn't find what may cause the misdetection of markers without external white border.
[...]

PR #3201, opened 2022-03-18
fix search for one contour in _filterTooCloseCandidates()

AleksandrPanov:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/965

This PR fixes:
- Unable to detect Aruco marker under some circumstances #3192 
- Aruco detection is worsened since OpenCV 4.1.2 #2492

The problem was in `_filterTooCloseCandidates()`. Fixed and added regression_3192, regression_2492 tests.

`opencv_extra=fix_filterTooCloseCandidates`

### DIFF
--- a/modules/aruco/test/test_arucodetection.cpp
+++ b/modules/aruco/test/test_arucodetection.cpp
@@ -644,4 +644,35 @@ TEST(CV_ArucoTutorial, can_find_gboriginal)
     }
 }
 
+TEST(CV_ArucoDetectMarkers, regression_3192)
+{
+    Ptr<aruco::Dictionary> dictionary = aruco::getPredefinedDictionary(aruco::DICT_4X4_50);
+    Ptr<aruco::DetectorParameters> detectorParams = aruco::DetectorParameters::create();
+    vector< int > markerIds;
+    vector<vector<Point2f> > markerCorners;
+    string imgPath = cvtest::findDataFile("aruco/regression_3192.png");
+    Mat image = imread(imgPath);
+    const size_t N = 2ull;
+    const int goldCorners[N][8] = { {345,120, 520,120, 520,295, 345,295}, {101,114, 270,112, 276,287, 101,287} };
+    const int goldCornersIds[N] = { 6, 4 };
+    map<int, const int*> mapGoldCorners;
+    for (size_t i = 0; i < N; i++)
+        mapGoldCorners[goldCornersIds[i]] = goldCorners[i];
+
+    aruco::detectMarkers(image, dictionary, markerCorners, markerIds, detectorParams);
+
+    ASSERT_EQ(N, markerIds.size());
+    for (size_t i = 0; i < N; i++)
+    {
+        int arucoId = markerIds[i];
+        ASSERT_EQ(4ull, markerCorners[i].size());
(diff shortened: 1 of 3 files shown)


modules/aruco/test/test_arucodetection.cpp
```diff
@@ -644,4 +644,36 @@ TEST(CV_ArucoTutorial, can_find_gboriginal)
...
 }
 
+TEST(CV_ArucoDetectMarkers, regression_3192)
+{
+    Ptr<aruco::Dictionary> dictionary = aruco::getPredefinedDictionary(aruco::DICT_4X4_50);
+    Ptr<aruco::DetectorParameters> detectorParams = aruco::DetectorParameters::create();
+    vector< int > markerIds;
+    vector<vector<Point2f> > markerCorners;
+    string imgPath = cvtest::findDataFile("aruco/regression_3192.png", true);
+    Mat image = imread(imgPath);
```
asmorkalov, line 654: Please add assertion, that image is not empty.

modules/aruco/test/test_arucodetection.cpp
```diff
@@ -644,4 +644,36 @@ TEST(CV_ArucoTutorial, can_find_gboriginal)
...
+    const int goldCorners[N][8] = { {345,120, 520,120, 520,295, 345,295}, {101,114, 270,112, 276,287, 101,287} };
+    const int goldCornersIds[N] = { 6, 4 };
+
+    aruco::detectMarkers(image, dictionary, markerCorners, markerIds, detectorParams);
+    ASSERT_EQ(N, markerIds.size());
+    map<int, const int*> mapGoldCorners;
+    for (size_t i = 0; i < N; i++)
+        mapGoldCorners[goldCornersIds[i]] = goldCorners[i];
+
+    ASSERT_EQ(N, markerIds.size());
```
asmorkalov, line 665: Duplicate.

modules/aruco/test/test_charucodetection.cpp
```diff
@@ -745,15 +745,16 @@ TEST(Charuco, issue_14014)
     ASSERT_EQ(corners.size(), 19ull);
     EXPECT_EQ(Size(4, 1), corners[0].size()); // check dimension of detected corners
 
-    ASSERT_EQ(rejectedPoints.size(), 21ull);
+    size_t numRejPoints = rejectedPoints.size();
+    ASSERT_TRUE(numRejPoints > 0ull);
```
asmorkalov, line 749: I recommend to stay with the original assertion, but change value.

asmorkalov (changes requested):

modules/aruco/test/test_charucodetection.cpp
AleksandrPanov: I think, that the exact number of rejected candidates should not be verified.

asmorkalov: Strong checks allow to highlight regressions. You can change particular constants each time when it's needed.

AleksandrPanov: check was added

asmorkalov (approved):

asmorkalov:
@alalek your turn.

PR #3207, opened 2022-03-19
replaced recursion DFS to stack DFS

aimbot6120:

    [ Y] I agree to contribute to the project under Apache 2 License.
    [ Y] To the best of my knowledge, the proposed patch is not based on a code under GPL or another license that is incompatible with OpenCV
    [ Y] The PR is proposed to the proper branch
    [ N] There is a reference to the original bug report and related work
    [ N] There is accuracy test, performance test and test data in opencv_extra repository, if applicable
    [ N] The feature is well documented and sample code can be built with the project CMake

The function `DFSUtil` in `opencv/opencv_contrib/modules/text/sample/text_detection_swt.cpp` implements Depth First Search (DFS) using recursion. This implementation uses the call stack of the program for recursion which makes it prone to stack overflow due to very deep recursion. This error was encountered for some images from my testset. The error was found through Valgrind.

Fix:
Implemented the `DFSUtil` function using external stack. This implementation was tested using the same images and the program is now working as intended.


modules/text/src/text_detector_swt.cpp
```diff
@@ -92,14 +93,25 @@ static
 void DFSUtil(int v, std::vector<bool> & visited, std::vector< std::vector<int> >& adj, int label, std::vector<int> &component_id)
 {
     // Mark the current node as visited and label it as belonging to the current component
```
alalek, line 95: This comment is related to removed/moved code.

aimbot6120: moved the comments

alalek (approved):

PR #3209, opened 2022-03-22
Set higher minimum number of decode surfaces when creating cv::cudacodec::VideoReader

cudawarped:
`cv::cudacodec::VideoReader` uses the minimum number of decode surfaces for correct decoding.  From the NVIDIA documentation:

> This guarantees correct functionality and optimal video memory usage but not necessarily the best performance, which depends on the design of the overall application. The optimal number of decode surfaces (in terms of performance and memory utilization) should be decided by experimentation for each application, but it cannot go below ..

Therefore this pull request allows the user to specify a higher minimum number of decode surfaces when creating `cv::cudacodec::VideoReader` to increase the decoding performace at the cost of GPU memory, if required.

<cut/>

```
force_builders=Custom
```


alalek (approved):

PR #3220, opened 2022-04-01
Fix infinite loop on ArUco apriltag refinement

buq2:
Software entered infinite loop when image height
was smaller than 10*cv::getNumThreads(). With high
core count machines this could happen with very
reasonable image sizes.

Fix is to ensure that chunksize is at least 1.

The bug can be triggered with following test program:

```
#include <opencv2/aruco.hpp>
#include <opencv2/core/utility.hpp>
#include <iostream>
#include <opencv2/imgcodecs.hpp>

int main(int argc, char *argv[]) {
    // Bug will happen even with 1 thread, but then the image
    // will be too small to detect corners.
    // cv::setNumThreads(1);
    const auto num_th = cv::getNumThreads();
    std::cout << "Number of threads: " << num_th << std::endl;

    // If 'freeze' is set to true, bug is triggered
    constexpr bool freeze = true;
    std::cout << "Should freeze: " << freeze << std::endl;

    // When triggering the bug, use image size that is
    // 1 pix smaller than 10*number of threads
    const auto height_img = freeze ? num_th*10-1 : num_th*10;
    std::cout << "Image height: " << height_img << std::endl;

    // Just to get nice white boarder
    const int shift = height_img > 10 ? 5 : 1;
    const auto height_marker = height_img-2*shift;

    // Create marker template
    cv::Mat img_marker;
    cv::Ptr<cv::aruco::Dictionary> dictionary = cv::aruco::getPredefinedDictionary(cv::aruco::DICT_4X4_50);
    cv::aruco::drawMarker(dictionary, 23, height_marker, img_marker, 1);

    // Copy to bigger image
    cv::Mat img(height_img, height_img, CV_8UC1);
    img.setTo(255);
    img_marker.copyTo(img(cv::Rect(shift, shift, height_marker, height_marker)));

    // Just for visualizing
    cv::imwrite("marker.png", img);

    // Bug only in apriltag refinement
    auto params = cv::aruco::DetectorParameters::create();
    params->cornerRefinementMethod = cv::aruco::CORNER_REFINE_APRILTAG;

    std::vector<std::vector<cv::Point2f>> corners;
    std::vector<int> ids;

    // If freeze=true, this will never return
    cv::aruco::detectMarkers(img, dictionary, corners, ids, params);

    // Ensure we get reasonable corners 
    // (not with this default test, white boarder needs to be bigger)
    for (const auto &marker : corners) {
        for (const auto &p : marker) {
            std::cout << p << std::endl;
        }
    }
}
```

### DIFF
--- a/modules/aruco/test/test_arucodetection.cpp
+++ b/modules/aruco/test/test_arucodetection.cpp
@@ -759,7 +759,7 @@ TEST_P(ArucoThreading, number_of_threads_does_not_change_results)
 
     params->cornerRefinementMethod = GetParam();
 
-    std::vector<std::vector<cv::Point2f>> original_corners;
+    std::vector<std::vector<cv::Point2f> > original_corners;
     std::vector<int> original_ids;
     {
         NumThreadsSetter thread_num_setter(1);
@@ -774,7 +774,7 @@ TEST_P(ArucoThreading, number_of_threads_does_not_change_results)
     for (size_t i_num_threads = 0; i_num_threads < sizeof(num_threads_to_test)/sizeof(int); ++i_num_threads) {
         NumThreadsSetter thread_num_setter(num_threads_to_test[i_num_threads]);
 
-        std::vector<std::vector<cv::Point2f>> corners;
+        std::vector<std::vector<cv::Point2f> > corners;
         std::vector<int> ids;
         cv::aruco::detectMarkers(img, dictionary, corners, ids, params);
 
--- a/modules/aruco/test/test_arucodetection.cpp
+++ b/modules/aruco/test/test_arucodetection.cpp
@@ -717,4 +717,93 @@ TEST(CV_ArucoDetectMarkers, regression_2492)
     }
 }
 
+TEST(CV_ArucoDetectMarkers, number_of_threads_does_not_change_results)
+{
+    struct NumThreadsSetter {
+        NumThreadsSetter(const int num_threads)
+        : original_num_threads_(cv::getNumThreads()) {
+            cv::setNumThreads(num_threads);
+        }
+
+        ~NumThreadsSetter() {
+            cv::setNumThreads(original_num_threads_);
+        }
+     private:
+        int original_num_threads_;
+    };
+    cv::Ptr<cv::aruco::DetectorParameters> params = cv::aruco::DetectorParameters::create();
+    // We are not testing against different dictionaries
+    // As we are interested mostly in small images, smaller
+    // markers is better -> 4x4
+    cv::Ptr<cv::aruco::Dictionary> dictionary = cv::aruco::getPredefinedDictionary(cv::aruco::DICT_4X4_50);
+
+    // Height of the test image can be chosen quite freely
+    // We aim to test against small images as in those the
+    // number of threads has most effect
+    const int height_img = 20;
+    // Just to get nice white boarder
+    const int shift = height_img > 10 ? 5 : 1;
+    const int height_marker = height_img-2*shift;
+
+    // Create a test image
+    cv::Mat img_marker;
+    cv::aruco::drawMarker(dictionary, 23, height_marker, img_marker, 1);
+
+    // Copy to bigger image to get a white border
+    cv::Mat img(height_img, height_img, CV_8UC1);
+    img.setTo(255);
+    img_marker.copyTo(img(cv::Rect(shift, shift, height_marker, height_marker)));
+
+    // Number of threads to be tested. OpenCV uses C++03 so no list-initialization
+    std::vector<int> num_threads_to_test;
+    num_threads_to_test.push_back(1);
+    num_threads_to_test.push_back(2);
+    num_threads_to_test.push_back(8);
+    num_threads_to_test.push_back(16);
+    num_threads_to_test.push_back(32);
+    num_threads_to_test.push_back(height_img-1);
+    num_threads_to_test.push_back(height_img);
+    num_threads_to_test.push_back(height_img+1);
+
+    std::vector<cv::aruco::CornerRefineMethod> methods_to_test;
+    methods_to_test.push_back(cv::aruco::CORNER_REFINE_NONE);
+    methods_to_test.push_back(cv::aruco::CORNER_REFINE_SUBPIX);
+    methods_to_test.push_back(cv::aruco::CORNER_REFINE_CONTOUR);
+    methods_to_test.push_back(cv::aruco::CORNER_REFINE_APRILTAG);
+
+    for (size_t i_method = 0; i_method < methods_to_test.size(); ++i_method) {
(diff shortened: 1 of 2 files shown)


asmorkalov:
@buq2 Thanks for the contribution. The patch looks reasonable for me. Could you add your reproducer with `setNumThreads` to the PR as test?

buq2:
@asmorkalov Do you mean adding the main function code from the initial PR description to (for example) https://github.com/opencv/opencv_contrib/blob/3.4/modules/aruco/test/test_arucodetection.cpp ? Sure! The "test" is not a very good test case as if it fails, it hangs indefinitely, but maybe it is better than nothing. I'll add it soonish and make it such that it will test also the other refinement methods with different initial number of threads with small aruco images.

modules/aruco/test/test_arucodetection.cpp
```diff
@@ -717,4 +717,71 @@ TEST(CV_ArucoDetectMarkers, regression_2492)
...
+                        EXPECT_NEAR(corners[i][j].x, original_corners[i][j].x, 0.1f);
+                        EXPECT_NEAR(corners[i][j].y, original_corners[i][j].y, 0.1f);
+                    }
+                }
+            }
+        }
+    }
+
+    // Reset number of threads
+    cv::setNumThreads(original_num_th);
```
asmorkalov, line 784: It does not happen, if at least one of asserts is raised. I propose to use holder class for it like this:
```
struct ThreadNumHolder
{
    explicit ThreadNumHolder(int threads):
        prevThreads(cv::getNumThreads())
    {
        cv::setNumThreads(threads);
    }
    ~ThreadNumHolder()
    {
        cv::setNumThreads(prevThreads);
    }
    int prevThreads;
};
```
And create local variable of `ThreadNumHolder` type in loop. It's destructor restores original number threads anyway.

modules/aruco/test/test_arucodetection.cpp
```diff
@@ -717,4 +717,71 @@ TEST(CV_ArucoDetectMarkers, regression_2492)
...
+    // Remember original number of threads
+    const auto original_num_th = cv::getNumThreads();
+    for (const auto refinement_method : {cv::aruco::CORNER_REFINE_NONE, cv::aruco::CORNER_REFINE_SUBPIX,
+        cv::aruco::CORNER_REFINE_CONTOUR, cv::aruco::CORNER_REFINE_APRILTAG})
+    {
+        params->cornerRefinementMethod = refinement_method;
+
+        std::vector<std::vector<cv::Point2f>> original_corners;
+        std::vector<int> original_ids;
+        for (const auto num_threads : {1, 2, 8, 16, 32, height_img-1, height_img, height_img+1}) {
```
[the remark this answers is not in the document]
buq2: Good point. Fixed

modules/aruco/test/test_arucodetection.cpp
```diff
@@ -717,4 +717,71 @@ TEST(CV_ArucoDetectMarkers, regression_2492)
...
+
+    // Copy to bigger image to get a white border
+    cv::Mat img(height_img, height_img, CV_8UC1);
+    img.setTo(255);
+    img_marker.copyTo(img(cv::Rect(shift, shift, height_marker, height_marker)));
+
+    // Remember original number of threads
+    const auto original_num_th = cv::getNumThreads();
+    for (const auto refinement_method : {cv::aruco::CORNER_REFINE_NONE, cv::aruco::CORNER_REFINE_SUBPIX,
+        cv::aruco::CORNER_REFINE_CONTOUR, cv::aruco::CORNER_REFINE_APRILTAG})
```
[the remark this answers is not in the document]
buq2: Not quite sure what you mean by `parametrized`. Not 100% sure but I guess catch generators use C++11 features

modules/aruco/test/test_arucodetection.cpp
```diff
@@ -717,4 +717,71 @@ TEST(CV_ArucoDetectMarkers, regression_2492)
...
+            std::vector<int> ids;
+            cv::aruco::detectMarkers(img, dictionary, corners, ids, params);
+
+            // If we don't find any markers, the test is broken
+            ASSERT_GE(ids.size(), 1);
+
+            if (original_corners.empty()) {
+                original_corners = std::move(corners);
+                original_ids = std::move(ids);
+            } else {
```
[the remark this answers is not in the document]
buq2: Now the first iterations is single threaded. This PR is not about absolute locations of the corners, but that result does not change based on number of threads. If absolute locations of the markers should be checked, I feel it be done by new PR.

modules/aruco/test/test_arucodetection.cpp
buq2: Good point. Fixed

asmorkalov:
@buq2 Thanks for the test. I refactored it a bit, to make it parameterized (obvious test report in case of failure) and extracted the iteration with `num_threads=1` as ground truth to make it obvious for reader.

asmorkalov (approved):

AleksandrPanov (approved):
LGTM, thx for fix and tests

buq2:
Thanks for the reviews! Next one will be less painful for everyone :)

asmorkalov:
@alalek I propose to merge the patch.

modules/aruco/test/test_arucodetection.cpp
```diff
@@ -717,4 +717,91 @@ TEST(CV_ArucoDetectMarkers, regression_2492)
...
+
+    std::vector<std::vector<cv::Point2f> > original_corners;
+    std::vector<int> original_ids;
+    {
+        NumThreadsSetter thread_num_setter(1);
+        cv::aruco::detectMarkers(img, dictionary, original_corners, original_ids, params);
+    }
+
+    ASSERT_EQ(original_ids.size(), 1);
+    ASSERT_EQ(original_corners.size(), 1);
```
AleksandrPanov, line 770: warning: comparison between signed and unsigned integer expressions

sorry, I overlooked

modules/aruco/test/test_arucodetection.cpp
```diff
@@ -717,4 +717,91 @@ TEST(CV_ArucoDetectMarkers, regression_2492)
...
+
+    for (size_t i_num_threads = 0; i_num_threads < sizeof(num_threads_to_test)/sizeof(int); ++i_num_threads) {
+        NumThreadsSetter thread_num_setter(num_threads_to_test[i_num_threads]);
+
+        std::vector<std::vector<cv::Point2f> > corners;
+        std::vector<int> ids;
+        cv::aruco::detectMarkers(img, dictionary, corners, ids, params);
+
+        // If we don't find any markers, the test is broken
+        ASSERT_EQ(ids.size(), 1);
```
AleksandrPanov, line 782: warning: comparison between signed and unsigned integer expressions

sorry, I overlooked

PR #3241, opened 2022-04-29
Workflow for Github Actions to build and test OpenCV on Linux for 3.4

asenyaev:
This PR adds a workflow for GitHub Actions to build and test OpenCV Contrib on ubuntu-20.04 for 3.4 branch.

### DIFF
--- a/.github/workflows/PR-3.4-U20.yaml
+++ b/.github/workflows/PR-3.4-U20.yaml
@@ -1,0 +1,214 @@
+name: PR:3.4 U20
+
+on:
+  pull_request:
+    branches:
+      - 3.4
+
+env:
+  EXTRA_CMAKE_OPTIONS: '-DBUILD_DOCS=ON -DPYTHON_DEFAULT_EXECUTABLE=/usr/bin/python3 -DBUILD_EXAMPLES=ON -DOPENCV_ENABLE_NONFREE=ON -DENABLE_CCACHE=OFF'
+  OPENCV_TEST_DATA_PATH: '/opencv_extra/testdata'
+  OPENCV_CONTRIB_DOCKER_WORKDIR: '/__w/opencv_contrib/opencv_contrib'
+  PR_AUTHOR: ${{ github.event.pull_request.user.login }}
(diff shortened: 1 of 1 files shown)


asmorkalov (approved):
:+1:

.github/workflows/PR-3.4-U20.yaml
```diff
@@ -0,0 +1,214 @@
...
+        fetch-depth: 0
+    - name: Merge opencv_contrib with ${{ env.SOURCE_BRANCH_NAME }} branch
+      run: |
+        cd ${{ env.OPENCV_CONTRIB_DOCKER_WORKDIR }}
+        git config --global --add safe.directory ${{ env.OPENCV_CONTRIB_DOCKER_WORKDIR }}
+        git config user.email "opencv.ci"
+        git config user.name "opencv.ci"
+        git pull -v "https://github.com/${{ env.PR_AUTHOR_FORK }}" "${{ env.SOURCE_BRANCH_NAME }}"
+    - name: Clone opencv
+      run: git clone --single-branch --branch ${{ env.TARGET_BRANCH_NAME }} https://github.com/opencv/opencv.git /opencv
```
fengyuentau, line 50: one quick question: do you find any difference between actions/checkout and git-clone?

asenyaev: Yes, only one difference - the path where actions/checkout locates a fetched repository can be only in `${{ github.workspace }}`, doesn't matter it's inside container or not, it can't fetch in a container on the path `/opencv` (opened issue exists for it). In addition, I would like to add git cache for workflows with self-hosted runners and clone repositories with `--reference` flag to reduce the network usage.

fengyuentau: Thanks for the explain. I thought about reducing network usage by mirroring git repos but was stuck on keeping the mirror update-to-date. Good to learn there is `--reference` serving same purposes but is easier than setting up a mirror.

fengyuentau (approved):

PR #3247, opened 2022-05-06
Add RTSP features to cudacodec::VideoReader

cudawarped:
Add the below two features to `cudacodec::VideoReader` to help when streaming from live sources.
1) Lift the limit on the maximum number of unparsed packets if streaming from a source over UDP, see https://github.com/opencv/opencv_contrib/issues/3225#issuecomment-1100206377
2) Internally drop frames if the rate which frames are requested by `nextFrame()/grab()` is less than the source FPS.  Although this is something which a user would eventually accomodate for themselves (by reading at the appropriate rate and choosing which packets to discard) it would be a useful for this to be "automatic" until they have implemented that functionality.

Testing - currently the additional tests only check that the new parameters have been set inside `VideoReader` and do not verify the functionality of (1) and (2).  This is because testing (1) requires a new test video file (I am not sure if that is overkill for this small feature) and testing (2) require a live RTSP source.  I can easily include an extra test file if required but I am not sure how I can simulate an live RTSP source?

### DIFF
--- a/modules/cudacodec/include/opencv2/cudacodec.hpp
+++ b/modules/cudacodec/include/opencv2/cudacodec.hpp
@@ -484,7 +484,7 @@ struct CV_EXPORTS_W_SIMPLE VideoReaderInitParams {
     CV_WRAP VideoReaderInitParams() : udpSource(false), liveSource(false), minNumDecodeSurfaces(0), rawMode(0) {};
     CV_PROP_RW bool udpSource;
     CV_PROP_RW bool liveSource;
-    CV_PROP_RW bool minNumDecodeSurfaces;
+    CV_PROP_RW int minNumDecodeSurfaces;
     CV_PROP_RW bool rawMode;
 };
 
--- a/modules/cudacodec/include/opencv2/cudacodec.hpp
+++ b/modules/cudacodec/include/opencv2/cudacodec.hpp
@@ -322,7 +322,7 @@ enum class VideoReaderProps {
     PROP_LRF_HAS_KEY_FRAME = 5, //!< FFmpeg source only - Indicates whether the Last Raw Frame (LRF), output from VideoReader::retrieve() when VideoReader is initialized in raw mode, contains encoded data for a key frame.
     PROP_COLOR_FORMAT = 6, //!< Set the ColorFormat of the decoded frame.  This can be changed before every call to nextFrame() and retrieve().
     PROP_UDP_SOURCE = 7, //!< Status of VideoReaderInitParams::udpSource initialization.
-    PROP_LIVE_SOURCE = 8, //!< Status of VideoReaderInitParams::liveSource initialization.
+    PROP_ALLOW_FRAME_DROP = 8, //!< Status of VideoReaderInitParams::allowFrameDrop initialization.
 #ifndef CV_DOXYGEN
     PROP_NOT_SUPPORTED
 #endif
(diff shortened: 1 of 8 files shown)


modules/cudacodec/src/frame_queue.hpp
```diff
@@ -63,7 +63,7 @@ class FrameQueue
 {
 public:
     ~FrameQueue();
-    void init(const int _maxSz);
+    void init(const int _maxSz, const bool force = false);
```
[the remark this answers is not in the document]
cudawarped: I agree thanks asmorkalov.

modules/cudacodec/test/test_video.cpp
```diff
@@ -70,6 +70,10 @@ PARAM_TEST_CASE(CheckDecodeSurfaces, cv::cuda::DeviceInfo, std::string)
 {
 };
 
+PARAM_TEST_CASE(CheckInitParams, cv::cuda::DeviceInfo, std::string, bool, bool, bool)
+{
+};
```
asmorkalov, line 75: The test is incomplete.

modules/cudacodec/include/opencv2/cudacodec.hpp
```diff
@@ -463,32 +465,42 @@ class CV_EXPORTS_W RawVideoSource
...
+@param minNumDecodeSurfaces Minimum number of internal decode surfaces used by the hardware decoder.  NVDEC will automatically determine the minimum number of
+surfaces it requires for correct functionality and optimal video memory usage but not necessarily for best performance, which depends on the design of the
+overall application. The optimal number of decode surfaces (in terms of performance and memory utilization) should be decided by experimentation for each application,
+but it cannot go below the number determined by NVDEC.
+@param rawMode Allow the raw encoded data which has been read up until the last call to grab() to be retrieved by calling retrieve(rawData,RAW_DATA_IDX).
+*/
+struct CV_EXPORTS_W_SIMPLE VideoReaderInitParams {
+    CV_WRAP VideoReaderInitParams() : udpSource(false), liveSource(false), minNumDecodeSurfaces(0), rawMode(0) {};
+    CV_PROP_RW bool udpSource;
+    CV_PROP_RW bool liveSource;
```
asmorkalov, line 481: I propose to have `allowFrameDrop` everywhere as variable name and point that it's designed for RTSP sreaming. The new name defines concrete functionality, but not abstract mode.

modules/cudacodec/src/frame_queue.cpp
```diff
@@ -55,18 +55,26 @@ cv::cudacodec::detail::FrameQueue::~FrameQueue() {
         delete[] isFrameInUse_;
 }
 
-void cv::cudacodec::detail::FrameQueue::init(const int _maxSz) {
+void cv::cudacodec::detail::FrameQueue::init(const int _maxSz, const bool force) {
     AutoLock autoLock(mtx_);
+    if (isFrameInUse_) {
+        if (force)
+            delete[] isFrameInUse_;
```
asmorkalov, line 62: Am I right that it cases memory leak, if `isFrameInUse_` defines alive frames?

cudawarped: Yes even though it is an internal class which is only called once it was vunerable to a memory leak if init() is called more than once.

modules/cudacodec/test/test_video.cpp
```diff
@@ -344,6 +351,22 @@ CUDA_TEST_P(CheckDecodeSurfaces, Reader)
         for (int i = 0; i < 100; i++) ASSERT_TRUE(reader->grab());
     }
 }
+
+CUDA_TEST_P(CheckInitParams, Reader)
```
cudawarped, line 355: The test is implemented here.  As mentioned in the description it currently only checks that the new parameters have been set.  To test the functionality an rstp streamer would be required.

modules/cudacodec/test/test_video.cpp
cudawarped: The test is implemented on line 355. As mentioned in the description it currently only checks that the new parameters have been set. To test the functionality an rstp streamer would be required.

modules/cudacodec/src/frame_queue.cpp
asmorkalov: Correct me, if I miss something, but the path and the module itself does not use `init` with `force=true`. I propose to remove it from `init` as it's dead code and dangerous flag.

asmorkalov (commented):
Looks good to me in general, besides `force` flag in init. I propose to remove it at all.
Tested on Ubuntu 18.04 with CUDA 10.2, NVIDIA Video Codec SDK 11.1.

modules/cudacodec/src/frame_queue.cpp
cudawarped: No you are correct.  Force is not used and can be removed because it could be dangerous to have the facility to initialize a frameque object twice.  It may be better if required in the future to have a specific reinitialization routine which performs safety checks, if the size of the que needs to be altered mid-decode.

cudawarped:
No you are correct. Force is not used and can be removed because it could be dangerous to have the facility to initialize a frameque object twice. It may be better if required in the future to have a specific reinitialization routine which performs safety checks, if the size of the que needs to be altered mid-decode.

asmorkalov (approved):

PR #3248, opened 2022-05-08
Force VideoReader to decode all YUV video formats to NV12

cudawarped:
Currently `VideoReader` only supports 8 bit YUV 422/420 video's.  This fix allows all supported YUV inputs 10 bit 420, and 8 and 16 bit 444 to be decoded.  The approach taken is to truncate all YUV formats to to NV12 instead of including CUDA kernels to convert from P016, YUV444 and YUV444_16Bit to BGR, which is probably overkill for OpenCV and can be included at a later date if desired.

The PR also allows the raw YUV frame to be output as well if desired.

### DIFF
--- a/modules/cudacodec/src/video_reader.cpp
+++ b/modules/cudacodec/src/video_reader.cpp
@@ -74,6 +74,9 @@ void videoDecPostProcessFrame(const GpuMat& decodedFrame, GpuMat& outFrame, int 
         outFrame.create(height, width, CV_8UC1);
         cudaMemcpy2DAsync(outFrame.ptr(), outFrame.step, decodedFrame.ptr(), decodedFrame.step, width, height, cudaMemcpyDeviceToDevice, stream);
     }
+    else if (colorFormat == ColorFormat::YUV) {
+        decodedFrame.copyTo(outFrame);
+    }
 }
 
 using namespace cv::cudacodec::detail;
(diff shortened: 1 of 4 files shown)


modules/cudacodec/src/video_reader.cpp
```diff
@@ -74,6 +74,9 @@ void videoDecPostProcessFrame(const GpuMat& decodedFrame, GpuMat& outFrame, int
         outFrame.create(height, width, CV_8UC1);
         cudaMemcpy2DAsync(outFrame.ptr(), outFrame.step, decodedFrame.ptr(), decodedFrame.step, width, height, cudaMemcpyDeviceToDevice, stream);
     }
+    else if (colorFormat == ColorFormat::YUV) {
+        decodedFrame.copyTo(outFrame);
```
asmorkalov, line 78: It should be done with stream too, I think.

cudawarped: Indeed it should, thanks.

@asmorkalov are CUDA builds/tests not working at the moment?

asmorkalov: Yes. We have issues with build farm right now, and have to rework it significantly.

asmorkalov:
I get build error:
```
/mnt/projects/Projects/OpenCV/opencv_contrib/modules/cudacodec/src/video_reader.cpp: At global scope:
/mnt/projects/Projects/OpenCV/opencv_contrib/modules/cudacodec/src/video_reader.cpp:102:14: error: extra qualification ‘{anonymous}::VideoReaderImpl::’ on member ‘set’ [-fpermissive]
         void VideoReaderImpl::set(const ColorFormat _colorFormat) CV_OVERRIDE;
```
System: Ubuntu 18.04, Video Codec SDK 11.1, Cuda 10.2.

cudawarped:
> I get build error:

Apologies, I thought I had changed that.  Fixed on both PR's.

asmorkalov (approved):
:+1: LGTM.
Tested on Ubuntu 18.04, CUDA 102, NVIDIA Video Codec SDK 11.1.

PR #3265, opened 2022-05-25
Update video_reader.cpp

TraubSimon:
Tried to build opencv with cmake and got that error:

openCV/opencv_contrib/modules/cudacodec/src/video_reader.cpp:99:14: error: extra qualification ‘{anonymous}::VideoReaderImpl::’ on member ‘set’ [-fpermissive]
   99 |         void VideoReaderImpl::set(const ColorFormat _colorFormat) CV_OVERRIDE;
      |              ^~~~~~~~~~~~~~~

Would propose do delete VideoReaderImpl:: in line 99


alalek (approved):

PR #3277, opened 2022-06-12
Add BAD descriptor to xfeatures2d module

iago-suarez:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/978

Box Average Difference (BAD) is a new descriptor [published last year](https://doi.org/10.1109/LRA.2021.3107024). The method improves efficient binary descriptors such as ORB or BEBLID with a new loss function and optimization method. It outperforms the other binary efficient descriptors in several tasks like image matching, structure from motion, or patch retrieval. It is also compact, being able to outperform BEBLID-512 with only 256 bits. Here we can see a comparison in HPatches dataset from the paper: 

image

Since the implementation is quite similar to the one of BEBLID, the same source code can be used. It is only necessary to add the new pre-trained weights and some minor modifications. The original BAD source can be found [here](https://github.com/iago-suarez/efficient-descriptors). More details in the [paper](https://arxiv.org/pdf/2108.08380.pdf) and the [project page](https://iago-suarez.com/efficient-descriptors/).

I tried to make the code clear, tested, and documented. If you find some possible improvements, please just let me know.

```
opencv_extra=4.x
force_builders=linux,docs
```


iago-suarez:
A missing file with pre-computed bad descriptors caused the tests to fail. I added it in [a opencv_extra fork](https://github.com/iago-suarez/opencv_extra/blob/4.x/testdata/cv/features2d/descriptor_extractors/descriptor-bad) and made the necessary [pull request](https://github.com/opencv/opencv_extra/pull/978). 

It should be fixed now. I would really appreciate it if you could rerun the GitHub actions.

alalek (commented):
Thank you for the great contribution!

modules/xfeatures2d/src/beblid.cpp
```diff
@@ -28,7 +28,8 @@ namespace xfeatures2d
 // Struct containing the 6 parameters that define an Average Box weak-learner
 struct ABWLParams
 {
-    int x1, y1, x2, y2, boxRadius, th;
+    int x1, y1, x2, y2, boxRadius;
+    float th;
```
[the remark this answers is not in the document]
iago-suarez: Thank you very much @alalek for your feedback. 

You are right. It is a better idea to preserve the original structure for BEBLID. In my last commit, I added a new ABWLParamsFloatTh for the new descriptor and undo the changes to the original ABWLParams.

I also made BEBLID_Impl a templated class that can be used with both structs.

modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
```diff
@@ -224,6 +224,48 @@ class CV_EXPORTS_W BEBLID : public Feature2D
...
+BAD obtains 621 (75.2%) with 256 bits. ORB obtains only 493 inliers (63%).
+
+If you find this code useful, please add a reference to the following paper:
+<BLOCKQUOTE> Iago Suárez, José M. Buenaposada, and Luis Baumela.
+Revisiting Binary Local Image Description for Resource Limited Devices.
+IEEE Robotics and Automation Letters, vol. 6, no. 4, pp. 8317-8324, Oct. 2021. </BLOCKQUOTE>
+
+The descriptor was trained in Liberty split of the UBC datasets \cite winder2007learning .
+*/
+class CV_EXPORTS_W BAD : public Feature2D
```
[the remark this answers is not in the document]
iago-suarez: Done.

After discussing with the other authors, we think that a better name would be TEBLID (Triplet-based Efficient Binary Local Image Descriptor), which makes it clear that it is an improvement over BEBLID and the name is better suited for googling. In any case, in the documentation, we indicate that this descriptor appears as BAD in the paper.

modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
```diff
@@ -224,6 +224,49 @@ class CV_EXPORTS_W BEBLID : public Feature2D
...
+/** @brief Class implementing TEBLID (Triplet-based Efficient Binary Local Image Descriptor),
+ * described in @cite Suarez2021TEBLID.
+
+TEBLID stands for Triplet-based Efficient Binary Local Image Descriptor, although originally it was called BAD
+\cite Suarez2021TEBLID. It is an improvement over BEBLID \cite Suarez2020BEBLID, that uses triplet loss,
+hard negative mining, and anchor swap to improve the image matching results.
+It is able to describe keypoints from any detector just by changing the scale_factor parameter.
+TEBLID is as efficient as ORB, BEBLID or BRISK, but the triplet-based training objective selected more
+discriminative features that explain the accuracy gain. It is also more compact than BEBLID,
+when running the [AKAZE example](https://raw.githubusercontent.com/opencv/opencv/master/samples/cpp/tutorial_code/features2D/AKAZE_match.cpp)
```
alalek, line 236: Could you please use this link? https://github.com/opencv/opencv/blob/4.x/samples/cpp/tutorial_code/features2D/AKAZE_match.cpp

- master -> 4.x
- use GitHub UI instead of raw code (some browsers open "download" UI)

modules/xfeatures2d/src/beblid.cpp
```diff
@@ -281,27 +327,54 @@ void BEBLID_Impl::compute(InputArray _image, vector<KeyPoint> &keypoints, Output
...
-    if (n_bits == SIZE_512_BITS)
-        wl_params_.assign(wl_params_512, wl_params_512 + sizeof(wl_params_512) / sizeof(wl_params_512[0]));
-    else if(n_bits == SIZE_256_BITS)
-        wl_params_.assign(wl_params_256, wl_params_256 + sizeof(wl_params_256) / sizeof(wl_params_256[0]));
+    WeakLearnerT * begin_ptr, * end_ptr;
+    if (n_bits == BEBLID::SIZE_512_BITS)
+    {
+        #include "beblid.p512.hpp"
+        begin_ptr = (WeakLearnerT *) std::begin(beblid_wl_params_512);
+        end_ptr = (WeakLearnerT *) std::end(beblid_wl_params_512);
```
alalek, line 343: > `(WeakLearnerT *)`

Using `WeakLearnerT = ABWLParamsFloatTh` and `BEBLID::SIZE_512_BITS)` is not a valid case and we should emit error.

Probably it makes sense to specialize ctor and/or add `wl_params` parameter.

```
template <class WeakLearnerT>
BEBLID_Impl<WeakLearnerT>::BEBLID_Impl(float scale_factor, int n_bits, std::vector<WeakLearnerT>&& wl_params)
```

and move parameters vector initialization on upper (caller) level (`::create()` calls).

---

Also these changes would not require special values like 102,103 for enum values.

alalek (commented):
Thank you for the update!

modules/xfeatures2d/include/opencv2/xfeatures2d.hpp
iago-suarez: Done!

modules/xfeatures2d/src/teblid.p256.hpp
```diff
@@ -0,0 +1,102 @@
...
+//     Iago Suarez, Jose M. Buenaposada, and Luis Baumela.
+//     Revisiting Binary Local Image Description for Resource Limited Devices.
+//     IEEE Robotics and Automation Letters, vol. 6, no. 4, pp. 8317-8324, Oct. 2021.
+
+// ABWLParams: x1, y1, x2, y2, boxRadius, th
+
+// Pre-trained parameters of TEBLID-256 trained in Liberty data set. 10K triplets are sampled per iteration. Each triplet
+// contains an anchor patch, a positive and a negative, selected as the hardest among 256 random negatives.
+static const std::vector<ABWLParamsFloatTh> teblid_wl_params_256 = {
+    {25, 14, 13, 15, 6, 21.65}, {16, 15, 14, 11, 1, 5.65}, {14, 14, 7, 8, 6, 4.95},
```
alalek, line 16: Could you please add `f` suffixes to floating-point numbers?

```.diff
-21.65
+21.65f
```

It is necessary to resolve build problem on Windows:

> c:\build\precommit-contrib_windows64\4.x\opencv_contrib\modules\xfeatures2d\src\teblid.p512.hpp(187): error C2397: conversion from 'double' to 'float' requires a narrowing conversion (compiling source file C:\build\precommit-contrib_windows64\4.x\opencv_contrib\modules\xfeatures2d\src\beblid.cpp)
```

File: modules/xfeatures2d/src/teblid.p256.hpp [hunk 6]
```diff
@@ -0,0 +1,102 @@
...
+// Implementation of the article:
+//     Iago Suarez, Jose M. Buenaposada, and Luis Baumela.
+//     Revisiting Binary Local Image Description for Resource Limited Devices.
+//     IEEE Robotics and Automation Letters, vol. 6, no. 4, pp. 8317-8324, Oct. 2021.
+
+// ABWLParams: x1, y1, x2, y2, boxRadius, th
+
+// Pre-trained parameters of TEBLID-256 trained in Liberty data set. 10K triplets are sampled per iteration. Each triplet
+// contains an anchor patch, a positive and a negative, selected as the hardest among 256 random negatives.
+static const std::vector<ABWLParamsFloatTh> teblid_wl_params_256 = {
```
alalek, line 15: > std::vector<ABWLParamsFloatTh>

Need to check generated binary code of this initializer. Array should be fine, but `std::vector` - doesn't.

Could you please use something like this:

```
static const ABWLParamsFloatTh teblid_wl_params_256_[] = {...};
static const std::vector<ABWLParamsFloatTh> teblid_wl_params_256(teblid_wl_params_256_, teblid_wl_params_256 + sizeof(teblid_wl_params_256_) / sizeof(teblid_wl_params_256_[0]));
```

iago-suarez: Sorry for late reply due to vacations and thanks again for your comments.

I have fixed it in my last commit.

modules/xfeatures2d/src/teblid.p256.hpp
iago-suarez: Done!

alalek (approved):

iago-suarez:
Thank you @alalek for your support and for maintaining OpenCV! It's a wonderful library, and I am always happy to contribute :smiley:

PR #3278, opened 2022-06-13
SetParameterization -> SetManifold to make SFM build with Ceres 2.1

yznovyak:
`SetParameterization` doesn't work with Ceres 2.1, it fails with an error message
```LocalParameterizations are deprecated. Use SetManifold instead```

- [+] I agree to contribute to the project under Apache 2 License.
- [+] To the best of my knowledge, the proposed patch is not based on a code under GPL or another license that is incompatible with OpenCV
- [+] The PR is proposed to the proper branch
- [-] There is a reference to the original bug report and related work
- [-] There is accuracy test, performance test and test data in opencv_extra repository, if applicable
- [+] The feature is well documented and sample code can be built with the project CMake


alalek:
Hello, Thank you for the fix!
We already have this fix on 3.4 branch here: #3274. Changes from 3.4 will be propagated to 4.x branch on this week during the regular merge (https://github.com/opencv/opencv/wiki/Branches)

PR #3294, opened 2022-06-27
Update test_hdf5.cpp

AaronWebster:
Uses a unique filename (via std::tmpnam) for test.  Allows unit test to be run in environments where the current working directory is not writable.

- [NA] There is accuracy test, performance test and test data in opencv_extra repository, if applicable

### DIFF
--- a/modules/hdf/test/test_hdf5.cpp
+++ b/modules/hdf/test/test_hdf5.cpp
@@ -7,6 +7,8 @@
  * @author Fangjun Kuang <csukuangfj dot at gmail dot com>
  * @date December 2017
  */
+#include <cstdio>
+
 #include "test_precomp.hpp"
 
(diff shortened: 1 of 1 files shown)


modules/hdf/test/test_hdf5.cpp
```diff
@@ -15,7 +17,7 @@ struct HDF5_Test : public testing::Test
 {
     virtual void SetUp()
     {
-        m_filename = "test.h5";
+        m_filename = std::tmpnam(nullptr) + "_test.h5";
```
asmorkalov, line 20: It's recommended to use `cv::tempfile` for tests: https://docs.opencv.org/4.x/db/de0/group__core__utils.html#ga9fcedbe6d0104d7ce012987478b060d4

PR #3323, opened 2022-08-09
Use T as temp var type in polyanticlockwise()

y-guyon:
To match the input data type.
Cast inner computation to double to prevent int overflows for any input type.

- [ ] There is a reference to the original bug report and related work (the input image triggering the overflow is not publicly available)

### DIFF
--- a/modules/mcc/src/common.hpp
+++ b/modules/mcc/src/common.hpp
@@ -79,10 +79,10 @@ void polyanticlockwise(std::vector<T> &points)
     // Sort the points in anti-clockwise order
     // Trace a line between the first and second point.
     // If the third point is at the right side, then the points are anti-clockwise
-    cv::Point v1 = points[1] - points[0];
-    cv::Point v2 = points[2] - points[0];
+    T v1 = points[1] - points[0];
+    T v2 = points[2] - points[0];
 
-    double o = (v1.x * v2.y) - (v1.y * v2.x);
+    double o = ((double)v1.x * v2.y) - ((double)v1.y * v2.x);
 
     if (o < 0.0) //if the third point is in the left side, then sort in anti-clockwise order
         std::swap(points[1], points[3]);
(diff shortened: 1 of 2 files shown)


modules/mcc/src/common.hpp
```diff
@@ -79,10 +79,10 @@ void polyanticlockwise(std::vector<T> &points)
     // Sort the points in anti-clockwise order
     // Trace a line between the first and second point.
     // If the third point is at the right side, then the points are anti-clockwise
-    cv::Point v1 = points[1] - points[0];
-    cv::Point v2 = points[2] - points[0];
+    T v1 = points[1] - points[0];
+    T v2 = points[2] - points[0];
 
-    double o = (v1.x * v2.y) - (v1.y * v2.x);
+    double o = ((double)v1.x * v2.y) - ((double)v1.y * v2.x);
```
asmorkalov, line 85: Why do you force the values to double. It could be done after computation to exclude compiler warnings. Also you can replace condition from `o < 0` to `(v1.x * v2.y) >= (v1.y * v2.x)` to not wary about expression return type.

modules/mcc/src/common.hpp
```diff
@@ -93,10 +93,10 @@ void polyclockwise(std::vector<T> &points)
     // Sort the points in clockwise order
     // Trace a line between the first and second point.
     // If the third point is at the right side, then the points are clockwise
-    cv::Point v1 = points[1] - points[0];
-    cv::Point v2 = points[2] - points[0];
+    T v1 = points[1] - points[0];
+    T v2 = points[2] - points[0];
 
-    double o = (v1.x * v2.y) - (v1.y * v2.x);
+    double o = ((double)v1.x * v2.y) - ((double)v1.y * v2.x);
```
asmorkalov, line 99: The same here.

modules/mcc/src/common.hpp
y-guyon: My reasoning was: `T` could be an integer type so `v1.x * v2.y` may overflow, so it is safer to cast to `double` earlier, since the result is stored in a `double o` anyway.

I looked at `polyanticlockwise()` and `polyclockwise()` uses and `T` is always `cv::Point2f`. I suggest we drop the `template` and `o`. What do you think?

asmorkalov: good idea

asmorkalov:
@y-guyon please take a look on CI. The patch breaks build.

y-guyon:
> @y-guyon please take a look on CI. The patch breaks build.

Yes thanks, I received the GitHub notification by email.
I moved the definitions to the .cc file now that they are not templated anymore.

asmorkalov (approved):
:+1:

PR #3324, opened 2022-08-10
Fix CUDA mem leak in GFTT and move CUDA malloc out of critical path

kevinchristensen1:
Fixes https://github.com/opencv/opencv_contrib/issues/3042, also mentioned in https://github.com/opencv/opencv_contrib/pull/2868
Also added a fix for https://github.com/opencv/opencv_contrib/issues/3194
Fixes cuda mem leak in `GoodFeaturesToTrackDetector` and moves cuda malloc/free into the ctor/dtor, out of critical path

### DIFF
--- a/modules/cudaimgproc/src/cuda/gftt.cu
+++ b/modules/cudaimgproc/src/cuda/gftt.cu
@@ -87,25 +87,22 @@ namespace cv { namespace cuda { namespace device
             }
         }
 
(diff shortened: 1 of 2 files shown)


asmorkalov (approved):
:+1: LGTM.

kevinchristensen1:
Thanks for reviewing Alex! It looks like the macOS-ARM64 CI build is failing due to a timeout issue unrelated to this patch. Is there anything else I need to do?

PR #3328, opened 2022-08-17
Fix array out of range when no edges found

fdivitto:
When input image is almost black edges.numOfEdges could be zero. In this case lines.sId size is zero and "detect lines" loop is never executed, so this array doesn't get chances to increase (see commit https://github.com/opencv/opencv_contrib/commit/7e14b4bba1766641975c88ca8a0bbf51511461f4).

Just before return the pointer to the beginning of lines.sId (pLineSID) is dereferenced at index 0, causing the AV because size of array is zero and pLineSID is invalid.

This fix simply make sure lines.sId contains at least one item.

### DIFF
--- a/modules/line_descriptor/src/binary_descriptor.cpp
+++ b/modules/line_descriptor/src/binary_descriptor.cpp
@@ -2250,7 +2250,7 @@ int BinaryDescriptor::EDLineDetector::EDline( cv::Mat &image, LineChains &lines 
   unsigned int linePixelID = edges.sId[edges.numOfEdges];
   lines.xCors.resize( linePixelID );
   lines.yCors.resize( linePixelID );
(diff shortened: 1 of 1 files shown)


asmorkalov:
If I understand the function correctly the best solution is to add early exit if `edges.numOfEdges == 0` like this:
```
if (edges.numOfEdges == 0)
{
  lines.numOfLines = 0;
  return 1; // need to check if it's expected return value if nothing found
}
```
It's obvious corner case handler for code reader, no questions about vectors size.

fdivitto:
As you prefer. I like a code with a few jumps or early exits from functions. "edges.numOfEdges" is implicitly checked in the main loop.

asmorkalov:
@fdivitto friendly reminder.

asmorkalov:
@fdivitto friendly reminder.

PR #3336, opened 2022-08-24
xphoto fix inpaint fsr

### LINKED ISSUE #2855: Bug in inpaint_fsr function for 16 bits / 32 bits / 64 bits images
- OpenCV => 4.5
- Operating System / Platform => Ubuntu 18.04
- Compiler => GCC
- 
##### Detailed description

When calling inpaint function from xphoto module with INPAINT_FSR_(BEST/FAST) parameter, there is an OpenCV exception if the provided src image is different than CV_8U.

This behavior is due to the call to : "src.convertTo(src, CV_8U, 1/257.0)" whereas src is a const reference argument.

##### Steps to reproduce

```cpp
#include <opencv2/xphoto/inpainting.hpp>

int main()
{
    cv::Mat input_16U_img = cv::Mat::zeros(cv::Size(500,500), CV_16U);
    cv::Mat mask = cv::Mat::zeros(input_16U_img.size(), CV_8U);
    cv::Mat out_img;
    cv::xphoto::inpaint(input_16U_img, mask, out_img, cv::xphoto::INPAINT_FSR_FAST);

    return 0;
}
```

berak:
resolves #2855

use a temporal Mat header for convertTo()

- [x] The PR is proposed to the proper branch ? ? 
(this goes to the 4.x branch, there is no fsr inpainting in 3.4)

### DIFF
--- a/modules/xphoto/src/inpainting_fsr.impl.hpp
+++ b/modules/xphoto/src/inpainting_fsr.impl.hpp
@@ -748,7 +748,7 @@ icvDetermineProcessingOrder(
 
 
 static
-void inpaint_fsr(const Mat &src, const Mat &mask, Mat &dst, const int algorithmType)
+void inpaint_fsr(Mat src, const Mat &mask, Mat &dst, const int algorithmType)
 {
     CV_Assert(algorithmType == xphoto::INPAINT_FSR_BEST || algorithmType == xphoto::INPAINT_FSR_FAST);
     CV_Check(src.channels(), src.channels() == 1 || src.channels() == 3, "");
@@ -767,7 +767,7 @@ void inpaint_fsr(const Mat &src, const Mat &mask, Mat &dst, const int algorithmT
                 CV_Error(Error::StsUnsupportedFormat, "Unsupported source image format!");
                 break;
             }
-            src.convertTo(src, CV_8U, 1/257.0);
+            src.convertTo(src, CV_8U, 1/256.0);
             break;
         }
         case CV_32FC1:


modules/xphoto/src/inpainting_fsr.impl.hpp
```diff
@@ -748,26 +748,28 @@ icvDetermineProcessingOrder(
...
             double minRange, maxRange;
-            minMaxLoc(src, &minRange, &maxRange);
+            minMaxLoc(src_, &minRange, &maxRange);
             if (minRange < 0 || maxRange > 65535)
             {
                 CV_Error(Error::StsUnsupportedFormat, "Unsupported source image format!");
                 break;
             }
-            src.convertTo(src, CV_8U, 1/257.0);
+            src_.convertTo(src, CV_8U, 1/257.0);
```
asmorkalov, line 772: 257 looks very strange here.

modules/xphoto/src/inpainting_fsr.impl.hpp
```diff
@@ -748,26 +748,28 @@ icvDetermineProcessingOrder(
 
 
 static
-void inpaint_fsr(const Mat &src, const Mat &mask, Mat &dst, const int algorithmType)
+void inpaint_fsr(const Mat &src_, const Mat &mask, Mat &dst, const int algorithmType)
```
asmorkalov, line 751: I propose to pass the matrix by value. cv::Mat is smart pointer, so no large overhead. It solves all potential issues in the pipeline.

modules/xphoto/src/inpainting_fsr.impl.hpp
```diff
@@ -767,7 +767,7 @@ void inpaint_fsr(const Mat &src, const Mat &mask, Mat &dst, const int algorithmT
                 CV_Error(Error::StsUnsupportedFormat, "Unsupported source image format!");
                 break;
             }
-            src.convertTo(src, CV_8U, 1/257.0);
+            src.convertTo(src, CV_8U, 1/255.0);
```
asmorkalov, line 770: In case if you want to get only  higher 8 bits of value, the multiplier should be 1/256.

asmorkalov (approved):

Issue #3330, opened 2022-08-20
NPP performance bottleneck with multiple streams


awarebayes:
##### System information (version)

- OpenCV => 4.6
- Operating System / Platform => Linux x64 with Nvidia Cuda 11.6
- Compiler => gcc 

##### Detailed description

Hello! I am working on a pipeline with multiple stream executing steps.
It looks something like that and flips every second image in the pipeline.
Everything works fine, however I have noticed a huge performance bottleneck when dealing with nvidia's npp.
For example in warpAffine, it takes ~20 times longer than your own implementation with WarpDispatcher::call.

Here is my pipeline

```
    cv::Mat trans_mat = get_affine_transform(cs, m_pixel_std, m_input_size);
    cv::cuda::warpAffine(
            *input.image,
            m_warped_resized_image,
            trans_mat,
            m_input_size,
            cv::INTER_LINEAR,
            cv::BORDER_CONSTANT,
            cv::Scalar(),
            m_stream
            );

    if (m_flip_images && m_index % 2 == 1)
    {
        cv::cuda::flip(m_warped_resized_image, m_flipped_image, 1, m_stream);
        m_flipped_image.convertTo(m_float_resized_image, CV_32FC3, 1.f, m_stream);
    }
    else
        m_warped_resized_image.convertTo(m_float_resized_image, CV_32FC3, 1.f, m_stream);
    cv::cuda::subtract(m_float_resized_image, cv::Scalar(127.5f, 127.5f, 127.5f),
                       m_float_resized_image, cv::noArray(), -1, m_stream);
    cv::cuda::divide(m_float_resized_image, cv::Scalar(128.0f, 128.0f, 128.0f),
                     m_float_resized_image, 1, -1, m_stream);
    cv::cuda::split(m_float_resized_image, m_chw, m_stream);
```

When I do no changes to your library, perf output looks something like that:

<img width="1111" alt="image" src="https://user-images.githubusercontent.com/42784580/185755151-e8c1ad63-d810-4643-bc94-812595d87eca.png">

as you can see, cudaWarpAffine is really performance hungry because of npp.

When i disable the npp, simply by setting [this if statement](https://github.com/opencv/opencv_contrib/blob/4.x/modules/cudawarping/src/warp.cpp#L247) to false, I see drastic performance improvement, like that:

<img width="1128" alt="image" src="https://user-images.githubusercontent.com/42784580/185754296-c0078771-d979-457c-9d2d-01f8cd178494.png">

[...]

PR #3338, opened 2022-08-24
disable npp in multistream context

awarebayes:

- [* ] I agree to contribute to the project under Apache 2 License.
- [* ] To the best of my knowledge, the proposed patch is not based on a code under GPL or another license that is incompatible with OpenCV
- [* ] The PR is proposed to the proper branch
- [* ] There is a reference to the original bug report and related work
- [*] There is accuracy test, performance test and test data in opencv_extra repository, if applicable
- [*] The feature is well documented and sample code can be built with the project CMake

This PR is to match #3330

The issue is simple to reproduce a code from [this popular tutorial](https://developer.ridgerun.com/wiki/index.php/How_to_use_OpenCV_CUDA_Streams) can be used, just replace cuda::resize with cuda::warpAffine and some random matrix to accompany.

To my best knowledge is does not make sense to use npp with multistream context and simple warp implementation works just fine.

You can see using nppSetStream uses a lot of ioctl as was referenced in the issue and thus this function should only be used with the default context. 

Alternatively there can be a static compilation warning in the NppWarp class, parametrized for calling it with StreamNull using static dispatch. I can probably try pulling that off, it may result in messier code, but the warning will be compile time.

### DIFF
--- a/modules/cudawarping/src/warp.cpp
+++ b/modules/cudawarping/src/warp.cpp
@@ -242,8 +242,7 @@ void cv::cuda::warpAffine(InputArray _src, OutputArray _dst, InputArray _M, Size
 
     bool useNpp = borderMode == BORDER_CONSTANT && ofs.x == 0 && ofs.y == 0 && useNppTab[src.depth()][src.channels() - 1][interpolation];
     // NPP bug on float data
-    useNpp = useNpp && src.depth() != CV_32F;
-
+    useNpp = useNpp && src.depth() != CV_32F && StreamAccessor::getStream(stream) == nullptr;
     if (useNpp)
     {
         typedef void (*func_t)(const cv::cuda::GpuMat& src, cv::cuda::GpuMat& dst, double coeffs[][3], int flags, cudaStream_t stream);


cudawarped:
@awarebayes Whilst I am no fan of NPP, I believe that they should offer the best performance going forward on newer architectures than the OpenCV hand rolled code which was optimized for hardware from 10 years ago.  Now if all the OpenCV CUDA routines could be updated for new hardware and CUDA SDK features when they are released then I would change my mind but OpenCV is open source and the CUDA modules are in the contrib repository so this is not realistic.

With that out of the way would it not be better to replace the existing global stream versions (e.g. [nppiWarpAffine_8u_C1R ](https://docs.nvidia.com/cuda/npp/group__affine__transform.html#gad379bb77066de4c8292dc8f20540910e))
with the newer ones which accept and NppStreamContext (e.g. [nppiWarpAffine_8u_C1R_ctx ](https://docs.nvidia.com/cuda/npp/group__affine__transform.html#ga58076b2d7d7c5cf8b0eefd7cec67cd06))?  I ran a quick test on a 2000x2000 matrix and the performance of the npp and OpenCV hand rolled version are about the same.  A futher improvement might be to have an additional flag which lets the user decide if they want to use the NPP routines or not.

awarebayes:
@cudawarped Should I be the one  to do it? I.E. replace with the contexted version

cudawarped:
@awarebayes Its your pull request, I'm mearly suggesting an alternative approach which in my opinion may be better, ultimately its up to the maintainers to decide.  If you do apply the suggested changes make sure to include the #define to only update for CUDA toolkit >= 10.1 see the [release notes ](https://docs.nvidia.com/cuda/archive/10.1/cuda-toolkit-release-notes/index.html#npp-new-features).

First I would check that the changes I suggested are as quick as your change for your specific problem.

awarebayes:
I did a compile time static dispatch for either _Ctx or non-_Ctx methods.

Tested, it performs not as good as opencv's because it uses cv::cuda::Mat::setTo to fill image with border color.

As for the details of the implemenation:
I basically check ,whether the library version supports _Ctx functions in a `#define canUseContext`
If it can, I concatenate _Ctx to the given function names in the function table. It it can't, I concatenate nothing to the names.
Then I need to issue a specific call to two different versions. I use static dispatch and Alexandrescu's [Type2Var](https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Int-To-Type) idiom with class templates to select the required version to call during the compile time.

So everything is static and should not affect the performance

awarebayes:
You may not like me using concatenation macros, it can be avoided

modules/cudawarping/src/warp.cpp
```diff
@@ -193,7 +260,8 @@ void cv::cuda::warpAffine(InputArray _src, OutputArray _dst, InputArray _M, Size
     CV_Assert( interpolation == INTER_NEAREST || interpolation == INTER_LINEAR || interpolation == INTER_CUBIC );
     CV_Assert( borderMode == BORDER_REFLECT101 || borderMode == BORDER_REPLICATE || borderMode == BORDER_CONSTANT || borderMode == BORDER_REFLECT || borderMode == BORDER_WRAP );
 
-    _dst.create(dsize, src.type());
+    if (_dst.size() != dsize)
+        _dst.create(dsize, src.type());
```
cudawarped, line 264: I think this is redundant as [void cv::cuda::GpuMat::create(int _rows, int _cols, int _type)](https://github.com/opencv/opencv/blob/c2c8da2517f57eae3d3db459ab36c3813692d87e/modules/core/src/cuda/gpu_mat.cu#L160) should automatically handle this for you in addition to checking the type.

modules/cudawarping/src/warp.cpp
```diff
@@ -144,11 +165,24 @@ namespace
...
-        typedef typename NppWarpFunc<DEPTH>::npp_type npp_type;
+
+        typedef typename NppWarpFunc<DEPTH, canUseContext>::npp_type npp_type;
 
         static void call(const cv::cuda::GpuMat& src, cv::cuda::GpuMat& dst, double coeffs[][3], int interpolation, cudaStream_t stream)
+        {
+            call_impl(src, dst, coeffs, interpolation, stream, Int2Type<canUseContext>());
+        }
+
+        template <int I>
```
cudawarped, line 178: Can we use cudev::Int2Type?

cudawarped:
@alalek I'm not great with templates is there a better way of implementing this change with them?  I'm asking because if this change is suitable it could probably be applied to all the CUDA NPP routines.

Just out of interest how long does support need to be offered to older CUDA releases?  `NppStreamContext` (_ctx) was added  3 years ago so if support only has to be for 3 years then the routines could be updated to just use the newer `NppStreamContext` versions instead of both?

modules/cudawarping/src/warp.cpp
awarebayes: far as I can remember, I checked the profiler and maybe there was some overhead? I am not sure...

modules/cudawarping/src/warp.cpp
awarebayes: sure thing

cudawarped:
@awarebayes where did you get to with this?

I finally had chance to take a look and I'm not sure if it makes sense to use templates in this case.  I may be wrong but the code we are trying to template is not another type but historic/unsafe functions, which should probably be removed in future, and are only included for users who are unable to upgrade to the lates version of CUDA.  

I think something similar to https://github.com/opencv/opencv_contrib/blob/ec9d7bca56a9f2645ecb96d9b51460b003e9b7a5/modules/cudawarping/src/warp.cpp#L152 would suffice, what do you think?

awarebayes:
I am pro removing old code and just going along with the new version.

Ifdefs can solve anything but those are just macros and do not provide static dispatch

awarebayes:
There is an obvious  problem for most people, using new versions of NPP, if it breaks older code, welp :(

cudawarped:
> Ifdefs can solve anything but those are just macros and do not provide static dispatch

Can you explain futher what you mean?  Both the templated and the #define version have one code path determined at compile time based on the version of NPP being used.  There is currently no flag to choose the older unsafe version.

> There is an obvious problem for most people, using new versions of NPP, if it breaks older code, welp :(

This could be an issue but then isn't that as an issue a user may/can/does unfortunately face every time they change the version of the CUDA SDK they are using?

modules/cudawarping/src/warp.cpp
```diff
@@ -45,8 +45,19 @@
 using namespace cv;
 using namespace cv::cuda;
 
+#define canUseContext NPP_VERSION >= (10 * 1000 + 1 * 100 + 0)
```
asmorkalov, line 48: please user uppercase for macros, e.g. `NPP_CONTEXT_AVAILABLE`

PR #3339, opened 2022-08-26
Fix memory leak in cudacodec

cudawarped:
In https://github.com/opencv/opencv_contrib/pull/3098 I carelessly introduced a memory leak, this fix should address that issue.

### DIFF
--- a/modules/cudacodec/src/frame_queue.hpp
+++ b/modules/cudacodec/src/frame_queue.hpp
@@ -50,11 +50,16 @@
 class RawPacket {
 public:
     RawPacket(const unsigned char* _data, const size_t _size = 0, const bool _containsKeyFrame = false);
-    unsigned char* Data() const { return *data; }
-    size_t size;
-    bool containsKeyFrame;
+    ~RawPacket();
+    RawPacket(const RawPacket& other);
+    RawPacket(RawPacket&& other) noexcept;
+    RawPacket& operator=(const RawPacket& other);
+    RawPacket& operator=(RawPacket&& other) noexcept;
+    unsigned char* Data() const { return data; }
+    size_t size = 0;
+    bool containsKeyFrame = false;
 private:
-    cv::Ptr<unsigned char*> data = 0;
+    unsigned char* data = 0;
 };
 
 namespace cv { namespace cudacodec { namespace detail {
--- a/modules/cudacodec/src/frame_queue.hpp
+++ b/modules/cudacodec/src/frame_queue.hpp
@@ -50,16 +50,12 @@
 class RawPacket {
 public:
     RawPacket(const unsigned char* _data, const size_t _size = 0, const bool _containsKeyFrame = false);
-    ~RawPacket();
-    RawPacket(const RawPacket& other);
-    RawPacket(RawPacket&& other) noexcept;
-    RawPacket& operator=(const RawPacket& other);
-    RawPacket& operator=(RawPacket&& other) noexcept;
-    unsigned char* Data() const { return data; }
-    size_t size = 0;
-    bool containsKeyFrame = false;
+    const unsigned char* Data() const noexcept { return data.data(); }
+    size_t Size() const noexcept  { return data.size(); }
+    bool ContainsKeyFrame() const noexcept  { return containsKeyFrame; }
 private:
-    unsigned char* data = 0;
+    std::vector<unsigned char> data;
+    bool containsKeyFrame = false;
 };
 
 namespace cv { namespace cudacodec { namespace detail {
--- a/modules/cudacodec/src/frame_queue.cpp
+++ b/modules/cudacodec/src/frame_queue.cpp
@@ -46,9 +46,31 @@
 #ifdef HAVE_NVCUVID
 
 RawPacket::RawPacket(const unsigned char* _data, const size_t _size, const bool _containsKeyFrame) : size(_size), containsKeyFrame(_containsKeyFrame) {
-    data = cv::makePtr<unsigned char*>(new unsigned char[size]);
-    memcpy(*data, _data, size);
+    data = new unsigned char[size];
+    memcpy(data, _data, size);
 };
+
+RawPacket::~RawPacket() {
+    if (data) delete[] data;
+}
+
+RawPacket::RawPacket(const RawPacket& other) : RawPacket(other.data, other.size, other.containsKeyFrame) {
+}
+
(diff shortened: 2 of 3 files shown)


modules/cudacodec/src/frame_queue.hpp
```diff
@@ -50,11 +50,16 @@
...
+    RawPacket(const RawPacket& other);
+    RawPacket(RawPacket&& other) noexcept;
+    RawPacket& operator=(const RawPacket& other);
+    RawPacket& operator=(RawPacket&& other) noexcept;
+    unsigned char* Data() const { return data; }
+    size_t size = 0;
+    bool containsKeyFrame = false;
 private:
-    cv::Ptr<unsigned char*> data = 0;
+    unsigned char* data = 0;
```
VadimLevin, line 62: If you make `data` - `std::unique_ptr<unsigned char[]>` the solution will be more robust.

Move operations and destructor can be defaulted:
```cpp

~RawPacket() = default;
RawPacket(RawPacket&& other) noexcept = default;
RawPacket& operator=(RawPacket&& other) noexcept = default;
```

Copy operations are quite simple to implement:
```cpp
RawPacket::RawPacket(const RawPacket& other)
    : RawPacket(other.data.get(), other.size, other.containsKeyFrame) {
}

RawPacket& operator=(const RawPacket& other) {
    *this = RawPacket(other); // use copy and move constructor
    return *this;
}
```

Small demo: [Godbolt link](https://godbolt.org/z/TP8dTGc8f)
```cpp
#include <iostream>
#include <memory>
#include <cstring>

std::ostream& print(std::ostream& stream, const unsigned char* data, size_t size) {
    stream << "[";
    if (size > 0 && data != nullptr) {
        stream << static_cast<int>(*data);
        for (size_t i = 1; i < size; ++i) {
            stream << ", " << static_cast<int>(data[i]);
        }
    }
    return stream << "]";
}

class RawPacket {
public:
    RawPacket(const unsigned char* data, size_t size) : size_{size} {
        std::cout << this << " from ";
        print(std::cout, data, size) << "\n";

        data_ = std::make_unique<unsigned char[]>(size_);
        std::memcpy(data_.get(), data, size_ * sizeof(unsigned char));
    }

    RawPacket(const RawPacket& other) : RawPacket(other.data_.get(), other.size_) {
        std::cout << this << " copy from " << &other << std::endl;
    }

    RawPacket& operator=(const RawPacket& other) {
        std::cout << this << " = copy from " << &other << std::endl;
        *this = RawPacket(other);
        return *this;
    }

    RawPacket(RawPacket&& other) noexcept = default;

    RawPacket& operator=(RawPacket&& other) noexcept = default;

    ~RawPacket() = default;

    size_t size() const noexcept {
        return size_;
    }

    const unsigned char* data() const noexcept {
        return data_.get();
    }

private:
    size_t size_{};
    std::unique_ptr<unsigned char[]> data_;
};

std::ostream& operator<<(std::ostream& stream, const RawPacket& packet) {
    stream << &packet << " contains ";
    return print(stream, packet.data(), packet.size());
}

int main() {
    const unsigned char source[] {1, 2, 3, 4, 5, 6, 7, 8};
    RawPacket packet(source, 3);
    std::cout << packet << std::endl;
    RawPacket packet2 = std::move(packet);
    std::cout << packet2 << std::endl;

    RawPacket packet3 = packet2;
    std::cout << packet3 << std::endl;
    RawPacket packet4(source + 2, 6);
    std::cout << packet4 << std::endl;
    packet4 = packet2;
    std::cout << packet4 << std::endl;
    return 0;
}
```

modules/cudacodec/src/frame_queue.hpp
```diff
@@ -50,11 +50,16 @@
...
-    unsigned char* Data() const { return *data; }
-    size_t size;
-    bool containsKeyFrame;
+    ~RawPacket();
+    RawPacket(const RawPacket& other);
+    RawPacket(RawPacket&& other) noexcept;
+    RawPacket& operator=(const RawPacket& other);
+    RawPacket& operator=(RawPacket&& other) noexcept;
+    unsigned char* Data() const { return data; }
+    size_t size = 0;
```
VadimLevin, line 59: Why is the `size` public? 
You shouldn't change the size of dynamic allocated memory without any restrictions.

modules/cudacodec/src/frame_queue.hpp
VadimLevin, line 62: Another option. Make `data` `std::vector<unsigned char>`.
Imho it's preferable:
[Godbolt link](https://godbolt.org/z/MvsMKMPnY). Everything can be defaulted.

```cpp
#include <iostream>
#include <vector>

std::ostream& print(std::ostream& stream, const unsigned char* data, size_t size) {
    stream << "[";
    if (size > 0 && data != nullptr) {
        stream << static_cast<int>(*data);
        for (size_t i = 1; i < size; ++i) {
            stream << ", " << static_cast<int>(data[i]);
        }
    }
    return stream << "]";
}

class RawPacket {
public:
    RawPacket(const unsigned char* data, size_t size) : data_{data, data + size} {
        std::cout << this << " from ";
        print(std::cout, data, size) << "\n";
    }

    RawPacket(const RawPacket& other) = default;

    RawPacket& operator=(const RawPacket& other) = default;

    RawPacket(RawPacket&& other) noexcept = default;

    RawPacket& operator=(RawPacket&& other) noexcept = default;

    ~RawPacket() = default;

    size_t size() const noexcept {
        return data_.size();
    }

    const unsigned char* data() const noexcept {
        return data_.data();
    }

private:
    std::vector<unsigned char> data_;
};

std::ostream& operator<<(std::ostream& stream, const RawPacket& packet) {
    stream << &packet << " contains ";
    return print(stream, packet.data(), packet.size());
}

int main() {
    const unsigned char source[] {1, 2, 3, 4, 5, 6, 7, 8};
    RawPacket packet(source, 3);
    std::cout << packet << std::endl;
    RawPacket packet2 = std::move(packet);
    std::cout << packet2 << std::endl;

    RawPacket packet3 = packet2;
    std::cout << packet3 << std::endl;
    RawPacket packet4(source + 2, 6);
    std::cout << packet4 << std::endl;
    packet4 = packet2;
    std::cout << packet4 << std::endl;
    return 0;
}
```

cudawarped: Thanks for the detailed review, I agree vector would be better than mixing smart pointers, I'll address this when I am back.

cudawarped:
@VadimLevin I've added the changes you suggested.

VadimLevin (approved):

asmorkalov:
Tested with Ubuntu 18.04 and Cuda 10.2

PR #3340, opened 2022-08-26
Add support for missing codecs to cudacodec

cudawarped:
Currently cudacodec does not support all the codecs supported by nvdec or all possible fourcc codes returned by FFmpeg.  This PR should address this by adding the av1 codec and dealing with all combinations of upper/lower case and 3 character fourcc combinations.


asmorkalov (commented):
:+1:

asmorkalov (approved):

PR #3355, opened 2022-09-16
Add scaling and cropping options to `cudacodec::VideoReader`

cudawarped:
Nivida's Video Codec SDK allows the user to specify the following parameters  to control the final output:
- Scaling dimension
- Cropping dimension
- Dimension if the user wants to change the aspect ratio

this pull request adds this capability to `cudacodec::VideoReader`.

From local experiments this does not effect the decoding utilization, however choosing for example a smaller ROI from the decoded video stream reduces the CUDA overhead from the post processing routines (ConvertNV12BLtoNV12).  

Additionaly [this ](https://github.com/cudawarped/opencv_contrib/blob/587207bd9bca9c555e4549ae8f8d4fad0de8275c/modules/cudacodec/src/video_parser.cpp#L128) addition makes the checks described in https://github.com/opencv/opencv_contrib/pull/3001 unecessary.

### DIFF
--- a/modules/cudacodec/CMakeLists.txt
+++ b/modules/cudacodec/CMakeLists.txt
@@ -6,7 +6,7 @@ set(the_description "CUDA-accelerated Video Encoding/Decoding")
 
 ocv_warnings_disable(CMAKE_CXX_FLAGS /wd4127 /wd4324 /wd4512 -Wundef -Wshadow)
 
-ocv_add_module(cudacodec opencv_core opencv_videoio opencv_cudaarithm OPTIONAL opencv_cudev WRAP python)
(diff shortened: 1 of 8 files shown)


cudawarped:
Sorry I missed that, that will teach me to be lazy and just build on Windows!

asmorkalov:
The issue is still there.

cudawarped:
Sorry again, I've added the other missing target but I am at a loss as to why it builds without it, I tried both make and ninja on WSL Ubuntu 20.04 and it built without the cudawapring target.  Maybe I'm missing something.

asmorkalov (approved):
:+1:

PR #3369, opened 2022-10-26
Update find ellipses

sturkmen72:

### DIFF
--- a/modules/ximgproc/samples/find_ellipses.py
+++ b/modules/ximgproc/samples/find_ellipses.py
@@ -1,0 +1,46 @@
+#!/usr/bin/python
+
+'''
+This example illustrates how to use cv.ximgproc.findEllipses function.
+
+Usage:
+    find_ellipses.py [<image_name>]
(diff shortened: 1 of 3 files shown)


asmorkalov:
@sturkmen72 Is the patch ready for review?

sturkmen72:
@asmorkalov the patch is ready for review except "trailing whitespace" issue ( I did not understand what is the problem )

modules/ximgproc/samples/find_ellipses.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/python
```
alalek, line 1: EOL (end-of-line) encoding of this file is CRLF (Windows).
Please switch to LF (UNIX style)

Some hints are here: https://stackoverflow.com/questions/27810758/how-to-replace-crlf-with-lf-in-a-single-file

alalek (approved):
LGTM :+1: Thank you for contribution!

PR #3370, opened 2022-11-09
C-API cleanup for OpenCV 5.x (imgproc, highgui)

mshabunin:
There were several places where obsolete API has been used. I could not fully test these changes (_cudalegacy_ and _rgbd_ example), but build process is fine.

This PR must be merged before opencv/opencv#22754 to preserve compatibility. Technically this PR can be backported to 4.x, I think.


asmorkalov (approved):
:+1:

PR #3371, opened 2022-11-09
More data types supported in cv::cuda::transpose()

chacha21:
proposal for #22782


chacha21:
proposal for  #22782

modules/cudaarithm/src/cuda/transpose.cu
```diff
@@ -60,34 +60,147 @@ void cv::cuda::transpose(InputArray _src, OutputArray _dst, Stream& stream)
...
+      (srcType == CV_16SC1) || (srcType == CV_16SC3) || (srcType == CV_16SC4) ||
+      (srcType == CV_32SC1) || (srcType == CV_32SC3) || (srcType == CV_32SC4) ||
+      (srcType == CV_32FC1) || (srcType == CV_32FC3) || (srcType == CV_32FC4);
+    const bool isElemSizeSupportedByNppi =
+      ((elemSize != 0) && !(elemSize%1) && ((elemSize/1)<=4)) ||
+      ((elemSize != 0) && !(elemSize%2) && ((elemSize/2)<=4)) ||
+      ((elemSize != 0) && !(elemSize%4) && ((elemSize/4)<=4)) ||
+      ((elemSize != 0) && !(elemSize%8) && ((elemSize/8)<=2));
+    if (src.empty())
+      dst.release();
```
alalek, line 85: Empty input should be a error as nobody want to process "nothing" in real use cases: https://github.com/opencv/opencv/issues/8300

chacha21: Ok, I thought it was better since [Allow empty matrices in most functions](https://github.com/opencv/opencv/wiki/2021#2021-11-17) (personnaly I prefer a no-op rather than an exception)

modules/cudaarithm/src/cuda/transpose.cu
```diff
@@ -60,34 +60,147 @@ void cv::cuda::transpose(InputArray _src, OutputArray _dst, Stream& stream)
...
     GpuMat src = getInputMat(_src, stream);
 
+    const int srcType = src.type();
+    const int srcDepth = src.depth();
+    const int srcCn = src.channels();
     const size_t elemSize = src.elemSize();
+    const size_t elemSize1 = src.elemSize1();
 
-    CV_Assert( elemSize == 1 || elemSize == 4 || elemSize == 8 );
+    //CV_Assert( elemSize == 1 || elemSize == 4 || elemSize == 8 );
```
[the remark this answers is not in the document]
chacha21: It is handled by 

```
...
else
      CV_Error(Error::StsUnsupportedFormat, "");
```

modules/cudaarithm/src/cuda/transpose.cu
chacha21: I will also add a fastpath for size == (1, 1), where transpose is just a copy

modules/cudaarithm/src/cuda/transpose.cu
```diff
@@ -60,34 +60,147 @@ void cv::cuda::transpose(InputArray _src, OutputArray _dst, Stream& stream)
...
 
-    if (elemSize == 1)
+    const bool isNppiNativelySupported =
+      (srcType == CV_8UC1)  || (srcType == CV_8UC3)  || (srcType == CV_8UC4)  ||
+      (srcType == CV_16UC1) || (srcType == CV_16UC3) || (srcType == CV_16UC4) ||
+      (srcType == CV_16SC1) || (srcType == CV_16SC3) || (srcType == CV_16SC4) ||
+      (srcType == CV_32SC1) || (srcType == CV_32SC3) || (srcType == CV_32SC4) ||
+      (srcType == CV_32FC1) || (srcType == CV_32FC3) || (srcType == CV_32FC4);
+    const bool isElemSizeSupportedByNppi =
+      ((elemSize != 0) && !(elemSize%1) && ((elemSize/1)<=4)) ||
```
cudawarped, line 80: Isn't elemSize%1 == 0 always?
When can elemSize == 0?

cudawarped:
Would it be possible to incorportate the newer stream api for newer versions of CUDA aswell?
https://github.com/opencv/opencv_contrib/pull/3338

modules/cudaarithm/src/cuda/transpose.cu
chacha21: You are right, but it was on purpose for code clarity/readibility
elemSize%1 and elemSize/1 will be optimized out by the compiler, this is just to balance with the 2, 4, 8 cases

Ok to get rid of the (elemSize != 0) check

chacha21:
> Would it be possible to incorportate the newer stream api for newer versions of CUDA aswell?
> See #3338

I am currently bases on 4.x branch, usage of _Ctx seems to be in master.
I would prefer deferring that work to another PR

modules/cudaarithm/src/cuda/transpose.cu
cudawarped: For clarity I would use elemSize1 and channels?
Under what circumstance will `isNppiNativelySupported  == false` and `isElemSizeSupportedByNppi == true` or `isElemSizeSupportedByGridTranspose  == true`?

chacha21: There is overlap between isNppiNativelySupported, isElemSizeSupportedByNppi  and isElemSizeSupportedByGridTranspose 
The idea is that if anything goes wrong, they can be disabled independently.

cudawarped: Is there a complete overlap, i.e. is `isElemSizeSupportedByNppi` and `isElemSizeSupportedByGridTranspose` redundant unless
`isNppiNativelySupported` is manually set to `false`?

chacha21: > Is there a complete overlap, i.e. is `isElemSizeSupportedByNppi` and `isElemSizeSupportedByGridTranspose` redundant unless
> `isNppiNativelySupported` is manually set to `false`?

There are cases where (isNppiNativelySupported == false) and (isElemSizeSupportedByNppi== true) (e.g. 16U_C2)

But there are currently no cases where (isElemSizeSupportedByNppi== false) and  (isElemSizeSupportedByGridTranspose == true). This could be useful if performance tests reveals that some NPP calls are slower than gridTranspose(). Specific cases could be excluded from isElemSizeSupportedByNppi.

cudawarped: Is `16UC2` is handled by the below block?
```
else if (!(elemSize%2) && ((elemSize/2)==2))
  nppSafeCall( nppiTranspose_8u_C4R(src.ptr<Npp8u>(), static_cast<int>(src.step),
    dst.ptr<Npp8u>(), static_cast<int>(dst.step), sz) );
```
If so wouldn't it be better to have this under 
```
else if (srcType == CV_16UC2)
```

modules/cudaarithm/src/cuda/transpose.cu
```diff
@@ -201,6 +199,9 @@ void cv::cuda::transpose(InputArray _src, OutputArray _dst, Stream& stream)
       else if (!(elemSize%8) && ((elemSize/8)==2))
         nppSafeCall( nppiTranspose_32f_C4R(src.ptr<Npp32f>(), static_cast<int>(src.step),
           dst.ptr<Npp32f>(), static_cast<int>(dst.step), sz) );
+
+      if (!stream)
```
cudawarped, line 203: Should this be applied to `gridTranspose()` aswell, I can't understand why it was previously missing, surely the result could have been in flight on returns from this function when no stream is passed?

modules/cudaarithm/src/cuda/transpose.cu
chacha21: Exactly, this is not a native nppi Call (isNppiNativelySupported == false) but it can be handled by cheatin on the data type (isElemSizeSupportedByNppi  == true)
That's why the tests inside `if (isElemSizeSupportedByNppi)` is explicitely focusing on `elemSize `rather than `srcType`

cudawarped: I'm not sure if this makes it easier or harder for the next person who looks at it to see whats going on, what do you think @alalek ?

Either way I think having overlap between both makes it difficult to understand, maybe if the "cheat" only included the cases where it was applicable it would be more obvious?

modules/cudaarithm/src/cuda/transpose.cu
chacha21: I have the opposite question : why is it ever needed ? I let them here, but I don't understand the purpose of this extra synchronization when the default stream (sync by default) is used.

cudawarped: Kernel launches are [asynchronous with respect](https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html) with respec to the host.  The [default stream](https://docs.nvidia.com/cuda/cuda-runtime-api/stream-sync-behavior.html) syncs with other streams by default in legacy mode.

My interpretation is that the OpenCV API works on the assumption that if a stream isn't passed the user wants synchronization.

chacha21: Do you mean that new tha exists both `cudaStreamPerThread` and `cudaStreamLegacy`, people using OpenCV always expect the `cudaStreamLegacy`, thus requiring the `cudaDeviceSynchronize()` ?
If so, I agree.

cudawarped: No users of the api will expect functions which they don't pass a stream to, to be synchronous with respect to the host when they return.  This would not be the case with either `cudaStreamPerThread` or `cudaStreamLegacy` as the kernel launch is asynchronous in both cases. 

In the example from the docs

```
1) k_1<<<1, 1, 0, s>>>();
2) k_2<<<1, 1>>>();
3) k_3<<<1, 1, 0, s>>>()
4) ...
```

k_2 waits on k_1 because k_2 is in the legacy default stream, then k_3 waits on the legacy stream.  Because of the specific way this has been set up k_1 and k_2 have finished executing before the call to `k_3<<<1, 1, 0, s>>>()` however the result from k_3 may still be in flight after control has returned to the host when you reach line 4.

Now I haven't used per thread default streams (I always use explicit streams) but my understanding is that if the CUDA_API_PER_THREAD_DEFAULT_STREAM macro was used to enable per thread default streams k_1 would run before k_3 but both would be asynchronous with respect to k_2.  Either way when control returns to the host on line 4 they may still all be in flight.

On the other hand if we have the following which is our case if no stream is passed

```
1) k_1<<<1, 1>>>();
2) k_2<<<1, 1>>>();
3) k_3<<<1, 1>>>();
4) 
```

in either case `cudaStreamPerThread` or `cudaStreamLegacy` all three kernels may still be in flight when control returns to the host on line 4 if we don't explicitly syncronize.

chacha21: What is the problem of the kernel on the default stream being in flight when returning ?

cudawarped: The OpenCV cuda API guarantees synchronization when the stream isn't passed to one of its functions.

chacha21: Ok, anyway, I just checked and `gridTranspose()` does ultimately call `cudaDeviceSynchronize()`(see [https://github.com/opencv/opencv_contrib/blob/4.x/modules/cudev/include/opencv2/cudev/grid/detail/transpose.hpp](https://github.com/opencv/opencv_contrib/blob/4.x/modules/cudev/include/opencv2/cudev/grid/detail/transpose.hpp))

asmorkalov: @cudawarped Do you have open issues with the PR. I want to merge, if you do not mind.

asmorkalov (approved):

modules/cudaarithm/src/cuda/transpose.cu
cudawarped: @asmorkalov I am not 100% convinced that `isNppiNativelySupported` and `isElemSizeSupportedByNppi` flags are needed.  I can understand that @chacha21 wants a fall back option in case something goes wrong and because of that wants to seperate out the two logic paths but I am not sure if the redundant calls under `isElemSizeSupportedByNppi` make the function harder to maintain, what do you think?

chacha21:
I could not test performance with other GPUs than mine (an old GTX 750). I am pretty confident, but it would be good if someone could have a look with more modern ones.

modules/cudaarithm/src/cuda/transpose.cu
chacha21: Whatever the decision, I will certainly agree : I explained my initial code structure, but I am OK to adapt for a more "OpenCV style"

asmorkalov:
Transpose perf test results for GF 1080 with CUDA 10.2, Ubuntu 18.04:
```
Geometric mean (ms)

             Name of Test               before     after     after   
                                       transpose transpose transpose 
                                                               vs    
                                                             before  
                                                           transpose 
                                                           (x-factor)
Transpose::Sz_Type::(1280x720, 8UC1)     0.024     0.024      1.00   
Transpose::Sz_Type::(1280x720, 32SC1)    0.050     0.037      1.36   
Transpose::Sz_Type::(1280x720, 64FC1)    0.072     0.073      0.99   
Transpose::Sz_Type::(1280x720, 16UC2)    0.050     0.037      1.36   
Transpose::Sz_Type::(1280x720, 16SC2)    0.051     0.037      1.36   
Transpose::Sz_Type::(1280x720, 32SC2)    0.072     0.073      0.98   
Transpose::Sz_Type::(1280x720, 8UC4)     0.050     0.037      1.36   
Transpose::Sz_Type::(1280x1024, 8UC1)    0.034     0.034      1.00   
Transpose::Sz_Type::(1280x1024, 32SC1)   0.076     0.050      1.53   
Transpose::Sz_Type::(1280x1024, 64FC1)   0.101     0.096      1.06   
Transpose::Sz_Type::(1280x1024, 16UC2)   0.076     0.050      1.53   
Transpose::Sz_Type::(1280x1024, 16SC2)   0.076     0.049      1.53   
Transpose::Sz_Type::(1280x1024, 32SC2)   0.101     0.095      1.06   
Transpose::Sz_Type::(1280x1024, 8UC4)    0.076     0.049      1.54   
Transpose::Sz_Type::(1920x1080, 8UC1)    0.050     0.050      1.00   
Transpose::Sz_Type::(1920x1080, 32SC1)   0.124     0.082      1.51   
Transpose::Sz_Type::(1920x1080, 64FC1)   0.170     0.169      1.01   
Transpose::Sz_Type::(1920x1080, 16UC2)   0.124     0.082      1.51   
Transpose::Sz_Type::(1920x1080, 16SC2)   0.124     0.082      1.50   
Transpose::Sz_Type::(1920x1080, 32SC2)   0.170     0.169      1.00   
Transpose::Sz_Type::(1920x1080, 8UC4)    0.124     0.082      1.51
```
The execution time is very small. Not sure if, it's real speedup or some system side effect. In any case, the PR does not introduce regressions.

modules/cudaarithm/src/cuda/transpose.cu
asmorkalov: 1. Conditions like  `if (!(elemSize%2) && ((elemSize/2)==2))` looks very cryptic. I understand the logic behind it, but  `if (elemSize==4)` definitely more readable.
2. The only reason for fallback I see is 2 channel matrix. All other cases are handled with regular types. I propose to merge "native" support and just "support" in single case and use relevant Npp call. No duplicated/dead branches, mor obvious testing.

modules/cudaarithm/src/cuda/transpose.cu
```diff
@@ -60,34 +60,202 @@ void cv::cuda::transpose(InputArray _src, OutputArray _dst, Stream& stream)
...
+              dst.ptr<Npp16u>(), static_cast<int>(dst.step), sz) );
+          else if (elemSize == 8)
+            nppSafeCall( nppiTranspose_16u_C4R(src.ptr<Npp16u>(), static_cast<int>(src.step),
+              dst.ptr<Npp16u>(), static_cast<int>(dst.step), sz) );
+          else if (elemSize == 12)
+            nppSafeCall( nppiTranspose_32s_C3R(src.ptr<Npp32s>(), static_cast<int>(src.step),
+              dst.ptr<Npp32s>(), static_cast<int>(dst.step), sz) );
+          else if (elemSize == 16)
+            nppSafeCall( nppiTranspose_32s_C4R(src.ptr<Npp32s>(), static_cast<int>(src.step),
+              dst.ptr<Npp32s>(), static_cast<int>(dst.step), sz) );
```
asmorkalov, line 164: The interpretation block does the transpose second time, if `elemSize` duplicates some of `srcType`

modules/cudaarithm/src/cuda/transpose.cu
```diff
@@ -60,34 +60,202 @@ void cv::cuda::transpose(InputArray _src, OutputArray _dst, Stream& stream)
...
+              dst.ptr<Npp16u>(), static_cast<int>(dst.step), sz, ctx) );
+          else if (elemSize == 8)
+            nppSafeCall( nppiTranspose_16u_C4R_Ctx(src.ptr<Npp16u>(), static_cast<int>(src.step),
+              dst.ptr<Npp16u>(), static_cast<int>(dst.step), sz, ctx) );
+          else if (elemSize == 12)
+            nppSafeCall( nppiTranspose_32s_C3R_Ctx(src.ptr<Npp32s>(), static_cast<int>(src.step),
+              dst.ptr<Npp32s>(), static_cast<int>(dst.step), sz, ctx) );
+          else if (elemSize == 16)
+            nppSafeCall( nppiTranspose_32s_C4R_Ctx(src.ptr<Npp32s>(), static_cast<int>(src.step),
+              dst.ptr<Npp32s>(), static_cast<int>(dst.step), sz, ctx) );
```
asmorkalov, line 242: The same here.

modules/cudaarithm/src/cuda/transpose.cu
```diff
@@ -60,34 +60,202 @@ void cv::cuda::transpose(InputArray _src, OutputArray _dst, Stream& stream)
...
+        }//end if (stream != 0)
 
         if (!stream)
-            CV_CUDEV_SAFE_CALL( cudaDeviceSynchronize() );
-    }
+          CV_CUDEV_SAFE_CALL( cudaDeviceSynchronize() );
```
asmorkalov, line 246: Syncronious NPP call is used if `!stream`. No need to sync again. Right?

modules/cudaarithm/src/cuda/transpose.cu
```diff
@@ -60,34 +60,202 @@ void cv::cuda::transpose(InputArray _src, OutputArray _dst, Stream& stream)
...
-        gridTranspose(globPtr<int>(src), globPtr<int>(dst), stream);
-    }
-    else // if (elemSize == 8)
-    {
-        gridTranspose(globPtr<double>(src), globPtr<double>(dst), stream);
-    }
+      gridTranspose(globPtr<signed int>(src), globPtr<signed int>(dst), stream);
+    else if (elemSize == 8)
+      gridTranspose(globPtr<double>(src), globPtr<double>(dst), stream);
+    */
```
asmorkalov, line 258: Please remove dead code.

modules/cudaarithm/src/cuda/transpose.cu
chacha21: > The interpretation block does the transpose second time, if `elemSize` duplicates some of `srcType`

Well... no ? This is still an "else if"

cudawarped:
@chacha21 If you use the new stream ctx api you should also include the old api for compatibility with pre CUDA 10.1, see https://github.com/opencv/opencv_contrib/pull/3338.

modules/cudaarithm/src/cuda/transpose.cu
```diff
@@ -56,40 +56,136 @@ using namespace cv;
...
+    else
+    {
+        #if USE_NPP_STREAM_CONTEXT
+        constexpr const bool useNppStreamCtx = true;
+        #else
+        constexpr const bool useNppStreamCtx = false;
+        #endif
+        cudaStream_t _stream = StreamAccessor::getStream(stream);
+
+        if (!_stream || !useNppStreamCtx)
```
cudawarped, line 146: Would it be better to always use the old api when CUDA SDK < 10.1 and the newer one otherwise regardless of whether there is a stream or not?

modules/cudaarithm/src/cuda/transpose.cu
```diff
@@ -56,40 +56,136 @@ using namespace cv;
 using namespace cv::cuda;
 using namespace cv::cudev;
 
-void cv::cuda::transpose(InputArray _src, OutputArray _dst, Stream& stream)
-{
-    GpuMat src = getInputMat(_src, stream);
+#define USE_NPP_STREAM_CONTEXT (NPP_VERSION >= (10 * 1000 + 1 * 100 + 0))
```
cudawarped, line 59: I think it would be better to define this globally somewhere although not sure where, maybe [modules/cudev/include/opencv2/cudev/common.hpp](https://github.com/cudawarped/opencv_contrib/blob/e40c43d96a22edaa3fbb880e957a3753938dc4f0/modules/cudev/include/opencv2/cudev/common.hpp#L61).

cudawarped:
@chacha21 I like the changes apart from my comments above.  I wouldn't necessarily act on them until @asmorkalov has taken a look as he will ultimately decide on the required format.

modules/cudaarithm/src/cuda/transpose.cu
```diff
@@ -56,40 +56,136 @@ using namespace cv;
...
+  };
+  #endif
+}
+
+void cv::cuda::transpose(InputArray _src, OutputArray _dst, Stream& stream)
+{
+    GpuMat src = getInputMat(_src, stream);
 
-    syncOutput(dst, _dst, stream);
+    const int srcType = src.type();
```
cudawarped, line 117: Is this redundant?

modules/cudaarithm/src/cuda/transpose.cu
```diff
@@ -56,40 +56,136 @@ using namespace cv;
...
+    const size_t elemSize = src.elemSize();
+
+    GpuMat dst = getOutputMat(_dst, src.cols, src.rows, src.type(), stream);
+
+    const bool isSupported =
+      (elemSize == 1) || (elemSize == 2) || (elemSize == 3) || (elemSize == 4) ||
+      (elemSize == 6) || (elemSize == 8) || (elemSize == 12) || (elemSize == 16);
+
+    if (!isSupported)
```
cudawarped, line 126: Would CV_Assert() be better here?

PR #3373, opened 2022-11-10
New rectification and reconstruction functions for omnidir model

stefan-spiss:
To make the 3D reconstruction workflow with the omnidir camera model similar to the one for the pinhole and fisheye model, new functions were implemented:
- estimateNewCameraMatrixForUndistortRectify -> estimates new projection matrices for the rectified output image (same as for the fisheye model)
- extended stereoRectify -> overloaded function that not only estimates the rotations between the cameras, but also new projection matrices after rectification
- reprojectImageTo3D -> function that depending on the rectification type (perspective or longilati) either calls the default reprojectImateTo3D function or calculates the 3D information itself (for the longilati rectification).

In addition, the documentation was updated and the naming style for function parameters was unified (new_size -> newSize).

Addresses parts of https://github.com/opencv/opencv/issues/16601#issue-566288527

Since there was no example code available for 3D reconstruction with the omnidir model, I did not add anything new yet. But if that is required, I of course can add an example and new data for it. For the same reason, no tests were added yet.


alalek:
Thank you for the contribution!

Could you add tests for the added functionality?

---

BTW, currently this module have zero tests. Please refer on test from other modules, e.g. from here: https://github.com/opencv/opencv_contrib/tree/4.6.0/modules/mcc/test

stefan-spiss:
@alalek I can add some tests, but similar to the issue https://github.com/opencv/opencv/issues/22537 it is not very easy to test all of it properly.

alalek:
I mean to add tests for the added functionality here (similar to your use case). No need to cover by tests already existed functionality - it is out of scope.

Also no need to parameterize inputs, one input is enough for good start.

stefan-spiss:
@alalek sounds good, I will do that!

asmorkalov:
@stefan-spiss friendly reminder about the test.

stefan-spiss:
@asmorkalov and @alalek: I implemented a test for the new function `omnidir::reprojectImageTo3D`

Regarding the two other functions I am unsure how to properly test them.

- `estimateNewCameraMatrixForUndistortRectify`  
As already mentioned before, testing the function is not that easy. In the fisheye model, the test of this function currently compares to some reference values, but to get these reference values, I would need to run the function itself. So it only would help to make sure, that future code changes do not alter the result of the function.  
I also had a look at the corresponding function of the pinhole model, `getOptimalNewCameraMatrix()` and I could not find any test for it.  
 Therefore, I am not sure how to test the function. Does someone has an idea? Or should I just do a similar test as in the fisheye model, using reference values?

- `stereoRectify`  
In the fisheye model, this function is tested using reference values, which were generated with the function itself (mentioned in a comment). So again, with that I would make sure that future changes produce the same result, but I would not check that the current results are correct.  
In the pinhole model, this function is tested in an integration test that covers the whole stereo calibration pipeline. Parts of it I could also use to test the `stereoRectify` function of the omnidir model, but with that I would only cover parts of the functionality. Do you have any suggestions regarding the test of this function?

PR #3380, opened 2022-11-22
Fix opencv issue #22721

cudawarped:
Fix https://github.com/opencv/opencv/issues/22721#issuecomment-1304837204

### DIFF
--- a/modules/cudacodec/src/video_writer.cpp
+++ b/modules/cudacodec/src/video_writer.cpp
@@ -61,12 +61,6 @@ GUID EncodingProfileGuid(const EncodeProfile encodingProfile);
 GUID EncodingPresetGuid(const EncodePreset nvPreset);
 bool Equal(const GUID& g1, const GUID& g2);
 
-EncoderParams::EncoderParams() : nvPreset(ENC_PRESET_P3), tuningInfo(ENC_TUNING_INFO_HIGH_QUALITY), encodingProfile(ENC_CODEC_PROFILE_AUTOSELECT),
-    rateControlMode(ENC_PARAMS_RC_VBR), multiPassEncoding(ENC_MULTI_PASS_DISABLED), constQp({ 0,0,0 }), averageBitRate(0), maxBitRate(0),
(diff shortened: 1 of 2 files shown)


modules/cudacodec/src/video_writer.cpp
```diff
@@ -61,12 +61,6 @@ GUID EncodingProfileGuid(const EncodeProfile encodingProfile);
 GUID EncodingPresetGuid(const EncodePreset nvPreset);
 bool Equal(const GUID& g1, const GUID& g2);
 
-EncoderParams::EncoderParams() : nvPreset(ENC_PRESET_P3), tuningInfo(ENC_TUNING_INFO_HIGH_QUALITY), encodingProfile(ENC_CODEC_PROFILE_AUTOSELECT),
```
alalek, line 64: There are should be implementation in the `#else` branch:

```
#else // !defined HAVE_NVCUVENC

... CV_Error(...NotImplemented...) implementation stubs ...
```

asmorkalov: `createVideoWriter` has this else branch. With the changes parameters structure could be created correctly without linker issue, but `createVideoWriter` raises the issue. I propose to merge the PR as is.

asmorkalov (approved):

modules/cudacodec/src/video_writer.cpp
cudawarped: I moved it to the header to mirror the way [cudacodec::FormatInfo()](https://github.com/opencv/opencv_contrib/blob/4.x/modules/cudacodec/include/opencv2/cudacodec.hpp#L304) is implemented.

Both structs can exist without the Nvidia Video Coded SDK being installed whether they should or not is a different matter.  Do you want me to move the implementation of both into the .cpp and prevent their use when `HAVE_NVCUVENC` and/or `HAVE_NVCUVID` are not defined?

asmorkalov: I prefer to have config structures always available how it's done in PR.

PR #3381, opened 2022-11-22
Update cudacodec to build against Nvidia Video Codec SDK 12

cudawarped:
Remove previously unsued variable which has been depreciated and prevented building against Nvidia Video Codec SDK 12.

### DIFF
--- a/modules/cudacodec/src/NvEncoder.cpp
+++ b/modules/cudacodec/src/NvEncoder.cpp
@@ -431,7 +431,7 @@ bool NvEncoder::Reconfigure(const NV_ENC_RECONFIGURE_PARAMS* pReconfigureParams)
 
 NV_ENC_REGISTERED_PTR NvEncoder::RegisterResource(void* pBuffer, NV_ENC_INPUT_RESOURCE_TYPE eResourceType,
     int width, int height, int pitch, NV_ENC_BUFFER_FORMAT bufferFormat, NV_ENC_BUFFER_USAGE bufferUsage,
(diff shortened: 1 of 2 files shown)


modules/cudacodec/src/NvEncoder.cpp
```diff
@@ -431,7 +431,7 @@ bool NvEncoder::Reconfigure(const NV_ENC_RECONFIGURE_PARAMS* pReconfigureParams)
 
 NV_ENC_REGISTERED_PTR NvEncoder::RegisterResource(void* pBuffer, NV_ENC_INPUT_RESOURCE_TYPE eResourceType,
     int width, int height, int pitch, NV_ENC_BUFFER_FORMAT bufferFormat, NV_ENC_BUFFER_USAGE bufferUsage,
-    NV_ENC_FENCE_POINT_D3D12* pInputFencePoint, NV_ENC_FENCE_POINT_D3D12* pOutputFencePoint)
+    NV_ENC_FENCE_POINT_D3D12* pInputFencePoint)
```
alalek, line 434: In general we should have `#if` guard with version check here.

cudawarped: Luckily we only call `RegisterResource` internally from `RegisterInputResources`.  `pOutputFencePoint` has always been a redundant function argument included because I mostly copied the implementation of NvEncoder from the Nvidia samples and missed the redundancy.

alalek (approved):

asmorkalov:
Tested manually with Ubuntu 18.04.

PR #3388, opened 2022-12-08
OdometryFrame: arg order changed

savuor:
Connected PR in main repo: [#22925@main](https://github.com/opencv/opencv/pull/22925)

### Changes

`OdometryFrame` constructor signature changed to more convenient where depth goes at 1st place, RGB image at 2nd.
This works better for depth-only `Odometry` algorithms.


asmorkalov:
@savuor Please take a look on failed tests.

savuor:
@asmorkalov The tests were broken because I pushed to the contrib repo before the main.
Pushing in the opposite order solved the problem.

PR #3389, opened 2022-12-10
build: fix/eliminate MSVC warnings

alalek:
**Main PR**: https://github.com/opencv/opencv/pull/22940

```
force_builders=Win64,Win64 OpenCL
```


alalek:
:+1:

PR #3393, opened 2022-12-18
Fix "operator[]" out of boundary access issue in PR#3322

misakicoca:
relates #3322

```
force_builders=Linux x64,Win64,docs,Linux x64 Debug
```

### DIFF
--- a/modules/ximgproc/src/find_ellipses.cpp
+++ b/modules/ximgproc/src/find_ellipses.cpp
@@ -1304,8 +1304,8 @@ void EllipseDetectorImpl::preProcessing(Mat1b &image, Mat1b &dp, Mat1b &dn) {
     // 2 - the pixel does belong to an edge
     for (int i = 0; i <= imgSize.height; i++) {
         int *tmpMag = magBuffer[(i > 0) + 1] + 1;
-        const short *tmpDx = (short *) (dx[i]);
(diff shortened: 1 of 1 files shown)


modules/ximgproc/src/find_ellipses.cpp
```diff
@@ -1304,8 +1304,8 @@ void EllipseDetectorImpl::preProcessing(Mat1b &image, Mat1b &dp, Mat1b &dn) {
     // 2 - the pixel does belong to an edge
     for (int i = 0; i <= imgSize.height; i++) {
         int *tmpMag = magBuffer[(i > 0) + 1] + 1;
-        const short *tmpDx = (short *) (dx[i]);
-        const short *tmpDy = (short *) (dy[i]);
+        const short *tmpDx = (short *) (dx.data + dx.step * i);
+        const short *tmpDy = (short *) (dy.data + dy.step * i);
```
alalek, line 1308: > `dx.data + dx.step * i`

Manual address arithmetic should be avoided.
Use Mat::ptr() instead (use 1 argument to get row start address).

BTW, If debug checks failed - then there is out of boundary access.

---

Also above:

> malloc

manual memory management should be avoided too. Use `std::vector` / `cv::AutoBuffer` instead.

PR #3399, opened 2022-12-21
Remove references to deprecated NumPy type aliases.

vrabaud:
This change replaces references to a number of deprecated NumPy type aliases (np.bool, np.int, np.float, np.complex, np.object, np.str) with their recommended replacement (bool, int, float, complex, object, str).

Those types were deprecated in 1.20 and are removed in 1.24, cf numpy/numpy#22607.


asmorkalov (approved):

vrabaud:
BTW, there is no code to fix on 3.4 in opencv_contrib

PR #3420, opened 2023-01-13
remove charuco duplication

AleksandrPanov:
Removed charuco duplication after move aruco to main repo.


asmorkalov (approved):

modules/aruco/src/aruco.cpp
```diff
@@ -59,64 +59,18 @@ int estimatePoseBoard(InputArrayOfArrays corners, InputArray ids, const Ptr<Boar
...
-        CV_Assert(currId >= 0 && currId < (int)board->getChessboardCorners().size());
-        objPoints.push_back(board->getChessboardCorners()[currId]);
-    }
-
-    // points need to be in different lines, check if detected points are enough
-    if(!_arePointsEnoughForPoseEstimation(objPoints)) return false;
+    // get object and image points for the solvePnP function
+    Mat objPoints, imgPoints;
+    board->matchImagePoints(charucoCorners, charucoIds, objPoints, imgPoints);
+    solvePnP(objPoints, imgPoints, cameraMatrix, distCoeffs, rvec, tvec, useExtrinsicGuess);
```
asmorkalov, line 71: solvePnP throws exception, if there are no enough points.

asmorkalov (changes requested):

modules/aruco/src/aruco.cpp
AleksandrPanov: added points check:
`if(charucoIds.getMat().total() < 4) return false;`

asmorkalov: I propose to add try-catch block and return false in case of exception. The function did not throw exceptions before and should not do it after the PR.

AleksandrPanov: Add try catch:
```
    try {
        solvePnP(objPoints, imgPoints, cameraMatrix, distCoeffs, rvec, tvec, useExtrinsicGuess);
    }
    catch (const cv::Exception& e) {
        CV_LOG_WARNING(NULL, "estimatePoseCharucoBoard: " << std::endl << e.what());
        return false;
    }
```

asmorkalov (approved):

rurban:
This caused a regression in interpolateCornersCharuco, which does not detect the markers anymore.
There need to be more testcases added, we can contribute some from https://github.com/SpexAI/calibpy/
See https://github.com/SpexAI/calibpy/issues/2

PR #3430, opened 2023-01-26
Omnidir calibration: uncertainty fixed

### LINKED ISSUE #3406: Uncertainties calculation for omnidir calibration can be incorrect

- OpenCV => 3.4
- Operating System / Platform => all
- Compiler => all

##### Detailed description

Related to [#19803@main](https://github.com/opencv/opencv/issues/19803)

According to the [MVG], various least squares method explanations (like [this](https://stats.stackexchange.com/a/64495)) and the standard camera calibration code in OpenCV ([here](https://github.com/opencv/opencv/blob/9208dcb07c015e1fda44e40bb07b43c700b4bf46/modules/calib3d/src/calibration.cpp#L1808)), the parameters uncertainty should be calculated with the "minus k" denominator correction:
$$\sigma^2 = \frac{1}{n - k} \sum^n_{i=1} \left( x_i - \bar{x} \right)^2$$
where k is a number of non-fixed model parameters.

The incorrect calculation appears in two places in omnidir calibration:
* `cv::omnidir::internal::estimateUncertainties()`: [here](https://github.com/opencv/opencv_contrib/blob/e247b680a6bd396f110274b6c214406a93171350/modules/ccalib/src/omnidir.cpp#L1785)
* `cv::omnidir::internal::estimateUncertaintiesStereo()`: [here](https://github.com/opencv/opencv_contrib/blob/e247b680a6bd396f110274b6c214406a93171350/modules/ccalib/src/omnidir.cpp#L1871)

***MVG:*** R. Hartley, A. Zisserman, Multiple View Geometry in Computer Vision, 2004, section 5.1.3, page 134

 - [ ] There is reproducer code and related data files: videos, images, onnx, etc

savuor:
Fixes #3406

This PR can be automatically merged into 4.x and 5.x

### DIFF
--- a/modules/ccalib/src/omnidir.cpp
+++ b/modules/ccalib/src/omnidir.cpp
@@ -1782,12 +1782,14 @@ void cv::omnidir::internal::estimateUncertainties(InputArrayOfArrays objectPoint
 
     Mat sigma_x;
     meanStdDev(reprojError.reshape(1,1), noArray(), sigma_x);
-    sigma_x *= sqrt(2.0*(double)reprojError.total()/(2.0*(double)reprojError.total() - 1.0));
-    double s = sigma_x.at<double>(0);
 
     Mat _JTJ_inv, _JTE;
     computeJacobian(objectPoints, imagePoints, parameters, _JTJ_inv, _JTE, flags, 0.0);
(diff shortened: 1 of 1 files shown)


asmorkalov (approved):
:+1:

PR #3446, opened 2023-02-27
changed CAP_OPENNI to CAP_OPENNI2 flag

vovka643:
Because CAP_OPENNI flag is obsolete. It was removed from openCV. 


alalek:
This patch should go 4.x branch as related functionality is dropped there.

This 5.x PR (branch) is needed for testing the main PR only.
Do **NOT** merge.

PR #3462, opened 2023-03-21
text: change default char_whitelist parameter.

Kumataro:
Fix https://github.com/opencv/opencv_contrib/issues/3457

This patch changes char_whitelist default from [0-9a-zA-Z] to all characters.

Current setting text module missrecognizes in CJK and other languages in default because it accepts only [0-9a-zA-Z]. 

And in "eng" language, wordlist contains some characters outside of [0-9a-zA-Z].
Those words are not recognized with default parameters.

So to disable whitelist is better than using "0-9a-zA-Z". 


n0099:
LGTM

asmorkalov (approved):

PR #3468, opened 2023-04-05
`cudacodec::VideoReader`: fix nv12 to bgr/bgra/grey conversion

cudawarped:
Fixes https://github.com/opencv/opencv_contrib/issues/3458.

Dependent on https://github.com/opencv/opencv_extra/pull/1051

### DIFF
--- a/modules/cudacodec/include/opencv2/cudacodec.hpp
+++ b/modules/cudacodec/include/opencv2/cudacodec.hpp
@@ -310,7 +310,7 @@ enum DeinterlaceMode
 struct CV_EXPORTS_W_SIMPLE FormatInfo
 {
     CV_WRAP FormatInfo() : nBitDepthMinus8(-1), ulWidth(0), ulHeight(0), width(0), height(0), ulMaxWidth(0), ulMaxHeight(0), valid(false),
-        fps(0), ulNumDecodeSurfaces(0) {};
+        fps(0), ulNumDecodeSurfaces(0), videoFullRangeFlag(false) {};
 
     CV_PROP_RW Codec codec;
     CV_PROP_RW ChromaFormat chromaFormat;
@@ -329,6 +329,7 @@ struct CV_EXPORTS_W_SIMPLE FormatInfo
     CV_PROP_RW cv::Size targetSz;//!< Post-processed size of the output frame.
     CV_PROP_RW cv::Rect srcRoi;//!< Region of interest decoded from video source.
     CV_PROP_RW cv::Rect targetRoi;//!< Region of interest in the output frame containing the decoded frame.
+    CV_PROP_RW bool videoFullRangeFlag;//!< Output value indicating if the black level, luma and chroma of the source are represented using the full or limited range (AKA TV or "analogue" range) of values as defined in Annex E of the ITU-T Specification.  Internally the conversion from NV12 to BGR obeys ITU 709.
 };
 
 /** @brief cv::cudacodec::VideoReader generic properties identifier.
--- a/modules/cudacodec/src/video_reader.cpp
+++ b/modules/cudacodec/src/video_reader.cpp
@@ -53,27 +53,54 @@ Ptr<VideoReader> cv::cudacodec::createVideoReader(const Ptr<RawVideoSource>&, co
 
 #else // HAVE_NVCUVID
 
-void nv12ToBgra(const GpuMat& decodedFrame, GpuMat& outFrame, int width, int height, cudaStream_t stream);
+void nv12ToBgra(const GpuMat& decodedFrame, GpuMat& outFrame, int width, int height, const bool videoFullRangeFlag, cudaStream_t stream);
 bool ValidColorFormat(const ColorFormat colorFormat);
 
-void videoDecPostProcessFrame(const GpuMat& decodedFrame, GpuMat& outFrame, int width, int height, const ColorFormat colorFormat,
+void cvtFromNv12(const GpuMat& decodedFrame, GpuMat& outFrame, int width, int height, const ColorFormat colorFormat, const bool videoFullRangeFlag,
     Stream stream)
 {
+    CV_Assert(decodedFrame.cols == width && decodedFrame.rows == height * 1.5f);
     if (colorFormat == ColorFormat::BGRA) {
-        nv12ToBgra(decodedFrame, outFrame, width, height, StreamAccessor::getStream(stream));
+        nv12ToBgra(decodedFrame, outFrame, width, height, videoFullRangeFlag, StreamAccessor::getStream(stream));
     }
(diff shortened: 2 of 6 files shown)


asmorkalov:
I see build issue with my CUDA 10.2 on Ubuntu:
```
      ^~~~~~~~~~~
In file included from /home/alexander/Projects/OpenCV/opencv_contrib/modules/cudacodec/src/precomp.hpp:57:0,
                 from /home/alexander/Projects/OpenCV/opencv_contrib/modules/cudacodec/src/video_reader.cpp:43:
/home/alexander/Projects/OpenCV/opencv_contrib/modules/cudacodec/src/video_reader.cpp:76:25: error: ‘nppiNV12ToBGR_709CSC_8u_P2C3R_Ctx’ was not declared in this scope
             nppSafeCall(nppiNV12ToBGR_709CSC_8u_P2C3R_Ctx(pSrc, decodedFrame.step, outFrame.data, outFrame.step, oSizeROI, nppStreamCtx));
                         ^
/home/alexander/Projects/OpenCV/opencv-master/modules/core/include/opencv2/core/private.cuda.hpp:162:52: note: in definition of macro ‘nppSafeCall’
 #define nppSafeCall(expr)  cv::cuda::checkNppError(expr, __FILE__, __LINE__, CV_Func)
                                                    ^~~~
/home/alexander/Projects/OpenCV/opencv_contrib/modules/cudacodec/src/video_reader.cpp:76:25: note: suggested alternative: ‘nppiNV12ToBGR_709HDTV_8u_P2C3R_Ctx’
             nppSafeCall(nppiNV12ToBGR_709CSC_8u_P2C3R_Ctx(pSrc, decodedFrame.step, outFrame.data, outFrame.step, oSizeROI, nppStreamCtx));
```

modules/cudacodec/include/opencv2/cudacodec.hpp
```diff
@@ -329,6 +329,7 @@ struct CV_EXPORTS_W_SIMPLE FormatInfo
     CV_PROP_RW cv::Size targetSz;//!< Post-processed size of the output frame.
     CV_PROP_RW cv::Rect srcRoi;//!< Region of interest decoded from video source.
     CV_PROP_RW cv::Rect targetRoi;//!< Region of interest in the output frame containing the decoded frame.
+    CV_PROP_RW bool videoFullRangeFlag;//!< indicates if the black level, luma and chroma are using the full range 0-255.
```
asmorkalov, line 332: The flag corresponds to BGR 709 HDTV (true) / RGB 709 CSC (false) conversion. I propose to use it naming or at least to define it in documentation string. Also, it makes sense to mention, that the flag is set automatically from stream parser.

cudawarped: > The flag corresponds to BGR 709 HDTV (true) / RGB 709 CSC (false) conversion. I propose to use it naming or at least to define it in documentation string. 

The flag indicates whether the luma and chroma samples are represented using the full range of 8 bit values (0-255) or a smaller range.  From my understanding the full range is usually but not limited to use for TV sources with the smaller range being the most common for video.  This is applicable to BGRA, BGR and Gray conversion not just BGRA.

The name is to be consistent with the definition in [Annex E of the ITU-T Specification](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&ved=2ahUKEwiD776-u6b-AhWyX_EDHc5LBc4QFnoECAoQAQ&url=https%3A%2F%2Fwww.itu.int%2Frec%2Fdologin_pub.asp%3Flang%3De%26id%3DT-REC-H.264-201602-S!!PDF-E%26type%3Ditems&usg=AOvVaw3vL-wKcGo4yatSlDHUtC5q) (video_full_range_flag) and the name given internally in the Nvidia Video Codec SDK (videoFullRangeFlag).  I can add these details to the documentation string but I think the name is valid?

> Also, it makes sense to mention, that the flag is set automatically from stream parser.

Agreed.

asmorkalov: Ok, let's presume naming, but add reference to BGR 709 HDTV / RGB 709 color spaces.

cudawarped:
Updated and tested on Windows 11 built with VS2019 against CUDA toolkit 10.2 and the latest version of Nvidia Video Codec SDK (12.0).  All cudacodec tests passing on RTX 3070 Ti.

modules/cudacodec/src/video_reader.cpp
```diff
@@ -53,27 +53,54 @@ Ptr<VideoReader> cv::cudacodec::createVideoReader(const Ptr<RawVideoSource>&, co
...
+#elif (CUDART_VERSION >= 10100)
         NppStreamContext nppStreamCtx;
         nppSafeCall(nppGetStreamContext(&nppStreamCtx));
         nppStreamCtx.hStream = StreamAccessor::getStream(stream);
-        nppSafeCall(nppiNV12ToBGR_8u_P2C3R_Ctx(pSrc, decodedFrame.step, outFrame.data, outFrame.step, oSizeROI, nppStreamCtx));
+        if (videoFullRangeFlag)
+            nppSafeCall(nppiNV12ToBGR_709HDTV_8u_P2C3R_Ctx(pSrc, decodedFrame.step, outFrame.data, outFrame.step, oSizeROI, nppStreamCtx));
+        else {
+#if (CUDART_VERSION < 11000)
+            CV_LOG_DEBUG(NULL, "Color reproduction may be inaccurate due CUDA version <= 11.0, for better results upgrade CUDA runtime or try ColorFormat::BGRA.");
```
asmorkalov, line 88: I propose to make the check in constructor, and use LOG_INFO to not generate spam. M.b. could you also add own conversion kernel for BGR in the same way as BGRA?

asmorkalov:
Works for me with CUDA 10.2 and Ubuntu 18.04.

modules/cudacodec/src/video_reader.cpp
cudawarped: > I propose to make the check in constructor, and use LOG_INFO to not generate spam.

Unfortunately that information isn't available until a frame has been run through the decoder.  This is something I've wanted to change for a while and was going to add in another PR.  Should I add that change to this PR or simply remove the debug output which is not really necessary?

> M.b. could you also add own conversion kernel for BGR in the same way as BGRA?

Actually I would prefer to remove the BGRA kernel.  Much as I am not a fan of the NPP libs (buggy, performance oscillates depending on the CUDA version etc.), they should be the best option for performance on future hardware + they should get more user testing than the OpenCV CUDA modules.

asmorkalov:
Let's stay it as is then. Let's merge things as is for  now and tune the behavior later.

asmorkalov (approved):

PR #3471, opened 2023-04-16
V4D - a high performance visualization module with GUI

kallaballa:

Pertaining Issue: https://github.com/opencv/opencv/issues/22923
Documentation: https://viel-zu.org/opencv/doxygen/html/d7/dfc/v4d.html

I am creating this (functional) draft to get V4D into review. Please shoot :)


kallaballa:
Builds are failing because GLFW is not installed, but afaik also Viz3D requires it. Could it be added to the dependencies?

PR #3477, opened 2023-04-21
`cudacodec::VideoReader` return luma hist from `nextFrame` if requested

cudawarped:
Histogram data is collected by NVDEC during the decoding process resulting in zero performance penalty, details can be found under [Getting Histogram Data](https://docs.nvidia.com/video-technologies/video-codec-sdk/12.0/nvdec-video-decoder-api-prog-guide/index.html#Getting-histogram-data) in the Nvidia Video SDK docs.

This PR allows that histogram to be requested and if supported returned from `VideoReader::nextFrame`.  Both C++ and python tests have been included.

Additionaly:
1. A utility function `MapHist()` has been included to remap the histogram when `FormatInfo::videoFullRangeFlag == false`
2. Errors in `test_cudacodec.py` have been addressed, this could be performed in an additional PR if necessary?

### DIFF
--- a/modules/cudacodec/include/opencv2/cudacodec.hpp
+++ b/modules/cudacodec/include/opencv2/cudacodec.hpp
@@ -295,7 +295,7 @@ enum ChromaFormat
 
 /** @brief Deinterlacing mode used by decoder.
 * @param Weave Weave both fields (no deinterlacing). For progressive content and for content that doesn't need deinterlacing.
-* Bob Drop one field.
+* @param Bob Drop one field.
 * @param Adaptive Adaptive deinterlacing needs more video memory than other deinterlacing modes.
 * */
 enum DeinterlaceMode
@@ -305,12 +305,22 @@ enum DeinterlaceMode
     Adaptive = 2
 };
 
+/** @brief Utility function demonstrating how to map the luma histogram when FormatInfo::videoFullRangeFlag == false
+    @param hist Luma histogram \a hist returned from VideoReader::nextFrame(GpuMat& frame, GpuMat& hist, Stream& stream).
+    @param histFull Host histogram equivelent to downloading \a hist after calling cuda::calcHist(InputArray frame, OutputArray hist, Stream& stream).
+
+    @note
+    -   This function demonstrates how to map the luma histogram back so that it is equivalent to the result obtained from cuda::calcHist()
+    if the returned frame was colorFormat::GRAY.
+ */
+CV_EXPORTS_W void MapHist(const GpuMat& hist, CV_OUT Mat& histFull);
+
 /** @brief Struct providing information about video file format. :
  */
 struct CV_EXPORTS_W_SIMPLE FormatInfo
 {
-    CV_WRAP FormatInfo() : nBitDepthMinus8(-1), nBitDepthChromaMinus8(-1), ulWidth(0), ulHeight(0), width(0), height(0), ulMaxWidth(0), ulMaxHeight(0), valid(false),
-        fps(0), ulNumDecodeSurfaces(0), videoFullRangeFlag(false) {};
+    CV_WRAP FormatInfo() : nBitDepthMinus8(-1), ulWidth(0), ulHeight(0), width(0), height(0), ulMaxWidth(0), ulMaxHeight(0), valid(false),
+        fps(0), ulNumDecodeSurfaces(0), videoFullRangeFlag(false), enableHistogram(false), nCounterBitDepth(0), nMaxHistogramBins(0){};
 
     CV_PROP_RW Codec codec;
     CV_PROP_RW ChromaFormat chromaFormat;
@@ -331,6 +341,9 @@ struct CV_EXPORTS_W_SIMPLE FormatInfo
     CV_PROP_RW cv::Rect srcRoi;//!< Region of interest decoded from video source.
     CV_PROP_RW cv::Rect targetRoi;//!< Region of interest in the output frame containing the decoded frame.
     CV_PROP_RW bool videoFullRangeFlag;//!< Output value indicating if the black level, luma and chroma of the source are represented using the full or limited range (AKA TV or "analogue" range) of values as defined in Annex E of the ITU-T Specification.  Internally the conversion from NV12 to BGR obeys ITU 709.
+    CV_PROP_RW bool enableHistogram;//!< Flag requesting histogram output if supported. Exception will be thrown when requested but not supported.
+    CV_PROP_RW int nCounterBitDepth;//!< Bit depth of histogram bins if histogram output is requested and supported.
+    CV_PROP_RW int nMaxHistogramBins;//!< Max number of histogram bins if histogram output is requested and supported.
 };
 
 /** @brief cv::cudacodec::VideoReader generic properties identifier.
@@ -375,6 +388,20 @@ public:
     The method throws an Exception if error occurs.
      */
     CV_WRAP virtual bool nextFrame(CV_OUT GpuMat& frame, Stream &stream = Stream::Null()) = 0;
+
+    /** @brief Grabs, decodes and returns the next video frame and frame luma histogram.
+
+    @param [out] frame The video frame.
+    @param [out] histogram Histogram of the luma component of the encoded frame, see note.
+    @param stream Stream for the asynchronous version.
+    @return `false` if no frames have been grabbed.
+
+    If no frames have been grabbed (there are no more frames in video file), the methods return false.
+    The method throws an Exception if error occurs.
+
+    @note Histogram data is collected by NVDEC during the decoding process resulting in zero performance penalty. NVDEC computes the histogram data for only the luma component of decoded output, not on post-processed frame(i.e. when scaling, cropping, etc. applied).  If the source is encoded using a limited range of luma values (FormatInfo::videoFullRangeFlag == false) then the histogram bin values will correspond to to this limited range of values and will need to be mapped to contain the same output as cuda::calcHist().  The MapHist() utility function can be used to perform this mapping on the host if required.
+     */
+    CV_WRAP_AS(nextFrameWithHist) virtual bool nextFrame(CV_OUT GpuMat& frame, CV_OUT GpuMat& histogram, Stream& stream = Stream::Null()) = 0;
 
     /** @brief Returns information about video file format.
     */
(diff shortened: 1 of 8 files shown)


modules/cudacodec/include/opencv2/cudacodec.hpp
```diff
@@ -373,7 +386,31 @@ class CV_EXPORTS_W VideoReader
...
+
+    /** @brief Bindings overload of VideoReader::nextFrame() which additionaly returns the frame luma histogram.
+    @param [out] frame The video frame.
+    @param [out] histogram Histogram of the luma component of the encoded frame, see note.
+    @param stream Stream for the asynchronous version.
+    @note Histogram data is collected by NVDEC during the decoding process resulting in zero performance penalty. NVDEC computes the histogram data for only the luma component of decoded output, not on post-processed frame(i.e. when scaling, cropping, etc. applied).  If the source is encoded using a limited range of luma values (FormatInfo::videoFullRangeFlag == false) then the histogram bin values will correspond to to this limited range of values and will need to be mapped to contain the same output as cuda::calcHist().  The MapHist() utility function can be used to perform this mapping on the host if required.
+     */
+    CV_WRAP bool nextFrameWithHist(CV_OUT GpuMat& frame, CV_OUT GpuMat& histogram, Stream& stream = Stream::Null()) {
+        return nextFrame(frame, histogram, stream);
+    }
```
asmorkalov, line 413: There is WRAP_AS macro that allows to define alternative name in bindings. No need to do it in API manually.

modules/cudacodec/src/video_reader.cpp
```diff
@@ -395,14 +417,26 @@ Ptr<VideoReader> cv::cudacodec::createVideoReader(const String& filename, const
...
+void cv::cudacodec::MapHist(const GpuMat& hist, Mat& histFull) {
+    Mat histHost; hist.download(histHost);
+    histFull.create(histHost.size(), histHost.type());
+    histFull = 0;
+    const float scale = 255.0f / 219.0f;
+    const int offset = 16;
+    for (int iScaled = 0; iScaled < histHost.cols; iScaled++) {
+        const int iHistFull = std::min(std::max(0, static_cast<int>(std::round((iScaled - offset) * scale))), static_cast<int>(histFull.total()) - 1);
+        histFull.at<int>(iHistFull) += histHost.at<int>(iScaled);
+    }
```
asmorkalov, line 439: Why do we need the function in API. Looks like the same result may be achieved with convertTo(): https://docs.opencv.org/4.x/d0/d60/classcv_1_1cuda_1_1GpuMat.html#a656ac1f6a1426527f838c19e2d677dc0. It's done on GPU and could be a part of user code, if it's required. You just need to mention the conversion parameters in  `nextFrame`.

cudawarped: Hi @asmorkalov, I'm not sure I understand.

If `cudacodec::FormatInfo::videoFullRangeFlag == false` the  hardware decoder generated histogram will have bins from 16-219 as the **encoded** luma component can only take these values.  When decoded to a gray image `img`, `img` will have a range of 0-255, therefore the histogram bins need to be mapped to 0-255 to be consistent with the result of calling `calcHist` on `img`.  Is there a way to use an existing function to perform this mapping?

cudawarped:
Looks like the Quadro P2000 doesn't support histogram output.  I'll have to change
```
ASSERT_TRUE(fmt.valid && histAvailable == fmt.histogramEnabled);
```
to a `SkipTestException`.

modules/cudacodec/src/video_reader.cpp
asmorkalov: My fault, I miss-interpreted your code. Please ignore my proposal.

cudawarped:
@asmorkalov Updated python binding and added `SkipTestException` for GPU's which don't support this feature.

cudawarped:
> [ RUN      ] CUDA_Codec/Histogram.Reader/2, where GetParam() = (Quadro P2000, ("highgui/video/big_buck_bunny_full_color_range.h264", true))
> unknown file: Failure
> C++ exception with description "Luma histogram output is not supported for current codec and/or on current device." thrown in the test body.
> [  FAILED  ] CUDA_Codec/Histogram.Reader/2, where GetParam() = (Quadro P2000, ("highgui/video/big_buck_bunny_full_color_range.h264", true)) (115 ms)

Is there a way to skip the test which won't cause the build bot to fail?

asmorkalov:
Test should throw `SkipTestException("Skip message")` exception. It's marked as skipped in GTest report and the message appears in logs and XML.

modules/cudacodec/src/video_decoder.cpp
```diff
@@ -127,15 +123,41 @@ void cv::cudacodec::detail::VideoDecoder::create(const FormatInfo& videoFormat)
...
+                videoFormat.histogramEnabled = false;
+            }
+            else {
+                videoFormat.histogramEnabled = true;
+                videoFormat.nCounterBitDepth = decodeCaps.nCounterBitDepth;
+                videoFormat.nMaxHistogramBins = decodeCaps.nMaxHistogramBins;
+            }
+        }
+        else {
+            CV_LOG_INFO(NULL, "Luma histogram output is not supported for current codec and/or on current device.");
```
cudawarped, line 142: I'm not seeing the log output I would expect when histogram output is requested but not available, should I be able to see it in the build bot output?

asmorkalov:
@cudawarped Thanks a lot for the patch. I apologize for the delay. The solution works well, besides histogram support flag. It's very non-intuitive thing and may cause issues with Java/python/etc bindings. I propose the following:
- restore const reference for parameters. In case if user app initialized several decoders initialization sequence will be broken without const.
- set histogram flag in the same value as provided by user (false by default) and raise exception, if it's not supported by hardware. It implements the same logic as before for all options without exception. It's the same, if I use ancient GPU and tries to decode 12-bit or h265 codec. User may use try-catch and initialize the pipeline without histogram part.
- Implement test with try catch and throw `SkipTestException` if histogram=true does not work.

cudawarped:
> @cudawarped Thanks a lot for the patch. I apologize for the delay. 

@asmorkalov  No worries I can see your busy by the number of issues and pr's that are coming in.

>    - set histogram flag in the same value as provided by user (false by default) and raise exception, if it's not supported by hardware. It implements the same logic as before for all options without exception. It's the same, if I use ancient GPU and tries to decode 12-bit or h265 codec. User may use try-catch and initialize the pipeline without histogram part.
>   -  Implement test with try catch and throw SkipTestException if histogram=true does not work.

I'll clean this up and just allow the SkipTestException for selected video files which are known not to have support on older hardware.

modules/cudacodec/src/video_parser.cpp
```diff
@@ -78,9 +78,18 @@ bool cv::cudacodec::detail::VideoParser::parseVideoData(const unsigned char* dat
     if (rawMode)
         currentFramePackets.push_back(RawPacket(data, size, containsKeyFrame));
 
-    if (cuvidParseVideoData(parser_, &packet) != CUDA_SUCCESS)
-    {
-        CV_LOG_ERROR(NULL, "Call to cuvidParseVideoData failed!");
+    CUresult retVal = CUDA_SUCCESS;
+    try {
```
cudawarped, line 82: @asmorkalov I had to change this because it appears [cuvidParseVideoData() does not propagate error codes](https://forums.developer.nvidia.com/t/errors-not-propagated-to-cuvidparsevideodata/259856) so the only way to signal failure is with an exception.

asmorkalov:
CUDA 10.2 + GF 1080:
```
[ RUN      ] CUDA_Codec/Histogram.Reader/1, where GetParam() = (NVIDIA GeForce GTX 1080, ("highgui/video/big_buck_bunny.h264", true))
[ERROR:4@3.052] global video_parser.cpp:86 parseVideoData OpenCV(4.8.0-dev) /home/alexander/Projects/OpenCV/opencv_contrib/modules/cudacodec/src/video_decoder.cpp:132: error: (-5:Bad argument) Luma histogram output is not supported for current codec and/or on current device. in function 'create'

unknown file: Failure
C++ exception with description "Luma histogram output is not supported for current codec and/or on current device." thrown in the test body.
[  FAILED  ] CUDA_Codec/Histogram.Reader/1, where GetParam() = (NVIDIA GeForce GTX 1080, ("highgui/video/big_buck_bunny.h264", true)) (73 ms)
[ RUN      ] CUDA_Codec/Histogram.Reader/2, where GetParam() = (NVIDIA GeForce GTX 1080, ("highgui/video/big_buck_bunny_full_color_range.h264", true))
[ERROR:5@3.201] global video_parser.cpp:86 parseVideoData OpenCV(4.8.0-dev) /home/alexander/Projects/OpenCV/opencv_contrib/modules/cudacodec/src/video_decoder.cpp:132: error: (-5:Bad argument) Luma histogram output is not supported for current codec and/or on current device. in function 'create'

unknown file: Failure
C++ exception with description "Luma histogram output is not supported for current codec and/or on current device." thrown in the test body.
[  FAILED  ] CUDA_Codec/Histogram.Reader/2, where GetParam() = (NVIDIA GeForce GTX 1080, ("highgui/video/big_buck_bunny_full_color_range.h264", true)) (125 ms)
```

modules/cudacodec/src/video_decoder.cpp
```diff
@@ -136,6 +152,7 @@ void cv::cudacodec::detail::VideoDecoder::create(const FormatInfo& videoFormat)
 #endif
```
asmorkalov, line 152: Need to raise exception for histogram option and old CUDA.

cudawarped:
> CUDA 10.2 + GF 1080:
> 
> ```
> [ RUN      ] CUDA_Codec/Histogram.Reader/1, where GetParam() = (NVIDIA GeForce GTX 1080, ("highgui/video/big_buck_bunny.h264", true))
> [ERROR:4@3.052] global video_parser.cpp:86 parseVideoData OpenCV(4.8.0-dev) /home/alexander/Projects/OpenCV/opencv_contrib/modules/cudacodec/src/video_decoder.cpp:132: error: (-5:Bad argument) Luma histogram output is not supported for current codec and/or on current device. in function 'create'
> 
> unknown file: Failure
> C++ exception with description "Luma histogram output is not supported for current codec and/or on current device." thrown in the test body.
> [  FAILED  ] CUDA_Codec/Histogram.Reader/1, where GetParam() = (NVIDIA GeForce GTX 1080, ("highgui/video/big_buck_bunny.h264", true)) (73 ms)
> [ RUN      ] CUDA_Codec/Histogram.Reader/2, where GetParam() = (NVIDIA GeForce GTX 1080, ("highgui/video/big_buck_bunny_full_color_range.h264", true))
> [ERROR:5@3.201] global video_parser.cpp:86 parseVideoData OpenCV(4.8.0-dev) /home/alexander/Projects/OpenCV/opencv_contrib/modules/cudacodec/src/video_decoder.cpp:132: error: (-5:Bad argument) Luma histogram output is not supported for current codec and/or on current device. in function 'create'
> 
> unknown file: Failure
> C++ exception with description "Luma histogram output is not supported for current codec and/or on current device." thrown in the test body.
> [  FAILED  ] CUDA_Codec/Histogram.Reader/2, where GetParam() = (NVIDIA GeForce GTX 1080, ("highgui/video/big_buck_bunny_full_color_range.h264", true)) (125 ms)
> ```

Isn't this expected behaviour, i.e. `SkipTestException` when histogram output is requested but not supported by the hardware?

modules/cudacodec/src/video_decoder.cpp
cudawarped: Do you mean throw an exception if  `CUDART_VERSION < 9000`.  I'm not sure if disabling histogram output for older versions of the CUDA toolkit is a bit extreme.  As CUDA 9 is over 5 years we could just leave this as it is?

asmorkalov:
`SkipTestException` exception marks the test es skipped, but not failed. The original exception was caught but GTest.

PR #3479, opened 2023-04-24
Fix invalid memory access in wechat_qrcode byte segment decoding

determ1ne:
A malicious crafted image will crash the wechat_qrcode module by invalid memory access.

Sample image:

qr

In `DecodedBitStreamParser::decodeByteSegment`, an attacker can build a qr code with a byte segment, in which the bits count is larger than actually available bits. Code in line 203 will try to read from an invalid memory region and crash the program. An additional sanity check on `count` should be added.

https://github.com/opencv/opencv_contrib/blob/960b3f685f39c0602b8a0dd35973a82ee72b7e3c/modules/wechat_qrcode/src/zxing/qrcode/decoder/decoded_bit_stream_parser.cpp#L197-L208


dddzg:
Thanks a lot. This PR could be contained with #3480 .

PR #3481, opened 2023-04-25
Fix Wechat QRcode bug

WanliZhong:
**Merged with test:** https://github.com/opencv/opencv_extra/pull/1060

This PR is based on https://github.com/opencv/opencv_contrib/pull/3479 and https://github.com/opencv/opencv_contrib/pull/3480 , and also fixes https://github.com/opencv/opencv_contrib/issues/3478

Thanks @zihaomu, @Konano and @GZTimeWalker.


GZTimeWalker:
Maybe we should use a more decent test case instead of "KFCVW50"? 😂

Konano:
This can also trigger the issue.

invalid_count_indicator

The decoded message is `I love OpenCV!!!`

zihaomu:
> Maybe we should use a more decent test case instead of "KFCVW50"? 😂

I think that's Ok.

PR #3485, opened 2023-04-27
Promote cv::VideoCapture properties in VideoReader

asmorkalov:

### DIFF
--- a/modules/cudacodec/src/ffmpeg_video_source.cpp
+++ b/modules/cudacodec/src/ffmpeg_video_source.cpp
@@ -192,6 +192,10 @@ void cv::cudacodec::detail::FFmpegVideoSource::updateFormat(const FormatInfo& vi
 
 bool cv::cudacodec::detail::FFmpegVideoSource::get(const int propertyId, double& propertyVal) const
 {
+    propertyVal = cap.get(propertyId);
+    if (propertyVal != 0.)
+        return true;
+
     CV_Assert(videoCaptureParams.size() % 2 == 0);
     for (std::size_t i = 0; i < videoCaptureParams.size(); i += 2) {
         if (videoCaptureParams.at(i) == propertyId) {
@@ -199,6 +203,7 @@ bool cv::cudacodec::detail::FFmpegVideoSource::get(const int propertyId, double&
             return true;
         }
     }
+
     return false;
 }
 
--- a/modules/cudacodec/test/test_video.cpp
+++ b/modules/cudacodec/test/test_video.cpp
@@ -430,6 +430,19 @@ CUDA_TEST_P(CheckParams, Reader)
     }
 }
 
+CUDA_TEST_P(CheckParams, CaptureProps)
+{
+    std::string inputFile = std::string(cvtest::TS::ptr()->get_data_path()) + "../highgui/video/big_buck_bunny.mp4";
+    cv::Ptr<cv::cudacodec::VideoReader> reader = cv::cudacodec::createVideoReader(inputFile);
+    double width, height, fps;
+    ASSERT_TRUE(reader->get(cv::VideoCaptureProperties::CAP_PROP_FRAME_WIDTH, width));
+    ASSERT_EQ(672, width);
+    ASSERT_TRUE(reader->get(cv::VideoCaptureProperties::CAP_PROP_FRAME_HEIGHT, height));
+    ASSERT_EQ(384, height);
+    ASSERT_TRUE(reader->get(cv::VideoCaptureProperties::CAP_PROP_FPS, fps));
+    ASSERT_EQ(24, fps);
+}
+
 CUDA_TEST_P(CheckDecodeSurfaces, Reader)
 {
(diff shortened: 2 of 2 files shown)


cudawarped:
Nice, I can't for the life of me remember why I would have done it the way but if this PR doesn't trip up on any tests
https://github.com/opencv/opencv_contrib/blob/5efb784670d97c9a942fd2b021d07f34663443ea/modules/cudacodec/src/ffmpeg_video_source.hpp#L74
can also be removed.

asmorkalov:
@cudawarped I reworked tests a bit. Could you take a look again?

modules/cudacodec/src/ffmpeg_video_source.cpp
```diff
@@ -192,14 +192,8 @@ void cv::cudacodec::detail::FFmpegVideoSource::updateFormat(const FormatInfo& vi
 
 bool cv::cudacodec::detail::FFmpegVideoSource::get(const int propertyId, double& propertyVal) const
 {
-    CV_Assert(videoCaptureParams.size() % 2 == 0);
```
cudawarped, line 195: Are `CAP_PROP_OPEN_TIMEOUT_MSEC` and `CAP_PROP_READ_TIMEOUT_MSEC` going to be queryable from `VideoCapture` going forward?  If not should they be stored here so the information is available to the user of the `VideoReader` class?

modules/cudacodec/src/ffmpeg_video_source.cpp
```diff
@@ -192,14 +192,8 @@ void cv::cudacodec::detail::FFmpegVideoSource::updateFormat(const FormatInfo& vi
...
-    for (std::size_t i = 0; i < videoCaptureParams.size(); i += 2) {
-        if (videoCaptureParams.at(i) == propertyId) {
-            propertyVal = videoCaptureParams.at(i + 1);
-            return true;
-        }
-    }
-    return false;
+    propertyVal = cap.get(propertyId);
```
cudawarped, line 195: Is `VideoCapture::get` thread safe?

modules/cudacodec/src/ffmpeg_video_source.cpp
asmorkalov: `CAP_PROP_OPEN_TIMEOUT_MSEC` and `CAP_PROP_READ_TIMEOUT_MSEC` are marked as open-only parameters of `VideoCapture`. It does not return value on get call. In case if open-only parameters are important for reader, I can restore open parameters handling.

modules/cudacodec/src/ffmpeg_video_source.cpp
asmorkalov: There is no such guarantee in OpenCV documentation, but I have not found suspicious places in FFmpeg getProperty implementation that could produce serious issues.

modules/cudacodec/src/ffmpeg_video_source.cpp
cudawarped: > It does not return value on get call. In case if open-only parameters are important for reader, I can restore open parameters handling.

I can't think of a specific reason for storing these two values but I also can't think of a reason why not if its just a simple case of restoring the parameter handling before querying `VideoCapture::get`.

modules/cudacodec/src/ffmpeg_video_source.cpp
cudawarped: > but I have not found suspicious places in FFmpeg getProperty implementation that could produce serious issues.

I thought the same but I wasn't sure if that could change in the future if `VideoCapture::get` adds any additional properties.  Would a mutex be overkill?

asmorkalov: Yes. I do not see use case where it helps.

modules/cudacodec/src/ffmpeg_video_source.cpp
asmorkalov: Restored back.

modules/cudacodec/test/test_video.cpp
```diff
@@ -397,39 +397,35 @@ CUDA_TEST_P(VideoReadRaw, Reader)
 CUDA_TEST_P(CheckParams, Reader)
 {
     std::string inputFile = std::string(cvtest::TS::ptr()->get_data_path()) + "../highgui/video/big_buck_bunny.mp4";
-    {
-        cv::Ptr<cv::cudacodec::VideoReader> reader = cv::cudacodec::createVideoReader(inputFile);
-        double msActual = -1;
-        ASSERT_FALSE(reader->get(cv::VideoCaptureProperties::CAP_PROP_OPEN_TIMEOUT_MSEC, msActual));
-    }
```
cudawarped, line 404: Should the tests which check for `CAP_PROP_OPEN_TIMEOUT_MSEC` etc. be added back as well?

asmorkalov:
@cudawarped I restored create-only parameters handling and related tests. Do you have any opens? May I merge the PR?

cudawarped:
LGTM 👍

PR #3486, opened 2023-04-28
`cudacodec::VideoReader` - ensure `FormatInfo.valid == true` on return from constructor

cudawarped:
As discussed in https://github.com/opencv/opencv_contrib/pull/3468#discussion_r1165613962 it would be useful if `VideoReader::FormatInfo` was valid in the constructor to make validation tasks simpler.

### DIFF
--- a/modules/cudacodec/src/video_decoder.hpp
+++ b/modules/cudacodec/src/video_decoder.hpp
@@ -70,6 +70,7 @@ public:
     void create(const FormatInfo& videoFormat);
     int reconfigure(const FormatInfo& videoFormat);
     void release();
+    bool inited() { AutoLock autoLock(mtx_); return decoder_; }
 
     // Get the codec-type currently used.
     cudaVideoCodec codec() const { return static_cast<cudaVideoCodec>(videoFormat_.codec); }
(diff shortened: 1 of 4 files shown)


modules/cudacodec/src/video_decoder.hpp
```diff
@@ -70,6 +70,7 @@ class VideoDecoder
     void create(const FormatInfo& videoFormat);
     int reconfigure(const FormatInfo& videoFormat);
     void release();
+    bool inited() { AutoLock autoLock(mtx_); return decoder_; }
```
asmorkalov, line 73: `decoder_` is not protected by `mtx_`. Only format modification does. Also decoder_value may be modified in the middle  of `cuvidCreateDecoder` call internally, e.g. just allocated memory for the structure and the check fails.  IMHO the original version was better, even if it's not truly thread safe.

asmorkalov:
I apologize for the large delay in review. OpenCV team moved barcode support to main repo. Please rebase your branch on top of current 4.x to continue development. Otherwise you'll see conflicts with current 4.x in main repo.

modules/cudacodec/src/video_decoder.hpp
cudawarped: What about placing the lock around the decoder creation as well so this version is thread safe?

asmorkalov (approved):
:+1:

PR #3490, opened 2023-05-10
fix(wechat_code): Modify `isnan` for compatibility with -ffast_math.

### LINKED ISSUE #3150: We chat QR Scan crashes with certain images.

- OpenCV => 4.5.4
- Operating System / Platform => Ubuntu 20.04
- Compiler => Visual Studio Code 1.63.2

##### Detailed description

I've implemented the WeChat QR Code Scanner in a similar fashion to the one included in guides such as [this one](https://learnopencv.com/wechat-qr-code-scanner-in-opencv/) and [this one](http://www.wechatcv.com/) (C++ implementations). The QR scanner seems to work in general, but we have noticed occasions where  the process just crashes silently. After launching the script in debugging mod we have cached the following errors:

`Object "/usr/local/lib/libopencv_wechat_qrcode.so.4.5", at 0x7fae230675ca, in cv::wechat_qrcode::WeChatQRCode::detectAndDecode[abi:cxx11](cv::_InputArray const&, cv::_OutputArray const&)`

`   Object "/usr/local/lib/libopencv_wechat_qrcode.so.4.5", at 0x7fae23066fdb, in cv::wechat_qrcode::WeChatQRCode::Impl::decode[abi:cxx11](cv::Mat const&, std::vector<cv::Mat, std::allocator<cv::Mat> >&, std::vector<cv::Mat, std::allocator<cv::Mat> >&)`

`Thread 1 "qr_scan" received signal SIGSEGV, Segmentation fault.
0x00007ffff7cc1097 in zxing::qrcode::Detector::sizeOfBlackWhiteBlackRun(int, int, int, int) () from target:/usr/local/lib/libopencv_wechat_qrcode.so.4.5
`

##### Steps to reproduce

For reproducing, you can use the scanner method `detectAndDecode` with the following image. We have found that this image manages to crash the scanner:
aiuda

Konano:
fix #3150
Merge with https://github.com/opencv/opencv/pull/23881

Reference: https://stackoverflow.com/questions/7263404/mingw32-stdisnan-with-ffast-math

### DIFF
--- a/modules/wechat_qrcode/src/zxing/zxing.hpp
+++ b/modules/wechat_qrcode/src/zxing/zxing.hpp
@@ -54,8 +54,14 @@ typedef unsigned char boolean;
 #include <cmath>
 
 namespace zxing {
-inline bool isnan(float v) { return std::isnan(v); }
-inline bool isnan(double v) { return std::isnan(v); }
+inline bool isnan(float v) { 
+    union { float v; uint32_t x; } u = { v };
+    return (u.x & 0x7fffffffu) > 0x7f800000u;
+}
+inline bool isnan(double v) { 
+    union { double v; uint64_t x; } u = { v };
+    return (u.x & ~0x8000000000000000uLL) > 0x7ff0000000000000uLL;
+}
 inline float nan() { return std::numeric_limits<float>::quiet_NaN(); }
 }  // namespace zxing
 
--- a/modules/wechat_qrcode/src/zxing/zxing.hpp
+++ b/modules/wechat_qrcode/src/zxing/zxing.hpp
@@ -11,6 +11,7 @@
 #ifndef __ZXING_ZXING_HPP__
 #define __ZXING_ZXING_HPP__
 
+#include "opencv2/core/fast_math.hpp"
 
 #define COUNTER_TYPE short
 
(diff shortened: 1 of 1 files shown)


modules/wechat_qrcode/src/zxing/zxing.hpp
```diff
@@ -54,8 +54,14 @@ typedef unsigned char boolean;
 #include <cmath>
 
 namespace zxing {
-inline bool isnan(float v) { return std::isnan(v); }
-inline bool isnan(double v) { return std::isnan(v); }
+inline bool isnan(float v) { 
+    union { float v; uint32_t x; } u = { v };
+    return (u.x & 0x7fffffffu) > 0x7f800000u;
+}
+inline bool isnan(double v) {
```
opencv-alalek, line 61: Lets reuse OpenCV implementation: https://github.com/opencv/opencv/blob/4.7.0/modules/core/include/opencv2/core/fast_math.hpp#L284

asmorkalov:
@Konano Friendly reminder.

modules/wechat_qrcode/src/zxing/zxing.hpp
```diff
@@ -54,8 +54,14 @@ typedef unsigned char boolean;
...
-inline bool isnan(float v) { return std::isnan(v); }
-inline bool isnan(double v) { return std::isnan(v); }
+inline bool isnan(float v) { 
+    union { float v; uint32_t x; } u = { v };
+    return (u.x & 0x7fffffffu) > 0x7f800000u;
+}
+inline bool isnan(double v) { 
+    union { double v; uint64_t x; } u = { v };
+    return (u.x & ~0x8000000000000000uLL) > 0x7ff0000000000000uLL;
+}
```
Konano, line 64:
```suggestion
inline bool isnan(float v) { return cvIsNaN(v); }
inline bool isnan(double v) { return cvIsNaN(v); }
```

I called `cvIsNaN()` directly but it doesn't seem to work, it still crashed in the local test.

Konano: @asmorkalov @opencv-alalek Need help.

asmorkalov (approved):
:+1:

Konano:
@asmorkalov @opencv-alalek Any new progress?

PR #3494, opened 2023-05-19
Several fixes for Python tests for CUDA modules

asmorkalov:


asmorkalov:
@cudawarped Could you take a look?

cudawarped:
I'll take a look next week.

modules/cudaoptflow/misc/python/test/test_nvidiaopticalflow.py
```diff
@@ -22,15 +22,15 @@ def test_calc(self):
         cuMat1 = cv.cuda_GpuMat(npMat1)
         cuMat2 = cv.cuda_GpuMat(npMat2)
         try:
-            nvof = cv.cuda_NvidiaOpticalFlow_1_0.create(cuMat1.shape[1], cuMat1.shape[0], 5, False, False, False, 0)
+            nvof = cv.cuda_NvidiaOpticalFlow_1_0.create((npMat1.shape[1], npMat1.shape[0]))
             flow = nvof.calc(cuMat1, cuMat2, None)
```
cudawarped, line 26: I think this also needs to be changed to
```
flow, cost = nvof.calc(cuMat1, cuMat2, None)
self.assertTrue(flow.size()[1] > 0 and flow.size()[0] > 0)
flowUpSampled = nvof.upSampler(flow, npMat1.shape[1], npMat1.shape[0], nvof.getGridSize(), None)
```
or similar.

PR #3496, opened 2023-05-24
Added .editorconfig

mshabunin:
Copy from the main OpenCV repository (https://github.com/opencv/opencv/blob/4.x/.editorconfig)


asmorkalov (approved):
:+1:

PR #3497, opened 2023-05-24
Moved barcode module to opencv/objdetect

mshabunin:


mshabunin:
@asmorkalov , @AleksandrPanov , should `barcode` test also be removed from infra scripts?

PR #3513, opened 2023-06-22
`cudacodec::VideoReader` - fix CUDART_VERSION tests

cudawarped:
The minor version which is compared to the `CUDART_VERSION` inside `cudacodec::VideoReader` is an order of magnitude too big.

### DIFF
--- a/modules/cudacodec/src/video_decoder.cpp
+++ b/modules/cudacodec/src/video_decoder.cpp
@@ -97,10 +97,10 @@ void cv::cudacodec::detail::VideoDecoder::create(const FormatInfo& videoFormat)
                             cudaVideoCodec_UYVY     == _codec;
 
 #if defined (HAVE_CUDA)
-#if (CUDART_VERSION >= 6500)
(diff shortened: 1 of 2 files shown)


asmorkalov:
@cudawarped the patch looks reasonable. Please enable more test cases then.

cudawarped:
> @cudawarped the patch looks reasonable. Please enable more test cases then.

@asmorkalov this shouldn't affect the test cases that were failing in [the other PR](https://github.com/opencv/opencv_contrib/pull/3488). 

From that PR I noticed that your error message was coming from the wrong place.  It should have been coming from the execution of functions which were only available in CUDA >= 10.2 but it was coming from those that were available in CUDA >=10.1 (the reverse would have been a problem but would have resulted in an error at compile time).  

The failure when decoding `sample_322x242_15frames.yuv420p.mpeg2video.mp4` should still exist as I suspect it is because the frame supplied to `cvtFromNv12` is corrupt.

asmorkalov (approved):
:+1:

PR #3516, opened 2023-06-29
`cuda`: add `moments`

cudawarped:
As requested by @cv3d this PR continues the work of adding `cuda::moments` started in https://github.com/opencv/opencv_contrib/pull/3500.

This PR introduces two functions for calculating image moments on the device, one asynchronous (`cuda::spatialMoments`) which only calculates the spatial moments and leaves the result on the device and one synchronous (`cuda::moments`) which calculates the spatial moments on the device, downloads the result to the host and then uses `cv::Moments()` to calculate the remaining centralized and normalized image moments.

The existing `cv::moments()` function returns the result in double precision due to the magnitude of the calculated image moments.  Unfortunately double precision performance on Nvidia GPU's is at least 32x (depending on the generation) slower than single precision making the perfomance of this asynchrounous device side version only marginally (~3x depending on CPU/GPU) faster than its CPU counterpart.  As a result the user can increase the performance by choosing a single precision result, and/or reducing the number of spatial image moments to calculate.

To give an indication of the performace resulting from the various parameter combinations, the table below has some preliminary timings of the CPU (i72700H) version vs the kernel execution time on a RTX 3070 Ti Laptop GPU.  The source image was uchar and contained a circle in the centre of radius 0.9*width.

  | CPU (us)|   |   | GPU (us)|   |   |  
-- | -- | -- | -- | -- | -- | -- | --
Size | `binaryImage == false`| `binaryImage == true`| GPU moments type | 1st order  | 2nd order | 3rd order  | speed up
640x480 | 117 | 472 | float | 4.61 | 3.71 | 3.81 | 31
640x480 | 117 | 472 | double | 13.63 | 26.05 | 39.62 | 3
1280x720 | 327 | 1288 | float | 5.95 | 5.09 | 5.09 | 64
1280x720 | 327 | 1288 | double | 21.25 | 45.73 | 70.79 | 5
1920x1080 | 572 | 4413 | float | 8.64 | 7.49 | 8.38 | 68
1920x1080 | 572 | 4413 | double | 55.68 | 128.55 | 204.77 | 3
4096x2160 | 2699 | 14674 | float | 22.66 | 22.62 | 24.26 | 111
4096x2160 | 2699 | 14674 | double | 165.89 | 388.74 | 628.62 | 4

**Note:**
1) The results on the GPU are with `binaryImage == false` as the performace is not affected by this flag, unlike on the CPU where the performance with `binaryImage == true` is significantly slower.
2) The GPU times measure the kernel execution only, not the launch latency.
3) The speed up column compares the faster CPU version (`binaryImage == false`) with the kernel execution time when calculating all the spatial image moments.

@cv3d I hope this is what you wanted?  Can you please take a look and let me know if this works correctly on the data you are using?  Additionally if you have any comments suggestions please let me know.


cv3d:
@cudawarped Thanks a lot for your efforts. I have performed preliminary tests, and it seems satisfying my use case. I might get back to you later on with more details or comments once I get better chance to dig on this further.

@alalek @asmorkalov Can you please consider this PR?

cv3d:
@cudawarped Perhaps an idea for a follow PR, but I wonder if we can have something powerful around binary images.  For example, with a multi-labels image, maybe we can compute moments of all shapes?
```
std::vector<cv::Moments> momentsOfLabels(InputArray src, ...);
```

Actually, we already have `cv::cuda::connectedComponents` (thanks @stal12), so having `cv::cuda::connectedComponentsWithStats` would be great, perhaps with expanded `ConnectedComponentsTypes` to contain more moments.

If you think this deserves pushing for, then maybe I can open an issue, and tag some people.

cudawarped:
> Actually, we already have `cv::cuda::connectedComponents` (thanks @stal12), so having `cv::cuda::connectedComponentsWithStats` would be great, perhaps with expanded `ConnectedComponentsTypes` to contain more moments.

I agree but it usually envolves several steps making it quite slow, i.e.
1) Getting a list (histogram if you want pixel counts or otherwise) of all the marker labels.  CCL usually labels each blob with the lowest image index of a connected region.
2) Compressing the list.
3) Relabeling the image using compressed labels.  We could stop here for your moments suggestion.
4) Processing the relabeled image calculating the bounds of each blob.

NPP has serveral routines to do this although I have found them to be buggy and very slow plus you need to use there slow labeling routine first before you can use [CompressedMarkerLabelsInfo](https://docs.nvidia.com/cuda/npp/group__image__filter__compresssed__marker__labels__info.html).

It would be interesting but unfortunately I don't have time in the near future to look at this.

cv3d:
@alalek @asmorkalov This PR is ready for a while now.
Thanks for your consideration~

cudawarped:
@asmorkalov Thank you for running the CI, sorry it looks like I've left the wrong sanity check in for the spatial moments test I'll address this when I can.

cudawarped:
@asmorkalov Fixed the sanity check, ready for your review.

cudawarped:
@asmorkalov Should this be included in 4.9.0?

cudawarped:
> AttributeError: module 'cv2.cuda' has no attribute 'MomentsOrder_THIRD'

@asmorkalov Should be fixed now, I changed from `enum class` to `enum` so I could use [CV_ENUM](https://github.com/opencv/opencv/blob/70e55fd7c6e88398a2f623aac5b97d0c05fdcbf7/modules/ts/include/opencv2/ts/ts_perf.hpp#L105) for clarity in the tests but forgot to retest in Python, sorry.

asmorkalov (approved):
:+1:

PR #3517, opened 2023-06-30
[cuda][cudacodec] Address build warnings

cudawarped:
`cudacodec` produces a number of build warnings, mainly introduced by myself :facepalm:.  

PR removes this pollution from the build output.

### DIFF
--- a/modules/cudacodec/include/opencv2/cudacodec.hpp
+++ b/modules/cudacodec/include/opencv2/cudacodec.hpp
@@ -385,6 +385,8 @@ struct CV_EXPORTS_W_SIMPLE FormatInfo
  */
 class CV_EXPORTS_W NVSurfaceToColorConverter {
 public:
+    virtual ~NVSurfaceToColorConverter() {}
+
     /** @brief Performs the conversion from the raw YUV Surface output from VideoReader to the requested color format. Use this function when you want to convert the raw YUV Surface output from VideoReader to more than one color format or you want both the raw Surface output in addition to a color frame.
      * @param yuv The raw YUV Surface output from VideoReader see @ref SurfaceFormat.
      * @param color The converted frame.
@@ -550,8 +552,7 @@ public:
      - Out: Value of the property.
     @return `true` unless the property is not supported.
     */
-    virtual bool get(const VideoReaderProps propertyId, double& propertyVal) const = 0;
-    CV_WRAP virtual bool getVideoReaderProps(const VideoReaderProps propertyId,  CV_OUT double& propertyValOut, double propertyValIn = 0) const = 0;
+    CV_WRAP_AS(getVideoReaderProps) virtual bool get(const VideoReaderProps propertyId, CV_OUT size_t& propertyVal) const = 0;
 
     /** @brief Retrieves the specified property used by the VideoSource.
 
@@ -562,6 +563,43 @@ public:
     @return `true` unless the property is unset set or not supported.
      */
     CV_WRAP virtual bool get(const int propertyId, CV_OUT double& propertyVal) const = 0;
+
+    /** @brief Determine whether the raw package at \a idx contains encoded data for a key frame.
+
+    @param idx Index of the encoded package to check.
+
+    @returns `true` if the raw package at \a idx contains encoded data for a key frame and `false` otherwise.
+
+    @note A typical use case is deciding to archive live video after streaming has been initialized. In this scenario you would not want to write any encoded frames before a key frame. This is simulated in the example below where VideoReader is initialized without enabling raw mode
+    ```
+    VideoReaderInitParams params;
+    params.rawMode = false;
+    Ptr<VideoReader> reader = createVideoReader(rtspUrl, {}, params);
+    ```
+    and then at some point in the future raw mode is enabled to enable the footage to be archived
+    ```
+    reader->set(VideoReaderProps::PROP_RAW_MODE, true);
+    ```
+    Because raw mode has been enabled mid stream the first raw package retrieved now is not guaranteed to contain a key frame. To locate the next raw package containing a key frame rawPackageHasKeyFrame() can then be used as shown below.
+    ```
+    double iRawPacketBase = -1;
+    reader->get(VideoReaderProps::PROP_RAW_PACKAGES_BASE_INDEX, iRawPacketBase);
+    GpuMat frame;
+    while (reader->nextFrame(frame)) {
+        double nRawPackets = -1;
+        reader->get(VideoReaderProps::PROP_NUMBER_OF_RAW_PACKAGES_SINCE_LAST_GRAB, nRawPackets);
+        for (int iRawPacketToWrite = static_cast<int>(iRawPacketBase); iRawPacketToWrite < static_cast<int>(iRawPacketBase + nRawPackets); iRawPacketToWrite++) {
+            if (reader->rawPackageHasKeyFrame(iRawPacketToWrite)) {
+                Mat packageToWrite;
+                reader->retrieve(packageToWrite, iRawPacketToWrite);
+                ...
+            }
+        }
+    }
+    ```
+    \sa retrieve
+    */
+    CV_WRAP virtual bool rawPackageHasKeyFrame(const size_t idx) const = 0;
 };
 
 /** @brief Interface for video demultiplexing. :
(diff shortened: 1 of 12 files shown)


modules/cudacodec/src/video_reader.cpp
```diff
@@ -345,14 +351,14 @@ namespace
...
-            propertyVal = rawPackets.size();
+            propertyVal = static_cast<double>(rawPackets.size());
             return true;
         case VideoReaderProps::PROP_RAW_MODE:
             propertyVal = videoSource_->RawModeEnabled();
             return true;
         case VideoReaderProps::PROP_LRF_HAS_KEY_FRAME: {
-            const int iPacket = propertyVal - rawPacketsBaseIdx;
-            if (videoSource_->RawModeEnabled() && iPacket >= 0 && iPacket < rawPackets.size()) {
+            const int iPacket = static_cast<int>(propertyVal) - rawPacketsBaseIdx;
```
asmorkalov, line 360: `propertyVal` is function output and could contain any value here. Looks like it's a bug.

modules/cudacodec/src/video_source.hpp
```diff
@@ -57,7 +57,7 @@ class VideoSource
 
     virtual FormatInfo format() const = 0;
     virtual void updateFormat(const FormatInfo& videoFormat) = 0;
-    virtual bool get(const int propertyId, double& propertyVal) const { return false; }
```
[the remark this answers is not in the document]
cudawarped: It was for `CuvidVideoSource` but I agree it will be much cleaner to move this to cuvid_video_source.h especially as `CUvideosource` is marked for removal in the next version of the Nvidia Video Codec SDK.

asmorkalov:
Tested manually with Ubuntu 18.04, CUDA 10.2 and GF 1080.

modules/cudacodec/src/video_reader.cpp
cudawarped: Its output and input, not sure what I was thinking here will alter this so that it makes more sense.

modules/cudacodec/src/NvEncoder.h
```diff
@@ -252,7 +252,7 @@ class NvEncoder
     /**
     * @brief This function returns the completion event.
     */
-    void* GetCompletionEvent(uint32_t eventIdx) { return (m_vpCompletionEvent.size() == m_nEncoderBuffer) ? m_vpCompletionEvent[eventIdx] : nullptr; }
+    void* GetCompletionEvent(uint32_t eventIdx) { return (m_vpCompletionEvent.size() == static_cast<size_t>(m_nEncoderBuffer)) ? m_vpCompletionEvent[eventIdx] : nullptr; }
```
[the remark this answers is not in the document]
cudawarped: We could but it may cause underflow errors without careful checking?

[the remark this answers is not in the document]
asmorkalov: The value is initialized at NvEncoder.cpp:220, but is not checked right now. I propose to migrate to size_t and add a check there.

modules/cudacodec/src/video_reader.cpp
asmorkalov: Please document "input" logic then.

modules/cudacodec/src/NvEncoder.h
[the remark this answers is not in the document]
asmorkalov: ok! Agreed!

modules/cudacodec/src/video_reader.cpp
cudawarped: @asmorkalov I have added a new method `rawPackageHasKeyFrame()` to avoid the input output argument confusion in the `get()` method but now I am wondering if it is the right time to make some more alterations.  Specifically those below which I "wrongly" introduced to mirror the `VideoCapture` class:

- remove `grab()/retrieve()` for the decoded frame.  This combination is completely unecessary `grab()` doesn't save any processing, it only remove the need to pass a `GpuMat` when you don't want to retrieve the decoded frame.
- `propertyVal` of `double` in `get(const VideoReaderProps propertyId, CV_OUT double& propertyVal)` results in a lot of needless casting when the value will always be an integer.

What do you think?

cudawarped: @asmorkalov I have pushed changes to demonstrate the removal of `grab()` and making `propertyVal` a `size_t` what do you think?

modules/cudacodec/include/opencv2/cudacodec.hpp
```diff
@@ -380,59 +378,18 @@ class CV_EXPORTS_W VideoReader
...
-    @param stream Stream for the asynchronous version.
-    @return `true` (non-zero) in the case of success.
-
-    The method/function grabs the next frame from video file or camera and returns true (non-zero) in
-    the case of success.
-
-    The primary use of the function is for reading both the encoded and decoded video data when rawMode is enabled.  With rawMode enabled
-    retrieve() can be called following grab() to retrieve all the data associated with the current video source since the last call to grab() or the creation of the VideoReader.
-     */
-    CV_WRAP virtual bool grab(Stream& stream = Stream::Null()) = 0;
```
asmorkalov, line 394: Why do you removed grab? It breaks compatibility and it's still mentioned in other methods.

cudawarped: > now I am wondering if it is the right time to make some more alterations. Specifically those below which I "wrongly" introduced to mirror the VideoCapture class:
> 
>    - remove grab()/retrieve() for the decoded frame. This combination is completely unecessary grab() doesn't save any processing, it only remove the need to pass a GpuMat when you don't want to retrieve the decoded frame.
>    -  propertyVal of double in get(const VideoReaderProps propertyId, CV_OUT double& propertyVal) results in a lot of needless casting when the value will always be an integer.
> 
> What do you think?

I removed `grab()` in that commit so you could take a look.  It was an unecessary addition which adds a lot of code mess to `VideoCapture` and was "wrongly" introduced to mirror `VideoCapture` without actually mirroring the funcitonality.  I can back it out and place it in another PR if you think agree with its removal?

cudawarped: @asmorkalov Added `grab()` back into to this PR.

cudawarped:
@asmorkalov Updated PR:
- Rebased onto 4.x
- Addressed new warnings
- Kept the new method `rawPackageHasKeyFrame()` to avoid the input output argument confusion in the `get()` method

Can you review when you have chance please?

cudawarped:
@asmorkalov Rebased onto 4.x.

asmorkalov:
```
[446/459] Building CXX object modules/cudacodec/CMakeFiles/opencv_cudacodec.dir/src/video_decoder.cpp.o
In function ‘const char* GetVideoCodecString(cudaVideoCodec)’,
    inlined from ‘void cv::cudacodec::detail::VideoDecoder::create(const cv::cudacodec::FormatInfo&)’ at /mnt/Projects/Projects/opencv_contrib/modules/cudacodec/src/video_decoder.cpp:180:9:
/mnt/Projects/Projects/opencv_contrib/modules/cudacodec/src/video_decoder.cpp:79:37: warning: array subscript 1230591318 is above array bounds of ‘GetVideoCodecString(cudaVideoCodec)::<unnamed struct> [18]’ [-Warray-bounds=]
   79 |             return aCodecName[eCodec].name;
      |                    ~~~~~~~~~~~~~~~~~^
/mnt/Projects/Projects/opencv_contrib/modules/cudacodec/src/video_decoder.cpp: In member function ‘void cv::cudacodec::detail::VideoDecoder::create(const cv::cudacodec::FormatInfo&)’:
/mnt/Projects/Projects/opencv_contrib/modules/cudacodec/src/video_decoder.cpp:53:7: note: while referencing ‘aCodecName’
   53 |     } aCodecName[] = {
      |       ^~~~~~~~~~
[459/459] Linking CXX executable bin/opencv_test_cudacodec
```

cudawarped:
@asmorkalov Fixed.

modules/cudacodec/src/video_decoder.cpp
```diff
@@ -76,7 +76,7 @@ static const char* GetVideoCodecString(cudaVideoCodec eCodec) {
     }
     for (int i = cudaVideoCodec_NumCodecs + 1; i < sizeof(aCodecName) / sizeof(aCodecName[0]); i++) {
         if (eCodec == aCodecName[i].eCodec) {
-            return aCodecName[eCodec].name;
```
asmorkalov: Need to check.

PR #3558, opened 2023-09-06
Add missing std namespace qualifiers

georgthegreat:
This is opencv_contrib part of [this PR](https://github.com/opencv/opencv/pull/24232) to opencv itself.

NB: we only build superres, optflow and ximgproc modules, so I can not be sure if all similar problems have been fixed.

### DIFF
--- a/modules/ximgproc/src/lsc.cpp
+++ b/modules/ximgproc/src/lsc.cpp
@@ -821,7 +821,7 @@ inline void SuperpixelLSCImpl::PostEnforceLabelConnectivity( int threshold )
 
           (*Stmp).Neighbor.insert( (*Stmp).Neighbor.end(), (*S).Neighbor.begin(), (*S).Neighbor.end() );
 
-          sort( (*Stmp).Neighbor.begin(), (*Stmp).Neighbor.end() );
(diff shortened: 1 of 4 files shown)


modules/ximgproc/src/lsc.cpp
```diff
@@ -821,7 +821,7 @@ inline void SuperpixelLSCImpl::PostEnforceLabelConnectivity( int threshold )
 
           (*Stmp).Neighbor.insert( (*Stmp).Neighbor.end(), (*S).Neighbor.begin(), (*S).Neighbor.end() );
 
-          sort( (*Stmp).Neighbor.begin(), (*Stmp).Neighbor.end() );
```
asmorkalov, line 824: Stmp->Neighbor.begin(), Stmp->Neighbor.end()

georgthegreat: I would like to avoid perfoming random refactorings in between.

This is almost automated change.

georgthegreat:
@asmorkalov, can we merge this without the above refactoring?

asmorkalov (approved):
:+1:

PR #3563, opened 2023-09-22
Update CMakeLists.txt to support static linking to HDF5

### LINKED ISSUE #3560: Statically linking to HDF5 library raises linker errors
- OpenCV => 4.8.0
- Compiler => Visual Studio 2022

##### Detailed description

We recently tried to compile OpenCV's HDF module, which only works by default while linking to a shared HDF5 libary. However, as it's a single-module dependendy, statically linking to HDF5 is advantageous here. Still, this results in linker errors caused by hardcoded configuration options for shared library linking. In particular, the flags inside the Windows branch in the `CMakeLists.txt` defined by
```
add_definitions(-DH5_BUILT_AS_DYNAMIC_LIB -D_HDF5USEDLL_)
```
result in linker errors. After removing them, only a linker error from the missing `ShLwApi.Lib` is raised.
These two issues can be resolved by extending above line to
```
if(HDF5_C_LIBRARY MATCHES "libhdf5.lib$")
    set(HDF5_LIBRARIES "${HDF5_LIBRARIES};ShLwApi.Lib")
else()
    add_definitions(-DH5_BUILT_AS_DYNAMIC_LIB -D_HDF5USEDLL_)
endif()
```
and compiling works with both, shared and static library linking to HDF5.
Still, we wanted to discuss this as an issue first before creating a respective PR.

##### Steps to reproduce

After cloning OpenCV's main and contrib repositories, run CMake with the following flags:
- `OPENCV_EXTRA_MODULES_PATH` pointing to the contrib's `modules` directory
- `HDF5_INCLUDE_DIRS` pointing to the `include` directory inside the HDF5 path (generated by its CMake install)
- `HDF5_C_LIBRARY` pointing to the static `lib/libhdf5.lib` in the same path

Now, building of the HDF module raises a bunch of linker error.

 - [X] I checked the problem with documentation, FAQ, open issues,
 - [X] I updated to the latest OpenCV version and the issue is still there
 - [ ] There is reproducer code and related data files: videos, images, onnx, etc

CSBVision:
Proposal to fix #3560.
In particular, the required library to fix the linker error while statically linking to HDF5 is included. Besides this, HDF5 might also depend on ZLIB. For this reasoning, it will be added to the dependent libraries (if available).
In this way, the same ZLIB build can be shared between OpenCV and HDF5 by first compiling ZLIB and, thereafter, use it for OpenCV using `BUILD_ZLIB=False`, `ZLIB_INCLUDE_DIR` and `ZLIB_LIBRARY_RELEASE` CMake flags. If HDF5 is linked dynamically, nothing changes.


CSBVision:
Closed as we want to clean our branches and this one is still unmerged. If anybody has the same problem, the code is still available. Besides, the issue can be resolved by modifying the CMake variables accordingly, too. Add `/UH5_BUILT_AS_DYNAMIC_LIB` to the C++ command line options and `ShLwApi.Lib` to the linker command line, e.g. by defining:
```
set CXXFLAGS=/UH5_BUILT_AS_DYNAMIC_LIB
set LDFLAGS=ShLwApi.Lib
```
before running CMake. Still, ideally OpenCV's CMake configuration should support both variants natively, without requiring the user to overwrite shared library-only flags.

PR #3567, opened 2023-09-29
Fixed compilation with SIMD disabled or with scalable SIMD

mshabunin:
related https://github.com/opencv/ci-gha-workflow/pull/111

One function in tsdf was not wrapped by an appropriate macro guard.


asmorkalov (approved):
:+1:

PR #3585, opened 2023-10-28
freetype: fix for huge CtoL param.

Kumataro:
The freetype module uses CtoL param to split bezier curve segment to polylines.
This patch makes local upper limitation for each segments. It is determined with total length of polylines at anchors.
And skip to push_back() same points, likes imgproc::ellipse2Poly().

### DIFF
--- a/modules/freetype/src/freetype.cpp
+++ b/modules/freetype/src/freetype.cpp
@@ -723,10 +723,28 @@ int FreeType2Impl::coFn( const FT_Vector *cnt,
 
     PathUserData *p = (PathUserData *)user;
 
+    // Calicurate upper limit of CtoL.
+    // It is same as total length of polylines{ mOldP, cnt, to }.
+
+    const double dx1 = p->mOldP.x  -  cnt->x;
+    const double dy1 = p->mOldP.y  -  cnt->y;
+    const double dx2 = cnt->x      -  to->x;
+    const double dy2 = cnt->y      -  to->y;
+
+    const int maxL = static_cast<int>(
+        std::pow( dx1 * dx1 + dy1 * dy1, 0.5 ) +
+        std::pow( dx2 * dx2 + dy2 * dy2, 0.5 )
+    );
+
+    const int iCtoL = std::min( maxL , p->mCtoL );
+
+    // Avoid to push_back() same as previous points.
+    Point prevPt(INT_MIN, INT_MIN);
+
     // Bezier to Line
-    for(int i = 0;i <= p->mCtoL; i++){
+    for(int i = 0;i <= iCtoL; i++){
         // Split Bezier to lines ( in FreeType coordinates ).
-        double u = (double)i * 1.0 / (p->mCtoL) ;
+        double u = (double)i * 1.0 / (iCtoL) ;
         double nu = 1.0 - u;
         double p0 =                  nu * nu;
         double p1 = 2.0 * u *        nu;
@@ -736,7 +754,12 @@ int FreeType2Impl::coFn( const FT_Vector *cnt,
         double Y = (p->mOldP.y) * p0 + cnt->y * p1 + to->y * p2;
 
         // Store points to draw( in OpenCV coordinates ).
-        p->mPts.push_back( Point ( ftd(X), ftd(Y) ) );
+        Point pt ( ftd(X), ftd(Y) );
+        if ( pt != prevPt )
+        {
+            p->mPts.push_back( pt );
+            prevPt = pt;
+        }
     }
     p->mOldP = *to;
     return 0;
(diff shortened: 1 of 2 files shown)


modules/freetype/src/freetype.cpp
```diff
@@ -723,10 +723,28 @@ int FreeType2Impl::coFn( const FT_Vector *cnt,
...
+    // It is same as total length of polylines{ mOldP, cnt, to }.
+
+    const double dx1 = p->mOldP.x  -  cnt->x;
+    const double dy1 = p->mOldP.y  -  cnt->y;
+    const double dx2 = cnt->x      -  to->x;
+    const double dy2 = cnt->y      -  to->y;
+
+    const int maxL = static_cast<int>(
+        std::pow( dx1 * dx1 + dy1 * dy1, 0.5 ) +
+        std::pow( dx2 * dx2 + dy2 * dy2, 0.5 )
```
asmorkalov, line 736: pow 0.5 is sqrt. I propose to use sqrt as it's faster and more obvious.

modules/freetype/src/freetype.cpp
```diff
@@ -723,10 +723,28 @@ int FreeType2Impl::coFn( const FT_Vector *cnt,
...
+
+    // Avoid to push_back() same as previous points.
+    Point prevPt(INT_MIN, INT_MIN);
+
     // Bezier to Line
-    for(int i = 0;i <= p->mCtoL; i++){
+    for(int i = 0;i <= iCtoL; i++){
         // Split Bezier to lines ( in FreeType coordinates ).
-        double u = (double)i * 1.0 / (p->mCtoL) ;
+        double u = (double)i * 1.0 / (iCtoL) ;
```
asmorkalov, line 747: *1.0 is redundant.

modules/freetype/src/freetype.cpp
```diff
@@ -754,10 +777,31 @@ int FreeType2Impl::cuFn( const FT_Vector *cnt1,
...
+    const double dy1 = p->mOldP.y  -  cnt1->y;
+    const double dx2 = cnt1->x     -  cnt2->x;
+    const double dy2 = cnt1->y     -  cnt2->y;
+    const double dx3 = cnt2->x     -  to->x;
+    const double dy3 = cnt2->y     -  to->y;
+
+    const int maxL = static_cast<int>(
+        std::pow( dx1 * dx1 + dy1 * dy1, 0.5 ) +
+        std::pow( dx2 * dx2 + dy2 * dy2, 0.5 ) +
+        std::pow( dx3 * dx3 + dy3 * dy3, 0.5 )
```
asmorkalov, line 793: the same for pow.

modules/freetype/src/freetype.cpp
```diff
@@ -754,10 +777,31 @@ int FreeType2Impl::cuFn( const FT_Vector *cnt1,
...
+
+    // Avoid to push_back() same as previous points.
+    Point prevPt(INT_MIN, INT_MIN);
+
     // Bezier to Line
-    for(int i = 0; i <= p->mCtoL ;i++){
+    for(int i = 0; i <= iCtoL ;i++){
         // Split Bezier to lines ( in FreeType coordinates ).
-        double u = (double)i * 1.0 / (p->mCtoL) ;
+        double u = (double)i * 1.0 / (iCtoL) ;
```
asmorkalov, line 804: *1.0 is redundant.

modules/freetype/src/freetype.cpp
Kumataro: Thank you for your review! Yes, it is better, I fixed it.

modules/freetype/src/freetype.cpp
Kumataro: Yes, I fixed it. And I replaced from ` (double)i` that is C-style-cast,  to `static_cast<double>(i)`.

modules/freetype/src/freetype.cpp
Kumataro: I fixed it, thank you!

modules/freetype/src/freetype.cpp
Kumataro: I fixed it, thank you!

Kumataro:
I believe that iCtoL is determined with font glyph path data, so it should be care about the range of it more than before.  

- if `iCtoL` == 0, dounle u is calicurated with i / 0.
- if `iCtoL` == INT_MAX, int i will be overflow.

I think that such a case is basically impossible. However, it's not absolute, so I'd like to put in some protection just in case.
So in additional patch, `iCtoL` is fitted to range from 1 to INT_MAX-1.

Kumataro:
Test is failed, but it is about unexpected module/function.  Maybe it is missjudgement.

I believe this file  doesn't reference to freetype module.
- https://github.com/opencv/opencv_contrib/blob/4.x/modules/ximgproc/misc/python/test/test_sparse_match_interpolator.py


Test result is here.

- https://github.com/opencv/opencv_contrib/actions/runs/6809972475/job/18519821850?pr=3585#step:88:350

```
======================================================================
FAIL: test_ric_interpolator (test_sparse_match_interpolator.Interpolator_test)
----------------------------------------------------------------------
  File "/home/ci/opencv_contrib/modules/ximgproc/misc/python/test/test_sparse_match_interpolator.py", line 67, in test_ric_interpolator
    self.assertTrue(cv.norm(dense_flow, ref_flow, cv.NORM_INF) <= MAX_DIF)
AssertionError: False is not true

----------------------------------------------------------------------
Ran 297 tests in 84.373s
```

PR #3586, opened 2023-10-30
freetype: Add function to load font from buffer

razaqq:
There currently is no way to load a freetype font from a memory buffer (for example embedded resources etc). This small PR adds a simple function adding that functionality.

### DIFF
--- a/modules/freetype/include/opencv2/freetype.hpp
+++ b/modules/freetype/include/opencv2/freetype.hpp
@@ -84,6 +84,19 @@ The function loadFontData loads font data.
 
     CV_WRAP virtual void loadFontData(String fontFileName, int idx) = 0;
 
+/** @brief Load font data.
+
+The function loadFontData loads font data.
+The data is not copied, the user needs to make sure the data lives at least as long as FreeType2.
+After the FreeType2 object is destroyed, the buffer can be safely deallocated.
+
+@param pBuf pointer to buffer containing font data
+@param bufSize size of buffer
+@param idx face_index to select a font faces in a single file.
+*/
+
+    CV_WRAP virtual void loadFontData(uchar* pBuf, size_t bufSize, int idx) = 0;
+
 /** @brief Set Split Number from Bezier-curve to line
 
 The function setSplitNumber set the number of split points from bezier-curve to line.
--- a/modules/freetype/src/freetype.cpp
+++ b/modules/freetype/src/freetype.cpp
@@ -67,6 +67,7 @@ public:
     FreeType2Impl();
     ~FreeType2Impl();
     void loadFontData(String fontFileName, int idx) CV_OVERRIDE;
+    void loadFontData(uchar* pBuf, size_t bufSize, int idx) CV_OVERRIDE;
     void setSplitNumber( int num ) CV_OVERRIDE;
     void putText(
         InputOutputArray img, const String& text, Point org,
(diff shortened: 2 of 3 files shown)


modules/freetype/include/opencv2/freetype.hpp
```diff
@@ -84,6 +84,19 @@ The function loadFontData loads font data.
...
+The function loadFontData loads font data.
+The data is not copied, the user needs to make sure the data lives at least as long as FreeType2.
+After the FreeType2 object is destroyed, the buffer can be safely deallocated.
+
+@param pBuf pointer to buffer containing font data
+@param bufSize size of buffer
+@param idx face_index to select a font faces in a single file.
+*/
+
+    CV_WRAP virtual void loadFontData(uchar* pBuf, size_t bufSize, int idx) = 0;
```
Kumataro, line 98: To avoid pyhton-binding error, it maybe better to use `char*` instead of `uchar*`.

modules/freetype/src/freetype.cpp
```diff
@@ -181,16 +182,39 @@ FreeType2Impl::~FreeType2Impl()
...
+    if ( mIsFaceAvailable  == true )
+    {
+        hb_font_destroy(mHb_font);
         CV_Assert(!FT_Done_Face(mFace));
     }
 
     mIsFaceAvailable = false;
-    CV_Assert( !FT_New_Face( mLibrary, fontFileName.c_str(), static_cast<FT_Long>(idx), &(mFace) ) );
+    FT_Open_Args args{ FT_OPEN_MEMORY, (FT_Byte*)pBuf, static_cast<FT_Long>(bufSize), nullptr, nullptr, nullptr, 0, nullptr };
+    CV_Assert( !FT_Open_Face(mLibrary, &args, idx, &mFace) );
```
Kumataro, line 215: Two functions reference different Freetype APIs. It would be better to have one or the other in common.
- Idea1) To use `FT_Open_Face()` for file and memory.
  - both functrions construct `FT_Open_Args` for file and memory.
  - new helper function receives `FT_Open_Args` and `idx`. 
- Idea1) To use `FT_New_Face()` for file, and to use `FT_New_Memory_Face()` for memory.

modules/freetype/src/freetype.cpp
```diff
@@ -181,16 +182,39 @@ FreeType2Impl::~FreeType2Impl()
...
+    {
+        CV_Assert(!FT_Done_Face(mFace));
+        return;
+    }
+    CV_Assert( mHb_font != NULL );
+    mIsFaceAvailable = true;
+}
+
+void FreeType2Impl::loadFontData(uchar* pBuf, size_t bufSize, int idx)
+{
```
Kumataro, line 205: I think to check `pBuf` and `bufSize` is needed.

Kumataro:
I believe to add some test is required to merge it. ( Opencv-extra contains font data.  )
Please could you try on `modules/freetype/test/` codes ?

asmorkalov:
@razaqq Friendly reminder.

modules/freetype/src/freetype.cpp
razaqq: yes but i do not want to assert.
i think both loading functions should return a result indicating if the loading was successful.
loading from file is an inherently failable i/o operation, so asserting there is very bad practice imo.

Kumataro: The reason why the `loadFontData()` function has no return value is because there is no way to recover if this function fails. If this function returns some errors, I believe it is difficult/hard to do anything. Is it better to continue with missing drawing ?

And If an inappropriate/invalid argument is set, it would be better to use CV_Assert() to indicate that there is a problem with the user program implementation/calling.
- If `pBuf == nullptr` and/or `bufSize == 0` are invalid input. (We can omit to check `bufSize`.)
- In other case, if `pBuf[0...bufSize]` are invalid/broken font data, we expects  `Face_Open_args()` will be return with failed.　It will be captured with `CV_Assert()`.
- In other case if `pBug[]` buffer is smaller than `bufSize`, we cannot do anything. Only what we can do is user-application handle it correctly.

modules/freetype/src/freetype.cpp
Kumataro: Thank you for your comment. It is OK. I think there is no problem.
In this pull-request, the FreeType function to be called was switched with an additional commit, so I confirmed the intention. thank you very much !!

asmorkalov:
@Kumataro I already added simple test for the new method.

Kumataro:
Thank you for your comment.

I created a patch on my own repository based on razaqq's repository and created a pull request ( https://github.com/opencv/opencv_contrib/pull/3593 ).

However, maybe I was able to commit directly to razaqq's repository?
In that case my pull request is unnecessary/redundant. I'm sorry if I'm wrong.

modules/freetype/src/freetype.cpp
razaqq: You could recover by trying to load another font. A lot of cpp programs (including all of mine) have exceptions disabled, by asserting you remove their option to handle it in a way they want to. If a user wants to, he can still assert the return value

Kumataro: 1. I think it is difficult to use another/alternative fonts.
- Cases where header data is corrupted can be caught at `loadFontData()` stage. This is a case where clearly incorrect data such as image data is specified.
- Cases where glyph data is corrupted cannot be caught until the `putText()` stage, which reads and renders the glyph data.(When glyph data of `A` is broken, if `A` is not contained strings for `putText()`, there are no problem.) 

With your design, that means if `putText()` failed, we should re-call `loadFontData()` with another font and retry to call `putText()`. I think it is too complex.

2. Exception handling is a prerequisite for using OpenCV. Following is [OpenCV Introduction Error Handling Section](https://docs.opencv.org/4.8.0/d1/dfb/intro.html) section.

> OpenCV uses exceptions to signal critical errors. When the input data has a correct format and belongs to the specified value range, but the algorithm cannot succeed for some reason.

PR #3623, opened 2024-01-19
Implement fast guided filter

w43322:
This PR implements fast guided filter. The concept is explained in this [paper](https://arxiv.org/pdf/1505.00996.pdf).

TLDR:
Added a 'scale' parameter to the current guided filter interface, allowing resampling inside the algorithm, to speed up computation. The resampling has to be done inside guided filter because the intermediate result has to be combined with original input to generate the final result.
The performance and behavior of the default code path (scale == 1.0) is unaltered. [before](https://pullrequest.opencv.org/buildbot/builders/precommit-contrib_linux64/builds/100981/steps/perf_ximgproc/logs/stdio) vs [after](https://pullrequest.opencv.org/buildbot/builders/precommit-contrib_linux64/builds/100980/steps/perf_ximgproc/logs/stdio)

~~- [ ] There is a reference to the original bug report and related work~~

### DIFF
--- a/modules/ximgproc/doc/ximgproc.bib
+++ b/modules/ximgproc/doc/ximgproc.bib
@@ -54,6 +54,13 @@
   publisher={Springer}
 }
 
+@article{Kaiming15,
+  title={Fast guided filter},
+  author={He, Kaiming and Sun, Jian},
+  journal={arXiv preprint arXiv:1505.00996},
+  year={2015}
+}
+
 @inproceedings{Lee14,
   title={Outdoor place recognition in urban environments using straight lines},
   author={Lee, Jin Han and Lee, Sehyung and Zhang, Guoxuan and Lim, Jongwoo and Chung, Wan Kyun and Suh, Il Hong},
--- a/modules/ximgproc/include/opencv2/ximgproc/edge_filter.hpp
+++ b/modules/ximgproc/include/opencv2/ximgproc/edge_filter.hpp
@@ -123,15 +123,15 @@ void dtFilter(InputArray guide, InputArray src, OutputArray dst, double sigmaSpa
 //////////////////////////////////////////////////////////////////////////
 //////////////////////////////////////////////////////////////////////////
 
-/** @brief Interface for realizations of Guided Filter.
+/** @brief Interface for realizations of (Fast) Guided Filter.
 
-For more details about this filter see @cite Kaiming10 .
+For more details about this filter see @cite Kaiming10 @cite Kaiming15 .
  */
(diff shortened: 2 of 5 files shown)


w43322:
last [github workflow](https://github.com/opencv/opencv_contrib/actions/runs/7609220813) was timed out during execution, kindly asking for a re-run :) @opencv-alalek

vpisarev:
@w43322, thanks for the contribution!

We would like to ask you to do a few things:
1. use reciprocal for the scale: 1 => 1, 2 => 0.5, 3 => 0.33333333333... etc. This is to match cv::resize() semantics.
1. replace INTER_NEAREST to INTER_LINEAR (also, please, try INTER_AREA) when downsampling image. I believe, it should improve the quality at the slight penalty in speed. We are going to optimize resize() even further, so the overhead should reduce with time.
2. Looks like the tests have been modified to check NORM_INF only when `scale==1`. Can we offer a stricter test for `scale != 1`?
3. Please, add `CV_Assert(scale <= 1);` to the function (once you convert `scale` to `1./scale`). 'scale' notion can be confusing sometimes, we need to add protective checks in order to provide proper diagnostic to user.

w43322:
@vpisarev Hi, I've updated the PR, please take a look :)

> 1. use reciprocal for the scale: 1 => 1, 2 => 0.5, 3 => 0.33333333333... etc. This is to match cv::resize() semantics.

✅

> 2. replace INTER_NEAREST with INTER_LINEAR (also, please, try INTER_AREA) when downsampling image. I believe, it should improve the quality at the cost of some decrease of speed. We are going to optimize resize() even further, so the overhead should reduce with time.

✅

Did some tests, seems like the runtime is mostly similar across three methods - although INTER_AREA is noticeably slower with `scale == 0.33`, so I opted for INTER_LINEAR for now.
[perf_test.txt](https://github.com/opencv/opencv_contrib/files/14233001/summary.txt)

> 3. Looks like the tests have been modified to check NORM_INF only when `scale==1`. Can we offer a stricter test for `scale != 1`?

Checking for NORM_INF only works for `scale == 1` because it only works when the two images are exactly the same. It was used to check against reference implementation. Since some information is lost during the resampling process, this check won't work on scaled images.

I've added a new check for scaled images, making sure that the FastGuidedFilter image is less blurry than a naive resampling. Please let me know if you think there are any other checks that are viable.

> 4. Please, add `CV_Assert(scale <= 1);` to the function (once you convert `scale` to `1./scale`). 'scale' notion can be confusing sometimes, we need to add protective checks in order to provide proper diagnostic to user.

✅

PR #3626, opened 2024-01-21
Fix gcc11 warning

zanaviska:
‘error’ may be used uninitialized. 
We used to declare error uninitialized, pass it to `LineFit`  as OUT parameter, hoping that every path will initialize `error` varible. But there is path, where `error` is kept uninitialized, so it is better to explicitly initialize it here


zanaviska:
There may be clang-tidy warning about reducing scope of the variable, but I decided that it goes outside of the change that I was trying to make, since I was just trying to get rid of console warning. If maintainers think, that I should reduce variable scope too, write here about it

asmorkalov (approved):
:+1:

PR #3630, opened 2024-01-28
Fix link error caused by lack of moc in the cvv module during opencv_…

GengGode:
…world build

issue : https://github.com/opencv/opencv_contrib/issues/3629

### but, A red prompt will appear,need opencv cmakelists fix
https://github.com/opencv/opencv/pull/24921#discussion_r1466140330
image


GengGode:
need https://github.com/opencv/opencv/pull/24934 To fix the issue where red prompts may appear

modules/cvv/CMakeLists.txt
```diff
@@ -15,8 +15,14 @@ endif()
 
 find_package(Qt${QT_VERSION_MAJOR} REQUIRED COMPONENTS ${CVV_QT_MODULES})
 
-set(CMAKE_AUTOMOC ON)
-set(CMAKE_INCLUDE_CURRENT_DIR ON)
```
asmorkalov, line 19: It does not work. See https://github.com/opencv/opencv/blob/eab21b61067385515c77783608f8c374bfe131c6/modules/highgui/CMakeLists.txt#L91

PR #3635, opened 2024-02-09
updated readme for several opencv_contrib modules (done by Gary)

vpisarev:

- [N/A] There is accuracy test, performance test and test data in opencv_extra repository, if applicable
- [+] The feature is well documented and sample code can be built with the project CMake


vpisarev:
Hello @garybradski, we noticed some changes that you've done in several opencv_contrib's READMEs. Our policy is to submit every change through pull request mechanisms. Also, "master" is not the branch that you should modify explicitly. It's supposed to be read-only branch that is currently automatically synchronized (in one direction) from 4.x and will be synchronized from 5.x after 5.0 is released.

PR #3641, opened 2024-02-20
Camera spatial velocity computation through interaction matrix

simutisernestas:
### Feature

The code computed camera spatial velocity given two images, pixels depth, camera matrix and the timestamp between the images. This is enabled by, so called, interaction matrix (usually utilized in visual servoing applications) relating pixel plane velocities to the camera spatial velocity in 3D i.e., twist - velocity and angular rate of the camera. The inverse problem can be solved by sampling pixel & their velocities to solve least-squares for twist. The relationship can be seen below in the picture.

image

The code does not include a proper example and is not tested but if there is interest I could contribute more appealing example and use case for camera velocity computation. However, I attach below a dummy example with random data simply to make sure that it's running as is. I have used this before in aiding UAV localization and thought someone else might benefit from it being integrated into `opencv`. 

```c++
#include "opencv2/rgbd/twist.hpp"
#include "opencv2/core.hpp"

int main()
{
    using namespace cv::rgbd;
    Twist t;

    // create two random image
    cv::Mat im0(480, 640, CV_8UC1);
    cv::Mat im1(480, 640, CV_8UC1);
    cv::Mat depths0(480, 640, CV_32F);
    cv::Mat K = cv::Mat::eye(3, 3, CV_32F);

    cv::theRNG().state = time(0);
    cv::randn(im0, 0, 255);
    cv::randn(im1, 0, 255);
    cv::randn(depths0, 0, 100);

    cv::Vec6d twist = t.compute(im0, im1, depths0, K, 0.1);

    return 0;
}

```

References
1. Chaumette, François, and Seth Hutchinson. "Visual servo control. I. Basic approaches." IEEE Robotics & Automation Magazine 13.4 (2006): 82-90.
2. https://robotacademy.net.au/lesson/image-based-visual-servoing/

### DIFF
--- a/modules/tracking/test/test_twist.cpp
+++ b/modules/tracking/test/test_twist.cpp
@@ -1,0 +1,42 @@
+#include "test_precomp.hpp"
+
+#include "opencv2/core.hpp"
+#include "opencv2/tracking/twist.hpp"
+
+namespace opencv_test
+{
+namespace
+{
+
+class TwistTest : public ::testing::Test
+{
+protected:
+    cv::detail::tracking::Twist twist;
+    cv::Mat uv, depth, K, J, duv;
+    cv::Vec6d result;
+
+    void SetUp() override
+    {
+        uv = (cv::Mat_<float>(2, 2) << 0.0, 0.0, 0.0, 0.0);
+        depth = (cv::Mat_<float>(1, 2) << 1.0, 1.0);
+        K = (cv::Mat_<float>(3, 3) << 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0);
+        duv = (cv::Mat_<float>(2, 2) << 0.0, 0.0, 0.0, 0.0);
+    }
+};
+
+TEST_F(TwistTest, TestInteractionMatrix)
+{
+    twist.interactionMatrix(uv, depth, K, J);
+    ASSERT_EQ(J.cols, 6);
+    ASSERT_EQ(J.rows, 4);
+}
+
+TEST_F(TwistTest, TestCompute)
+{
+    result = twist.compute(uv, duv, depth, K);
+    for (int i = 0; i < 6; i++)
+        ASSERT_NEAR(result[i], 0.0, 1e-6);
+}
+
+} // namespace
+} // namespace opencv_test
--- a/modules/tracking/test/test_twist.cpp
+++ b/modules/tracking/test/test_twist.cpp
@@ -3,22 +3,24 @@
 #include "opencv2/core.hpp"
 #include "opencv2/tracking/twist.hpp"
 
-#define eps 1e-4
-
 namespace opencv_test
 {
 namespace
 {
 
+using namespace cv::detail::tracking;
+
+float const eps = 1e-4f;
+
 class TwistTest : public ::testing::Test
 {
 protected:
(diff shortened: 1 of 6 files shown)


savuor (commented):
To my mind:
1. This should go into the `tracking` module
1. We should use sparse set of point correspondences, not dense and definitely not based on a specific optical flow algorithm.
1. The problem as it is formulated now can be solved by a combination of `findEssentialMat` + `decomposeEssentialMat` or `recoverPose` and Rt-to-twist decomposition available in `DualQuat` class from 5.x, even without using additional Z coordinates. And it's more preferable since the existing functions are quite robust and fast, thanks to RANSAC.
1. This function would make more sense as a solution to a different problem if it is changed to take pixel velocities as input instead of images or points.
1. The function should be documented
1. The function should be covered by at least one test. You can transform your sample code to a test.

simutisernestas:
>     1. This should go into the `tracking` module
moved
>     2. We should use sparse set of point correspondences, not dense and definitely not based on a specific optical flow algorithm.
removed the the flow calculation from withing the class
>     3. The problem as it is formulated now can be solved by a combination of `findEssentialMat` + `decomposeEssentialMat` or `recoverPose` and Rt-to-twist decomposition available in `DualQuat` class from 5.x, even without using additional Z coordinates. And it's more preferable since the existing functions are quite robust and fast, thanks to RANSAC.
I suppose this doesn't provide information about the scale of the scene? Having the depth input here, the velocity information is in absolute units right away.
>     4. This function would make more sense as a solution to a different problem if it is changed to take pixel velocities as input instead of images or points.
I've refactored the code to explicitly take in the pixel velocities.
>     5. The function should be documented
Done
>     6. The function should be covered by at least one test. You can transform your sample code to a test.
Done

Also made the interaction matrix computation as public method. Thus one could use that for visual servoing applications, like driving a camera with specific pixel velocities, or up to a pixel location.

modules/tracking/test/test_twist.cpp
```diff
@@ -0,0 +1,106 @@
...
+    // cam = mv.CentralCamera()
+    // print(cam.K)
+    // print(cam.visjac_p([1, 1], 2.0))
+    // [[1. 0. 0.]
+    //  [0. 1. 0.]
+    //  [0. 0. 1.]]
+    // [[-0.5  0.   0.5  1.  -2.   1. ]
+    //  [ 0.  -0.5  0.5  2.  -1.  -1. ]]
+
+    cv::Mat uv = (cv::Mat_<float>(2, 1) << 1.0, 1.0);
```
savuor, line 37: Again, for such cases it's better to use `std::vector<float>` or `std::vector<cv::Vec3f>` or `cv::Mat` or similar and initialize it with { curly braced initializer list} and then convert it to `Mat`

savuor (changes requested):
I think you're right, we need Z values to get an absolute scene scale. This means that we do need that functionality.
However, there are things to fix before it's merged:

simutisernestas:
Addressed all of the comments

savuor (approved):

simutisernestas:
@savuor Could you rerun the CI, I've removed some dead comments in tests and fixed miss-matched docs parameter. Should pass all the checks now : )

simutisernestas:
LGTM

PR #3645, opened 2024-02-29
added getColorCharts()

AleksandrPanov:
merge this PR after https://github.com/opencv/opencv_contrib/pull/3647
Computes and returns the coordinates of the central parts of the charts modules.

### DIFF
--- a/modules/mcc/include/opencv2/mcc/checker_model.hpp
+++ b/modules/mcc/include/opencv2/mcc/checker_model.hpp
@@ -89,6 +89,15 @@ public:
 
     CV_WRAP virtual TYPECHART getTarget() = 0;
     CV_WRAP virtual std::vector<Point2f> getBox() = 0;
+
+    /** @brief Computes and returns the coordinates of the central parts of the charts modules.
+     *
+     * This method computes transformation matrix from the checkers's coordinates (`cv::mcc::CChecker::getBox()`)
+     * and find by this the coordinates of the central parts of the charts modules.
+     * It is used in `cv::mcc::CCheckerDraw::draw()` and in `ChartsRGB` calculation.
+     */
+    CV_WRAP virtual std::vector<Point2f> getColorCharts() = 0;
+
     CV_WRAP virtual Mat getChartsRGB() = 0;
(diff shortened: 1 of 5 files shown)


modules/mcc/include/opencv2/mcc/checker_model.hpp
```diff
@@ -89,6 +89,15 @@ class CV_EXPORTS_W CChecker
...
     CV_WRAP virtual TYPECHART getTarget() = 0;
     CV_WRAP virtual std::vector<Point2f> getBox() = 0;
+
+    /** @brief Computes and returns the coordinates of the central parts of the charts modules.
+     *
+     * This method computes transformation matrix from the checkers's coordinates (`cv::mcc::CChecker::getBox()`)
+     * and find by this the coordinates of the central parts of the charts modules.
+     * It is used in `cv::mcc::CCheckerDraw::draw()` and in `ChartsRGB` calculation.
+     */
+    CV_WRAP virtual std::vector<Point2f> getComputedCharts() = 0;
```
asmorkalov, line 99: M.b. `getColorCharts()` is more relevant.

asmorkalov:
Could you add the new method to test(s) and at least one of the samples.

modules/mcc/include/opencv2/mcc/checker_model.hpp
AleksandrPanov: Mb `getComputedColorCharts()`?

AleksandrPanov:
> Could you add the new method to test(s) and at least one of the samples.

Method `CCheckerDraw::draw()` call method `getComputedCharts()`and `CCheckerDraw::draw()` tested in `CV_mccRunCCheckerDrawTest` tests

But now I'm adding a stronger test.

modules/mcc/include/opencv2/mcc/checker_model.hpp
asmorkalov: I propose to name it `getColorCharts` without `computed`. The function behaviour may change.

AleksandrPanov: fixed

asmorkalov (approved):
:+1:

PR #3646, opened 2024-03-03
fixes #25081:Latch input corruption fix

Dhanwanth1803:
Fixes https://github.com/opencv/opencv/issues/25081

As mentioned by WennPaper making it `grayImage = image.clone();` will make a deep copy. It prevents the `InputArray` from being modified.

### DIFF
--- a/modules/xfeatures2d/src/latch.cpp
+++ b/modules/xfeatures2d/src/latch.cpp
@@ -519,7 +519,7 @@ namespace cv
             switch (image.type())
             {
             case CV_8UC1:
-                grayImage = image.clone();
+                grayImage = sigma_ ? image.clone() : image;
                 break;
             case CV_8UC3:
                 cvtColor(image, grayImage, COLOR_BGR2GRAY);
--- a/modules/xfeatures2d/src/latch.cpp
+++ b/modules/xfeatures2d/src/latch.cpp
@@ -519,7 +519,7 @@ namespace cv
             switch (image.type())
             {
             case CV_8UC1:
-                grayImage = image;
+                grayImage = image.clone();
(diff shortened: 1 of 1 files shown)


dkurt:
add a link to the issue

asmorkalov (approved):
:+1:

dkurt (dismissed):

modules/xfeatures2d/src/latch.cpp
```diff
@@ -519,7 +519,7 @@ namespace cv
             switch (image.type())
             {
             case CV_8UC1:
-                grayImage = image;
+                grayImage = image.clone();
```
dkurt, line 522: GaussianBlur is called only for `sigma != 0.0`. So clone only then.

dkurt (changes requested):
GaussianBlur is called only for `sigma != 0.0`. So clone only then.

Dhanwanth1803:
> GaussianBlur is called only for `sigma != 0.0`. So clone only then.

Sir, in that case will this work? 
`case CV_8UC1:
                if (sigma_ != 0.) grayImage = image.clone();
                break;`

Dhanwanth1803:
> GaussianBlur is called only for `sigma != 0.0`. So clone only then.

Sir, In that case will this work?
`case CV_8UC1:
                if (sigma_ != 0.) grayImage = image.clone(); 
                break; 
`

modules/xfeatures2d/src/latch.cpp
dkurt, line 522:
```suggestion
                grayImage = sigma_ ? image : image.clone();
```

PR #3699, opened 2024-03-15
Mcc add perf tests improve performance

AleksandrPanov:
Added perf tests to mcc module.
Also these optimizations have been added:

- added `parallel_for_` to `performThreshold()`
- removed `toL`/`fromL` and added `dst` to avoid copy data
- added `parallel_for_` to `elementWise()` ("batch" optimization improves performance of Windows version, Linux without changes).

Configuration:
Ryzen 5950X, 2x16 GB 3000 MHz DDR4
OS: Windows 10, Ubuntu 20.04.5 LTS

Performance results in milliseconds:

| OS and alg version   | process, ms | infer, ms |
| -------------------- | ----- | ------ |
| win_default          | 63.09 | 457.57 |
| win_optimized_without_batch       | 48.69 | 111.78 |
| win_optimized_batch  | 48.42 | 47.28  |
| linux_default        | 50.88 | 300.7  |
| linux_optimized_batch| 36.06 | 41.62  |

### DIFF
--- a/modules/mcc/perf/perf_precomp.hpp
+++ b/modules/mcc/perf/perf_precomp.hpp
@@ -1,0 +1,19 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#ifndef __OPENCV_PERF_PRECOMP_HPP__
+#define __OPENCV_PERF_PRECOMP_HPP__
+
+#include "opencv2/ts.hpp"
+#include "opencv2/ts/cuda_test.hpp"
+#include "opencv2/mcc.hpp"
+#include "opencv2/mcc/ccm.hpp"
+
+namespace opencv_test
+{
+using namespace cv::mcc;
+using namespace cv::ccm;
+}
+
+#endif
--- a/modules/mcc/perf/perf_precomp.hpp
+++ b/modules/mcc/perf/perf_precomp.hpp
@@ -1,19 +1,14 @@
-// This file is part of OpenCV project.
-// It is subject to the license terms in the LICENSE file found in the top-level directory
-// of this distribution and at http://opencv.org/license.html.
-
 #ifndef __OPENCV_PERF_PRECOMP_HPP__
 #define __OPENCV_PERF_PRECOMP_HPP__
 
 #include "opencv2/ts.hpp"
-#include "opencv2/ts/cuda_test.hpp"
 #include "opencv2/mcc.hpp"
-#include "opencv2/mcc/ccm.hpp"
 
 namespace opencv_test
 {
 using namespace cv::mcc;
 using namespace cv::ccm;
+using namespace perf;
 }
 
 #endif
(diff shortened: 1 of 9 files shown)


modules/mcc/src/utils.hpp
```diff
@@ -77,10 +78,26 @@ Mat rgb2gray(const Mat& rgb);
...
 {
-    Mat dst = src.clone();
+    if (dst.empty() || !dst.isContinuous() || dst.total() != src.total() || dst.type() != src.type())
+        dst = Mat(src.rows, src.cols, src.type());
     const int channel = src.channels();
+    if (src.isContinuous()) {
+        const int num_elements = (int)src.total()*channel;
+        const double *psrc = (double*)src.data;
+        double *pdst = (double*)dst.data;
+        const int batch = 128;
```
AleksandrPanov, line 90: This "batch" optimization improve performance in Windows

dkurt: Which are common values of `num_elements`? We can make `batch` dependent on number of threads:
```cpp
const int batch = num_elements / max(1, getNumThreads());
```
or
```cpp
const int batch = num_elements / (getNumThreads() > 1 ? getNumThreads() * 4 : 1);
```
instead of 4 you may choose another constant to get128 in you configuration.

modules/mcc/perf/perf_precomp.hpp
```diff
@@ -0,0 +1,20 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#ifndef __OPENCV_PERF_PRECOMP_HPP__
+#define __OPENCV_PERF_PRECOMP_HPP__
+
+#include "opencv2/ts.hpp"
+#include "opencv2/ts/cuda_test.hpp"
```
dkurt, line 9: Is that header required?

AleksandrPanov: 1. The header `"opencv2/ts/cuda_test.hpp"` is needed to use `randomMat`
2. `#include "opencv2/ts/ts_perf.hpp"` instead `"opencv2/ts.hpp"` does not work

modules/mcc/src/utils.hpp
AleksandrPanov: Now:
`const int batch = std::max(1, getNumThreads() > 1 ? num_elements / (1024*getNumThreads()) : num_elements);`
Now each thread processes a minimum of 1024 array elements.

With case `const int batch = num_elements / max(1, getNumThreads());` a regression in performance appears (from 47 ns ti 57 ms).

AleksandrPanov: I would suggest using batch 128, but your second sample would also work.

AleksandrPanov: Batch - the minimum required number of consecutive elements in an array that a thread can process at one time.

modules/mcc/perf/perf_precomp.hpp
dkurt: Please use cv::randu instead

dkurt (approved):

PR #3701, opened 2024-03-20
4.x: Fix data corruption in WeChatQRCode::impl::decode

elatkin:
Data corruption may lead to incorrect results if multiple QR found at image

```
allow_multiple_commits=1
```


mshabunin:
Is it possible to add simple test case reproducing the issue? E.g. with some synthetic image or one of existing test images?

mshabunin:
I suggest also changing method signature to use `const vector<Mat>& candidate_points` and for-loop to use `const auto& point` for consistency.

mshabunin (approved):

PR #3703, opened 2024-03-21
Use proper C++ types.

vrabaud:
This is necessary to get https://github.com/opencv/opencv/pull/25248 working.

### DIFF
--- a/modules/videostab/src/precomp.hpp
+++ b/modules/videostab/src/precomp.hpp
@@ -59,7 +59,7 @@
 
 inline float sqr(float x) { return x * x; }
 
-inline float intensity(const cv::Point3_<unsigned char> &bgr)
+inline float intensity(const cv::Point3_<uint8_t> &bgr)
 {
     return 0.3f*bgr.x + 0.59f*bgr.y + 0.11f*bgr.z;
 }
--- a/modules/videostab/src/precomp.hpp
+++ b/modules/videostab/src/precomp.hpp
@@ -59,7 +59,7 @@
 
 inline float sqr(float x) { return x * x; }
 
-inline float intensity(const cv::Point3_<uchar> &bgr)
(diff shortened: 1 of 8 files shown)


modules/videostab/src/precomp.hpp
```diff
@@ -59,7 +59,7 @@
 
 inline float sqr(float x) { return x * x; }
 
-inline float intensity(const cv::Point3_<uchar> &bgr)
+inline float intensity(const cv::Point3_<unsigned char> &bgr)
```
mshabunin, line 62: Why not `uint8_t`?

vrabaud: I am trying to stay close to the current OpenCV typedefs at https://github.com/opencv/opencv/blob/3aefd4862c5ee92918e15d96e9973cbfde48e5c2/modules/core/include/opencv2/core/hal/interface.h#L51
But indeed, we might as well use the proper types from `<cstdint>`. Should I switch to that?

mshabunin: We have this proposal for 5.x -opencv/opencv#24994 , so eventually, we will switch to the `<cstdint>` types. Since changes in this PR do not affect public interfaces, I suggest using these types here. I'm not sure regarding opencv/opencv#25248 though, perhaps it can be partially converted to `uint8_t` (internal usage, C++ code) and partially to `unsigned char` (C-API).

vrabaud: Done: switched to uint_8_t.
For the other PR, let's not change the typedefs right now. All it does is encapsulate the OpenCV typedefs in the `cv` namespace, that's it. As this header is public, it prevents conflicts with other libraries that also define it in the global namespace (like SWIG), but differently (int64 is a `long long int` and not an `int64_t` like OpenCV).

mshabunin (approved):

PR #3719, opened 2024-04-11
fix not merging when there are only two segs on one line

Roger8:
Fixes https://github.com/opencv/opencv_contrib/issues/3715

This bug can be recurrent when there are only two segments on one line , if `do_merge` is set to be true. 
The `while` loop will not enter in , then output has still two segments but not one merged one segment.


Roger8:
fix #3719

asmorkalov (approved):

PR #3724, opened 2024-04-21
Rename getInteractionMatrix() to computeInteractionMatrix()

catree:

---

see https://github.com/opencv/opencv_contrib/pull/3641

cc @simutisernestas @savuor

I prefer to use `computeInteractionMatrix()` instead of `getInteractionMatrix()` naming.
I have added references to some classical papers on this subject. Even if it is out of scope, they can be useful.

Warning, one bibtex reference already exist in the main `opencv` repo. I don't know if we can use `@cite` defined in the main repo?


savuor (approved):
OK, see no problem

PR #3734, opened 2024-05-02
Fixed Wredundant-move produced by GCC 13.2 (Ubuntu 24.04).

asmorkalov:

### DIFF
--- a/modules/face/src/facemarkLBF.cpp
+++ b/modules/face/src/facemarkLBF.cpp
@@ -667,7 +667,7 @@ Mat FacemarkLBFImpl::BBox::project(const Mat &shape) const {
         res(i, 0) = (shape_(i, 0) - x_center) / x_scale;
         res(i, 1) = (shape_(i, 1) - y_center) / y_scale;
     }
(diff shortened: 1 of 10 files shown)


modules/face/src/facemarkLBF.cpp
```diff
@@ -667,7 +667,7 @@ Mat FacemarkLBFImpl::BBox::project(const Mat &shape) const {
         res(i, 0) = (shape_(i, 0) - x_center) / x_scale;
         res(i, 1) = (shape_(i, 1) - y_center) / y_scale;
     }
-    return std::move(res);
```
opencv-alalek, line 670: MacOS compiler emits warnings without that: https://github.com/opencv/opencv_contrib/actions/runs/8928671648/job/24524735399?pr=3734#step:16:66

Need to choose one variant and adjust compilers settings for other.

asmorkalov:
@opencv-alalek I fixed warnings on Mac. Could you take a look again?

mshabunin (approved):

PR #3755, opened 2024-06-07
Remap relative2

chacha21:
Git got crazy and cannot pull [https://github.com/opencv/opencv_contrib/pull/3600](https://github.com/opencv/opencv_contrib/pull/3600) properly

Attempt to resubmit


chacha21:
#3600 works again

PR #3777, opened 2024-08-14
Adds support for CUDA MHT 16-bit demosaicking

llschloesser:

This pull request adds support for CUDA 16-bit MHT demosaicking handled identically to 8-bit input. This means the output can be 1, 3, or 4 channels. The same template is used for 8-bit, so there is no need for additional tests unless desired and there is 16-bit test data available.

credit @shane-nichols

- [N/A] There is accuracy test, performance test and test data in opencv_extra repository, if applicable
- [N/A] The feature is well documented and sample code can be built with the project CMake


llschloesser:
@asmorkalov I'd like to see these changes merged if you can get to it. We have a 16-bit image processing pipeline for our ROV cameras operating in low-light and due to the size of the images we prefer to use CUDA processing. This debayering method is higher quality that the default linear alternative. If you need a 16-bit raw image for a test I can provide. Thanks!

PR #3782, opened 2024-08-20
speedup random forest getVotes method.

ivafanas:
Port of https://github.com/opencv/opencv/pull/26046

Speedup `DTreesImplForRTrees::getVotes` method in about x2 times.

`Mat` objects are usually not expensive for creation or copying. But if algorithm itself is fast, operations on `Mat` objects start to take significant time. Naive extraction of loop invariant temporary `Mat` wrapper object creation allows to decrease `getVotes` method run time in about x2 times.


asmorkalov (approved):
:+1:

Issue #3773, opened 2024-08-06
CUDA 12.6 build errors


ghfue:
##### System information (version)

- OpenCV => master
- Operating System / Platform => Ubuntu 22.04
- Compiler => GCC 11.4

##### Detailed description

The build breaks when upgrading from CUDA 12.5 to CUDA 12.6.

Here is the full build error:
```
[  9%] Building NVCC (Device) object modules/cudaarithm/CMakeFiles/cuda_compile_1.dir/src/cuda/cuda_compile_1_generated_polar_cart.cu.o
/opencv_contrib/modules/cudev/include/opencv2/cudev/grid/detail/transform.hpp(518): error: incomplete type "cuda::std::__4::tuple_size<cv::cudev::PtrTraitsBase<cv::cudev::ZipPtrSz<thrust::THRUST_200500_500_520_600_610_700_750_800_860_890_900_NS::tuple<cv::cudev::PtrTraitsBase<cv::cudev::GpuMat_<float>, cv::cudev::GlobPtr<float>>::ptr_type, cv::cudev::PtrTraitsBase<cv::cudev::GpuMat_<float>, cv::cudev::GlobPtr<float>>::ptr_type>>, cv::cudev::ZipPtr<thrust::THRUST_200500_500_520_600_610_700_750_800_860_890_900_NS::tuple<cv::cudev::PtrTraitsBase<cv::cudev::GpuMat_<float>, cv::cudev::GlobPtr<float>>::ptr_type, cv::cudev::PtrTraitsBase<cv::cudev::GpuMat_<float>, cv::cudev::GlobPtr<float>>::ptr_type>>>::ptr_type>" (aka "cuda::std::__4::tuple_size<cv::cudev::ZipPtr<thrust::THRUST_200500_500_520_600_610_700_750_800_860_890_900_NS::tuple<cv::cudev::GlobPtr<float>, cv::cudev::GlobPtr<float>>>>") is not allowed
          Unroll<tuple_size<DstPtrTuple>::value>::transform(srcVal, dst, op, y, x);
                 ^
          detected during:
[...]

PR #3785, opened 2024-09-03
Workaround for CUDA 12.6 tuple_size issue #3773.

asmorkalov:
Fixes https://github.com/opencv/opencv_contrib/issues/3773

### DIFF
--- a/modules/cudaarithm/src/cuda/polar_cart.cu
+++ b/modules/cudaarithm/src/cuda/polar_cart.cu
@@ -159,25 +159,15 @@ void cv::cuda::cartToPolar(InputArray _xy, OutputArray _mag, OutputArray _angle,
     GpuMat_<float> magc(mag);
     GpuMat_<float> anglec(angle);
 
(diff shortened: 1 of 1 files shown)


asmorkalov:
cc @cudawarped @chacha21. I want to drop gridTransformTuple and some hacky things on top of tuple, because it's hard to maintain it with different CUDA versions. What do you think about the solution and further cleanup?

cudawarped:
@asmorkalov Unfortunatley it removes the fusion as the source has to be read from global memory twice but I can't think of a less involved fix.  The alternative of adding two ops to `gridTransformUnary` as you did with `gridTransformBinary` is probably overkill as we should probably be moving away from `cudev` and towards Thrust for these operations if the CUDA namespace has a future.

asmorkalov:
@cudawarped Thanks for the opinion.

chacha21:
Another solution would be to drop the cudev/functional abstraction for polarToCart/cartToPolar and use dedicated CUDA kernels to handle the different cases

PR #3800, opened 2024-09-26
Get CUDA code to compile with clang CUDA and without CUDA

vrabaud:
Changelist:
- there are some syntactic changes: `<< <` -> `<<<`. For some reason, I do not need to change all those in the code.
-  `::min` -> `std::min` in `__host__` code
- `modules/cudaimgproc/src/moments.cpp` needs to have the CUDA code in the `#ifdef`
- The signature of `cv::cuda::swapChannels` is not exactly the same as the C++ one in `modules/cudaimgproc/src/color.cpp` 
- `cv::cuda::FarnebackOpticalFlow::create` needs to be explicit about which FarnebackOpticalFlow it returns

### DIFF
--- a/modules/cudaarithm/src/arithm.cpp
+++ b/modules/cudaarithm/src/arithm.cpp
@@ -54,6 +54,8 @@ void cv::cuda::mulAndScaleSpectrums(InputArray, InputArray, OutputArray, int, fl
 
 void cv::cuda::dft(InputArray, OutputArray, Size, int, Stream&) { throw_no_cuda(); }
 
+Ptr<DFT> cv::cuda::createDFT(Size, int) { throw_no_cuda(); return Ptr<DFT>(); }
+
 Ptr<Convolution> cv::cuda::createConvolution(Size) { throw_no_cuda(); return Ptr<Convolution>(); }
 
 #else /* !defined (HAVE_CUDA) */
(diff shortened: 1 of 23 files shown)


cudawarped:
> * there are some syntactic changes: `<< <` -> `<<<`. For some reason, I do not need to change all those in the code.

On Windows this is an annoying default in Visual Studio, what errors does clang give from this?  I'm concerned that if any new kernel's are added by users of VS then it will brake the clang build.

Strangely it build fine for me on Ubuntu 20:04 - clang version 10.0.0-4ubuntu1.

vrabaud:
Thx @cudawarped . I am ok fixing the `<< <` as it goes with whatever new code comes in. This is the kind of error I get:
```
polar_cart.cu.cc:195:43: error:
 expected expression                                                                                                      
  195 |             polarToCartImpl_<T, false> << <grid, block, 0, stream >> >(shrinkPtr(magc), shrinkPtr(anglec), shrinkP
tr(xc), shrinkPtr(yc), scale, anglec.rows, anglec.cols);
      |                                           ^                                                                       
```

asmorkalov:
@vrabaud Could you add some notes how you build OpenCV + CUDA + Clang. Just to be sure that we are talking about the same thing.

vrabaud:
@asmorkalov , for my purpose, I do not use nvcc but clang cuda instead. This is all internal to Google so I actually have no idea of the command lines unfortunately.

vrabaud:
I am done adding missing implementations, PTAL

asmorkalov (approved):

PR #3813, opened 2024-10-21
Update radon_transform.cpp

### LINKED ISSUE #3739: Radon Transform
### System Information

openCV version 4.9.0-dev
Ubuntu 22.04.4 LTS 

### Detailed description

The Radon Transform provided by openCV changes its src contrary to its documentation. I fixed that behavior locally by changing: 

```Mat _srcMat = src.getMat() ---> Mat _srcMat = src.getMat().clone();```

in the "radon_transform.cc" file.

### Steps to reproduce

After the call 
```
double thetastep = 1.0
double  startAngle = -20;
double endAngle = 20;
cv::ximgproc::RadonTransform(matRef, radMatRef, thetaStep, startAngle, endAngle, true);
```
the cv::Mat matRef seess to be chagen. More detailed: I was using the Radon Transform within a loop when i recognized the behavior. 

- [ ] I checked the problem with documentation, FAQ, open issues, forum.opencv.org, Stack Overflow, etc and have not found any solution
- [ ] I updated to the latest OpenCV version and the issue is still there

sturkmen72:
closes #3739


sturkmen72:
@asmorkalov ( just a question ) i was expecting https://github.com/opencv/opencv_contrib/pull/3813/commits/fb292b6fb43ec7ed0b61fe9bceb198d3c4ef705f should work without problem. but https://github.com/opencv/opencv_contrib/pull/3813/commits/2f9d0a6aff02a40edcae3d7054865569af7d2b98 seems needed.
i wonder (out of this PR scope) if we need to check why `normalize(radon, radon, 0, 255, NORM_MINMAX, CV_8UC1);` did not work here.

PR #3828, opened 2024-11-22
`cudacodec::Videoreader` fix YUV color conversion

cudawarped:
#### Currently `VideoReader` color conversion:
1. Forces quantization to 8 bits by using the NV12 surface format for all inputs regardless of the encoded bit depth or chroma subsampling.
2. Does not respect the color space standard if specified in the video source.
3. Only outputs BGR, BGRA, GRAY and NV12 frames.

**Note:** Previous attempts to fix this https://github.com/opencv/opencv_contrib/pull/3468 didn't address (2).

#### This PR uses the Nvidia Video Codec SDK ColorSpace.cu sample as a guide to:
1. Use the color space standard if it exists and default back to a suitable format based on the video resolution if not.
2. Decode YUV420 10/12 bit and YUV444 8/10/12 bit sources.
3. Allow the user to choose the desired output bitdepth.  e.g. Reqest 8 bit output from a 16 bit source.
5. Add RGB/RGBA and planar outputs.


modules/cudacodec/src/video_reader.cpp
```diff
@@ -191,9 +150,17 @@ namespace
         videoSource_->setVideoParser(videoParser_);
         videoSource_->start();
         waitForDecoderInit();
+        FormatInfo format = videoDecoder_->format();
+        if (format.colorSpaceStandard == ColorSpaceStandard::Unspecified) {
+            if (format.width > 1280 || format.height > 720)
+                format.colorSpaceStandard = ColorSpaceStandard::BT709;
+            else
+                format.colorSpaceStandard = ColorSpaceStandard::BT601;
+        }
```
asmorkalov, line 159: Is it some heuristic? Do you have any reference for the decision?

asmorkalov (approved):

modules/cudacodec/src/video_reader.cpp
cudawarped: Yes, from my understanding 709 is for HD but I could't find a definitive source to say if there was an exact definition of what resolution this starts at.  Although many sources mentioned 720p as the minimum resolution in the end I chose 720p as it replicated the color format behaviour from sources decoded with the `VideoCapture` class..

PR #3829, opened 2024-11-23
getDefaultName() added

abbaselmas:
all other contrib feature has this function but tbmr


sturkmen72:
@abbaselmas missing `CV_WRAP String getDefaultName() const CV_OVERRIDE;` at https://github.com/abbaselmas/opencv_contrib/blob/patch-1/modules/xfeatures2d/include/opencv2/xfeatures2d.hpp#L1196-L1212

abbaselmas:
> missing `CV_WRAP String getDefaultName() const CV_OVERRIDE;` at

@sturkmen72 good catch thanks for the heads up

asmorkalov (approved):

PR #3836, opened 2024-11-28
Extend cudaimgproc::demosaicing for f32

apbr:
Extend `cudaimgproc::demosaicing` to also allow images with f32 values.

- [x] There is a reference to the original bug report and related work -> not applicable

### DIFF
--- a/modules/cudaimgproc/src/color.cpp
+++ b/modules/cudaimgproc/src/color.cpp
@@ -72,6 +72,8 @@ namespace cv { namespace cuda {
 
         template <int cn>
         void MHCdemosaic(PtrStepSzb src, int2 sourceOffset, PtrStepSzb dst, int2 firstRed, cudaStream_t stream);
+
+        void MHCdemosaic_float3(PtrStepSzf src, int2 sourceOffset, PtrStepSzf dst, int2 firstRed, cudaStream_t stream);
     }
 }}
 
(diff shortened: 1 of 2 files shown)


llschloesser:
@apbr please see https://github.com/opencv/opencv_contrib/pull/3777

It seems like adding 32f to that PR would be easier than deconflicting with this one and we might get all functionality merged sooner. Can I ask what sensor you are using where you have 32f raw pixels?

apbr:
@llschloesser #3777 was merged 1 hour after your comment. So I'll just rebase my PR.
> Can I ask what sensor you are using where you have 32f raw pixels?     

We do some image processing steps before the debayering. Their resulting datatype is f32. So it's not directly from the sensor.

llschloesser:
If you follow the template pattern of the earlier PR you'll get 32f to gray and 4 channel, there is no need to restrict to 3 channel.

apbr:
> If you follow the template pattern of the earlier PR you'll get 32f to gray and 4 channel, there is no need to restrict to 3 channel.

Good point, I'll rework and test it.

apbr:
Reworked the changes.

asmorkalov:
@apbr Could you add simple test for the new feature?

apbr:
> @apbr Could you add simple test for the new feature?

@asmorkalov I could give it a try. Should they be in here?

asmorkalov:
Yes, as shortcut you can use existing demosaic test and reference data for u8, but scale it for larger range. I just want to be sure that the template implementation does not lead to overflows or other arithmetical issues.

PR #3864, opened 2025-01-06
Fix Unused Variable Warning in `test_rotation_and_scale_invariance.cpp`

shyama7004:
### Description:
This PR resolves a warning caused by an unused variable `IMAGE_BIKES` in the file `test_rotation_and_scale_invariance.cpp` within the `xfeatures2d` module of the OpenCV contrib repository.

- **Issue:** The variable `IMAGE_BIKES` was declared but not used in the test file, leading to a compiler warning (`-Wunused-const-variable`).

```
warning: unused variable 'IMAGE_BIKES' [-Wunused-const-variable]
static const char* const IMAGE_BIKES = "detectors_descriptors_evaluation/
```

### DIFF
--- a/modules/xfeatures2d/test/test_rotation_and_scale_invariance.cpp
+++ b/modules/xfeatures2d/test/test_rotation_and_scale_invariance.cpp
@@ -10,8 +10,9 @@
 namespace opencv_test { namespace {
 
 static const char* const IMAGE_TSUKUBA = "features2d/tsukuba.png";
-// static const char* const IMAGE_BIKES = "detectors_descriptors_evaluation/images_datasets/bikes/img1.png";
-
+#ifdef OPENCV_ENABLE_NONFREE
+static const char* const IMAGE_BIKES = "detectors_descriptors_evaluation/images_datasets/bikes/img1.png";
+#endif // OPENCV_ENABLE_NONFREE
 // ========================== ROTATION INVARIANCE =============================
 
 #ifdef OPENCV_ENABLE_NONFREE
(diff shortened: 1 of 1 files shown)


modules/xfeatures2d/test/test_rotation_and_scale_invariance.cpp
```diff
@@ -10,7 +10,7 @@
 namespace opencv_test { namespace {
 
 static const char* const IMAGE_TSUKUBA = "features2d/tsukuba.png";
-static const char* const IMAGE_BIKES = "detectors_descriptors_evaluation/images_datasets/bikes/img1.png";
```
asmorkalov, line 13: It's used inside of `#ifdef OPENCV_ENABLE_NONFREE` block. It makes sense to move the constant there.

asmorkalov (changes requested):

modules/xfeatures2d/test/test_rotation_and_scale_invariance.cpp
shyama7004: ok I will

shyama7004: I have one concern , in many OpenCV test cases, INSTANTIATE_TEST_CASE_P is used, but it’s deprecated. Can we replace it with INSTANTIATE_TEST_SUITE_P? Please share your thoughts.

asmorkalov: Let stay it as is.

asmorkalov (approved):

PR #3868, opened 2025-01-09
`cudacodec::VideoWriter` add flag for writing full range luma chroma video

cudawarped:
Currently the meta data in all encoded video's is not updated to reflect the range (full of limited as defined in Annex E of the ITU-T Specification) of encoded luma/chroma values.  With all video's indicating that they use the limited range.  This means that when a full range YUV source is encoded the decoder will incorrectly decode it to a limited range resulting in poor color reproduction.

This PR correctly sets this meta data field for all encoded video's.

### DIFF
--- a/modules/cudacodec/src/video_writer.cpp
+++ b/modules/cudacodec/src/video_writer.cpp
@@ -315,6 +315,13 @@ GUID EncodingPresetGuid(const EncodePreset nvPreset) {
     CV_Error(Error::StsUnsupportedFormat, msg);
 }
 
+std::string GetVideoCodecString(const GUID codec) {
+    if (codec == NV_ENC_CODEC_H264_GUID) return "AVC/H.264";
+    else if (codec == NV_ENC_CODEC_HEVC_GUID) return "H.265/HEVC";
+    else if (codec == NV_ENC_CODEC_AV1_GUID) return "AV1";
+    else return "Unknown";
+}
+
 void VideoWriterImpl::InitializeEncoder(const GUID codec, const double fps)
 {
     NV_ENC_INITIALIZE_PARAMS initializeParams = {};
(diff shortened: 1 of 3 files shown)


modules/cudacodec/src/video_writer.cpp
```diff
@@ -342,10 +342,20 @@ void VideoWriterImpl::InitializeEncoder(const GUID codec, const double fps)
...
+            initializeParams.encodeConfig->encodeCodecConfig.h264Config.h264VUIParameters.videoSignalTypePresentFlag = 1;
+        }
+    }
+    else if (codec == NV_ENC_CODEC_HEVC_GUID) {
         initializeParams.encodeConfig->encodeCodecConfig.hevcConfig.idrPeriod = encoderParams.idrPeriod;
+        if (encoderParams.videoFullRangeFlag) {
+            initializeParams.encodeConfig->encodeCodecConfig.hevcConfig.hevcVUIParameters.videoFullRangeFlag = 1;
+            initializeParams.encodeConfig->encodeCodecConfig.hevcConfig.hevcVUIParameters.videoSignalTypePresentFlag = 1;
+        }
+    }
```
asmorkalov, line 358: I propose to print warning, if full range luma is used with other codec.

cudawarped: Good call especially as the Blackwell will add support for AV1 encoding.

cudawarped:
@asmorkalov I've added a warning when the codec is not h264 or h265.

**Note:** Because I can't test AV1 encoding (I don't have access to a card with 8th Gen NVDEC) the cudacodec module does not currently support AV1 encoding.  This is why I haven't also added AV1 support for the `videoFullRangeFlag` (`NV_ENC_CONFIG::encodeConfig->encodeCodecConfig.av1Config.colorRange`) flag in this PR.

asmorkalov:
You actually can submit the PR  with GF 4x support. I have a device for test.

cudawarped:
That's good to know although I'll need access to one first to test everything locally first before submitting a PR.

asmorkalov:
@cudawarped There is conflict with the previous PR please rebase. Also `videoFullRangeFlag` looks ugly for me. It's bool and it's obviously flag. I propose to rename it to something like `encodeVideoFullRange`.

cudawarped:
@asmorkalov Correct, I copied the name from the SDK where its a flag (`unsigned char video_full_range_flag` in NVDEC and `uint32_t  videoFullRangeFlag` in NVENC).  Additionaly as I mentioned above this is refered to as `uint32_t colorRange` for AV1 encoding so a differently named bool here would make sense.  I tried to keep the naming the same as the SDK for transparency but I can change this if its less confusing?

For consistency should I also rename it in `VideoReader` to `videoFullRangeSource` or similar?

asmorkalov:
`videoFullRangeSource` looks good.

asmorkalov:
Oh, got the issue. Let it be videoFullRangeFlag everywhere to not break code and confuse users.

PR #3872, opened 2025-01-16
Updated version of README

ronaldosc:
This PR is a correction in some commands in main documentation.

Were edited the commands for bash for default and deleted the "$" for best copy & paste.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,25 +15,25 @@ provides production-quality support for this module.
 
 You can build OpenCV, so it will include the modules from this repository. Contrib modules are under constant development and it is recommended to use them alongside the master branch or latest releases of OpenCV.
 
-Here is the CMake command for you:
+  >Here is the CMake command line for you:
 
-```
-$ cd <opencv_build_directory>
-$ cmake -DOPENCV_EXTRA_MODULES_PATH=<opencv_contrib>/modules <opencv_source_directory>
-$ make -j5
+```bash
+cd <opencv_build_directory>
+cmake -D OPENCV_EXTRA_MODULES_PATH=<opencv_contrib>/modules <opencv_source_directory>
+make -j5
 ```
 
 As the result, OpenCV will be built in the `<opencv_build_directory>` with all
 modules from `opencv_contrib` repository. If you don't want all of the modules,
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -15,25 +15,26 @@ provides production-quality support for this module.
...
+```bash
+cd <opencv_build_directory>
+cmake -D OPENCV_EXTRA_MODULES_PATH=<opencv_contrib>/modules <opencv_source_directory>
+make -j5
 ```
 
 As the result, OpenCV will be built in the `<opencv_build_directory>` with all
 modules from `opencv_contrib` repository. If you don't want all of the modules,
-use CMake's `BUILD_opencv_*` options. Like in this example:
+use CMake's `BUILD_opencv_*` option where * is the given name of a module as <reponame> in this example:
```
opencv-alalek, line 28: > reponame

This is not accurate. This just confuses user.
`<module>` should be used instead.

Ref: https://docs.opencv.org/4.x/db/d05/tutorial_config_reference.html#tutorial_config_reference_general_modules

README.md
```diff
@@ -15,25 +15,26 @@ provides production-quality support for this module.
...
-Here is the CMake command for you:
+  >***Here is the CMake command line for you:***
 
-```
-$ cd <opencv_build_directory>
-$ cmake -DOPENCV_EXTRA_MODULES_PATH=<opencv_contrib>/modules <opencv_source_directory>
-$ make -j5
+```bash
+cd <opencv_build_directory>
+cmake -D OPENCV_EXTRA_MODULES_PATH=<opencv_contrib>/modules <opencv_source_directory>
```
opencv-alalek, line 22: This is not a mistake.

opencv-alalek (commented):
We don't accept PRs with massive "changes for changes" without significant "added value".

README.md
ronaldosc: I know that. But for the best practices and DX the -D separated is more readable.

README.md
ronaldosc: Sure, you're correct. The name of repo consists in the same name of the repo, right?  
Can you suggest more clear text explanation for this line ?

PR #3906, opened 2025-03-16
serialization for mcc

kmatzen:
I added serialization support for ColorCorrectionModel.

https://github.com/opencv/opencv_contrib/issues/3905

### DIFF
--- a/modules/mcc/src/ccm.cpp
+++ b/modules/mcc/src/ccm.cpp
@@ -27,15 +27,28 @@
 
 #include "opencv2/mcc/ccm.hpp"
 #include "linearize.hpp"
+
 namespace cv {
 namespace ccm {
 class ColorCorrectionModel::Impl
 {
 public:
     Mat src;
+
     std::shared_ptr<Color> dst = std::make_shared<Color>();
+    // Track initialization parameters for serialization
+    Mat dst_colors;
+    COLOR_SPACE dst_cs_enum;
+    Mat dst_colored;
+    CONST_COLOR dst_constcolor;
+    bool dst_use_constcolor;
+
     Mat dist;
+
     RGBBase_& cs;
+    // Track initialization parameters for serialization
+    COLOR_SPACE cs_enum;
+
     Mat mask;
 
     // RGBl of detected data and the reference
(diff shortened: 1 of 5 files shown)


modules/mcc/src/ccm.cpp
```diff
@@ -433,5 +462,134 @@ Mat ColorCorrectionModel::getMask() const{
...
+       << "gamma" << p->gamma
+       << "deg" << p->deg
+       << "saturated_threshold" << p->saturated_threshold
+       << "initial_method_type" << p->initial_method_type
+       << "weights_coeff" << p->weights_coeff
+       << "masked_len" << p->masked_len
+       << "max_count" << p->max_count
+       << "epsilon" << p->epsilon
+       << "dst_use_constcolor" << p->dst_use_constcolor;
+
```
gursimarsingh, line 494: Thank you for the contribution. The code looks good overall, but I feel we don’t need to write/save this many details.
I think keeping [ccm, loss, colorchart, ccm_type, gamma, deg, saturated-threshold, linear-type, distance-type] should be sufficient.

kmatzen: That makes sense.  My reasoning for serializing everything was (1) reloading the model in another process should enable all interfaces on the class to function as if it were the original and (2) if state inside the class is non-essential, then should it be stored in the class in the first place?  Are some of the fields only cached state for the others?

gursimarsingh: (1) For inference, only _p->ccm_, _p->linear,_ _p->cs_, and _p->shape_ (which can also be derived from _ccm_type_) are actually needed. You can see this directly in the _infer()_ function in **ccm.cpp**.
(2) Most of the other fields are indeed just cached values to speed up CCM computation and aren't necessary for inference.

kmatzen: You don't want to support use cases like a model being saved, loaded elsewhere, and then accessors like `getWeights()` being called?

gursimarsingh: CCM calculation is not really something similar to a deep learning model where we want to save all the parameters and resume training. 

The application is to just save the matrix and some other metadata for informative purpose and then use it for inference in an application

@asmorkalov what are your thoughts? Your comments are welcomed

PR #3936, opened 2025-05-15
Add warpAffine and resizeDown APIs in FastCV Extension

quic-apreetam:
- Added warpAffine function to apply affine transformations.
2x3 affine transformations for both CV_8UC1 and CV_8UC3 input
2x2 matrix-based patch extraction for grayscale images, with ROI.

- Deprecated resizeDownBy2 and resizeDownBy4 functions.
- Introduced resizeDown function to down-scale images using specified scaling factors or dimensions, supporting both single-channel (CV_8UC1) and two-channel (CV_8UC2) images.

### DIFF
--- a/modules/fastcv/include/opencv2/fastcv/warp.hpp
+++ b/modules/fastcv/include/opencv2/fastcv/warp.hpp
@@ -45,55 +45,44 @@ CV_EXPORTS_W void warpPerspective2Plane(InputArray _src1, InputArray _src2, Outp
     InputArray _M0, Size dsize);
 
 /**
- * @brief Applies an affine transformation to a grayscale image using a 2x3 matrix.
- * @param _src              Input grayscale image of type 'CV_8UC1'.
- * @param _dst              Output image after transformation.
- * @param _M                2x3 affine transformation matrix.
- * @param dsize             Size of the output image.
- * @param interpolation     Interpolation method, Supported methods include:
- *                          'INTER_NEAREST': Nearest neighbor interpolation.
- *                          'INTER_LINEAR': Bilinear interpolation.
- *                          'INTER_AREA': Area interpolation.
- * @param borderValue       Constant pixel value for border pixels.
+ * @brief Performs an affine transformation on an input image using a provided transformation matrix.
+ * 
+ * This function performs two types of operations based on the transformation matrix:
+ * 
+ * 1. Standard Affine Transformation (2x3 matrix):
+ *    - Transforms the entire input image using the affine matrix
+ *    - Supports both CV_8UC1 and CV_8UC3 types
+ * 
+ * 2. Patch Extraction with Transformation (2x2 matrix):
+ *    - Extracts and transforms a patch from the input image
+ *    - Only supports CV_8UC1 type
+ *    - If input is a ROI: patch is extracted from ROI center in the original image
+ *    - If input is full image: patch is extracted from image center
+ * 
+ * @param _src              Input image. Supported formats:
+ *                          - CV_8UC1: 8-bit single-channel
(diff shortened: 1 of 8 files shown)


quic-apreetam:
hi @asmorkalov, i think we don't need dummy PR for this

modules/fastcv/include/opencv2/fastcv/warp.hpp
```diff
@@ -44,6 +44,57 @@ CV_EXPORTS_W void warpPerspective(InputArray _src, OutputArray _dst, InputArray
...
+ * @brief Warps a patch centered at a specified position in the input image using an affine transformation.
+ * @param src          Input grayscale image of type 'CV_8UC1'.
+ * @param position     Position in the image where the patch is centered.
+ * @param affine       2x2 affine transformation matrix of type 'CV_32FC1'.
+ * @param patch        Output image patch after transformation.
+ * @param patchSize    Size of the output patch.
+ *
+ * @return Returns 0 if the transformation is valid.
+ */
+CV_EXPORTS_W void warpAffineROI(InputArray _src, const cv::Point2f& position, InputArray _affine, OutputArray _patch, Size patchSize);
```
asmorkalov, line 96: Technically, the API is redundant. OpenCV cv::Mat has locateRoi method that may extract details of the original matrix, if roi is provided as a parameter. You can just handle the case in existing warpAffine.

modules/fastcv/src/warp.cpp
```diff
@@ -175,5 +175,109 @@ void warpPerspective(InputArray _src, OutputArray _dst, InputArray _M0, Size dsi
...
+                (uint8_t*)dst.data, dst.cols, dst.rows, dst.step, NULL, fcvInterpolation, FASTCV_BORDER_CONSTANT, borderValue);
+
+    if (status != FASTCV_SUCCESS)
+    {
+        std::string s = fcvStatusStrings.count(status) ? fcvStatusStrings.at(status) : "unknown";
+        CV_Error( cv::Error::StsInternal, "FastCV error: " + s);
+    }
+}
+
+void warpAffine3Channel(InputArray _src, OutputArray _dst, InputArray _M, Size dsize, OutputArray _dstBorder)
```
asmorkalov, line 226: Technically, you do not need extra API call. You can handle amount of channels in warpAffine in run time and select proper FastCV call.

modules/fastcv/src/scale.cpp
```diff
@@ -8,55 +8,68 @@
...
+        dsize = Size(saturate_cast<int>(ssize.width*inv_scale_x),
+                     saturate_cast<int>(ssize.height*inv_scale_y));
+        CV_Assert( !dsize.empty() );
+    }
+    else
+    {
+        inv_scale_x = static_cast<double>(dsize.width) / ssize.width;
+        inv_scale_y = static_cast<double>(dsize.height) / ssize.height;
+        CV_Assert(inv_scale_x > 0);
+        CV_Assert(inv_scale_y > 0);
```
asmorkalov, line 32: The limitation is not documented in the function header.

modules/fastcv/include/opencv2/fastcv/warp.hpp
quic-apreetam: in warpAffineROI, we use fcvTransformAffineu8_v2 which uses 2x2 Affine transformation and also uses fixed point precision, that's why it was added separately

modules/fastcv/src/scale.cpp
quic-apreetam: updated, thanks

quic-apreetam:
@asmorkalov unified warpAffine, warpAffine3Channel, and warpAffineROI into a single warpAffine function

asmorkalov (approved):

PR #3937, opened 2025-05-17
tracking: moved CSRT to main repository, removed legacy implemetation

mshabunin:
merge with opencv/opencv#27328

* removed `cv::tracking::trackerCSRT` and `cv::legacy::tracking::trackerCSRT` in favor of the one from _video_ module
* removed from samples and tests
* removed from docs (moved to main)

**Note:** I tried to preserve cv::legacy::tracking interface by redirecting calls to the new interface, but it was tied with private implementation and didn't go as planned. Perhaps it is still possible to do it with slightly more effort, but I decided to remove it. We are allowed to break source compatibility of _opencv_contrib_ modules.


mshabunin:
Need to be reworked

PR #3939, opened 2025-05-20
Properly fix inf/inf issue in MCC

vrabaud:
Properly fix https://github.com/opencv/opencv_contrib/issues/3316

Without this fix, we could end up with 4 corners, which might seem valid, while some of those are (0,0).

### DIFF
--- a/modules/mcc/src/bound_min.cpp
+++ b/modules/mcc/src/bound_min.cpp
@@ -167,7 +167,7 @@ void CBoundMin::calculate()
         j = (i + 1) % 4;
         Vcart = lines[i].cross(lines[j]);
         if (fabs(Vcart.z) <= 1e-6){
-            continue;
+            return;
         }
         Vhom.x = Vcart.x / Vcart.z;
         Vhom.y = Vcart.y / Vcart.z;


asmorkalov:
@gursimarsingh please test and review.

abhishek-gola:
Hi @vrabaud,
The change looks good to me. Using _return_ instead of _continue_ seems preferable here. Although _continue_ still produces the correct output—since the specific checker box is ultimately rejected during the CCM evaluation due to a high error:-
```
   if (!cccm.evaluate(scm, offset, iTheta, error))
            continue;
```
using _return_ will help avoiding unnecessary computations by exiting early.

vrabaud:
It actually crashes for me: having two points that are the same find a matrix in https://github.com/opencv/opencv_contrib/blob/eebd8977e158d81a4aaf41653a038b82cf6507d4/modules/mcc/src/checker_detector.cpp#L802 that is not invertible, creating Inf values again in https://github.com/opencv/opencv_contrib/blob/eebd8977e158d81a4aaf41653a038b82cf6507d4/modules/mcc/src/checker_detector.cpp#L868

PR #3943, opened 2025-05-22
Add Otsu's method to cv::cuda::threshold

troelsy:
I implemented Otsu's method in CUDA for a separate project and want to add it to cv::cuda::threshold

I have made an effort to use existing OpenCV functions in my code, but I had some trouble with `ThresholdTypes` and `cv::cuda::calcHist`. I couldn't figure out how to include `precomp.hpp` to get the definition of `ThresholdTypes`. For `cv::cuda::calcHist` I tried adding `opencv_cudaimgproc`, but it creates a circular dependency on `cudaarithm`. I have include a simple implementation of `calcHist` so the code runs, but I would like input on how to use `cv::cuda::calcHist` instead. 

### DIFF
--- a/modules/cudaarithm/src/cuda/threshold.cu
+++ b/modules/cudaarithm/src/cuda/threshold.cu
@@ -95,11 +95,229 @@ namespace
     }
 }
 
-double cv::cuda::threshold(InputArray _src, OutputArray _dst, double thresh, double maxVal, int type, Stream& stream)
+template <uint n_bins>
+__global__ void otsu_mean(uint *histogram, uint *threshold_sums, unsigned long long *sums)
+{
+    extern __shared__ unsigned long long shared_memory_u64[];
+
+    uint bin_idx = blockIdx.x * blockDim.x + threadIdx.x;
+    uint threshold = blockIdx.y * blockDim.y + threadIdx.y;
+
+    uint threshold_sum_above = 0;
+    unsigned long long sum_above = 0;
+
+    if (bin_idx >= threshold)
+    {
+        uint value = histogram[bin_idx];
+        threshold_sum_above = value;
+        sum_above = value * bin_idx;
+    }
+
+    blockReduce<n_bins>((uint *)shared_memory_u64, threshold_sum_above, bin_idx, plus<uint>());
+    blockReduce<n_bins>((unsigned long long *)shared_memory_u64, sum_above, bin_idx, plus<unsigned long long>());
+
+    if (bin_idx == 0)
+    {
+        threshold_sums[threshold] = threshold_sum_above;
+        sums[threshold] = sum_above;
+    }
+}
+
+template <uint n_bins>
+__global__ void
+otsu_variance(longlong2 *variance, uint *histogram, uint *threshold_sums, unsigned long long *sums)
+{
+    extern __shared__ signed long long shared_memory_i64[];
+
+    uint bin_idx = blockIdx.x * blockDim.x + threadIdx.x;
+    uint threshold = blockIdx.y * blockDim.y + threadIdx.y;
+
+    uint n_samples = threshold_sums[0];
+    uint n_samples_above = threshold_sums[threshold];
+    uint n_samples_below = n_samples - n_samples_above;
+
+    unsigned long long total_sum = sums[0];
+    unsigned long long sum_above = sums[threshold];
+    unsigned long long sum_below = total_sum - sum_above;
+
+    int threshold_variance_above = 0;
+    int threshold_variance_below = 0;
+    if (bin_idx >= threshold)
+    {
+        uint mean = sum_above / n_samples_above;
+        int sigma = bin_idx - mean;
+        threshold_variance_above = sigma * sigma;
+    }
+    else
+    {
+        uint mean = sum_below / n_samples_below;
+        int sigma = bin_idx - mean;
+        threshold_variance_below = sigma * sigma;
+    }
+
+    uint bin_count = histogram[bin_idx];
+    signed long long threshold_variance_above64 = threshold_variance_above * bin_count;
+    signed long long threshold_variance_below64 = threshold_variance_below * bin_count;
+    blockReduce<n_bins>((signed long long *)shared_memory_i64, threshold_variance_above64, bin_idx, plus<signed long long>());
+    blockReduce<n_bins>((signed long long *)shared_memory_i64, threshold_variance_below64, bin_idx, plus<signed long long>());
+
+    if (bin_idx == 0)
+    {
+        variance[threshold] = make_longlong2(threshold_variance_above64, threshold_variance_below64);
+    }
+}
+
+template <uint n_thresholds>
+__global__ void
+otsu_score(uint *otsu_threshold, uint *threshold_sums, longlong2 *variance)
+{
+    extern __shared__ float shared_memory_f32[];
+
+    uint threshold = blockIdx.x * blockDim.x + threadIdx.x;
+
+    uint n_samples = threshold_sums[0];
+    uint n_samples_above = threshold_sums[threshold];
+    uint n_samples_below = n_samples - n_samples_above;
+
+    float threshold_mean_above = (float)n_samples_above / n_samples;
+    float threshold_mean_below = (float)n_samples_below / n_samples;
+
+    longlong2 variances = variance[threshold];
+    float variance_above = (float)variances.x / n_samples_above;
+    float variance_below = (float)variances.y / n_samples_below;
+
+    float above = threshold_mean_above * variance_above;
+    float below = threshold_mean_below * variance_below;
+    float score = above + below;
+
+    float original_score = score;
+
+    blockReduce<n_thresholds>((float *)shared_memory_f32, score, threshold, minimum<float>());
+
+    if (threshold == 0)
+    {
+        shared_memory_f32[0] = score;
+    }
+    __syncthreads();
+
+    score = shared_memory_f32[0];
+
+    // We found the minimum score, but we need to find the threshold. If we find the thread with the minimum score, we
+    // know which threshold it is
+    if (original_score == score)
+    {
+        *otsu_threshold = threshold - 1;
+    }
+}
+
+template <uint n_bins, uint n_thresholds>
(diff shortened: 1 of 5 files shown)


cudawarped:
@troelsy Your test currently doesn't use the Otsu method and when it does it fails the assert.  Once you've fixed the test I will review your implementation.

How does the performance of this compare to the CPU version?

cudawarped:
> See also the screenshot of Nsight Systems for a visualization of the execution timeline. Mind that it says ~80µs, but that is due to profilling overhead.

You're measuring the API calls there not the execution time of the kernels/memory allocs etc..  As you are including synchronization this should be about the same but in general it is not the same thing.  If there wasn't any synchronization the time for the API calls would be much much smaller than the execution time.

> From my measurements, about half of the time is spend on allocating and deallocating memory, so if there is a way to reuse the allocations we could reduce the execution time significantly.

You should be able to reduce the allocations/deallocations by using [`BufferPool`](https://docs.opencv.org/3.4/d5/d08/classcv_1_1cuda_1_1BufferPool.html).  The other main bottleneck is the copy of the threshold from device to host but because the NPP routines use a host threshold there isn't anything you can do about this.

modules/cudaarithm/src/cuda/threshold.cu
```diff
@@ -95,12 +95,232 @@ namespace
     }
 }
 
-double cv::cuda::threshold(InputArray _src, OutputArray _dst, double thresh, double maxVal, int type, Stream& stream)
+template <uint n_bins>
+__global__ void otsu_mean(uint *histogram, uint *threshold_sums, unsigned long long *sums)
+{
+    extern __shared__ unsigned long long shared_memory_u64[];
+
+    uint bin_idx = blockIdx.x * blockDim.x + threadIdx.x;
```
cudawarped, line 103: This is confusing.  As you have 256 threads per block and 256 blocks I would suggest the following launch parameters
```
dim3 block_all = dim3(n_bins);
dim3 grid_all = dim3(n_thresholds);
dim3 block_score = dim3(n_thresholds);
dim3 grid_score = dim3(1);
```
with
```
uint bin_idx = threadIdx.x;
uint threshold = blockIdx.x;
```
`int`is usually more performant than `uint`.

modules/cudaarithm/src/cuda/threshold.cu
```diff
@@ -95,12 +95,232 @@ namespace
...
+    cudaStream_t cuda_stream = StreamAccessor::getStream(stream);
+
+    dim3 block_all = dim3(n_bins, 1, 1);
+    dim3 grid_all = dim3(divUp((int)n_bins, block_all.x), divUp((int)n_thresholds, block_all.y), 1);
+    dim3 block_score = dim3(n_thresholds, 1, 1);
+    dim3 grid_score = dim3(divUp((int)n_thresholds, block_score.x), 1, 1);
+
+    GpuMat gpu_threshold_sums = GpuMat(1, n_bins, CV_32SC1);
+    GpuMat gpu_sums = GpuMat(1, n_bins, CV_64FC1);
+    GpuMat gpu_variances = GpuMat(1, n_bins, CV_32SC4);
```
cudawarped, line 230: Can this be `CV_64C2`?

modules/cudaarithm/src/cuda/threshold.cu
```diff
@@ -95,12 +95,232 @@ namespace
...
+    otsu_variance<256><<<grid_all, block_all, shared_memory, cuda_stream>>>(gpu_variances.ptr<longlong2>(), histogram, gpu_threshold_sums.ptr<uint>(), gpu_sums.ptr<unsigned long long>());
+
+    shared_memory = n_bins * sizeof(float);
+    otsu_score<256><<<grid_score, block_score, shared_memory, cuda_stream>>>(
+        otsu_threshold, gpu_threshold_sums.ptr<uint>(), gpu_variances.ptr<longlong2>());
+}
+
+// TODO: Replace this is cv::cuda::calcHist
+template <uint n_bins>
+__global__ void histogram_kernel(
```
cudawarped, line 244: @asmorkalov Due to a circular dependency when adding the **opencv_cudaimgproc** to **opencv_cudaarithm** in _CMakeLists.txt_ `cv::cuda::calcHist` cannot be called from here.  To avoid refactoring would it make sence at a minimum to only have a single implementation of `calcHist` even if it is duplicated or do you have a better suggestion? i.e. copy the implemntation from  `cv::cuda::calcHist` here.

modules/cudaarithm/src/cuda/threshold.cu
```diff
@@ -95,12 +95,232 @@ namespace
...
+
+        const uint n_bins = 256;
+        const uint n_thresholds = 256;
+
+        // Find the threshold using Otsu and then run the normal thresholding algorithm
+        GpuMat gpu_histogram = GpuMat(n_bins, 1, CV_32SC1);
+        calcHist(src, gpu_histogram, stream);
+
+        GpuMat gpu_otsu_threshold(1, 1, CV_32SC1);
+        compute_otsu_async<n_bins, n_thresholds>(gpu_histogram.ptr<uint>(), gpu_otsu_threshold.ptr<uint>(), stream);
```
cudawarped, line 313: All kernel launches are async so I would rename this `compute_otsu`.

modules/cudaarithm/src/cuda/threshold.cu
troelsy: I can't find `CV_64C2`. Is the name correct?

modules/cudaarithm/src/cuda/threshold.cu
```diff
@@ -95,12 +95,232 @@ namespace
...
+        float mean = (float) sum_below / n_samples_below;
+        float sigma = bin_idx - mean;
+        threshold_variance_below_f32 = sigma * sigma;
+    }
+
+    uint bin_count = histogram[bin_idx];
+    signed long long threshold_variance_above_i64 = (signed long long)(threshold_variance_above_f32 * bin_count);
+    signed long long threshold_variance_below_i64 = (signed long long)(threshold_variance_below_f32 * bin_count);
+    blockReduce<n_bins>((signed long long *)shared_memory_i64, threshold_variance_above_i64, bin_idx, plus<signed long long>());
+    __syncthreads();
```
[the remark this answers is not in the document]
troelsy: You're right, it doesn't affect occupancy and is easier to read

`otsu_sums` became 5.13% faster
`otsu_variance` became 3.96% faster

modules/cudaarithm/src/cuda/threshold.cu
```diff
@@ -95,12 +95,232 @@ namespace
...
 
     const int depth = src.depth();
 
+    // TODO: How can we access precomp.hpp to avoid defining THRESH_OTSU?
+    const int THRESH_OTSU = 8;
+    if ((type & THRESH_OTSU) == THRESH_OTSU)
+    {
+        CV_Assert(depth == CV_8U);
+        CV_Assert(src.channels() == 1);
+        CV_Assert(maxVal == 255.0);
```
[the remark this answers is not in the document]
troelsy: I removed the assert as we are not using `maxVal` for Otsu

modules/cudaarithm/test/test_element_operations.cpp
```diff
@@ -2577,6 +2577,54 @@ INSTANTIATE_TEST_CASE_P(CUDA_Arithm, Threshold, testing::Combine(
...
+{
+    cv::Mat src = randomMat(size, CV_MAKE_TYPE(type, channel));
+
+    cv::cuda::GpuMat dst = createMat(src.size(), src.type(), useRoi);
+    double otsu_gpu = cv::cuda::threshold(loadMat(src, useRoi), dst, 0, 255, threshOp);
+
+    cv::Mat dst_gold;
+    double otsu_cpu = cv::threshold(src, dst_gold, 0, 255, threshOp);
+
+    EXPECT_NEAR(otsu_gpu, otsu_cpu, 1e-5);
```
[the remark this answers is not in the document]
troelsy: I went with `ASSERT_DOUBLE_EQ`

troelsy:
@cudawarped What's the next step? The test fails, but it says `(409) Conflict: an artifact with this name already exists on the workflow run`, so I'm guessing it doesn't have anything to do with my changes.

Should I do something about the histogram implementation or should we let it be for now?

modules/cudaarithm/src/cuda/threshold.cu
cudawarped: @asmorkalov any thoughts on this?

asmorkalov:
Please ignore the error. Looks like some bug in our CI scripts. I'll take a look.

modules/cudaarithm/src/cuda/threshold.cu
```diff
@@ -95,12 +95,235 @@ namespace
...
+    dim3 grid_all(n_thresholds);
+    dim3 block_score(n_thresholds);
+    dim3 grid_score(1);
+
+    BufferPool pool(stream);
+    GpuMat gpu_threshold_sums(1, n_bins, CV_32SC1, pool.getAllocator());
+    GpuMat gpu_sums(1, n_bins, CV_64FC1, pool.getAllocator());
+    GpuMat gpu_variances(1, n_bins, CV_32SC4, pool.getAllocator());
+
+    uint shared_memory;
```
cudawarped, line 237: This is now redundant.

modules/cudaarithm/src/cuda/threshold.cu
```diff
@@ -95,12 +95,235 @@ namespace
...
+        sums[threshold] = sum_above;
+    }
+}
+
+__global__ void
+otsu_variance(longlong2 *variance, uint *histogram, uint *threshold_sums, unsigned long long *sums)
+{
+    const uint32_t n_bins = 256;
+
+    __shared__ signed long long shared_memory_a[n_bins / WARP_SIZE];
```
cudawarped, line 134: You are relying on warp reduction being used in `blockReduce` before the write to shared memory, given that you have a significant amount of available shared memory its safer to allocate enough memory
```
__shared__ signed long long shared_memory_a[n_bins];
```
in case the underlying implementation changes.

PR #3949, opened 2025-06-08
Feature #27286: Add ColoredHashTSDFVolume support in rgbd module(CPU)

huiyu-mcse:
Introduced 'ColoredHashTSDFVolume', a new CPU-only volume type extending 'HashTSDFVolume' with color integration support. This feature enables reconstruction of colored 3D scenes using aligned depth and RGB images, while benefiting from the memory and performance advantages of hashed volumetric fusion.

Main contributions:
- New data structure to store both TSDF and RGB color information
- Extended raycasting and fetch functions to support color output

This enhancement follows the direction proposed in issue #25155 and addresses feature request #27286.

### DIFF
--- a/modules/rgbd/include/opencv2/rgbd/colored_kinfu.hpp
+++ b/modules/rgbd/include/opencv2/rgbd/colored_kinfu.hpp
@@ -75,9 +75,14 @@ struct CV_EXPORTS_W Params
     CV_WRAP static Ptr<Params> hashTSDFParams(bool isCoarse);
 
     /** @brief ColoredTSDF parameters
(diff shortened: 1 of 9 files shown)


savuor:
I'm for integrating this new functionality into 5.x first. Here are my pros:
* We already have ordinary TSDF with color support in 5.x, but buggy
* TSDF in 5.x has better test coverage and the tests require minimum change to support new TSDF classes
* The TSDF interface in 5.x was largely reworked, it's more uniform and among other things supports Python bindings

Cons are:
* This code is already written
* 5.x is not released yet

@asmorkalov what are your thoughts on this?

PR #3953, opened 2025-06-12
bgsegm: Enable __popcnt intrinsic for MSVC on ARM to fix LSBP test failure

pratham-mcw:
This PR enables use of the `__popcnt` intrinsic for ARM/ARM64 targets when building with MSVC. Previously, these targets(ARM/ARM64) were explicitly excluded using macro guards.

When building OpenCV on Windows on ARM64 with MSVC (cl.exe), the BackgroundSubtractor_LSBP.Accuracy test fails because the code falls into the else block using different bit counting method, which leads to incorrect results and causes the test to fail.

### DIFF
--- a/modules/bgsegm/src/bgfg_gsoc.cpp
+++ b/modules/bgsegm/src/bgfg_gsoc.cpp
@@ -70,7 +70,7 @@ const float LSBPtau = 0.05f;
 inline int LSBPDist32(unsigned n) {
 #if defined(__GNUC__) || defined(__clang__)
     return __builtin_popcount(n);
-#elif defined(_MSC_VER) && !(defined(_M_ARM) || defined(_M_ARM64))
+#elif defined(_MSC_VER)
     return __popcnt(n);
 #else
     // Taken from http://graphics.stanford.edu/~seander/bithacks.html#CountBitsSetParallel


asmorkalov:
I do not see __popcnt support for ARM in the Microsoft documentation: https://learn.microsoft.com/en-us/cpp/intrinsics/popcnt16-popcnt-popcnt64?view=msvc-170. It describes x86 specific.

Kumataro:
Hello, at Microsoft blog, Visual Studio 2022 v17.11 and later seems to support `__popcnt` family instructions for ARM64 platform (not ARM32).

See https://devblogs.microsoft.com/cppblog/msvc-backend-updates-in-visual-studio-2022-version-17-11/

> Support popcnt instructions, __popcnt, popcnt16, and popcnt64, thanks to our friends at ARM. 

To relax this macro guards for ARM64, We have to validate MSVC version.

asmorkalov:
ok, msvs version check should be added then

modules/bgsegm/src/bgfg_gsoc.cpp
```diff
@@ -70,7 +70,7 @@ const float LSBPtau = 0.05f;
 inline int LSBPDist32(unsigned n) {
 #if defined(__GNUC__) || defined(__clang__)
     return __builtin_popcount(n);
-#elif defined(_MSC_VER)
```
asmorkalov: Why "less" in `_MSC_VER < 1941`? Most probably it's a mistake.

asmorkalov (approved):

PR #3960, opened 2025-06-19
sfm: support ng-log

### LINKED ISSUE #3946: Move from glog to ng-log
https://github.com/google/glog?tab=readme-ov-file#google-logging-library -> https://github.com/ng-log/ng-log

glog is no longer maintained and was renamed to ng-log by the original developers.

opencv_contrib currently fails to build with ng-log.

Kumataro:
Close #3946 


Kumataro:
Next ng-log ver 0.9.0 will drop compativility to Google glog. So we have to support ng-log instead of it.

https://github.com/ng-log/ng-log

PR #3963, opened 2025-06-25
cudev: Add __shfl_down implementation for long long and unsigned long for CUDA Tookit < 9.0

cudawarped:
Draft fix for https://github.com/opencv/opencv_contrib/issues/3962.

Support for `__shfl_down` on `long long` was not introduced until CUDA Toolkit 9.0.  I don't know if this is just software support or if hardware support was added as well.  Its a long shot but it may be the reason that the tests are failing on Compute Capbability 5.3 devices.  

### DIFF
--- a/modules/cudev/include/opencv2/cudev/warp/shuffle.hpp
+++ b/modules/cudev/include/opencv2/cudev/warp/shuffle.hpp
@@ -334,12 +334,28 @@ __device__ __forceinline__ uint shfl_down(uint val, uint delta, int width = warp
 
 __device__ __forceinline__ signed long long shfl_down(signed long long val, uint delta, int width = warpSize)
 {
+#if defined __CUDACC_VER_MAJOR__ < 9
+    union { long long ll; int2 i2; } u;
+    u.ll = val;
+    u.i2.x = __shfl_down(u.i2.x, delta, width);
+    u.i2.y = __shfl_down(u.i2.y, delta, width);
+    return u.ll;
+#else
     return __shfl_down(val, delta, width);
+#endif
 }
 
 __device__ __forceinline__ unsigned long long shfl_down(unsigned long long val, uint delta, int width = warpSize)
 {
-    return (unsigned long long) __shfl_down(val, delta, width);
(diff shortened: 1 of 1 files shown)


troelsy:
Hi, I thought I would chime in as it relates to my recent PR. The company I work for uses Jetson TX2 with CC=6.2, CUDA Toolkit 10.2 and everything seems to work, so I took a look at it. It looks like all devices that support warp shuffle (CC≥3.0) will support shuffle with `long long` as long as the CUDA Toolkit ≥ 9.0.

To the question about if it is implemented in software or hardware, I think warp shuffle are always done 32 bit at a time because the registers are limited to 32 bit. It will just be two shuffles for 64 bit types. The PTX also indicate this in Compiler Explorer: https://godbolt.org/z/nxdYcqoWe. If the PTX view doesn't show, try opening a new compiler window.

I think the if-statement should be changed to check the CUDA Toolkit version instead. The current code will change the behavior on Jetson TX2 even though it should be supported. Does OpenCV specify a minimum version of CUDA Toolkit?

cudawarped:
@troelsy The flag was just a test to try and fix the crash on CC 5.3 devices.  Do you have access to CUDA toolkit < 9.0 to test whether `_shufl_down` compiles or not on CUDA toolkit < 9.0?  Godbolt doesn't have NVCC <= 9.1.85.

troelsy:
TX2 should be able to run CUDA Toolkit 8.0, but my department doesn't have access to the firmware, so I can't try it out

cudawarped:
@asmorkalov Do you have a machine with CUDA Toolkit < 9.0 on it to check this?

asmorkalov:
No, unfortunately. I want to deploy something with desktop PC, but after the 4.12 release. It's almost ready.

cudawarped:
Would you like me to kill this PR or change the `#define` from `__CUDA_ARCH__ < 700` to `__CUDACC_VER_MAJOR__ < 9`?

asmorkalov:
@cudawarped Is it ready for merge?

cudawarped:
@asmorkalov  👍

PR #3964, opened 2025-06-27
Fixed out-of-bound access in CUDA Otsu threshold implementation.

asmorkalov:
The issue introduced in https://github.com/opencv/opencv_contrib/pull/3943
Fixes https://github.com/opencv/opencv_contrib/issues/3962


asmorkalov:
cc @cudawarped

cudawarped:
@cudawarped Nice catch I didn't notice @troelsy hadn't fixed everywhere! 🥳

PR #3980, opened 2025-07-29
ximgproc: optimize Adaptive Manifold function for ARM64

pratham-mcw:

- This PR introduces an ARM64-specific performance optimization in AdaptiveManifoldFilter::h_filter by applying loop unrolling.
- The optimization is guarded with #if defined(_M_ARM64) to ensure it only affects ARM64 builds.
- The optimization does not affect accuracy and maintains the same numerical behavior as the original scalar implementation.

Performance Improvements :

- The optimization significantly improves the performance of adaptive Manifold function on Windows ARM64 targets.
- The table below shows timing comparisons before and after the optimization:
image

### DIFF
--- a/modules/ximgproc/src/adaptive_manifold_filter_n.cpp
+++ b/modules/ximgproc/src/adaptive_manifold_filter_n.cpp
@@ -521,7 +521,7 @@ void AdaptiveManifoldFilterN::h_filter(const Mat1f& src, Mat& dst, float sigma)
 
         dst_row[0] = src_row[0];
         int x = 1;
-    #if defined(_M_ARM64)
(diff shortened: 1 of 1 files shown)


pratham-mcw:
Hi @asmorkalov, just following up on this PR. The loop unrolling in the h_filter function has shown significant performance improvements on Windows-ARM64 targets, while maintaining functional correctness across other architectures. 
Please let me know if there are any additional changes or updates you’d like me to make.

asmorkalov:
I do not see stable improvement on Linux ARM (GCC, Jetson Orin). SO let's merge as is without check extension.

PR #3983, opened 2025-07-31
Add assert to ensure using non-overlapping memory regions

mitresthen:
This pr addresses this issue: https://github.com/opencv/opencv/issues/27429
where the user did not realize that the warpaffine function requires non-overlapping src and dst memory regions. 
The code now compares the input memory regions and asserts that they do not overlap. There is also a test for this functionality.

### DIFF
--- a/modules/cudawarping/src/warp.cpp
+++ b/modules/cudawarping/src/warp.cpp
@@ -208,18 +208,7 @@ void cv::cuda::warpAffine(InputArray _src, OutputArray _dst, InputArray _M, Size
     _dst.create(dsize, src.type());
     GpuMat dst = _dst.getGpuMat();
 
-    const uchar* src_start = src.data;
-    const uchar* dst_start = dst.data;
-
-    size_t src_size = src.step * src.rows;
-    size_t dst_size = dst.step * dst.rows;
-
-    const uchar* src_end = src_start + src_size;
-    const uchar* dst_end = dst_start + dst_size;
-
-    bool overlap = src_end > dst_start && dst_end > src_start;
-
-    CV_Assert(!overlap && "In-place operation not supported for cv::cuda::warpAffine");
+    CV_Assert( src.data != dst.data && "In-place operation not supported for cv::cuda::warpAffine" );
 
     Size wholeSize;
     Point ofs;


cudawarped:
@asmorkalov Should this be a documentation issue? e.g. This function does not support in place operations.

I've been looking through to find a convention for dealing with this.  I've found functions which
1. explicitly document they support in place
2. explicitly document they don't support in place but don't assert on it
3. don't document in place operations
4. clone the source if the source and destination memory locations are equal. e.g. [cv::warpPerspective](https://github.com/opencv/opencv/blob/8dc4ad3ff38ebe998dff29b322f4c3797061220a/modules/imgproc/src/imgwarp.cpp#L2847)
5. use a buffer if the source and destination memory locations are equal. e.g [cv::cuda::SeparableLinearFilter::apply](https://github.com/opencv/opencv_contrib/blob/cb495d9b798a4bb3fed0cc76afe2e90dbf41e799/modules/cudafilters/src/filtering.cpp#L466)

But I haven't found the consistent approach that the original issue implies

> If it does not support inplace operations an error should appear similar to what happens with other functions.

Would it be better in this PR to [document ](https://docs.opencv.org/4.x/db/d29/group__cudawarping.html#ga9e8dd9e73b96bdc8e27d85c0e83f1130)that the function does not support in place operations with a simple assert
```
CV_Assert( src.data == dst.data );
```

mitresthen:
I think the suggestion to document that this function is not meant to be used inplace and combining it with an assertion is the best way forward. 
That way it will be very clear how the function should be used, even those who don't read the function docs will be made aware of the correct usage. The assert also serves as a contract so the function can assume  src.data != dst.data. 
Using source clone or buffering (options 4 and 5) adds unnecessary overhead and hidden complexity to the function. For a low-level library like OpenCV which is often used in realtime applications, this is probably not the best approach.

cudawarped:
> For a low-level library like OpenCV which is often used in realtime applications, this is probably not the best approach.

I agree but I would take it one step futher.  I don't think we should be checking to see if the source and destination are overlapping just that they are not the same for the reason you mentioned

> even those who don't read the function docs will be made aware of the correct usage.

If a user has a source and destination which are not equal but are overlapping then they likely have far bigger issues with their program and an assert on overlap is not going to help them out.

mitresthen:
Fair point! I've simplified it to be an equality check now, and removed all the overlap logic.

asmorkalov:
@cudawarped Thanks a lot for the investigation. You are right, there is no generic rule. Several asserts and copyTo calls we added as ad-hoc solution for some tickets. Different decisions were done depending on context: what the function logic and how it's already used to not break existing code. I would say that current solution with docs and assert is good enough.

asmorkalov (approved):

PR #3992, opened 2025-09-01
structured_light: GreyCodePattern 16-bit support and performance improvements

Atari2:
This PR aims to greatly improve performance of the GreyCodePattern algorithm in the StructuredLight module. It does so by caching and pre-calculating a bunch of stuff needed for the computation and instead of doing manual looping and indexing over the matrices it leverages built-in opencv functions to automatically parallelize work (as well as doing some manual parallelization). Additionally, thanks to the fact that it is using opencv functions, it gains the ability to correctly process 16-bit images, which is a plus.

Work done in more detail:
- Vastly improve performance of the GreyCodePattern structured light algorithm by leveraging built-in OpenCV functions and parallel code instead of raw loops, at the cost of a bit more memory (decoding a set of 96 5320x4600 images acquired using stereo cameras goes from taking ~100 seconds to taking ~2.5 seconds, in release mode)
- Automatically gain the ability to decode 16-bit images (switching from raw loops and uchar-indexing to OpenCV functions gives this for free)

Manual testing was done to ensure that the calculated results are correct, both for generation of the patterns and for decoding, in both Debug and Release mode, using real world data.

### DIFF
--- a/modules/structured_light/test/test_graycodepattern_fast.cpp
+++ b/modules/structured_light/test/test_graycodepattern_fast.cpp
@@ -1,0 +1,396 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
+ //
+ //  IMPORTANT: READ BEFORE DOWNLOADING, COPYING, INSTALLING OR USING.
+ //
+ //  By downloading, copying, installing or using the software you agree to this license.
+ //  If you do not agree to this license, do not download, install,
+ //  copy or use the software.
+ //
+ //
+ //                           License Agreement
+ //                For Open Source Computer Vision Library
+ //
+ // Copyright (C) 2015, OpenCV Foundation, all rights reserved.
+ // Third party copyrights are property of their respective owners.
+ //
+ // Redistribution and use in source and binary forms, with or without modification,
+ // are permitted provided that the following conditions are met:
+ //
+ //   * Redistribution's of source code must retain the above copyright notice,
+ //     this list of conditions and the following disclaimer.
+ //
+ //   * Redistribution's in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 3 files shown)


Atari2:
@asmorkalov sorry to ping but I wanted to know if there was any chance for this PR to get looked at or not. I would be very interested in this code ending up in the official release of opencv_contrib

modules/structured_light/src/graycodepattern.cpp
```diff
@@ -125,184 +125,152 @@ GrayCodePattern_Impl::GrayCodePattern_Impl( const GrayCodePattern::Params &param
...
+                }
+            }
+        }
+    });
+    parallel_for_(Range(0, params.height), [&](const Range& range) {
+        for (int i = range.start; i < range.end; ++i)
+        {
+            for (size_t k = 0; k < numOfRowImgs; ++k)
+            {
+                uchar flag = (((i >> k) & 1) ^ ((i >> (k + 1)) & 1));
```
asmorkalov: The similar comment here.

asmorkalov:
@Atari2 Could you also add some test(s) for the new feature?

Atari2:
Thank you for the review @asmorkalov, sorry again to disturb 😅 
I have clarified a bit the two code snippets that you highlighted by adding a small comment and changing a few variable names to make them clearer, hopefully that is enough. 

I also added a few tests to verify that the 16-bit images and data are correctly handled. Tell me if you need anything different and if what I added is OK. I see in the github CI that there is 1 run failing but I don't think that it is related to anything that I changed, because I only touched the structured_light submodule.

Atari2:
Any news on this? @asmorkalov

PR #3999, opened 2025-09-10
add missing cudaFree

incubus-ank:
I found missing memory cudaFree in cv::cuda::FAST_Impl::detectAsync()

I created an issue, but did not receive a response.
https://github.com/opencv/opencv_contrib/issues/3994

### DIFF
--- a/modules/cudafeatures2d/src/fast.cpp
+++ b/modules/cudafeatures2d/src/fast.cpp
@@ -141,6 +141,7 @@ namespace
         if (count == 0)
         {
             _keypoints.release();
+            cudaSafeCall( cudaFree(d_counter) );
             return;
         }
 


cudawarped:
@incubus-ank Would it be more performant to move the allocation into the constructor and free into the destructor of `FAST_Impl` to save the allocation on every invocatoin of `FAST_Impl::detectAsync`?

incubus-ank:
> @incubus-ank Would it be more performant to move the allocation into the constructor and free into the destructor of `FAST_Impl` to save the allocation on every invocatoin of `FAST_Impl::detectAsync`?

Yes, I think it's a good idea.

I'm not a pro at working with CUDA. Should I define/remove copy and move constructors to avoid funny situations and create a better interface?

cudawarped:
I would say no because this class can only be created with the factory method `cv::cuda::FastFeatureDetector::create` and accessed through the resulting smart pointer, @asmorkalov keep me honest here?

Ideally I would suggest using a `GpuMat` member variable for `d_counter` which would handle the memory management for you but that would involve altering the type from `unsigned int` to `int` and I not sure if this has any implications.

PR #4033, opened 2025-11-08
Add cv::cuda::cvtColorTwoPlane

srcejon:
This PR adds cv::cuda::cvtColorTwoPlane, similar to cv::cvtColorTwoPlane. 

Currently it supports COLOR_YUV2BGR_NV12 and COLOR_YUV2RGB_NV12 for CV_8U only. Unfortunately there doesn't appear to be a npp function supporting RGBA.

It requires NPP v13 or greater, as nppiNV12ToRGB_8u_ColorTwist32f_P2C3R_Ctx is buggy in earlier versions of NPP.

- [] There is a reference to the original bug report and related work
- [] There is accuracy test, performance test and test data in opencv_extra repository, if applicable

No patch to opencv_extra - but test is included that compares cv::cuda::cvtColorTwoPlane to cv::cvtColorTwoPlane.

### DIFF
--- a/modules/cudaimgproc/include/opencv2/cudaimgproc.hpp
+++ b/modules/cudaimgproc/include/opencv2/cudaimgproc.hpp
@@ -90,6 +90,24 @@ performance.
  */
 CV_EXPORTS_W void cvtColor(InputArray src, OutputArray dst, int code, int dcn = 0, Stream& stream = Stream::Null());
 
+/** @brief Converts an image from one color space to another where the source image is stored in two planes.
+
+This function currently only supports YUV420 in NV12 format to RGB/BGR conversion. The conversion used is:
+- R=1.163999557*(Y-16) + 1.5959997177(V-128)
+- G=1.163999557*(Y-16) -0.812999725(V-128) -0.390999794(U-128)
+- B=1.163999557*(Y-16) + 2.017999649(U-128)
+
+@param src1 8-bit image (CV_8U) of the Y plane.
+@param src2 image containing interleaved (CV_8UC2) U/V plane.
+@param dst Destination image.
+@param code Color space conversion code. It can take any of the following values:
+- #COLOR_YUV2BGR_NV12
+- #COLOR_YUV2RGB_NV12
+@param stream Stream for the asynchronous version.
+
+ */
+CV_EXPORTS_W void cvtColorTwoPlane(InputArray src1, InputArray src2, OutputArray dst, int code, Stream& stream = Stream::Null());
+
 enum DemosaicTypes
 {
     //! Bayer Demosaicing (Malvar, He, and Cutler)
--- a/modules/cudaimgproc/src/color.cpp
+++ b/modules/cudaimgproc/src/color.cpp
@@ -49,6 +49,8 @@ using namespace cv::cuda;
 
 void cv::cuda::cvtColor(InputArray, OutputArray, int, int, Stream&) { throw_no_cuda(); }
 
+void cv::cuda::cvtColorTwoPlane(InputArray, InputArray, OutputArray, int, Stream&) { throw_no_cuda(); }
+
 void cv::cuda::demosaicing(InputArray, OutputArray, int, int, Stream&) { throw_no_cuda(); }
 
 void cv::cuda::swapChannels(InputOutputArray, const int[4], Stream&) { throw_no_cuda(); }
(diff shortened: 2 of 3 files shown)


cudawarped:
@srcejon Would it be better to use [`cv::cudacodec::NVSurfaceToColorConverter`](https://docs.opencv.org/4.x/d9/dd0/classcv_1_1cudacodec_1_1NVSurfaceToColorConverter.html) as NV12 is a video format?  The only difference being `NVSurfaceToColorConverter` accepts NV12 as a single plane, however it supports
1. conversion to BGR, RGB, BGRA, RGBA and gray,
2. 8 and 16 bit input/output,
3. packed and planar formats,
4.  full and limited ranges, and
5. different color space standards.

e.g.
```
int rows = 1080;
int cols = 1920;
int type = CV_8U;
cv::Mat nv12(rows*1.5,cols, type);
cv::randu(nv12, 16, 235);
cv::cuda::GpuMat bgr;
Ptr<cv::cudacodec::NVSurfaceToColorConverter> yuvConverter = cudacodec::createNVSurfaceToColorConverter(cv::cudacodec::ColorSpaceStandard::BT601);
yuvConverter->convert(nv12, bgr, cv::cudacodec::SurfaceFormat::SF_NV12, cv::cudacodec::ColorFormat::BGR);
Mat bgrHost;
bgr.download(bgrHost);
```

If you agree then the requirement for Nvidia Video Codec SDK to be installed for this function to be available should be removed.

srcejon:
> @srcejon Would it be better to use [`cv::cudacodec::NVSurfaceToColorConverter`](https://docs.opencv.org/4.x/d9/dd0/classcv_1_1cudacodec_1_1NVSurfaceToColorConverter.html) as NV12 is a video format? The only difference being `NVSurfaceToColorConverter` accepts NV12 as a single plane

In my application (Which is using FFMPEG with HW decoder (pixfmt=AV_PIX_FMT_CUDA, so should be using NVDEC)), it appears to be giving me two plane NV12. As far as I can see, the UV plane isn't contiguous with the Y plane - there's a small gap inbetween.

(Why aren't I using cv::cudacodec? Because I have an mp4 with multiple video + audio streams, which doesn't appear to be supported)

Both cudacodec and FFPMEG are using NVDEC, so presumably there should be a way for it to work, but can't currently see it.

Maybe it's a useful function anyway - as it's in the main opencv library.

cudawarped:
I guess that's how FFMpeg wants to present the output, the Nvidia decoder uses

> cudaVideoSurfaceFormat_NV12=0,          /**< Semi-Planar YUV [Y plane followed by interleaved UV plane]     */

Anyway what's your use case for outputing NV12 and then converting to BGR/RGB and why only from SD files using BT.601 color space?

@asmorkalov Should we have additional CUDA functions for different data layouts?

srcejon:
> I guess that's how FFMpeg wants to present the output, the Nvidia decoder uses
> 
> > cudaVideoSurfaceFormat_NV12=0,          /**< Semi-Planar YUV [Y plane followed by interleaved UV plane]     */

I will dig in to it more to try to find out why it appears different and if really necessary to have two plane conversion.

> Anyway what's your use case for outputing NV12 and then converting to BGR/RGB and why only from SD files using BT.601 color space?

Good question, that's possibly not what I want. I just blindly copied the BT.601 coeffs from cv::cvtColorTwoPlane so the functionality is the same. (cv::cvtColor() and cv::cvtColorTwoPlane() don't have color space & range as parameters).

The source mp4 is actually created using cudacodec::VideoWriter with 4096x3000 BGRA input, h264 and mostly default EncoderParams. It's not clear from the OpenCV or NVEnc docs what the default color space used is. It looks like NvEncoder.cpp doesn't set h264VUIParameters explictly.

edit: Running a few tests shows that it defaults to the BT.709 color space. [EncoderParams](https://docs.opencv.org/4.x/d3/db6/structcv_1_1cudacodec_1_1EncoderParams.html#aeb2b4fad493b471a9e9f6d55b3881e35).[videoFullRangeFlag](https://docs.opencv.org/4.x/d3/db6/structcv_1_1cudacodec_1_1EncoderParams.html#a23f9da61ec6790d261b5d21473045c80) determines whether full or limited range is used for the YUV values.

srcejon:
> I will dig in to it more to try to find out why it appears different and if really necessary to have two plane conversion.

The issue is that my frame height is 3000, which isn't divisible by 16, so it allocates a 3008 line YUV buffer. Those extra 8 lines account for the gap between the Y and UV when viewed as a 3000 line frame.

So, it seems it is useful to have a two plane conversion function, but also should have the other functionality of NVSurfaceToColorConverter. Perhaps NVSurfaceToColorConverter::convert could have an overloaded version with separate y and uv parameters? This would be preferable to using the NPP functions as it could support RGBA and also wouldn't depend on NPP v13.

cudawarped:
Not sure I understand.  You are encoding a 4096x3000 video to h264 using `cudacodec::VideoWriter`, then encapsulating it in an mp4 with audio and other encoded streams.  Then you are decoding using FFmpeg because OpenCV only supports a single video stream to two plane NV12 which you then want to convert back to BGR.  Why are you decoding to NV12 with FFmpeg and not directly decoding to BGR?

> The issue is that my frame height is 3000, which isn't divisible by 16, so it allocates a 3008 line YUV buffer. Those extra 8 lines account for the gap between the Y and UV when viewed as a 3000 line frame.

If its a single plane you should still be able to convert it using the function I mentioned where the NV12 input frame has1.5*3008 rows.

srcejon:
> Why are you decoding to NV12 with FFmpeg and not directly decoding to BGR?

I can't see an option to get BGR when the decoder is NVDEC. To clarify, I'm using libavcodec directly rather than the ffmpeg executable.

> > The issue is that my frame height is 3000, which isn't divisible by 16, so it allocates a 3008 line YUV buffer. Those extra 8 lines account for the gap between the Y and UV when viewed as a 3000 line frame.
> 
> If its a single plane you should still be able to convert it using the function I mentioned where the NV12 input frame has1.5*3008 rows.

Yes, if I oversize the buffers with the extra lines, then it works. Thanks.

asmorkalov:
CI:
```
[ RUN      ] CUDA_ImgProc/CvtColorTwoPlane.YUV2BGR_NV12/0, where GetParam() = (Quadro P2000, 128x128, CV_8U, whole matrix)
unknown file: Failure
C++ exception with description "OpenCV(4.13.0-dev) /home/ci/opencv_contrib/modules/cudaimgproc/src/color.cpp:2125: error: (-206:Bad flag (parameter or structure field)) cvtColorTwoPlane not supported with NPP <= 12 as nppiNV12ToRGB_8u_ColorTwist32f_P2C3R_Ctx is buggy in function 'cvtColorTwoPlane'
" thrown in the test body.
[  FAILED  ] CUDA_ImgProc/CvtColorTwoPlane.YUV2BGR_NV12/0, where GetParam() = (Quadro P2000, 128x128, CV_8U, whole matrix) (81 ms)
[ RUN      ] CUDA_ImgProc/CvtColorTwoPlane.YUV2BGR_NV12/1, where GetParam() = (Quadro P2000, 128x128, CV_8U, sub matrix)
[       OK ] CUDA_ImgProc/CvtColorTwoPlane.YUV2BGR_NV12/1 (59 ms)
[ RUN      ] CUDA_ImgProc/CvtColorTwoPlane.YUV2RGB_NV12/0, where GetParam() = (Quadro P2000, 128x128, CV_8U, whole matrix)
unknown file: Failure
C++ exception with description "OpenCV(4.13.0-dev) /home/ci/opencv_contrib/modules/cudaimgproc/src/color.cpp:2125: error: (-206:Bad flag (parameter or structure field)) cvtColorTwoPlane not supported with NPP <= 12 as nppiNV12ToRGB_8u_ColorTwist32f_P2C3R_Ctx is buggy in function 'cvtColorTwoPlane'
" thrown in the test body.
[  FAILED  ] CUDA_ImgProc/CvtColorTwoPlane.YUV2RGB_NV12/0, where GetParam() = (Quadro P2000, 128x128, CV_8U, whole matrix) (21 ms)
```

srcejon:
> : error: (-206:Bad flag (parameter or structure field)) cvtColorTwoPlane not supported with NPP <= 12 as nppiNV12ToRGB_8u_ColorTwist32f_P2C3R_Ctx is buggy in function 'cvtColorTwoPlane'" thrown in the test body.

Yep, NPP 13 is required, as there is a bug in nppiNV12ToRGB_8u_ColorTwist32f_P2C3R_Ctx in version 12.

Should the test be marked as expected to fail or similar?

asmorkalov:
I'm looking on options proposed by @cudawarped and alternatives.

PR #4052, opened 2025-12-11
Stream waits in corners.cu before textures destruction

gagicb:
Using a texture handle after calling cudaDestroyTextureObject results in undefined behavior.
In GFTT/corners, we had the following sequence:

    Enter function F
    Create texture T
    Launch kernel K on stream S
    Return from function F, which destroys texture T

At step 4, we must guarantee that kernel K launched on stream S has completed and that T is no longer used before it is destroyed. Thats why I added stream synchronization before function return.

Relevant description from the CUDA SDK documentation:

https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TEXTURE__OBJECT.html#group__CUDART__TEXTURE__OBJECT__1_gae41ef138ba69d54b91068247f8f6c4e8:~:text=Use%20of%20the%20handle%20after%20this%20call%20is%20undefined%20behavior.

### DIFF
--- a/modules/cudaimgproc/src/cuda/corners.cu
+++ b/modules/cudaimgproc/src/cuda/corners.cu
@@ -154,7 +154,9 @@ namespace cv { namespace cuda { namespace device
             cudaSafeCall( cudaGetLastError() );
 
             if (stream == 0)
-                cudaSafeCall( cudaDeviceSynchronize() );
+                cudaSafeCall(cudaDeviceSynchronize());
+            else
+                cudaSafeCall(cudaStreamSynchronize(stream));
         }
 
         /////////////////////////////////////////// Corner Min Eigen Val /////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


asmorkalov:
cc @cudawarped

asmorkalov:
@gagicb Could you add some issue description and why you add the synchronization for non-default stream. Some reference to documentation is useful too.

gagicb:
> @gagicb Could you add some issue description and why you add the synchronization for non-default stream. Some reference to documentation is useful too.

Hi @asmorkalov,

Using a texture handle after calling cudaDestroyTextureObject results in undefined behavior.
In GFTT/corners, we had the following sequence:

1. Enter function F
2. Create texture T
3. Launch kernel K on stream S
4. Return from function F, which destroys texture T

At step 4, we must guarantee that kernel K launched on stream S has completed and that T is no longer used before it is destroyed. Thats why I added stream synchronization before function return.

Relevant description from the CUDA SDK documentation:

https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TEXTURE__OBJECT.html#group__CUDART__TEXTURE__OBJECT__1_gae41ef138ba69d54b91068247f8f6c4e8:~:text=Use%20of%20the%20handle%20after%20this%20call%20is%20undefined%20behavior.

Best regards,
gagik

PR #4063, opened 2026-01-03
Clarify disparity_map_right type in DisparityFilter documentation

muskan-101:
The documentation of cv::ximgproc::DisparityFilter::filter()
did not specify the required type of the optional right disparity map.

In practice, the implementation requires a single-channel CV_32F matrix,
otherwise a runtime assertion is triggered.

This PR clarifies the expected type in the Doxygen comment.


muskan-101:
Thank you so much for considering my PR.

PR #4065, opened 2026-01-08
quality: Optimize SSIM computation for ~2x performance improvement

cl445:
## Summary

Optimizes `QualitySSIM` (the SSIM computation in the `quality` module) and adds performance
tests for it. The largest, most reproducible gain is on the **CPU path** (up to ~2.3x for the
precomputed-reference use case); the OpenCL path is also faster on most sizes.

Related: #3797 (earlier buffer-reduction proposal for SSIM).

### What changed

- **Single fused pass for the final SSIM formula.** The score/quality-map step used to run as
  ~8 separate element-wise OpenCV calls, each a full read/write pass over the image (memory
  bandwidth bound). It is now a single pass: an OpenCL kernel on the GPU path (`src/opencl/ssim.cl`)
  and a `cv::parallel_for_` loop on the CPU path. `sigma12 = blur(I1·I2) − mu1·mu2` is recovered
  inline in both.
- **CV_32F throughout.** For floating-point inputs this replaces the previous CV_64F path
  (≈2x on those inputs). For 8-bit inputs the module already worked in CV_32F (via `expand_mat`),
  so there the gain comes from the items above, not from precision.
- **`sepFilter2D` with a cached 1D Gaussian kernel** for the blur (maps to a faster path than
  `GaussianBlur` on the OpenCL backend; on par on CPU).
- **`need_quality_map`** internal flag so the quality map isn't copied back to the caller when
  not requested.
- **Performance tests** (`perf/perf_ssim.cpp`) — the `quality` module had none.

### Measured performance

Apple M4 Pro, OpenCV 4.x, Release, statically linked benchmark binaries. Times are the perf
framework's median; the speedup is the ratio of best-of-4 runs (old implementation → this PR),
interleaved to control for thermal drift.

**CPU path** (most reproducible):

| Size | Channels | before | after | speedup |
|---|---|---|---|---|
| VGA 640×480 | 1 | 2.21 ms | 1.64 ms | 1.35x |
| 720p 1280×720 | 3 | 20.0 ms | 15.2 ms | 1.32x |
| 1080p 1920×1080 | 1 | 8.82 ms | 5.08 ms | 1.74x |
| 1080p 1920×1080 | 3 | 24.1 ms | 12.7 ms | 1.89x |
| 1080p, precomputed ref | 1 | 7.14 ms | 3.27 ms | 2.18x |
| 1080p, precomputed ref | 3 | 20.2 ms | 8.94 ms | 2.26x |

**OpenCL path**: also faster on most sizes (roughly 1.2–1.9x for VGA/720p in my runs), but the
measurements are noisy on this hardware (run-to-run swings of ±20–40% even for identical code),
so I'm not quoting precise per-size numbers — CI on other GPUs is the better signal here.

### Accuracy

The math is unchanged (same formula, FP32). All existing `opencv_test_quality` SSIM cases pass
on both the CPU and OpenCL paths; SSIM scores stay well within the module's test tolerance
(`QUALITY_ERR_TOLERANCE = 0.002`).

### API / compatibility

- Public API signatures unchanged.
- The quality map is now `CV_32F` (was `CV_64F`); for 8-bit inputs this matches the previous
  behavior, for floating-point inputs it is the intended precision/perf trade-off.

- [x] The PR is proposed to the proper branch (4.x)
  - Performance tests added (`opencv_perf_quality`); accuracy is covered by the existing `test_ssim` cases (no new test data required).


cl445:
The macOS-X64 CI failure appears to be unrelated to this PR:

- **Failed step:** `Performance:imgcodecs` (exit code 139 = SIGSEGV)
- **This PR changes:** `modules/quality/` (SSIM optimization)

Other recent PRs (e.g., #4057) also show macOS failures in different modules, suggesting infrastructure issues.

Could a maintainer please re-run the failed CI job? Thank you!

cl445:
Rebased onto the latest `4.x` and addressed a couple of self-review points:

- Documented the `CV_64F` → `CV_32F` precision change explicitly (the quality map type changes to `CV_32F`; SSIM scores stay within ~1e-4, well below the test tolerance of `0.002`, so all existing `test_ssim` cases pass).
- Minor cleanup: the cached 1D Gaussian kernel is now returned by `const&` to avoid a per-call header copy.

The earlier macOS CI failure was unrelated (a SIGSEGV in `Performance:imgcodecs`); the required precommit builds are green now.

@asmorkalov would you (or another maintainer of the quality module) have a chance to take a look when time permits? The change is backward compatible at the API level and gives roughly a 2x speedup. Happy to adjust anything. Thanks!

cl445:
Force-pushed an update:

- **New:** the final SSIM formula now runs as a single fused pass on the CPU path too (`cv::parallel_for_`), mirroring the existing OpenCL kernel. Previously it was ~8 separate element-wise calls (memory-bandwidth bound). This is the main new win.
- Reworked the benchmark numbers in the description to measured, reproducible values (statically-linked binaries, best-of-4 interleaved runs on an Apple M4 Pro). The CPU path is now ~1.3x (VGA/720p) up to ~2.2–2.3x (precomputed reference, 1080p).
- Corrected the description: for 8-bit inputs the module already worked in CV_32F, so the precision change only affects floating-point inputs; the 8-bit gains come from the fused pass and `sepFilter2D`.
- All existing `opencv_test_quality` SSIM tests still pass on both CPU and OpenCL paths.

OpenCL timings are quite noisy on this hardware, so I'm relying on CI on other GPUs for that path.

PR #4076, opened 2026-02-07
cudacodec: fix black/white/uvWhite of BT2020/BT2020C

quink-black:
This patch addresses two critical issues in color conversion:

- The previous implementation caused severe color shifts because uvWhite was handled as an 8-bit value while the luminance was processed as 10-bit P010.

- Additionally, the code was hardcoded to Limited Range.

- x There is a reference to the original bug report and related work
- x There is accuracy test, performance test and test data in opencv_extra repository, if applicable


asmorkalov:
cc @cudawarped

asmorkalov:
@quink-black Could you a simple test for the case? E.g. encode syntetic image as video, decode with the cudacodecs and check pixel colors?

quink-black:
> @quink-black Could you a simple test for the case? E.g. encode syntetic image as video, decode with the cudacodecs and check pixel colors?

Done. colorspace conversion doesn't need encoder/decoder.

quink-black:
ping.

asmorkalov (approved):

PR #4086, opened 2026-02-26
add CV_16F support in hdf5

### LINKED ISSUE #4075: [feature] CV_16F support in HDF5
HDF5 v2.0.0 finally introduced Float16 support. So we should take the advantage to allow save/load CV_16F in opencv.

Right now we have to cast it to CV_16S then save as int16 in HDF5, at the load side we have to do the cast again.

https://github.com/opencv/opencv_contrib/pull/4074 allow us to directly save CV_16F data as it is with proper metadata

Kethan1701:
HDF5 >= 1.14.4 introduced IEEE float16 predefined datatypes. This change adds the ability to directly save/load CV_16F matrices in HDF5 without casting to CV_16S.

- Add CV_16F case to GetH5type() using H5T_IEEE_F16LE
- Add float16 detection to GetCVtype() via H5T_FLOAT class + 2-byte size
- Add float16 detection to dsgettype(), dsread(), atread() to prevent H5Tget_native_type from promoting float16 to float32
- Handle both single-channel and multi-channel (H5T_ARRAY) cases
- All changes guarded with #if H5_VERSION_GE(1,14,4)
- Add CV_16F to existing attribute depth tests
- Add dedicated float16 dataset and attribute write/read tests

Fixes #4075

### DIFF
--- a/modules/hdf/src/hdf5.cpp
+++ b/modules/hdf/src/hdf5.cpp
@@ -38,1399 +38,1395 @@
 
 using namespace std;
 
-namespace cv
-{
-namespace hdf
-{
+namespace cv {
+namespace hdf {
 
-class HDF5Impl CV_FINAL : public HDF5
-{
+class HDF5Impl CV_FINAL : public HDF5 {
 public:
+  HDF5Impl(const String &HDF5Filename);
 
-    HDF5Impl( const String& HDF5Filename );
+  virtual ~HDF5Impl() CV_OVERRIDE { close(); };
 
-    virtual ~HDF5Impl() CV_OVERRIDE { close(); };
+  // close and release
+  virtual void close() CV_OVERRIDE;
 
-    // close and release
-    virtual void close( ) CV_OVERRIDE;
+  /*
+   * h5 generic
+   */
 
-    /*
-     * h5 generic
-     */
+  // check if object / link exists
+  virtual bool hlexists(const String &label) const CV_OVERRIDE;
 
-    // check if object / link exists
-    virtual bool hlexists( const String& label ) const CV_OVERRIDE;
+  virtual bool atexists(const String &atlabel) const CV_OVERRIDE;
+  virtual void atdelete(const String &atlabel) CV_OVERRIDE;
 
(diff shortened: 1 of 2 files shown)


Utkarshkhandka:
I reviewed PR #4086 and have a few points to address before merging.

1. Big-endian detection: The PR maps `CV_16F` to `H5T_IEEE_F16LE` (little-endian) unconditionally. On big-endian systems (some HPC platforms do use OpenCV + HDF5), this will silently write data in the wrong byte order. The CMake detection step should also check endianness:

```
cpp
// In GetH5type():
#if H5_VERSION_GE(1,14,4)
case CV_16F:
#if defined(WORDS_BIGENDIAN) || defined(__BIG_ENDIAN__)
    return H5T_IEEE_F16BE;
#else
    return H5T_IEEE_F16LE;
#endif
#endif
```

2. NaN/Inf round-trip correctness: CV_16F can hold NaN and Inf values (used in some DNN outputs). The existing tests only check normal float16 values. It would be worth adding:

```cpp
// Test NaN and Inf survive round-trip
Mat nanmat(1, 1, CV_16F);
nanmat.at(0,0) = std::numeric_limits::infinity();
// write → read → compare bits exactly
```

3. H5T_ARRAY path for multi-channel
In `GetCVtype()`, when the HDF5 type is `H5T_ARRAY` (multi-channel), does the 2-byte float detection still fire correctly after `H5Tget_super()`? Worth confirming that the multi-channel test actually exercises this path (by checking the test output with `--gtest_filter=*float16*` and a debug print of the HDF5 type class).

4. Minor: version guard consistency
PR uses `#if H5_VERSION_GE(1,14,4)` but the issue description references HDF5 v1.14.3 as the first version with float16. Worth verifying which is actually the first stable release with `H5T_IEEE_F16LE` defined - could be 1.14.3.

Good work on the implementation overall — the `dsgettype()` / `dsread()` / `atread()` native type promotion fix is the right approach and easy to miss.

braindevices:
I already fully implemented https://github.com/opencv/opencv_contrib/pull/4078

PR #4088, opened 2026-03-09
ximgproc: fix angle difference normalization in FastLineDetector merging

### LINKED ISSUE #3956: fast_line_detector do not merge segments oriented in the opposite direction
The function ```bool FastLineDetectorImpl::mergeSegments``` allows to merge small segments into a longer one based on the distance between these segments and their orientation.

Currently, two segments on the same line with two different directions (angle of 180° and -180° for example) are not merge as the difference between their two angles is computed as:

```c
float angdiff = fabs(seg1.angle - seg2.angle);
```
See https://github.com/opencv/opencv_contrib/blob/1e4d4e0f3e4c7d6d7ab9d738026fe13d3cd85cf4/modules/ximgproc/src/fast_line_detector.cpp#L268C5-L268C51

The difference should be computed as:
```c
float angdiff = fabs(seg1.angle - seg2.angle) % 180.0f ;
```

AYELET0527661122:
Fixes https://github.com/opencv/opencv_contrib/issues/3956
Fix angle difference computation in FastLineDetector::mergeSegments.

Previously the angular difference was computed using a direct absolute
difference between segment angles. As a result, collinear segments with
opposite directions (~pi radians apart) could fail to merge.

The angle difference is now normalized so that segments differing by pi
radians are treated as equivalent.

A regression test (mergeOppositeDirections) was added to verify that
touching parallel segments are merged correctly for both same and
opposite directions.

### DIFF
--- a/modules/ximgproc/test/test_fld.cpp
+++ b/modules/ximgproc/test/test_fld.cpp
@@ -199,6 +199,64 @@ TEST_F(ximgproc_FLD, mergeLines)
     ASSERT_EQ(EPOCHS, passedtests);
 }
 
+TEST_F(ximgproc_FLD, mergeOppositeDirections)
+{
+    std::vector<std::pair<cv::Vec4f, cv::Vec4f>> touchingParallelPairs;
+
+    RNG rng(12345);
+    std::vector<cv::Point2f> dirs;
+    for (int i = 0; i < 10; i++)
+    {
+        float x = rng.uniform(1.f, 99.f);
+        float y = 100.f - x;
+        dirs.emplace_back(x, y);
+    }
+    
+    // --- Add the four fixed vectors ---
+    dirs.emplace_back(100.f,   0.f);
+    dirs.emplace_back(  0.f, 100.f);
+    dirs.emplace_back(-100.f,  0.f);
+    dirs.emplace_back(  0.f, -100.f);
+    // --- Construct segment pairs ---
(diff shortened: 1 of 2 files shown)


AlexanderHaritonov:
Please remove trailing whitespaces.
See https://pullrequest.opencv.org/buildbot/builders/precommit-contrib_docs/builds/101694/steps/whitespace%20opencv_contrib/logs/stdio

asmorkalov:
Rebased to 4.x to trigger CI.

modules/ximgproc/test/test_fld.cpp
```diff
@@ -199,6 +199,64 @@ TEST_F(ximgproc_FLD, mergeLines)
     ASSERT_EQ(EPOCHS, passedtests);
 }
 
+TEST_F(ximgproc_FLD, mergeOppositeDirections)
+{
+    std::vector<std::pair<cv::Vec4f, cv::Vec4f>> touchingParallelPairs;
+
+    RNG rng(12345);
```
asmorkalov, line 206: Please just use theRNG to fix warning:
```
  Warning: /home/ubuntu/opencv_contrib/modules/ximgproc/test/test_fld.cpp:206:9: warning: declaration of 'rng' shadows a member of 'opencv_test::{anonymous}::ximgproc_FLD_mergeOppositeDirections_Test' [-Wshadow]
```
The OpenCV test system initializes theRNG for each test individually.

PR #4107, opened 2026-04-15
Fix wechat qrcode 3519

### LINKED ISSUE #3519: QR code recognition module has bugs
### System Information

OpenCV version: 4.7.0
Operating System / Platform: win11
Compiler & compiler version:  C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.29.30133/bin/Hostx64/x64/cl.exe

### Detailed description

ok，Above is the compilation environment and runtime environment，When using QR code recognition, no error is reported under release, but under debug, the error reported is array out of bounds, below I attach the code and run the results through the image

When scale=2, an error is reported, from the above figure you can find that the index has a value, but the data is empty, the reason is unknown

### Steps to reproduce

ok，Above is the compilation environment and runtime environment，When using QR code recognition, no error is reported under release, but under debug, the error reported is array out of bounds, below I attach the code and run the results through the image

When scale=2, an error is reported, from the above figure you can find that the index has a value, but the data is empty, the reason is unknown

- [ ] I checked the problem with documentation, FAQ, open issues, forum.opencv.org, Stack Overflow, etc and have not found any solution
- [ ] I updated to the latest OpenCV version and the issue is still there

Nikhil-jaiswal007:
### Description
Fix potential out-of-bounds / invalid access in WeChatQRCode::detectAndDecode
when no QR candidates are detected.

When detector returns empty candidate_points (e.g. scaled images with scale=2),
decode() was still called. This may lead to inconsistent state and crashes in
debug builds.

### Fix
Add early return when candidate_points is empty before calling decode().

### Patch
modules/wechat_qrcode/src/wechat_qrcode.cpp

Before:
auto candidate_points = p->detect(input_img);
auto res_points = vector<Mat>();
auto ret = p->decode(input_img, candidate_points, res_points);

After:
auto candidate_points = p->detect(input_img);

if (candidate_points.empty())
{
    if (points.needed())
        points.release();
    return vector<string>();
}

auto res_points = vector<Mat>();
auto ret = p->decode(input_img, candidate_points, res_points);

### Issue
Fixes #3519

### DIFF
--- a/modules/wechat_qrcode/src/wechat_qrcode.cpp
+++ b/modules/wechat_qrcode/src/wechat_qrcode.cpp
@@ -20,6 +20,8 @@ public:
     Impl() {}
     ~Impl() {}
     /**
+     * 
+     * 
      * @brief detect QR codes from the given image
      *
      * @param img supports grayscale or color (BGR) image.
@@ -96,9 +98,20 @@ vector<string> WeChatQRCode::detectAndDecode(InputArray img, OutputArrayOfArrays
     } else {
         input_img = img.getMat();
     }
-    auto candidate_points = p->detect(input_img);
-    auto res_points = vector<Mat>();
-    auto ret = p->decode(input_img, candidate_points, res_points);
+   auto candidate_points = p->detect(input_img);
+
+// No QR candidates detected (can happen for scaled images, e.g. scale=2).
+// Avoid calling decode() with empty input which may lead to out-of-bounds.
+if (candidate_points.empty())
+{
+    if (points.needed())
+        points.release();
+    return vector<string>();
+}
+
+   auto res_points = vector<Mat>();
+   auto ret = p->decode(input_img, candidate_points, res_points);
+   
     // opencv type convert
     vector<Mat> tmp_points;
     if (points.needed()) {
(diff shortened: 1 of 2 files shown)


modules/wechat_qrcode/src/wechat_qrcode.cpp
```diff
@@ -20,6 +20,8 @@ class WeChatQRCode::Impl {
     Impl() {}
     ~Impl() {}
     /**
+     *
```
vrabaud: can you remove those ? Thx

Nikhil-jaiswal007: Done

modules/wechat_qrcode/src/wechat_qrcode.cpp
```diff
@@ -96,9 +97,20 @@ vector<string> WeChatQRCode::detectAndDecode(InputArray img, OutputArrayOfArrays
...
+   auto candidate_points = p->detect(input_img);
+
+// No QR candidates detected (can happen for scaled images, e.g. scale=2).
+// Avoid calling decode() with empty input which may lead to out-of-bounds.
+if (candidate_points.empty())
+{
+    if (points.needed())
+        points.release();
+    return vector<string>();
+}
```
asmorkalov, line 109: Formatting is broken.

PR #4109, opened 2026-04-16
imgaug: fix RNG consistency and add reproducibility tests

CassieuiL:
This PR is intentionally scoped to RNG behavior and reproducibility verification in `imgaug`.

### What this PR does
- centralizes internal `imgaug` RNG access without changing the public API shape
- keeps `cv::imgaug::setSeed()`, `cv::imgaug::rng`, and `cv::imgaug::state` intact
- routes `ColorJitter` random ordering through the unified RNG source
- clarifies deterministic behavior of `setSeed()` in Doxygen
- adds 5 self-contained reproducibility tests

### Local validation

The RNG behavior is independently validated:

- the 5 new reproducibility tests pass locally
- they cover:
  - same seed ⇒ same output
  - same seed + same call sequence ⇒ same output sequence
  - different seed ⇒ different output
  - reproducibility for detection transforms
  - same seed ⇒ same bbox results

### Note on existing `opencv_test_imgaug`

- I also ran the full `opencv_test_imgaug.exe`
- 17/22 tests fail immediately due to:
  `findDataFile("imgaug/...")` not locating external test assets
- these are legacy data-driven tests depending on `opencv_extra/testdata/imgaug`
- failures are unrelated to this RNG change

 Therefore, these should not be considered regressions introduced by this PR.

### Follow-up work

- restore matching `opencv_extra/testdata/imgaug`, or
- migrate legacy tests to self-contained tests

This will be handled in a separate PR.


CassieuiL:
Note:
The 17 failing tests in opencv_test_imgaug are due to missing external imgaug test assets (opencv_extra/testdata/imgaug), not caused by this change.
The new reproducibility tests are self-contained and validate RNG behavior directly.

PR #4144, opened 2026-06-08
cudev: define ulong on Windows so vec_traits compiles with CUDA

ssam18:
### Fixes opencv/opencv#29265

Building the CUDA enabled core module on Windows with Visual Studio fails to compile cudev `vec_traits.hpp` with `identifier "ulong" is undefined`. The header specializes `MakeVec` and `VecTraits` for the scalar `ulong` type, but `ulong` only exists on glibc based platforms through `<sys/types.h>`. On Windows it is not a builtin and CUDA only ships the vector types `ulong1..4`, not the scalar, so the build breaks. This was introduced when the `long` and `ulong` rows were added on the 5.x branch, which is why the same module builds fine on Linux but not on Windows.

The change defines `ulong` as `unsigned long` on Windows, which matches the element type CUDA uses for `ulong1..4` on each platform, so the existing specializations compile. It is guarded to `_WIN32` so platforms that already provide `ulong` are untouched.

I could not run a full nvcc build locally since this machine has no CUDA toolkit, and the error is Windows only so it does not reproduce on Linux where `ulong` is already defined. I reproduced the exact failure with a minimal program that mimics the macro pattern without a scalar `ulong`, confirmed it fails the same way, and confirmed the typedef makes it compile.

<details>
<summary>Pull Request Readiness Checklist</summary>

</details>

### DIFF
--- a/modules/cudev/include/opencv2/cudev/util/vec_traits.hpp
+++ b/modules/cudev/include/opencv2/cudev/util/vec_traits.hpp
@@ -49,6 +49,19 @@
 #include "../common.hpp"
 #include "opencv2/core/cuda/cuda_compat.hpp"
 
+// The 64-bit integer vector-trait rows use the platform's 64-bit integer scalar
+// together with CUDA's matching vector types. On LP64 platforms (e.g. Linux)
+// that is `long`/`ulong` with CUDA's `long1..4`/`ulong1..4`. On Windows `long`
+// is 32-bit, so the 64-bit rows must instead use `long long`/`unsigned long
+// long` with CUDA's `longlong1..4`/`ulonglong1..4`; otherwise
+// VecTraits<int64_t>/<uint64_t> - which core's gpu_mat.cu instantiates - have no
+// matching specialization and the CUDA core module fails to build. CUDA ships
+// the vector types but not the scalar aliases, so define them here.
+#if defined(_WIN32)
+typedef long long          longlong;
+typedef unsigned long long ulonglong;
+#endif
+
 namespace cv {
 
     using cv::cuda::device::compat::double4;
@@ -76,8 +89,14 @@ CV_CUDEV_MAKE_VEC_INST(int)
 CV_CUDEV_MAKE_VEC_INST(uint)
 CV_CUDEV_MAKE_VEC_INST(float)
 CV_CUDEV_MAKE_VEC_INST(double)
+// 64-bit integer rows: width-correct scalar + CUDA vector types per platform.
+#if defined(_WIN32)
+CV_CUDEV_MAKE_VEC_INST(longlong)
+CV_CUDEV_MAKE_VEC_INST(ulonglong)
+#else
 CV_CUDEV_MAKE_VEC_INST(long)
 CV_CUDEV_MAKE_VEC_INST(ulong)
+#endif
 
 #undef CV_CUDEV_MAKE_VEC_INST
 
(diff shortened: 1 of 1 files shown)


asmorkalov (approved):

modules/cudev/include/opencv2/cudev/util/vec_traits.hpp
```diff
@@ -49,6 +49,15 @@
 #include "../common.hpp"
 #include "opencv2/core/cuda/cuda_compat.hpp"
 
+// `ulong` is provided by <sys/types.h> on glibc-based platforms, but it is not
```
O-HIGELIN: This doesn't work.
In Windows long is 32 bits so long ulong will be equivalent type to int uint.
What I suggest:
```
#if defined(_WIN32)
typedef long long longlong;
typedef unsigned long long ulonglong;
#endif
```
In head of file then in macro defined types
```
//For int 64 types
#if defined(_WIN32)
CV_CUDEV_MAKE_VEC_INST(longlong)
CV_CUDEV_MAKE_VEC_INST(ulonglong)
#else
CV_CUDEV_MAKE_VEC_INST(long)
CV_CUDEV_MAKE_VEC_INST(ulong)
#endif

```
and...
```
//For int 64 types
#if defined(_WIN32)
CV_CUDEV_VEC_TRAITS_INST(longlong)
CV_CUDEV_VEC_TRAITS_INST(ulonglong)
#else
CV_CUDEV_VEC_TRAITS_INST(long)
CV_CUDEV_VEC_TRAITS_INST(ulong)
#endif
```

ssam18:
Thanks for the review. The `ulong` here isn't meant to be a 64-bit type, it's just the scalar that CUDA's ulong1..4 vector types are built on, and CUDA makes those unsigned long (which is 32-bit on Windows), so unsigned long is the width that actually matches. The only thing missing on Windows was that scalar, which is why the build broke on the ulong specialization. 64-bit Mats go through int64_t/uint64_t in core rather than this row, so I'd keep long/ulong as is to stay in line with the Linux side.

O-HIGELIN:
> Thanks for the review. The `ulong` here isn't meant to be a 64-bit type, it's just the scalar that CUDA's ulong1..4 vector types are built on, and CUDA makes those unsigned long (which is 32-bit on Windows), so unsigned long is the width that actually matches. The only thing missing on Windows was that scalar, which is why the build broke on the ulong specialization. 64-bit Mats go through int64_t/uint64_t in core rather than this row, so I'd keep long/ulong as is to stay in line with the Linux side.

I understand it very well. But if you only solve undefined ulong typedef, you will solve compilation error on vec_traits.hpp, but further you will have compilation error for incomplete definition type in gpu_mat.cu expecting a definition for 64 template.
I don't really know about Linux, but under Windows uint or ulong will produce the same template type.
If you want 64 bits integer template type you have to declare longlong and ulonglong template type.
For me this work fine and solve my problem.

ssam18:
You're right, thanks for catching this and for the suggestion.

I dug into it and here is what is happening. Defining ulong does get vec_traits.hpp to compile, but core's gpu_mat.cu instantiates VecTraits<int64_t> and VecTraits<uint64_t> for the CV_64S and CV_64U convert paths. On Windows those two are long long and unsigned long long, which are not the same type as the 32 bit unsigned long, so they never get a specialization and you hit the incomplete type error you described.

So I redid the patch the way you suggested. On _WIN32 the 64 bit rows now use long long and unsigned long long with CUDA's longlong1..4 and ulonglong1..4, and I added the scalar typedefs since CUDA gives you the vector types but not the scalars. Linux stays on long and ulong like before.

Credit to you for spotting this and for the fix. When you get a chance, could you re-test on your Windows CUDA build? Thanks.

asmorkalov:
@ssam18 There are many merge conflicts in the PR. Please rebase and fix them.

ssam18:
> @ssam18 There are many merge conflicts in the PR. Please rebase and fix them.

done

asmorkalov:
@ssam18 is the problem specific to 5.x or reproducible with 4.x too?

ssam18:
> @ssam18 is the problem specific to 5.x or reproducible with 4.x too?

It's specific to 5.x and 4.x isn't affected.

The long/ulong rows in cudev/util/vec_traits.hpp only exist on 5.x. They were added in 9bff619b (cudev types long, ulong) to back the new CV_64S/CV_64U depths. On 4.x, the MakeVec/VecTraits instantiation lists stop at double, so ulong is never named in that header, and the Windows build can't hit the "identifier ulong is undefined" error.

So there's nothing to backport, and fixing it only makes sense on 5.x.

PR #4175, opened 2026-07-11
xfeatures2d: fix DAISY roi descriptor buffer overflow

shimat:
`ComputeDescriptorsInvoker` computed the output row index as an absolute image offset (`y*image->cols + x`), but the descriptors buffer passed to `DAISY::compute(image, roi, descriptors)` is only `roi.width*roi.height` rows, addressed relative to the roi. Unless roi covers the whole image, this writes past the end of the allocated buffer.

Store the roi's y offset alongside the existing x offset and compute the row index relative to the roi. The absolute `(y, x)` passed to `get_unnormalized_descriptor()` is unchanged, since sampling is done against the full-image gradient layers regardless of roi.

Added a regression test in `test_daisy.cpp` comparing descriptors computed via a roi against the same pixels computed via the full image.

### DIFF
--- a/modules/xfeatures2d/src/daisy.cpp
+++ b/modules/xfeatures2d/src/daisy.cpp
@@ -1033,6 +1033,7 @@ struct ComputeDescriptorsInvoker : ParallelLoopBody
     {
       x_off = _roi->x;
       x_end = _roi->x + _roi->width;
+      y_off = _roi->y;
       image = _image;
       layers = _layers;
       th_q_no = _th_q_no;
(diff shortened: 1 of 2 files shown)


fallenmi (approved):
AI disclosure: I used OpenAI Codex to assist this review; I verified the result locally against exact head `47e345fb70f4f515195423605ad158654ce2416f`.

Looks good. The ROI-sized descriptor matrix now uses ROI-relative row indices while descriptor sampling remains in absolute image coordinates. I built the exact head against current OpenCV 5.x core and the new ROI/full-image equivalence regression passes.

As a deterministic control, I rolled back only the three production hunks while retaining the submitted test. The same test then fails across the ROI (maximum descriptor differences around `0.06` to `0.11`); restoring the exact-head implementation makes it pass again. `git diff --check` is clean.

The retained upstream reds do not implicate this change: the Windows job exits before checkout because its support script is missing, while Ubuntu 26.04 fails unrelated GStreamer/G-API and BRISQUE tests. The other ten reported checks are green.

PR #4182, opened 2026-07-24
xfeatures2d: fix matchLOGOS crash with fewer than N neighbors

### LINKED ISSUE #4174: xfeatures2d: matchLOGOS crashes with out-of-bounds iterator access when fewer than N candidate points exist

- OpenCV => 5.x (opencv_contrib)
- Operating System / Platform => any
- Compiler => any

##### Detailed description

`Point::nearestNeighboursNaive` in `modules/xfeatures2d/src/logos/Point.cpp` loops with `count < N` only, without checking whether the `minMatch` iterator has reached `minMatch.end()`:

```cpp
int count = 0;
for (std::vector<MatchPoint>::const_iterator mmit = minMatch.begin(); count < N; ++mmit, count++)
{
    nnVector[static_cast<size_t>(count)] = vP[static_cast<size_t>(mmit->index)];
}
```

`N` defaults to 5 (`LogosParams::NUM1`/`NUM2`). `minMatch` only contains the *other* points (the point itself is excluded), so whenever fewer than `N` other points are available, `mmit` advances past `minMatch.end()` and dereferencing it is undefined behavior; in practice this crashes.

Note that simply stopping the loop early is not enough on its own: `nnVector` is pre-sized to `N` via `nnVector.resize(N)`, so any untouched trailing slots stay as null pointers, and `PointPair::computeLocalSupport()` later iterates the full `nnVector` and dereferences every entry (`(*nnIterator)->getLabel()`), so a fix needs to also shrink `nnVector` to the number of neighbours actually found, not just guard the loop.

##### Steps to reproduce

Call `cv::xfeatures2d::matchLOGOS` with fewer than 5 keypoints sharing the same visual-word label:

```cpp
std::vector<cv::KeyPoint> keypoints1, keypoints2;
for (int i = 0; i < 3; i++)
{
    keypoints1.emplace_back(cv::KeyPoint(10.f * i, 10.f * i, 1));
    keypoints2.emplace_back(cv::KeyPoint(10.f * i, 10.f * i, 1));
}
std::vector<int> nn1 = {0, 0, 0};
std::vector<int> nn2 = {0, 0, 0};
std::vector<cv::DMatch> matches;
cv::xfeatures2d::matchLOGOS(keypoints1, keypoints2, nn1, nn2, matches); // crashes
```

JaimeFine:
# Fixes #4174

- Fix out-of-bounds access in `Point::nearestNeighboursNaive()` when fewer than `N` neighbors are available.
- Resize `nnVector` to the actual neighbor count to avoid invalid trailing neighbor entries.
- Add a regression test for the 3-keypoint same-label `matchLOGOS` reproducer.

## Tested:
- Built `opencv_test_xfeatures2d` on Windows with OpenCV 5.x + opencv_contrib 5.x
- Ran LOGOS-related tests with `OPENCV_TEST_DATA_PATH` set to `C:\dev\opencv_extra\testdata`

> **NOTE:** No new test data or performance test is needed for this bug fix; the regression test uses synthetic keypoints.

### DIFF
--- a/modules/xfeatures2d/src/logos/Point.cpp
+++ b/modules/xfeatures2d/src/logos/Point.cpp
@@ -75,11 +75,12 @@ void Point::nearestNeighboursNaive(const std::vector<Point*>& vP, int index, int
     }
 
     std::sort(minMatch.begin(), minMatch.end(), cMP);
-    nnVector.resize(static_cast<size_t>(N));
-    int count = 0;
-    for (std::vector<MatchPoint>::const_iterator mmit = minMatch.begin(); count < N; ++mmit, count++)
+    // A safer handle:
+    const size_t nnCount = std::min(static_cast<size_t>(N), minMatch.size());
+    nnVector.resize(nnCount);
+    for (size_t count = 0; count < nnCount; count++)
     {
-        nnVector[static_cast<size_t>(count)] = vP[static_cast<size_t>(mmit->index)];
+        nnVector[count] = vP[static_cast<size_t>(minMatch[count].index)];
     }
 
     nnFound = true;
--- a/modules/xfeatures2d/test/test_logos_matcher.cpp
+++ b/modules/xfeatures2d/test/test_logos_matcher.cpp
@@ -149,4 +149,24 @@ TEST(XFeatures2d_LogosMatcher, logos_matcher_regression)
             << " ; correctMatches: " << correctMatches;
 }
 
+TEST(XFeatures2d_LogosMatcher, regression_4174_small_label_group) {
+    std::vector<cv::KeyPoint> keypoints1, keypoints2;
+    for (int i = 0; i < 3; i++) {
+        keypoints1.emplace_back(cv::KeyPoint(10.f * i, 10.f * i, 1));
+        keypoints2.emplace_back(cv::KeyPoint(10.f * i, 10.f * i, 1));
+    }
+    std::vector<int> nn1 = {0, 0, 0};
+    std::vector<int> nn2 = {0, 0, 0};
+    std::vector<cv::DMatch> matches;
+
+    cv::xfeatures2d::matchLOGOS(keypoints1, keypoints2, nn1, nn2, matches);
+
+    for (size_t i = 0; i < matches.size(); i++) {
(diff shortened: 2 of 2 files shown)


JaimeFine:
The failing GitHub Actions run appears to be unrelated to this patch.

The workflow parser fails before running tests because
`.github/workflows/PR-5.x.yaml` references a reusable workflow with `@geometry_module`,
which does not appear to be a valid branch/tag/commit in `opencv/ci-gha-workflow`?

Local validation for this patch completed successfully.

JaimeFine:
Hi OpenCV maintainers, gentle follow-up on this small `xfeatures2d` bug fix.

This PR fixes #4174 by avoiding an out-of-bounds access in `matchLOGOS` when fewer than `N` candidate neighbors are available, and adds a synthetic regression test for the small-neighbor case.

I noticed the GitHub Actions workflow did not reach the test stage because of what looks like an unrelated reusable-workflow reference issue. I validated locally on Windows with OpenCV 5.x + opencv_contrib 5.x by building `opencv_test_xfeatures2d` and running the LOGOS-related tests.

Please let me know if there is anything I should adjust to make the PR easier to review. Also happy to retarget or rebase if `5.x` is not the preferred branch for this fix.

Thanks!

fallenmi (approved):
AI disclosure: I used OpenAI Codex to assist this review; I verified the results locally against exact head `4273f41ddd6e216cccbe6d5eb8dec15034e70622`.

The bounds fix is correct. An independent three-point / `N=150` oracle compiled with ASan/UBSan aborts on exact base `57fc85479e94411ae7c004fe1aec8beca8682ec5` with a container-overflow in `Point::nearestNeighboursNaive()`. The same oracle on this head exits cleanly and returns exactly the two available neighbours in distance order.

I also built `opencv_test_xfeatures2d` from the exact PR head against current OpenCV 5.x core `390c4fdcb9`; the new `regression_4174_small_label_group` test passes. The older data-backed LOGOS test could not run locally because `opencv_extra` test data was not present, but the new synthetic regression and the independent sanitizer oracle require no external data. The patch applies cleanly to current opencv_contrib 5.x, and the two touched files have not changed since the PR base.

Looks good to me.

PR #4184, opened 2026-07-31
rgbd: restore fetching colors for pointcloud in coloredkinfu

PeterBowman:
Cherry-picked from 04cafeb2447ce86cbc9bd291f59aefdbeafc032e (authored by @paroj). As pointed out by @alalek in https://github.com/opencv/opencv_contrib/pull/3121#issuecomment-1003111648, this 4.x patch never reached 5.x because of merge conflicts.

### DIFF
--- a/modules/rgbd/include/opencv2/rgbd/colored_kinfu.hpp
+++ b/modules/rgbd/include/opencv2/rgbd/colored_kinfu.hpp
@@ -225,15 +225,16 @@ public:
 
     CV_WRAP virtual void render(OutputArray image, const Matx44f& cameraPose) const = 0;
 
-    /** @brief Gets points and normals of current 3d mesh
+    /** @brief Gets points, normals and colors of current 3d mesh
 
       The order of normals corresponds to order of points.
       The order of points is undefined.
 
         @param points vector of points which are 4-float vectors
         @param normals vector of normals which are 4-float vectors
+        @param colors vector of colors which are 4-float vectors
      */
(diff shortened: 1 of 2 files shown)


fallenmi (changes requested):
This replaces a released public pure-virtual method rather than extending it, which breaks source compatibility for external `ColoredKinFu` implementations. OpenCV 5.0.0 and this PR's exact base expose `getCloud(OutputArray, OutputArray) const`. A minimal derived class that implements that method with `override` compiles against both headers; against exact head `6154ffca6036bdf16b5fd06b5049ea3c50fbe507` it fails because the only virtual now has three parameters. The default argument preserves ordinary two-argument calls, but it cannot preserve an override's signature.

Please keep the released two-output virtual intact and add the color-producing API additively (an overload or a separately named method), then add a compile-time compatibility probe for the existing derived-class signature. A controlled old-header/new-library vtable probe also shows the practical mismatch: the exact-base implementation ignores the unused fourth argument, while this head treats it as the new `colors` output—filling a valid sentinel with 4,389 entries and segfaulting when that stale argument is invalid.

The intended feature itself works in an exact-head build. An external synthetic-plane oracle returns 4,389 aligned points, normals, and `CV_32FC4` colors; the legacy two-argument call still returns the same point/normal counts. The complete enabled RGB-D test binary passes 7/7 tests, `git diff --check` is clean, and the patch merges cleanly with current `5.x`. GitHub exposes no checks for this head.

Reviewed with OpenAI Codex assistance; I independently built the exact revisions and verified the public API, source-compatibility, mixed-vtable, runtime color-output, and test results.

PR #4187, opened 2026-08-09
rgbd: avoid unaligned SIMD stores in LineMOD

Nikhi00718:
## Summary

- avoid aligned SIMD destination accesses when LineMOD response-map rows are not 16-byte aligned
- retain the aligned path only when both source and destination row strides preserve alignment
- add a regression test using a 218-pixel, 3-channel image

Fixes #29559

## Testing

- `git diff --check`
- The native OpenCV C++ test suite could not be built in this Windows environment because no C++ compiler is installed. The added `rgbd` regression test is intended to run in repository CI.

## AI assistance disclosure

I identified and reproduced the issue, then used OpenAI Codex to assist with the implementation and regression-test work. I personally reviewed the resulting diff and ran the validation commands listed above to verify the fix. Any full-suite or local-environment limitations are documented in the validation section.

### DIFF
--- a/modules/rgbd/src/linemod.cpp
+++ b/modules/rgbd/src/linemod.cpp
@@ -923,6 +923,9 @@ static void orUnaligned8u(const uchar * src, const int src_stride,
   volatile bool haveSSE3 = checkHardwareSupport(CPU_SSE3);
 #endif
   bool src_aligned = reinterpret_cast<unsigned long long>(src) % 16 == 0;
+  bool dst_aligned = reinterpret_cast<unsigned long long>(dst) % 16 == 0;
+  src_aligned = src_aligned && src_stride % 16 == 0;
+  dst_aligned = dst_aligned && dst_stride % 16 == 0;
 #endif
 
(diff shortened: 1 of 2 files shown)


Nikhi00718:
I reviewed the SIMD change: the aligned path is now selected only when both source and destination base addresses and row strides preserve 16-byte alignment, while unaligned loads/stores cover the fallback. The added 218-pixel, 3-channel regression targets the reported row-stride case. Native rgbd CI is still required because this host has no C++ build toolchain; please request maintainer review on the current head.

Nikhi00718:
Hi maintainers — this PR is still open and ready for review. No CI check result is currently reported on the PR; I would appreciate a maintainer review when convenient. Thanks!

Nikhi00718:
Hi maintainers — gentle follow-up on the LineMOD unaligned SIMD store fix. The workflow shows no executed jobs on this fork PR, so I’d appreciate CI authorization and a review when convenient. Thanks!

PR #4188, opened 2026-08-11
viz: Add iostream include to precomp.hpp

CSBVision:
With upgrading to OpenCV 5, compiling opencv_viz with MSVC 19.51 results in build errors: `std::cout` and `std::cerr` are undefined, as well as two `std::` prefixes are missing. This minor patch resolves both.


asmorkalov:
Rebased to 5.x to retrigger CI.

asmorkalov (approved):

PR #4190, opened 2026-08-13
face: fix facemark landmarks returned as all zeros on OpenCV 5

### LINKED ISSUE #29703: Face Landmark Detector with LBF fails on 5.0.0 but works on 4.14.0
### System Information

C++

OpenCV Version: 5.0.0  and OpenCV Version: 4.14.0

Operating Systems: MacOS 26.6.1 and Linux Ubuntu 24.04 LTS

GCC 15 on Linux , GCC 17 on MacOS

### Detailed description

Hello
We have been struggling with figuring this out on our end with version 5.0.0 for several days already, hence opening a bug report.

When using `cv::face::FacemarkLBF` to perform face landmark detection, using the pretrained LBF model, we see success with OpenCV versions 4.10.0 (default install on Ubuntu) and a self-compiled version of 4.14.0. But when we build our own version of OpenCV 5.0.0 or even 5.x branch, the face landmark detection for the same code returns all `0`s.

We have attached our C++ file to reproduce this.

[face_landmark_bug_version5.cpp](https://github.com/user-attachments/files/30999598/face_landmark_bug_version5.cpp)

You can download the LBF model from Github:
[lbfmodel.yaml](https://github.com/kurnianggoro/GSOC2017/raw/refs/heads/master/data/lbfmodel.yaml)

You can download the LBP Cascade front face file from Github itself:
[lbpcascade_frontalface_improved.xml](https://github.com/opencv/opencv/blob/4.x/data/lbpcascades/lbpcascade_frontalface_improved.xml)

You can download `lena.jpg` or any front facing image and try.

Please advise on what can be done here to allow usage of the OpenCV 5.x branch.

Thanks.

### Steps to reproduce

Compile the C++ code attached with any openCV either 4.14.0 or 5.0.0  either as below, or use CMake

```
$ g++ -o facelandmark.out -std=c++11 face_landmark_bug_version5.cpp $(pkg-config --cflags --libs opencv4)

```

Run the test as below:
```
$ ./facelandmark.out --input=lena.jpg --face_cascade=lbpcascade_frontalface_improved.xml --lbf_model=lbfmodel.yaml
```

Test Data is Lena image
Image 

### Screen Output with OpenCV 5.0.0
```
Faces: [125 x 125 from (212, 214)]
Number of landmarks: 68
landmarks: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```

Image

### Screen Output with OpenCV 4.14.0

```
Faces: [125 x 125 from (212, 214)]
Number of landmarks: 68
 landmarks: [189.85692, 244.99849;
 189.573, 265.23071;
 190.28471, 285.68533;
 192.16396, 304.77322;
 199.92172, 321.61148;
 213.44188, 334.55539;
 229.90805, 344.25223;
 247.21469, 350.08112;
 261.91473, 351.49884;
 273.12766, 348.0722;
 280.45651, 337.73193;
 287.40784, 326.68713;
 294.07806, 314.58868;
 300.16626, 301.78378;
 305.61111, 289.78311;
 310.00641, 277.69308;
 311.77563, 264.89987;
 221.51239, 235.25133;
 233.99059, 229.8634;
 248.06059, 229.12758;
 261.37134, 232.98105;
 273.08405, 239.95854;
 293.03967, 243.79985;
 299.82352, 241.5486;
 307.23904, 241.1391;
 313.29672, 242.40109;
 315.51544, 247.94843;
 282.38507, 253.79198;
 283.29684, 264.84595;
 284.64902, 275.81897;
 285.97562, 287.20401;
 265.06906, 292.7955;
 272.082, 295.05756;
 278.93143, 297.60449;
 284.56372, 296.2334;
 288.87366, 293.44086;
 235.69563, 249.32544;
 245.03613, 248.25198;
 253.23431, 249.43468;
 259.18671, 254.55368;
 251.79207, 255.42607;
 243.47229, 253.99423;
 289.37369, 258.71997;
 296.58646, 255.80627;
 302.93579, 256.15747;
 307.00156, 259.59174;
 302.6824, 261.957;
 296.25293, 261.17407;
 244.21379, 313.19656;
 258.38275, 311.51071;
 269.67856, 309.72714;
 275.0871, 311.89273;
 279.67282, 310.50317;
 283.18167, 312.59058;
 284.49457, 315.88657;
 280.36493, 323.34406;
 276.34192, 326.49203;
 271.16849, 327.42108;
 265.16592, 326.75275;
 255.27713, 323.33264;
 247.66557, 313.82654;
 268.45746, 314.21292;
 274.15723, 315.18939;
 278.80133, 314.71707;
 281.80692, 316.51697;
 277.89505, 318.18228;
 272.99765, 319.35822;
 267.27936, 318.86191]
```

Image

- [x] There is reproducer code and related data files (videos, images, onnx, etc)

aarochu:
Fixes https://github.com/opencv/opencv/issues/29703

## Problem

`Facemark::fit()` returns the correct number of landmarks with every point left at `(0,0)` on OpenCV 5, while the same code works on 4.x. The reporter hit this with `FacemarkLBF` and the pretrained `lbfmodel.yaml`.

It only happens when the output is a `std::vector<std::vector<Point2f>>` — which is what the documentation, the samples and the reporter's code all pass — so `landmarks[i].size()` looks right (68) and every coordinate is zero.

## Root cause

`_copyVector2Output()` writes each result through a `Mat` header obtained from `OutputArray::getMat(i)`. That header only *aliases* the destination vector's storage, so the source has to match its shape already: `copyTo()` calls `create()` first, and when the shape differs the header is reallocated, so the copy lands in a fresh buffer that is discarded instead of in the caller's vector.

The source was built as `Mat(Mat(vec[i]).t())` to match the `1 x N` header OpenCV 4 hands out for a `vector<vector<T>>`:

```cpp
// modules/core/src/matrix_wrap.cpp, _InputArray::getMat_()
// 4.x
return v->empty() ? Mat() : Mat(1, int(v->size()), t, v->data());
// 5.x
int v_sz = int(v->size());
return v->empty() ? Mat() : Mat(1, &v_sz, t, v->data());
```

OpenCV 5 returns a **1-D** header there instead, so the transposed source no longer matches and every landmark is silently dropped.

## Fix

Build the source as a `1 x N` two-channel row with `reshape()`, which matches the destination header on both 4.x and 5.x, and size the destination from the actual landmark count instead of a hard-coded `68` (so models with a different number of landmarks work too).

The helper is duplicated in the LBF, AAM and Kazemi implementations and all three were affected, so all three are fixed.

The `Mat`/`UMat` output branches are left alone: they bind the real destination object rather than an aliasing header, so a reallocating `copyTo()` still delivers the data there. Changing them would alter their output shape without fixing a bug.

## Testing

Verified on a local Windows/MSVC build of 5.x + contrib (`5.1.0-dev`), using the reporter's exact sequence (cascade → `detectMultiScale` → `FacemarkLBF::create`/`loadModel`/`fit`) with the pretrained `lbfmodel.yaml` from the issue:

Before:
```
Faces: [136 x 136 from (235, 237)]
Number of landmarks: 68
Non-zero landmarks: 0 / 68
First 5 landmarks: [0, 0] [0, 0] [0, 0] [0, 0] [0, 0]
```

After:
```
Faces: [136 x 136 from (235, 237)]
Number of landmarks: 68
Non-zero landmarks: 68 / 68
First 5 landmarks: [213.177, 280.037] [213.544, 301.395] [214.891, 322.97] [218.014, 343.195] [227.687, 361.095]
```

I also reproduced the underlying shape mismatch in isolation against `opencv_core` 5.x and confirmed the replacement copies correctly for 5, 68 and 194 landmarks as well as the empty case.

The existing `CV_Face_FacemarkLBF.test_workflow` and `CV_Face_FacemarkAAM.test_workflow` only asserted the landmark *count*, which stayed correct throughout — that is why this went unnoticed. They now also assert the points actually reached the caller. Confirmed this is a real regression test: with the source fix reverted and only the test change applied, it fails with

```
  nonzero
    Which is: 0
  facial_points[0].size()
    Which is: 68
```

Full `opencv_test_face` suite passes with the fix (19/19, with `OPENCV_TEST_DATA_PATH` plus the `face_landmark_model.dat` the module downloads at configure time).

### DIFF
--- a/modules/face/src/facemarkAAM.cpp
+++ b/modules/face/src/facemarkAAM.cpp
@@ -348,7 +348,6 @@ static void _copyVector2Output(std::vector< std::vector< Point2f > > &vec, Outpu
         for (unsigned int i = 0; i < vec.size(); i++) {
             out.create((int)vec[i].size(), 1, CV_32FC2, i);
             Mat m = out.getMat(i);
-            // Source must match m's shape or copyTo() reallocates m instead of writing through it.
             Mat(vec[i]).reshape(2, 1).copyTo(m);
         }
     }
--- a/modules/face/src/facemarkAAM.cpp
+++ b/modules/face/src/facemarkAAM.cpp
@@ -346,9 +346,16 @@ static void _copyVector2Output(std::vector< std::vector< Point2f > > &vec, Outpu
     }
     else if (out.kind() == _OutputArray::STD_VECTOR_VECTOR) {
         for (unsigned int i = 0; i < vec.size(); i++) {
-            out.create(68, 1, CV_32FC2, i);
+            out.create((int)vec[i].size(), 1, CV_32FC2, i);
             Mat m = out.getMat(i);
-            Mat(Mat(vec[i]).t()).copyTo(m);
+            // `m` is only a header over the destination vector's storage, so the
+            // source has to match its shape already: a copyTo() that needs to
+            // resize reallocates `m` instead of writing through it, and the
+            // landmarks are silently dropped. OpenCV 4 handed out a 1 x N header
+            // here while OpenCV 5 hands out a 1-D one, which is why transposing
+            // to a column left every landmark at (0,0) on 5.x -- see
+            // https://github.com/opencv/opencv/issues/29703
+            Mat(vec[i]).reshape(2, 1).copyTo(m);
         }
     }
     else {
--- a/modules/face/src/facemarkAAM.cpp
+++ b/modules/face/src/facemarkAAM.cpp
@@ -348,13 +348,7 @@ static void _copyVector2Output(std::vector< std::vector< Point2f > > &vec, Outpu
         for (unsigned int i = 0; i < vec.size(); i++) {
             out.create((int)vec[i].size(), 1, CV_32FC2, i);
             Mat m = out.getMat(i);
-            // `m` is only a header over the destination vector's storage, so the
-            // source has to match its shape already: a copyTo() that needs to
-            // resize reallocates `m` instead of writing through it, and the
-            // landmarks are silently dropped. OpenCV 4 handed out a 1 x N header
-            // here while OpenCV 5 hands out a 1-D one, which is why transposing
-            // to a column left every landmark at (0,0) on 5.x -- see
-            // https://github.com/opencv/opencv/issues/29703
+            // Source must match m's shape or copyTo() reallocates m instead of writing through it.
             Mat(vec[i]).reshape(2, 1).copyTo(m);
         }
     }
--- a/modules/face/src/facemarkLBF.cpp
+++ b/modules/face/src/facemarkLBF.cpp
@@ -395,7 +395,6 @@ static void _copyVector2Output(std::vector< std::vector< Point2f > > &vec, Outpu
         for (unsigned int i = 0; i < vec.size(); i++) {
             out.create((int)vec[i].size(), 1, CV_32FC2, i);
             Mat m = out.getMat(i);
-            // Source must match m's shape or copyTo() reallocates m instead of writing through it.
             Mat(vec[i]).reshape(2, 1).copyTo(m);
         }
     }
--- a/modules/face/src/facemarkLBF.cpp
+++ b/modules/face/src/facemarkLBF.cpp
@@ -395,13 +395,7 @@ static void _copyVector2Output(std::vector< std::vector< Point2f > > &vec, Outpu
         for (unsigned int i = 0; i < vec.size(); i++) {
             out.create((int)vec[i].size(), 1, CV_32FC2, i);
             Mat m = out.getMat(i);
-            // `m` is only a header over the destination vector's storage, so the
-            // source has to match its shape already: a copyTo() that needs to
-            // resize reallocates `m` instead of writing through it, and the
-            // landmarks are silently dropped. OpenCV 4 handed out a 1 x N header
-            // here while OpenCV 5 hands out a 1-D one, which is why transposing
-            // to a column left every landmark at (0,0) on 5.x -- see
-            // https://github.com/opencv/opencv/issues/29703
+            // Source must match m's shape or copyTo() reallocates m instead of writing through it.
             Mat(vec[i]).reshape(2, 1).copyTo(m);
         }
     }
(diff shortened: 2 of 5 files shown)


asmorkalov:
Please drop irrelevant comments from the code.

aarochu:
Comments should be gone!

fallenmi (approved):
Reviewed with OpenAI Codex assistance; I independently built and tested the exact revision and verified the results below.

The OpenCV 5 output conversion is fixed. Keeping the new regression assertions while rolling back only the three production conversion hunks made both `FacemarkAAM.test_workflow` and `FacemarkLBF.test_workflow` fail deterministically: each returned 68 landmarks and all 68 were `(0, 0)`. Restoring exact head `05f89d83f99373869fd9f4200443a2b2b7cc5c03` made both tests pass with every returned landmark nonzero.

I built the face module against current OpenCV 5.x core and ran the complete face test binary with the matching `opencv_extra` data: 19/19 tests passed. The dynamic `vec[i].size()` allocation also removes the hard-coded 68-point assumption while the two-channel reshape preserves the actual `Point2f` layout. `git diff --check` is clean.

The published Actions matrix is still red (3 success, 9 failure), so those jobs should be resolved or rerun before merge; my approval is based on the exact local base/head regression and full face-module suite above.

stealthylabs:
Hi I am the reporter of the original bug, and was wondering what is needed  to move this forward, or how can we help further. 

Thank you @aarochu

PR #4192, opened 2026-08-13
fix(rgbd): prevent SIGSEGV in linemod::orUnaligned8u from stale alignment fast path

### LINKED ISSUE #29559: [Bug] linemod::match crashes with SIGSEGV due to misaligned store in orUnaligned8u
### System Information

- **OpenCV version**: 5.0.0 (contrib)
- **Operating System**: Arch Linux
- **Compiler**: GCC 14
- **CPU**: Intel Core i7-9750H(SSE2 / SSE3 supported)

### Detailed description

When calling `cv::linemod::Detector::match()`, the process crashes with a segmentation fault deep inside the SIMD-optimized helper function `orUnaligned8u`. The crash occurs because the function uses an **aligned store instruction** (`_mm_or_si128` with dereferenced `__m128i*`) on a destination pointer that is **not guaranteed to be 16-byte aligned**.

### Steps to reproduce

1. Create a `linemod` detector, add at least one template using a reference image and a binary mask.
2. Prepare a test image whose width (in bytes) is **not a multiple of 16** (e.g., width = 218 for a 3-channel image, giving a stride of 654).
3. Call `detector->match(sources, threshold, matches)`.

SomSamantray:
## Problem

`cv::linemod::Detector::match()` could crash the process with SIGSEGV whenever the input image's width (or its size after one `pyrDown()` pyramid level) wasn't a multiple of 16 — a size restriction nowhere documented or enforced, so ordinary images could trigger it unpredictably.

The crash was inside `orUnaligned8u()` (`modules/rgbd/src/linemod.cpp`), which computes whether its `src` pointer is 16-byte aligned **once, before its row loop**, and reuses that single decision for every row. But `src`/`dst` advance by `src_stride`/`dst_stride` each row, and those strides aren't guaranteed to be multiples of 16 (`spread()`'s destination stride is exactly the image's row width in bytes). Once a stride isn't 16-aligned, the pointer drifts in and out of alignment from row to row — so the branch decision made at row 0 goes stale, and a later row can hit a plain `__m128i*` dereference (which the compiler lowers to an aligned `movdqa`) on a pointer that is no longer actually aligned.

## Fix

Removes the once-per-call alignment fast path entirely and always accesses `src`/`dst` with unaligned SSE intrinsics (`_mm_loadu_si128`/`_mm_lddqu_si128`/`_mm_storeu_si128`). The remaining SSE3-vs-SSE2 branch reflects genuine hardware capability, not a (now provably unsound) alignment assumption. Output is unchanged for every previously-working input; only the crash is eliminated.

Fixes opencv/opencv#29559
Related: opencv/opencv#29576 (same crash signature via Python bindings; very likely the same root cause, not independently verified here)

## Validation

Since building the full `opencv` + `opencv_contrib` dependency chain just to run one test wasn't practical for this fix's turnaround, the fix was verified with a standalone harness that reproduces `orUnaligned8u()`'s exact pre-fix and post-fix logic outside the build (same compiler/flags OpenCV uses for its SSE2/SSE3 code paths):

- Pre-fix logic **crashes** on a destination row stride that isn't a multiple of 16 (the real production trigger — matched to how `spread()` actually calls this function), on an initially-offset pointer, and on mixed unaligned `src`/`dst`.
- Post-fix logic **completes correctly** (bit-identical to an independent scalar reference) on all of the above, across both the SSE3 and SSE2-fallback code paths.

A regression test (`modules/rgbd/test/test_linemod.cpp`) is included, exercising the fix through the public `Detector::addTemplate()`/`match()` API with a synthetic image sized so its row width isn't 16-byte-aligned at every pyramid level — reproducing the original crash through the real call path, then asserting an actual correct match (not just crash-freedom). It has not been compiled locally in this environment; it follows the module's existing test conventions and its logic mirrors the already-verified standalone harness.

One known coverage gap: the SSE2-only (no-SSE3) fallback branch this fix also touches isn't independently exercised by the new test on typical SSE3-capable CI hardware, since `checkHardwareSupport(CPU_SSE3)` is true almost everywhere the test would run. The SSE3 and SSE2 branches are identical in shape (only the load intrinsic differs), so risk is low, but flagging for visibility.


asmorkalov:
The PR contains not relevant files and LLM prompt.

PR #4196, opened 2026-08-22
ccalib: fix misleading assertion when no calibration views survive

### LINKED ISSUE #28462: cv2.omnidir.calibrate Assertion Failure
### System Information

- OpenCV: 4.13.0 (opencv-python + opencv-contrib-python)
- OS: Windows 11
- Python: 3.10
- CPU-only build

### Detailed description

cv2.omnidir.calibrate consistently fails with the following assertion:
```
OpenCV(4.13.0) ... omnidir.cpp:854: error: (-215:Assertion failed)
!objectPoints.empty() && objectPoints.type() == CV_64FC3
in function 'cv::omnidir::internal::computeJacobian'
```
This happens even when:
- object points are valid NumPy arrays of shape (N, 1, 3)
- image points are valid NumPy arrays of shape (N, 1, 2)
- all arrays are float32 or float64
- the lists contain multiple valid entries (e.g., 15 images × 48 points)
- flags = 0 (no CALIB_USE_GUESS)
- K, xi, D are provided as valid initial matrices
- no arrays are empty
The error occurs before optimization begins, inside the Jacobian computation.

### Steps to reproduce

```
import cv2
import numpy as np

# Example: 15 images, each with 48 detected corners
all_obj_pts = [np.random.rand(48, 1, 3).astype(np.float32) for _ in range(15)]
all_img_pts = [np.random.rand(48, 1, 2).astype(np.float32) for _ in range(15)]

image_size = (1920, 1080)

K_init = np.eye(3, dtype=np.float32)
xi_init = np.array([1.0], dtype=np.float32)
D_init = np.zeros((1, 4), dtype=np.float32)

criteria = (cv2.TERM_CRITERIA_EPS + cv2.TERM_CRITERIA_MAX_ITER, 100, 1e-6)

ret, K, xi, D, rvecs, tvecs, idx = cv2.omnidir.calibrate(
    all_obj_pts,
    all_img_pts,
    image_size,
    K_init,
    xi_init,
    D_init,
    0,  # no CALIB_USE_GUESS
    criteria
)
```

### Expected Behavior

cv2.omnidir.calibrate should:
- accept a Python list of (N,1,3) object point arrays
- accept a Python list of (N,1,2) image point arrays
- accept float32 or float64
- initialize K, xi, D internally when flags = 0
- proceed to optimization without assertion failure

### Actual Behavior
The function fails immediately with:
!objectPoints.empty() && objectPoints.type() == CV_64FC3

```
!objectPoints.empty() && objectPoints.type() == CV_64FC3
```
even though:
- objectPoints is not empty
- the type is correct
- the shape is correct
- the list contains valid entries

This suggests an internal inconsistency in the omnidir module, likely in:
```
cv::omnidir::internal::computeJacobian
```
where the type check does not match the actual type produced earlier in the pipeline.
### 
Notes
- The same data works correctly with cv2.calibrateCamera and cv2.fisheye.calibrate.
- The failure occurs regardless of float32/float64.
- The failure occurs regardless of whether K/xi/D are None or valid matrices.
- The failure occurs with real checkerboard data and synthetic data.
- I've successfully calibrated with Scaramuzza's OCamCalib (v3.0 November 2013)

- [x] There is reproducer code and related data files (videos, images, onnx, etc)

SomSamantray:
`cv2.omnidir.calibrate` failed with a confusing assertion that looked like a type bug even when the input arrays had the correct type, shape, and were non-empty:

```
CV_Assert(!objectPoints.empty() && objectPoints.type() == CV_64FC3)
```

The actual cause has nothing to do with types. `omnidir::calibrate()` filters out any calibration view whose initial-pose reprojection error exceeds an internal threshold; if every view fails that filter (e.g. the object/image point correspondences don't come from a real calibration target), the point vectors end up empty and the code still proceeds into `computeJacobian()`, tripping the assertion above on its emptiness half only — the type half was never actually false.

This PR replaces that misleading assertion with two explicit, actionable errors:

- `omnidir::calibrate()` now reports "no calibration views survived initial pose estimation" when every view is rejected.
- `omnidir::stereoCalibrate()` has an independent failure mode not covered by the first fix: two per-camera views can each individually survive filtering, yet their surviving-view sets can fail to intersect. That path used to crash even earlier, via `Mat::copyTo()` on a fixed-size `OutputArray::getMat()` temporary that refuses to "reallocate" an empty array even to the same size. Both issues are fixed together — the intersection-empty case is now a clear "no calibration views are valid for both cameras" error.

`modules/ccalib` had no test suite at all before this change; this PR adds its first one, covering the error path for both functions plus a synthetic-projection-based happy-path regression test proving valid calibration is unaffected.

Fixes opencv/opencv#28462

### Validation

Built and ran the new test suite locally (`opencv_test_ccalib`, scoped build of `core, imgproc, geometry, calib, objdetect, features, xfeatures2d, highgui, imgcodecs, ccalib, ts`): all 4 tests pass, including the happy-path tests that confirm valid single-camera and stereo calibration are unaffected.

### DIFF
--- a/modules/ccalib/src/omnidir.cpp
+++ b/modules/ccalib/src/omnidir.cpp
@@ -766,9 +766,20 @@ void cv::omnidir::internal::initializeStereoCalibration(InputArrayOfArrays objec
     Mat interIdx1, interIdx2, interOri;
 
     getInterset(idx1, idx2, interIdx1, interIdx2, interOri);
+
+    if (interOri.empty())
+    {
+        CV_Error(cv::Error::StsBadArg,
+            "omnidir::stereoCalibrate: no calibration views are valid for both cameras. "
+            "Each camera individually may have views that survived initial pose estimation, "
+            "but the sets of valid views for camera 1 and camera 2 do not overlap (their "
+            "intersection is empty). Check that the same views are usable for both cameras, "
+            "or verify the object/image point correspondences for every input view.");
+    }
+
     if (idx.empty())
         idx.create(1, (int)interOri.total(), CV_32S);
-    interOri.copyTo(idx.getMat());
+    interOri.copyTo(idx);
 
     int n_inter = (int)interIdx1.total();
 
@@ -1112,6 +1123,18 @@ double cv::omnidir::calibrate(InputArrayOfArrays patternPoints, InputArrayOfArra
         _imagePoints.push_back(_imagePointsTmp[_idx.at<int>(i)]);
     }
 
+    if (_patternPoints.empty())
+    {
+        CV_Error(cv::Error::StsBadArg,
+            "omnidir::calibrate: no calibration views survived initial pose estimation. "
+            "Every input view was rejected because its estimated initial pose reprojection "
+            "error exceeded the internal threshold. This usually means the object points and "
+            "image points for every view do not correspond to a consistent, real calibration "
+            "target (e.g. mismatched or randomly generated correspondences), or the image size "
+            "passed to calibrate() does not match the points. Check the object/image point "
+            "correspondences for every input view.");
+    }
+
     int n = (int)_patternPoints.size();
     Mat finalParam(1, 10 + 6*n, CV_64F);
     Mat currentParam(1, 10 + 6*n, CV_64F);
@@ -1252,7 +1275,7 @@ double cv::omnidir::stereoCalibrate(InputOutputArrayOfArrays objectPoints, Input
     if(idx.needed())
     {
         idx.create(1, (int)_idx.total(), CV_32S);
-        _idx.copyTo(idx.getMat());
+        _idx.copyTo(idx);
     }
 	for (int i = 0; i < (int)_idx.total(); ++i)
 	{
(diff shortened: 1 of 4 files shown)


fallenmi (approved):
AI disclosure: I used OpenAI Codex to assist this review; I verified the results locally against exact head `13e3f598ec8c740af32bf2b58955e15cc8781f70`.

The new guards and `OutputArray` copies fix the reported failure modes without disturbing valid calibration. Using the same deterministic invalid-view input against exact base `17af220dd982eb56d647da56032aaebf2caf374f` reaches the misleading `CV_64FC3` assertion in `computeJacobian()`. The exact head instead raises `StsBadArg` with the intended `no calibration views survived` explanation.

I built `opencv_test_ccalib` from this exact contrib head against current OpenCV 5.x core `390c4fdcb9`. All four new tests pass: the single-camera and stereo empty-survivor errors, plus both valid synthetic calibration controls. The patch also applies cleanly to current opencv_contrib 5.x; the touched production path has not changed since the PR base.

Looks good to me.

PR #4201, opened 2026-08-31
Fix out-of-bounds access in EdgeAwareInterpolator when k exceeds match count

### LINKED ISSUE #4195: optflow: calcOpticalFlowSparseToDense segfaults when k exceeds the number of sparse matches
### System information

- OpenCV: 4.10.0 with opencv_contrib, built from source, static
- Platform: Ubuntu 22.04, x86_64
- Compiler: GCC 11.4, TBB parallel backend
- Also seen on Windows 10 x64 with an MSVC build of the same version

The same code is on the `4.x` and `5.x` branches of opencv_contrib today, so
this is not specific to 4.10.

### Detailed description

`cv::optflow::calcOpticalFlowSparseToDense` reads a null pointer and the process
dies with SIGSEGV when `k` is large relative to the number of sparse matches the
image can supply. The call below uses the documented default parameters on a
legal input.

The function samples the image on a grid of `grid_step` pixels. It tracks those
points with `calcOpticalFlowPyrLK` and keeps the ones that survive. The survivors
go to `EdgeAwareInterpolator`, configured with `setK(k)`.

A 64x48 image at the default `grid_step` of 8 yields 48 samples, and the tracker
drops some of them. The default `k` is 128. So the interpolator is asked to fit a
model over the 128 nearest matches when fewer than 48 exist.

In `modules/optflow/src/sparsetodenseflow.cpp` the only check on `k` is:

```cpp
CV_Assert( grid_step>1 && k>3 && sigma>0.0001f && fgs_lambda>1.0f && fgs_sigma>0.01f );
```

Nothing compares `k` with `points.size()` or `points_filtered.size()` before
`gd->setK(k)`.

The crash is not deterministic. The routine parallelizes internally, so the same
input dies on some runs and completes on others. That is why a small image with
the defaults can look fine in a quick test.

### Steps to reproduce

```cpp
#include <opencv2/core.hpp>
#include <opencv2/optflow.hpp>

#include <cstdio>
#include <cstdlib>

{
    const int seed = argc > 1 ? std::atoi(argv[1]) : 1;
    cv::setRNGSeed(seed);

    cv::Mat from(48, 64, CV_8UC1);
    cv::Mat to(48, 64, CV_8UC1);
    cv::Mat flow;
    cv::randu(from, cv::Scalar(0), cv::Scalar(255));
    cv::randu(to,   cv::Scalar(0), cv::Scalar(255));

    // grid_step 8 and k 128 are the documented defaults
    cv::optflow::calcOpticalFlowSparseToDense(from, to, flow, 8, 128);

    std::printf("ok\n");
    return 0;
}
```

Run one seed per process, because a crash ends the process:

```
for s in $(seq 1 300); do ./repro $s || echo "crashed on seed $s"; done
```

Measured on the build above, 300 runs per row:

```
image     k     grid_step   crashes
64x48     128   8           5 / 300
64x48     48    8           0 / 300
64x48     44    8           0 / 300
64x48     32    8           0 / 300
320x240   128   8           0 / 300
```

On Windows with an MSVC build the rate is higher and the threshold is lower.
With one fixed input, `k=128` crashed 23 times in 100. `k=46` crashed 6 times in
60. `k=44` and below stayed clean in 100 runs each.

So the crash needs two things together: a small image, and a `k` close to or
above the number of matches. Either one alone is fine. Lowering `k` or enlarging
the image removes it.

### Suggested fix

Clamp `k` to what the image can supply before configuring the interpolator, or
refuse with a message:

```cpp
if ((int)points_filtered.size() <= k)
    k = std::max(4, (int)points_filtered.size() - 1);
gd->setK(k);
```

A caller who leaves the defaults alone on a small image should get a result or an
error, not a dead process.

Huxingyu:
### Pull Request Checklist

- [x] There are no new and unexpected changes
- [x] All existing and new tests pass
- [x] Code complies with [OpenCV coding style](https://github.com/opencv/opencv/wiki/Coding_Style_Guide)
- [x] This PR is submitted to the correct branch (`4.x`)

### Summary

Fixes #4195: `calcOpticalFlowSparseToDense` (or direct `EdgeAwareInterpolator` use) crashes when the configured `k` exceeds the actual number of sparse matches.

### Root cause

`EdgeAwareInterpolatorImpl::interpolate` allocates `NNlabels` as `match_num x k` and pre-fills it with the `-1` sentinel. The KNN pass only fills as many entries as it can actually expand, and the RANSAC pass reads exactly `k` entries per row. When `k > match_num` (e.g. the default `k=128` with the 48x64 / `grid_step=8` case from the issue, which samples only 48 grid points), the RANSAC pass reads the `-1` sentinels and dereferences `matches[-1]` / `transforms[-1]`, crashing.

### Fix

- Use `num_neighbors = min(k, match_num)` for the current call only; the user's `k` (and `getK()`) stays unchanged.
- Operate all KNN/RANSAC passes on `num_neighbors`, so the `-1` sentinels are never consumed.
- Reject degenerate input with a clear `cv::Exception`: fewer than 3 unique matches, or `k < 3` (the RANSAC affine fit needs 3 points).

### Tests

- `DenseOpticalFlow_SparseToDenseFlow.Regression_4195` (optflow): default 48x64 input no longer crashes; 8x8 input throws `cv::Exception`.
- `InterpolatorTest.RejectsTooFewMatches` (ximgproc): 0/1/2 matches throw `StsBadArg`.
- `InterpolatorTest.KExceedsMatchCount` (ximgproc): `k=128` with 5 matches produces valid flow and leaves `getK()==128` unchanged.

### Validation

Built `opencv_test_optflow` / `opencv_test_ximgproc` locally (ASan build) and ran the new regression tests plus the existing `InterpolatorTest.MultiThreadReproducibility` suite — all pass. The standalone repro from the issue no longer crashes under ASan. Data-dependent reference tests require `opencv_extra` and run in CI.

### DIFF
--- a/modules/optflow/test/test_OF_accuracy.cpp
+++ b/modules/optflow/test/test_OF_accuracy.cpp
@@ -287,6 +287,25 @@ TEST(DenseOpticalFlow_SparseToDenseFlow, ReferenceAccuracy)
     EXPECT_LE(calcRMSE(GT, flow), target_RMSE);
 }
 
+TEST(DenseOpticalFlow_SparseToDenseFlow, Regression_4195)
+{
+    // https://github.com/opencv/opencv_contrib/issues/4195
+    RNG rng(1);
+    Mat from(48, 64, CV_8UC1);
+    rng.fill(from, RNG::UNIFORM, 0, 255);
+    Mat to = from.clone();
+    Mat flow;
+
+    // The default k of 128 exceeds the 48 points sampled with grid_step 8.
+    ASSERT_NO_THROW(calcOpticalFlowSparseToDense(from, to, flow));
+    EXPECT_EQ(from.size(), flow.size());
+    EXPECT_EQ(CV_32FC2, flow.type());
+    EXPECT_TRUE(checkRange(flow));
+
+    Mat small = from(Rect(0, 0, 8, 8)).clone();
+    EXPECT_THROW(calcOpticalFlowSparseToDense(small, small, flow), cv::Exception);
+}
+
 TEST(DenseOpticalFlow_PCAFlow, ReferenceAccuracy)
 {
     Mat frame1, frame2, GT;
--- a/modules/ximgproc/src/sparse_match_interpolators.cpp
+++ b/modules/ximgproc/src/sparse_match_interpolators.cpp
@@ -173,10 +173,11 @@ Ptr<EdgeAwareInterpolatorImpl> EdgeAwareInterpolatorImpl::create()
 void EdgeAwareInterpolatorImpl::interpolate(InputArray from_image, InputArray from_points, InputArray, InputArray to_points, OutputArray dense_flow)
 {
     CV_Assert( !from_image.empty() && (from_image.depth() == CV_8U) && (from_image.channels() == 3 || from_image.channels() == 1) );
-    CV_Assert( !from_points.empty() && !to_points.empty() && from_points.sameSize(to_points) );
+    if (from_points.empty() || to_points.empty())
+        CV_Error(Error::StsBadArg, "EdgeAwareInterpolator requires at least 3 matches");
+    CV_Assert(from_points.sameSize(to_points));
     CV_Assert((from_points.isVector() || from_points.isMat()) && from_points.depth() == CV_32F);
     CV_Assert((to_points.isVector() || to_points.isMat()) && to_points.depth() == CV_32F);
-    CV_Assert(from_points.sameSize(to_points));
 
     w = from_image.cols();
     h = from_image.rows();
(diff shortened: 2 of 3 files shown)


Huxingyu:
Hi @asmorkalov, this PR fixes #4195 (segfault in `calcOpticalFlowSparseToDense` when `k` exceeds the number of sparse matches). It clamps the effective neighbor count to `min(k, match_num)` per call without mutating the user's `k`, and rejects fewer than 3 matches/neighbors with a clear `cv::Exception`, plus regression tests.

This is my first contribution to opencv_contrib, so the CI workflow needs a maintainer to approve it before it will run. Could you approve the workflow when you have a moment? I've already built `opencv_test_optflow`/`opencv_test_ximgproc` locally (ASan) and all new tests plus the existing `MultiThreadReproducibility` suite pass.

Huxingyu:
Hi @asmorkalov,

All CI jobs have now completed. The failures appear to be existing baseline or infrastructure failures rather than regressions from this PR.

Most importantly, both test modules affected by this change pass on all four x86 Linux configurations:

- Ubuntu 20.04: opencv_test_optflow and opencv_test_ximgproc passed
- Ubuntu 22.04: opencv_test_optflow and opencv_test_ximgproc passed
- Ubuntu 24.04: opencv_test_optflow and opencv_test_ximgproc passed
- Ubuntu 26.04: opencv_test_optflow and opencv_test_ximgproc passed

The successful macOS, ARM64, and RISC-V jobs provide additional coverage.

The red checks are unrelated:

- Windows fails before checkout because the self-hosted runner cannot find print_system_information.sh. The same failure is present in an unrelated run: https://github.com/opencv/opencv_contrib/actions/runs/33073351880/job/98521559744
- CUDA fails in CornerHarris and CornerMinEigen with ApertureSize(0), following opencv commit 6e4c669be805: https://github.com/opencv/opencv/commit/6e4c669be805167fbbe2d43e75770e8ffdc3667e
- Ubuntu 20.04 and 22.04 fail in the existing OpenCL DualTVL1 performance sanity test. The same test fails with the same values in unrelated PR runs, including https://github.com/opencv/opencv_contrib/actions/runs/33073351880
- Ubuntu 22.04 and 24.04 time out in opencv_test_shape on the same self-hosted runner, opencv-cn-lin2-x86-64.
- Ubuntu 26.04 fails the unrelated CV_Quality_BRISQUE.multi_channel floating-point tolerance test. The identical failure was already present in an unrelated July run.

Could you please review the PR and confirm whether these unrelated checks can be treated as baseline failures? If a rerun is preferred, it would probably be more useful after the shared CI issues are repaired.

The PR-specific regression tests pass, and the change was also validated locally with ASan.

PR #4208, opened 2026-09-02
ximgproc: fix uninitialised read of ellipse periphery point in ValidateCircles

### LINKED ISSUE #4207: EdgeDrawing::ValidateCircles() reads an uninitialised ellipse point for odd perimeters
- OpenCV => 4.x head (reproduced on 4.14.0)
- Operating System / Platform => Ubuntu (Linux 5.15) 64 Bit
- Compiler => gcc 9.4, valgrind 3.24

##### Detailed description
In `EdgeDrawingImpl::ValidateCircles()` (modules/ximgproc/src/edge_drawing.cpp:3254
on 4.x), `ComputeEllipsePoints(circle->eq.coeff, px, py, noPoints)` is called with
`noPoints = (int)computeEllipsePerimeter(...)`, which is odd about half the time.
`ComputeEllipsePoints` writes exactly `2 * (noPoints/2)` entries
(`px[j-1]` and `px[j-1+npts]` for j = 1..npts; edge_drawing.cpp:4975-4989), so for an
odd `noPoints` the last slot `px/py[noPoints-1]` is **never written**. The validation
loop `for (j = 0; j < noPoints; j++)` (:3269) nevertheless reads it:

- for the **first** candidate the `px/py` buffers are freshly `new`-allocated, so this
  is a read of **uninitialised** memory (an indeterminate value — undefined behaviour);
- for **later** candidates the buffers are reused, so it reads a **stale** value left
  by a previous candidate.

That phantom point is then fed into the NFA validation: it is counted as a periphery
sample (`noPeripheryPixels++`, :3276), and depending on its indeterminate coordinates
is **either** skipped by the bounds check (:3278-3281, when it falls outside the image)
**or** fully processed (when it happens to fall inside, where it can also change
`aligned`). Either way the decision `checkValidationByNFA(noPeripheryPixels, aligned)`
(:3440) is computed from corrupted statistics for every odd-perimeter ellipse, and the
direction of the perturbation is not predictable.

Impact: reading the unwritten slot is undefined behaviour and is flagged by memory
sanitizers. The behavioural effect is usually small: in our testing over 93 OpenCV
sample images plus a synthetic case, applying the fix changed the number of detected
ellipses on 6 of them (in both directions) and left the other 87 unchanged. Where it
does change, the post-fix computation uses only the real periphery points (the
never-written slot is no longer read); we did not assess which output is closer to
ground truth. (Results were deterministic across repeated runs on this build.)

The reference implementation guards the call with `if (noPoints % 2) noPoints--;`
(CihanTopal/ED_Lib). The OpenCV port adopted the `points_buffer_size` bound from that
same fix but missed the odd-count guard.

##### Steps to reproduce (stock OpenCV, no source changes)
Run `detectEllipses()` on an image with several ellipses (some will have an odd
computed perimeter) under valgrind:

```.cpp
#include <opencv2/imgproc.hpp>
#include <opencv2/ximgproc.hpp>
#include <vector>
using namespace cv;

int main() {
    Mat img(400, 400, CV_8UC1, Scalar(255));
    ellipse(img, Point(120,120), Size(70,45),  20, 0, 360, Scalar(0), 2);
    ellipse(img, Point(280,150), Size(55,55),   0, 0, 360, Scalar(0), 2);
    ellipse(img, Point(200,300), Size(90,40),  60, 0, 360, Scalar(0), 2);
    ellipse(img, Point(320,320), Size(33,50), 130, 0, 360, Scalar(0), 2);
    ellipse(img, Point( 90,320), Size(48,27),  95, 0, 360, Scalar(0), 2);
    Ptr<ximgproc::EdgeDrawing> ed = ximgproc::createEdgeDrawing();
    ed->detectEdges(img);
    std::vector<Vec6d> ellipses;
    ed->detectEllipses(ellipses);
    return 0;
}
```

```
$ valgrind --track-origins=yes ./a.out
==...== Conditional jump or move depends on uninitialised value(s)
==...==    at 0x...: cv::ximgproc::EdgeDrawingImpl::ValidateCircles(bool)
==...==  Uninitialised value was created by a heap allocation
==...==    at 0x...: cv::ximgproc::EdgeDrawingImpl::ValidateCircles(bool)   // the px/py new double[]
==...== ERROR SUMMARY: 2 errors from 2 contexts
```

With the fix applied this is `ERROR SUMMARY: 0 errors`.

Note on scope: valgrind flags only the genuinely-uninitialised reads (the first
candidate, before that slot has ever been written). For later candidates the slot
holds a stale value that valgrind considers initialised, so it does not report those;
the write-count-vs-read-count argument above is what covers that case.

##### Deterministic confirmation (requires a temporary source edit — diagnostic only, NOT the fix)
To confirm, layout-independently, that the slot is never written, temporarily
instrument the call site (edge_drawing.cpp:3254):

```.cpp
// right before ComputeEllipsePoints(circle->eq.coeff, px, py, noPoints):
if (noPoints > 0) { px[noPoints-1] = py[noPoints-1] = -1e300; }   // sentinel
ComputeEllipsePoints(circle->eq.coeff, px, py, noPoints);
// right after it:
if ((noPoints & 1) && px[noPoints-1] == -1e300)
    fprintf(stderr, "odd noPoints=%d: px/py[noPoints-1] left UNWRITTEN\n", noPoints);
```

For every odd `noPoints` this fires (e.g. noPoints = 375, 229 on the image above),
proving `ComputeEllipsePoints` never wrote the slot that the validation loop reads.
This instrumentation is only for diagnosis and is not part of the proposed fix.

- [x] There is reproducer code and related data files

purehol:
### Root cause
`ValidateCircles()` sets `noPoints = computeEllipsePerimeter(...)` (odd about half
the time) and calls `ComputeEllipsePoints(px, py, noPoints)`, which writes exactly
`2 * (noPoints/2)` points. For an odd `noPoints` the last slot `px/py[noPoints-1]`
is never written, yet the validation loop `for (j = 0; j < noPoints; j++)` reads
it — an uninitialised read for the first candidate, or a stale value from a
previous candidate (the buffers are reused). That phantom point is then fed into
`checkValidationByNFA(noPeripheryPixels, aligned)`.

### Fix
Drop the odd point before sampling (`if (noPoints % 2) noPoints--;`), matching the
reference implementation ED_Lib. The OpenCV port had adopted the
`points_buffer_size` bound from the same upstream fix but missed this odd-count
guard.

### Behavior change
This is not a neutral change of output — it removes a phantom sample from the NFA
statistics for odd-perimeter ellipses. The phantom point holds an indeterminate
value (uninitialised for the first candidate, stale for later ones); it is counted
as a periphery sample and, depending on that value, is either skipped by the bounds
check (when it falls outside the image) or fully processed (when it falls inside,
where it can also change `aligned`), so it perturbs the validation of every
odd-perimeter ellipse in an unpredictable direction. The effect is usually small:
over 93 OpenCV sample images plus a synthetic case, the fix changed the
detected-ellipse count on 6 of them (in both directions) and left the other 87
unchanged. Where an ellipse does change, the post-fix computation uses only the real
periphery points (the never-written slot is no longer read); we did not assess which
output is closer to ground truth. Reading the unwritten slot is also undefined
behaviour, flagged by memory sanitizers.

### Verification
On stock OpenCV, `valgrind --track-origins` on `detectEllipses()` over an image
with several ellipses reports "uninitialised value" in `ValidateCircles`
(`ERROR SUMMARY: 2 errors`) before the change and `0 errors` after it. A
deterministic, layout-independent confirmation (temporary source instrumentation,
not part of this change) writes a sentinel to `px/py[noPoints-1]` before the call
and observes it left unwritten for odd `noPoints` (e.g. 375, 229). Added
`TEST_F(ximgproc_ED, detectEllipsesOddPerimeterNoUninitRead)` to exercise the
odd-perimeter path under memory-checking CI. (valgrind flags only the
first-candidate uninitialised reads; the stale-reuse case is covered by the
write-count-vs-read-count argument above.)

Fixes #4207

- [x] There is accuracy test, performance test and test data in the repository, if applicable

### DIFF
--- a/modules/ximgproc/src/edge_drawing.cpp
+++ b/modules/ximgproc/src/edge_drawing.cpp
@@ -3251,6 +3251,11 @@ void EdgeDrawingImpl::ValidateCircles(bool validate)
 
         if (circle->isEllipse)
         {
+            // ComputeEllipsePoints() fills exactly 2*(noPoints/2) points, so for an
+            // odd noPoints the last slot px/py[noPoints-1] is never written; drop it
+            // so the validation loop below does not read an uninitialised/stale value.
+            if (noPoints % 2)
+                noPoints--;
             ComputeEllipsePoints(circle->eq.coeff, px, py, noPoints);
         }
         else
--- a/modules/ximgproc/test/test_fld.cpp
+++ b/modules/ximgproc/test/test_fld.cpp
@@ -328,4 +328,24 @@ TEST_F(ximgproc_ED, detectLinesAndEllipses)
     EXPECT_GE(ellipses.size(), ellipses_size);
     EXPECT_LE(ellipses.size(), ellipses_size + 2);
 }
+
+// Exercises ValidateCircles() over ellipses of varied size/orientation so that
+// candidates with an ODD computed perimeter are processed. For those,
+// ComputeEllipsePoints() leaves px/py[noPoints-1] unwritten while the validation
+// loop reads it; before the fix this is an uninitialised/stale read (flagged by
+// valgrind/MSan at edge_drawing.cpp's ValidateCircles branches). This test guards
+// that path under memory-checking CI.
+TEST_F(ximgproc_ED, detectEllipsesOddPerimeterNoUninitRead)
+{
+    Mat img(400, 400, CV_8UC1, Scalar(255));
+    ellipse(img, Point(120, 120), Size(70, 45),  20, 0, 360, Scalar(0), 2, LINE_8);
+    ellipse(img, Point(280, 150), Size(55, 55),   0, 0, 360, Scalar(0), 2, LINE_8);
+    ellipse(img, Point(200, 300), Size(90, 40),  60, 0, 360, Scalar(0), 2, LINE_8);
+    ellipse(img, Point(320, 320), Size(33, 50), 130, 0, 360, Scalar(0), 2, LINE_8);
+    ellipse(img, Point(90,  320), Size(48, 27),  95, 0, 360, Scalar(0), 2, LINE_8);
+    detector->detectEdges(img);
+    vector<Vec6d> ellipses;
+    ASSERT_NO_THROW(detector->detectEllipses(ellipses));
+    EXPECT_GE(ellipses.size(), 1u);
+}
 }} // namespace


purehol:
Pushed a follow-up commit (b93a0eb) updating the expected ellipse counts in the existing `ximgproc_ED.detectLinesAndEllipses` test (2449 → 2450, 2446 → 2447). This was the cause of the `opencv_test_ximgproc` failure in the previous CI run.

The previous values were computed with the never-written `px/py[noPoints-1]` slot included in the NFA validation for odd-perimeter candidates. With the fix that slot is no longer read, and one additional ellipse passes validation in each mode. The new counts were identical on every CI configuration that ran the test (Ubuntu 20.04/22.04/24.04/26.04, Linux ARM64, ARM64-FastCV, macOS x64/ARM64), so I kept the exact `EXPECT_EQ` checks rather than loosening them. The Windows job did not reach the build step in that run, so I have no Windows numbers yet; if they differ I will switch these two checks to the same `GE/LE` tolerance already used for `lines` in this test.

